PR #9, opened 2018-02-18
Docs: Move .corkboard/ to docs/ and isolate components

chrislloyd:
Context: Upgrading docs, see #6, #7 & #13.

I'm starting to isolate and split up the components that we use in the docs. The idea here is that they should be individually importable from a "template" package or at least baked into the docs site. I've also moved the `.corkboard/` folder to `docs/` to setup future work. This probably should have been done in two separate commits but I was in flow over the weekend so, uh, I guess here it is.

~~I was naughty and there are a few style changes in this PR too.~~ Reverted style changes to make this easier to land.


docs/src/components/Card.js
```diff
@@ -0,0 +1,68 @@
+// @flow
+import React, { isValidElement } from 'react';
+import Box from '../../../src/Box/Box';
```
christianvuerings: @chrislloyd do we need the extra `components` directory? Could we use `docs/src` instead?

`../../..`

docs/src/components/Example.js
```diff
@@ -0,0 +1,88 @@
...
+import Mask from '../../../src/Mask/Mask';
+import Heading from '../../../src/Heading/Heading';
+import { LiveProvider, LiveEditor, LiveError, LivePreview } from 'react-live';
+
+type Props = {|
+  defaultCode: string,
+  scope: Object,
+|};
+
+export default ({ defaultCode, scope }: Props) => (
```
christianvuerings: @chrislloyd I also added `stacked` to `Example` - could we add it back in here?

We use this for the `Container` example which needs a minWidth of `800`

chrislloyd: Sure thing

docs/src/components/Card.js
chrislloyd: Ah yes, this isn’t obvious in this PR - my next PR moves the docs into the src folder. This allows them to be organized in a DAG.

chrislloyd:
@christianvuerings decided to try and land this and do the next steps in a follow-up PR. The netlify build isn't working because the build directory has changed. I need to update that manually in their UI once this diff has landed. I tried looking for a file-base config but no luck?

christianvuerings:
@chrislloyd Did you try their `netlify.toml` file? https://www.netlify.com/docs/continuous-deployment/#deploy-contexts

Issue #5, opened 2018-02-17
Update documentation to use new example format


chrislloyd:
We introduced a new live editable example format a while ago, but not every example has been updated. This issue is to keep track of our progress to converting all the examples.

- [x] Container
- [x] ErrorFlyout
- [x] Flyout
- [x] Heading
- [x] IconButton
- [x] Image
- [x] Label
- [x] Letterbox
- [x] Mask
- [x] Masonry
- [x] Modal
- [x] Pog
- [x] Pulsar
- [x] SearchField
- [x] SegmentedControl
- [x] SelectList
- [x] Spinner
- [x] Tabs
- [x] Text
- [x] TextArea
- [x] TextField
- [x] Toast
- [x] Tooltip

blairanderson:
Looks great. 

Not sure if this is a relevant place to mention, individual component URL paths are broken on github pages https://pinterest.github.io/gestalt/

If you click a component it changes to https://pinterest.github.io/COMPONENT (example https://pinterest.github.io/Container ) which is not a valid path and redirects to the pinterest engineering page.

chrislloyd:
@blairanderson sorry for the slow reply - great catch! It was fixed by @serranoarevalo in #88.

chrislloyd:
All done.

PR #10, opened 2018-02-18
Updates Heading, Image, Label & Text docs to use Example

chrislloyd:
Partial fix for #5.

Screenshots:

Heading

Text

Image


christianvuerings:
@chrislloyd could you add some screenshots?

chrislloyd:
@christianvuerings good call - updated the description. Going to merge anyway because I think it's better than before (even if there are still small errors pending review). If any, would rather fix those in follow up PRs because of some of the other work I'm doing.

Also, can't wait for the per-PR docs!

PR #11, opened 2018-02-18
Internal: Webpack 3 upgrade

christianvuerings:
Before we got an error when trying to build the docs:

screen shot 2018-02-18 at 9 59 02 am


chrislloyd:
Hey this looks good. Just a heads up if you do too much more work on this -
my other PR is aiming on getting us to a place where we can just use
create-react-app (or other system) for the docs. I think we’re a hop, skip
and a jump away from being able to not worry about webpack and use prod
builds in dev :)

PR #14, opened 2018-02-19
ExperimentalMasonry: Reference measurementStore from props instead of instance

bishwei:
This PR updates Masonry so that it always references `measurementStore` from props instead of `this.measurementStore`.  

Currently the measurementStore is bound to the instance in the constructor.  Because of this, if a new measurementStore is passed in during `componentWillReceiveProps`, we actually end up referencing the wrong measurementStore.

### DIFF
--- a/src/ExperimentalMasonry/Masonry.js
+++ b/src/ExperimentalMasonry/Masonry.js
@@ -32,7 +32,7 @@ type Props<T> = {|
   flexible?: boolean,
   gutterWidth?: number,
   items: Array<T>,
-  measurementStore?: Cache<T, *>,
+  measurementStore: Cache<T, *>,
   minCols: number,
   layout?:
     | DefaultLayoutSymbol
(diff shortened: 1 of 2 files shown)


src/ExperimentalMasonry/Masonry.js
```diff
@@ -32,7 +32,7 @@ type Props<T> = {|
   flexible?: boolean,
   gutterWidth?: number,
   items: Array<T>,
-  measurementStore?: Cache<T, *>,
+  measurementStore: Cache<T, *>,
```
bishwei: FYI, I opted not to make the same fix for the control Masonry since that should be going away soon

src/ExperimentalMasonry/Masonry.js
```diff
@@ -406,10 +410,11 @@ export default class ExperimentalMasonry<T> extends React.Component<
 
   render() {
     const {
-      columnWidth = 236,
+      columnWidth,
```
chrislloyd: Good catch!

bishwei:
@chrislloyd - yeah this change would be transparent to the consumer.  The only difference is that flow will scream at you if you try to pass in `null` for `measurementStore`

PR #16, opened 2018-02-20
Create custom width property for flyout

JediMasterRemington:
Sometimes design specifications could require the Flyout to extend pass the five standardized sizes. Because the flyout has no method for extending beyond the size, this PR adds a width property to the Flyout component in order to allow the flyout to have custom sizes for when the Flyout must be a different size other than the standardized sizes.

_Pros_
- allows for flexibility in designing components that use the fly-out
- avoids having to add extra sizes for larger fly-outs (i.e. xxl, xxxl, xxxxl)

_Cons_
- less consistency in fly-outs if developers and engineers choose not to use the standardized sizes

custom flyout width screenshot

### DIFF
--- a/src/Flyout/Flyout.js
+++ b/src/Flyout/Flyout.js
@@ -9,7 +9,7 @@ type Props = {|
   idealDirection?: 'up' | 'right' | 'down' | 'left',
   onDismiss: () => void,
   positionRelativeToAnchor?: boolean,
-  size?: 'xs' | 'sm' | 'md' | 'lg' | 'xl',
+  size?: 'xs' | 'sm' | 'md' | 'lg' | 'xl' | number,
 |};
 
 export default class Flyout extends React.PureComponent<Props> {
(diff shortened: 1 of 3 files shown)


chrislloyd:
Hey @JediMasterRemington - this looks good. However, I think having both a `width` & a `size` property might be a little bit confusing. What would happen if somebody did the following?

```
<Flyout size="sm" width={800} ... />
```

Could you perhaps merge both properties together so that `size` also takes a number?

JediMasterRemington:
I've made it so size can now take a number

src/Flyout/Flyout.js
```diff
@@ -52,5 +52,5 @@ Flyout.propTypes = {
   idealDirection: PropTypes.oneOf(['up', 'right', 'down', 'left']),
   onDismiss: PropTypes.func.isRequired,
   positionRelativeToAnchor: PropTypes.bool,
-  size: PropTypes.oneOf(['xs', 'sm', 'md', 'lg', 'xl']), // default: sm
+  size: PropTypes.oneOf(['xs', 'sm', 'md', 'lg', 'xl', PropTypes.number]), // default: sm
```
chrislloyd: I don't think this will do what you want it to. It's looking for the exact usage of `PropTypes.number` rather than

```
PropTypes.oneOfType(PropTypes.number, PropTypes.oneOf([ ... ]))
```

PR #17, opened 2018-02-20
SearchField: Add custom `onBlur` prop / Rename syntheticEvent => event / Use stricter flowtype on event to remove if check

kennandavison:
Example Use Case: <SearchField> with autocomplete items menu below. Would be nice to close this menu when onBlur is called.

Also renames syntheticEvent to event and uses a stricter flowtype to remove if check.


src/SearchField/SearchField.js
```diff
@@ -63,7 +65,19 @@ export default class SearchField extends React.Component<Props, State> {
       });
     }
   };
-  handleBlur = () => this.setState({ focused: false });
+  handleBlur = (syntheticEvent: SyntheticEvent<>) => {
```
chrislloyd: Can we type the actual element that this is called on here?

src/SearchField/SearchField.doc.js
```diff
@@ -23,6 +23,10 @@ card(
         type: 'string',
         required: true,
       },
+      {
+        name: 'onBlur',
+        type: '({ value: string }) => void',
```
chrislloyd: Do we need to pass the value in here? It hasn't changed with the event and should be the same as what's held by the parent.

kennandavison:
Updated. For anyone wondering, it is suggested that you utilize onChange's last value instead of relying on the event passed from onBlur.

src/SearchField/SearchField.doc.js
```diff
@@ -23,14 +23,24 @@ card(
         type: 'string',
         required: true,
       },
+      {
+        name: 'onBlur',
+        type: '(event: SyntheticEvent<{ target: HTMLInputElement }>) => void',
```
chrislloyd: I think this type has parenthesis `({ event: ... })`

src/SearchField/SearchField.doc.js
```diff
@@ -23,14 +23,24 @@ card(
...
+      },
       {
         name: 'onChange',
-        type: '({ value: string }) => void',
+        type: `({
+          value: string,
+          event: SyntheticEvent<{ target: HTMLInputElement }>
```
chrislloyd: I don't think you need to add `<{ target: HTMLInputElement }>` here - just `<HTMLInputElement>` will match the code.

src/SearchField/SearchField.doc.js
```diff
@@ -23,14 +23,24 @@ card(
...
+        }) => void`,
         required: true,
       },
       {
         name: 'onFocus',
-        type: '({ value: string }) => void',
+        type: `({
+          value: string,
+          event: SyntheticEvent<{ target: HTMLInputElement }>
```
chrislloyd: Same as above.

src/SearchField/SearchField.js
```diff
@@ -9,8 +9,15 @@ import Icon from '../Icon/Icon';
 type Props = {|
   accessibilityLabel: string,
   id: string,
-  onChange: ({ value: string, syntheticEvent: SyntheticEvent<> }) => void,
-  onFocus?: ({ value: string, syntheticEvent: SyntheticEvent<> }) => void,
+  onBlur?: ({ syntheticEvent: SyntheticEvent<HTMLInputElement> }) => void,
```
chrislloyd: The key should be named `event` instead of `syntheticEvent`.

src/SearchField/SearchField.doc.js
chrislloyd: Also, the key `event` should be `syntheticEvent`.

PR #20, opened 2018-02-21
Avatar: Add optional children prop

kennandavison:


chrislloyd:
What's the use case for changing the color of the icon? I'd also prefer to not add the extra prop that's just passed through to the Icon - could we pull the Icon out of Avatar instead?

kennandavison:
@chrislloyd 

Use Case: I thought it'd be nice to change the color of the verified check for people using Gestalt OS. 

I do agree that it'd make more sense to pass the Icon through children. I'll update the PR to add an optional children prop to Avatar and people can mirror Avatar's verified code if they want a different color.

chrislloyd:
@kennandavison I appreciate the enthusiasm but I'm hesitant to merge changes without a solid use-case. Extra props are something we'll have to support over time.

PR #21, opened 2018-02-21
Sticky: Add zIndex support

bishwei:
Currently, if any element that appears after `Sticky` sets a `position` attribute, it will be stacked on top.  This makes `Sticky` somewhat useless since, most of the time, it will be followed by some sort of scrollable content.  

This PR updates `Sticky` so that, by default, it sets a z-index of 1.  This should handle the 99% case where elements that appear after Sticky set a position attribute, but no z-index.  For the other 1% case where we might have to set a z-index, I've added a `dangerouslySetZIndex` prop.  

cc @PFarejowicz

### DIFF
--- a/src/Sticky/Sticky.doc.js
+++ b/src/Sticky/Sticky.doc.js
@@ -18,7 +18,7 @@ card(
         type: 'number',
       },
       {
-        name: 'children',
+        name: 'children?',
         type: 'React.Node',
       },
       {
@@ -40,9 +40,6 @@ card(
       },
     ]}
   />,
-  md`
-    **Note**: if no threshold is passed to Sticky, it will behave the same as a relatively positioned element
-  `,
   { stacked: true }
 );
 
--- a/src/Sticky/Sticky.doc.js
+++ b/src/Sticky/Sticky.doc.js
@@ -19,7 +19,7 @@ card(
       },
       {
         name: 'children',
-        type: 'any',
+        type: 'React.Node',
       },
       {
         name: 'left',
@@ -32,6 +32,11 @@ card(
       {
         name: 'top',
         type: 'number',
+      },
+      {
+        name: 'dangerouslySetZIndex',
+        type: 'number',
+        defaultValue: 1,
       },
     ]}
   />,
@@ -53,15 +58,15 @@ card(
           <Text>This should stick</Text>
         </Box>
       </Sticky>
-      <Box marginTop={10}>
+      <Box marginTop={10} position="relative">
         <Text>Scroll</Text>
         <Text>Keep scrolling</Text>
         <Text>Scroll more</Text>
       </Box>
     </Box>
     <Box>
-      <Sticky top={0}>
-      <Box alignItems="center" color="lightGray" display="flex" height={40}>
+      <Sticky top={0} dangerouslySetZIndex={3}>
+      <Box alignItems="center" color="lightGray" display="flex" height={40} position="relative" dangerouslySetInlineStyle={{ __style: { zIndex: 2 } }}>
           <Text>This should also stick</Text>
         </Box>
       </Sticky>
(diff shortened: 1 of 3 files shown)


src/Sticky/Sticky.js
```diff
@@ -13,20 +13,23 @@ type Threshold =
...
 
 type Props = {|
-  children: any,
+  children: React.Node,
+  dangerouslySetZIndex?: number,
   ...Threshold,
 |};
 
 export default function Sticky(props: Props) {
+  const { dangerouslySetZIndex = 1, children } = props;
```
bishwei: couldn't destructure this above without flow screaming at me

src/Sticky/Sticky.doc.js
```diff
@@ -53,15 +58,15 @@ card(
           <Text>This should stick</Text>
         </Box>
       </Sticky>
-      <Box marginTop={10}>
+      <Box marginTop={10} position="relative">
```
bishwei: updated the docs - first example is the 90% case where you shouldn't need to set zindex manually.  second example is one where you would

PFarejowicz:
Awesome, this looks good by me!

Something else I noticed, the docs say “**Note**: if no threshold is passed to Sticky, it will behave the same as a relatively positioned” but the Flow typing for Threshold requires at least one of the configurations to exist, meaning you can’t pass no threshold.

src/Sticky/Sticky.doc.js
```diff
@@ -19,7 +19,7 @@ card(
       },
       {
         name: 'children',
```
chrislloyd: This should probably be `children?` if we're going to change it.

chrislloyd:
I wonder if we should follow the same lead of `dangerouslySetInnerHtml={{ __html: 'asdf' }}` and `dangerouslySetInlineStyle={{ __style: { ... } }}` and pass an object here?

```
<Sticky dangerouslySetZIndex={{ __index: 999 }} />
```

The deterrent with the dangerous property is both in the name and the amount of characters that it takes to write.

bishwei:
@chrislloyd i'll defer to you on that one.  i think having `dangerouslySet` in the prop name is valuable since it does make you question your choice here.  however, changing it from a string to an object isn't going to make a dev think more about using it.

src/Sticky/Sticky.doc.js
```diff
@@ -18,8 +18,8 @@ card(
         type: 'number',
       },
       {
-        name: 'children',
-        type: 'any',
+        name: 'children?',
```
PFarejowicz: Was the "?" on purpose?

PFarejowicz:
@chrislloyd I thought about that when I was reading through this, but in the end came to the same conclusion as @bishwei. To me, the z index in and of itself isn't dangerous like letting the programmer implement any css in `style` or potentially introducing XSS in `html`. Z index is only disastrous when devs start using arbitrary values with no explanation. Now that I think about it, is it worth it to still set it to 1 by default, or just have it completely up to the dev so they are fully aware of her/his actions?

src/Sticky/Sticky.doc.js
bishwei: children shouldn't be optional for sticky imo

bishwei:
my reasoning for defaulting zIndex to 1 is that i think it should cover the majority of use cases, where the content you want to scroll may be relatively or absolutely positioned (or both) but have no z-index set.  in that scenario, `Sticky` would still work without having to set zIndex

chrislloyd:
Yeah I think the default here makes sense. However zIndexes are “dangerous” in the sense that there are lots of trade offs that you need to bear in mind when using them and we want to remind people of that before they use them. We could definitely have another label other than “dangerous” because there are no security trade offs, however I’d prefer to change that in one go so we’re consistent (change dangerouslySetInlineStyle too).

Issue #13, opened 2018-02-19
Modernize documentation & project structure


chrislloyd:
The docs are currently used for:

* Fast development of new components
* Specialized education about components

While the docs have been good so far, there have been a few niggling points that haven't been addressed:

* The docs build off their own interpretation of source files. The can lead to differences occurring between the production builds and documentation site.
* The reference source files directly. By reaching into `src/Box/Box.js` they can bypass testing wether the component is actually exported. This has lead to bugs.
* The static build doesn't server render any components
* It's difficult to group components together (or make any sort of logical changes to navigation)

I've tried addressing some of these issues in the past but have always found difficulties. I'm proposing a solutions that I think will iteratively move the docs to a better place.

1. Rename `.corkboard/` to `docs/`.
1. Move docs files inline into that folder. This is probably controvercial because it's easy and nice being able to have docs co-located with components. However, I think moving them away helps enforce their isolation from the source. It also gives us the flexibility to write documentation that isn't directly tied to a component (i.e. What are all the colors we use? How do I install Gestalt?)
1. Move the docs to their own [yarn workspace](https://yarnpkg.com/lang/en/docs/workspaces) that is linked against the production build. The biggest downside to this step is that to use the docs you may need to run `yarn watch` and `cd docs && yarn start` at the same time. I don't think this is too much of an inconvenience but we can tool around it if necessary.
1. Upgrade corkboard. `corkboard@0.6.0` basically reduces the API down to `group` and `card`. This means we can self-contain all styling in the docs folder itself. We've made really good progress towards this previously but I'd like to finish the job.
1. Use `create-react-app` for the docs. This will allow us to remove most of the config that we currently have for them and give us some nice extra tooling. I don't see this as a forever solution (it doesn't solve SSR) but it will take some of the immediate pain away.

There are a couple of assumptions I'm making here which are probably worth calling out:

1. The docs co-located next to their component source files aren't worth the convenience. This probably means we'll remove corkboard at some point as it's not really needed if the docs co-located.
1. We should start leveraging yarn workspaces. The docs, grid playground + integration tests & gestalt itself are all probably separate packages. This also aligns with future work.
1. `create-react-app` is a good stepping stone to better docs. We could use [next](https://github.com/zeit/next.js/) or [gatsby](https://www.gatsbyjs.org/) but before we move there, CRA is a good environment that similar to our current setup but can be totally isolated.

chrislloyd:
Done now across several issues above.

PR #22, opened 2018-02-21
Internal: Workspaces

chrislloyd:
This PR uses Yarn workspaces to split Gestalt into 3 separate packages: `/docs`, `/tests` (masonry integration tests) & `gestalt` itself. This PR looks massive but really just a lot of files have changed locations.

For context & backstory please see #13.

NOTES:

- The Dockerfile isn't awesome yet, there a bunch of duplication involved. I think this could be optimized in a future PR.
- `codemods` & `scripts` are kind of on their own. More infra around those in the future.
- I'm like 60% confident publishing works - the next person to do it might need to do a little bit of massaging (and I'm happy to help out with that).
- Flow doesn't run in the docs and integration tests. This isn't awesome but isn't tooo terrible either.
- This error isn't killable until we update the integration tests.

screen shot 2018-02-28 at 10 40 40 am

TODO:

- [x] Export list of icons available
- [x] Update README
- [x] Update CHANGELOG
- [x] Verify publishing works
- [x] Verify hot reloading works
- [x] Take another stab at having flow run in docs/integration tests


pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

packages/gestalt/src/Icon/Icon.js
```diff
@@ -58,6 +60,8 @@ export default function Icon(props: IconProps) {
   );
 }
 
+Icon.icons = IconNames;
```
bishwei: hmmm can we only expose this in dev/testing?
something like: 
```
if (__DEV__) {
  Icon.icons = IconNames
}
```

docs/src/index.js
```diff
@@ -5,8 +5,9 @@ import App from './components/App';
 import CardPage from './components/CardPage';
 import { render } from 'react-dom';
 import { getCards } from 'corkboard/init';
+import '!style-loader!css-loader!gestalt/dist/gestalt.css';
```
bishwei: don't feel strongly either way, but i think we could also just add `noParse: /gestalt/` to the webpack config

docs/package.json
```diff
@@ -0,0 +1,45 @@
+{
+  "name": "gestalt-docs",
+  "version": "0.0.0",
```
christianvuerings: @chrislloyd Should these be in sync with the Gestalt version? Or will we always leave it at `0.0.0`?

docs/src/Avatar.doc.js
```diff
@@ -1,10 +1,8 @@
 // @flow
 import * as React from 'react';
 import PropTypes from 'prop-types';
-import Avatar from './Avatar';
-import Box from '../Box/Box';
-import Text from '../Text/Text';
-import { ns, card, md, PropTable, Example } from '../../docs/src/cards';
+import { Avatar, Box, Text } from 'gestalt';
```
christianvuerings: 💯

yarn.lock
```diff
@@ -7333,7 +7172,7 @@ react-test-renderer@^15.3.2:
     fbjs "^0.8.9"
     object-assign "^4.1.0"
 
-react@^15.1.0:
+react@^15.1.0, react@^15.4.1:
```
christianvuerings: @chrislloyd are we sure that this won't load 2 versions of React when people use Gestalt?

docs/src/index.js
chrislloyd: My next step is to just have the docs use CRA so no webpack at all! Didn’t want to spent too much time tidying this up so just followed conventions.

chrislloyd: My next step is to just have the docs use CRA so no webpack at all! Didn’t want to spent too much time tidying this up so just followed conventions.

packages/gestalt/src/Icon/Icon.js
chrislloyd: Yeah… Is your concern here bundle size? We’re not currently building multiple envs of Gestalt so adding something like this would be more work than it looks.

I don’t feel strongly about this one - was a request from @christianvuerings to save some typing in the docs.

docs/package.json
chrislloyd: The version here is mostly just to suppress warnings as it’s a private package. The version is really the Git SHA of whatever you have it checked out at. I think adding a version here and pegging it to Gestalt could just be busywork.

chrislloyd: The version here is mostly just to suppress warnings as it’s a private package. The version is really the Git SHA of whatever you have it checked out at. I think adding a version here and pegging it to Gestalt could just be busywork.

yarn.lock
chrislloyd: Good catch! However I don’t think this affects package installing as only the package.json version ranges are respected there for libs. What this does mean however is that either the docs or tests are using a different version of React which is problematic.

packages/gestalt/src/Icon/Icon.js
bishwei: not bundle size, just pointing it out since it does seem like something that is specifically done for testing.  my first thought when i looked at this was "why is this here" and had to go look around for how it was used so, that's why i raised it.

definitely not a concern for this pr.  happy to take this as a follow up action item for myself as well

chrislloyd:
Agreed yeah. Let’s revisit once it’s landed - it’s awesome but probably not
terrible either.

PR #31, opened 2018-02-23
Touchable full height

pfarejowicz:
`Touchable` has prop to control the width of the element relative to its parent, I ran into a situation where I needed to access this same control for the height - there was a fixed height being set on a parent and a child of `Touchable` needed to be full height of this parent that was up the chain but the underlying `div` element would not allow a `height: 100%` to be set. Usually you can move the `Touchable` itself higher in the tree to wrap more components, but in this situation the event handlers were defined at the child level and did not allow us to simply move it all up.

Also fixed a redundant existence check on the `onTouch` prop in an event handler method.

New prop table:
screen shot 2018-02-23 at 11 10 44 am
New example:
screen shot 2018-02-23 at 11 32 03 am


christianvuerings:
@pfarejowicz How do you feel about updating the example to make it a bit more obvious?

image

```
<Box color="white" display="flex" width={500} height={250}>
  <Box column={6}>
    <Touchable fullHeight>
      <Box height="100%" color="lightGray">
        <Text align="center">
          Full parent height
        </Text>
      </Box>
    </Touchable>
  </Box>
  <Box column={6}>
    <Touchable>
      <Box height="100%" color="lightGray">
        <Text align="center">
          Child height only
        </Text>
     </Box> 
    </Touchable>
  </Box>
</Box>
```

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

PR #41, opened 2018-03-04
Removes corkboard from docs

chrislloyd:
Context: #13

Corkboard was a pretty useful project when we wanted to keep the docs close to the components and very light-weight. However, we're now moving towards documentation which is more heavy-weight and closer to a real, fully fledged site. In addition I also don't think corkboard ever quite lived up to its promise of REPL-like documentation next to components (more on that later). This PR removes corkboard from the docs project so help with future upgrades and changes.

My approach here was to do the minimum possible amount of changes that struck a balance between leaving the project in a good state and having to totally re-write all the docs. I settled on each `.doc.js` file export it's own component, which each file having a little shim to support the corkboard syntax:

```jsx
let cards = [];
```

At the end of the file we collect those cards and export a component:

```jsx
export default () => <CardPage cards={cards} />;
```

In my mind, this approached helped maintain the writing order of all the docs, but made it so that we could migrate (and tend to) each doc file individually. Now their only thing each file cards about is exporting a component so internal changes are easy.

_Note:_ most of the changes in this diff were achieved by code-mods which I haven't committed seeing as they're a one-time thing. While it may look like a big diff, I took extra care to make sure each individual commit doesn't break/change the docs so stepping through commit-wise is recommended to review.


pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

docs/package.json
```diff
@@ -5,17 +5,17 @@
...
   "dependencies": {
-    "corkboard": "0.3.3",
     "gestalt": ">0.0.0",
     "highlight.js": "^9.12.0",
     "marked": "^0.3.17",
-    "react": "^15.4.1",
     "react-dom": "^15.4.1",
-    "react-live": "^1.10.1"
+    "react-live": "^1.10.1",
+    "react-router": "^2.0.1",
```
bishwei: doesn't really matter since this is docs, but just curious why 2?

chrislloyd: Because it was the same version as corkboard had. Can do an upgrade in another PR.

docs/src/components/Navigation.js
```diff
@@ -2,10 +2,9 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { Text, Box, SelectList, Link, Icon, Heading } from 'gestalt';
+import routes from '../routes';
 
-type Props = {|
-  cards: *,
-|};
+type Props = {||};
```
christianvuerings: @chrislloyd What does this mean? That no props should be passed to it?

chrislloyd: yeah, could probably use void here. Will revisit

PR #42, opened 2018-03-04
Uses create-react-app for the docs

chrislloyd:
Context: #13 
Depends on: #41 

This PR adds [create-react-app](https://github.com/facebook/create-react-app) to manage the build of the docs. It's a faster, better maintained version of what we had already that lets us focus more on writing great docs instead of messing with build configs.

I don't see us using this forever (I think a system like [Gatsby](https://www.gatsbyjs.org/) is better in the long run) but this was an easy PR and will save some time in the meantime.

Note: I had to downgrade Jest because of an issue with how Jest resolves dependencies. [It looks like it's fixed](https://github.com/facebook/jest/issues/5294) however we need to wait for CRA to update to the version with the fix. Jest 22 provided support for puppeteer, which is a shame to loose but seeing as we're not using it right now I think it's a good tradeoff.


pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

gaearon:
I published `react-scripts@1.1.3` which should fix `environment.dispose is not a function` for you.

https://github.com/facebook/create-react-app/releases/tag/v1.1.3

chrislloyd:
@gaearon you're the best, thanks!

PR #51, opened 2018-03-06
IE11 fixes: make docs work in IE11 / fallback to position relative on Sticky / fix images in docs / fix scrollbar always showing on proptable

christianvuerings:
Before / After always scrollbar:
screen shot 2018-03-06 at 9 46 52 am
screen shot 2018-03-06 at 9 47 14 am

Before / After images not showing:
screen shot 2018-03-06 at 9 41 14 am
screen shot 2018-03-06 at 9 43 07 am

Docs not working at all:
screen shot 2018-03-06 at 9 35 30 am


packages/gestalt/src/index.js
```diff
@@ -1,4 +1,5 @@
 // @flow
+import 'babel-polyfill';
```
christianvuerings: @chrislloyd @bishwei This is probably the most controversial change, wdyt? Any suggestions of an alternative?

chrislloyd: Yeah, I'm a hard 👎  on this - we should either update `browsers` in `package.json` to ship with IE11 support or we should figure out something else. This brings all of the regenerator runtime into Gestalt.

bishwei: imo, this should be on the consumer, not on gestalt.

also, this would definitely bloat our js size a lot

chrislloyd:
@christianvuerings can you perhaps land this without the babel-runtime change and we can discuss a good option for that going forward? Seems like this PR has lots of "ands" :P

PR #53, opened 2018-03-06
React 16 support

chrislloyd:
This updates Gestalt's compatibility to include React 16.2.0. It's pretty straight forward (most of the heavy lifting has been done in previous PRs). The thing to call out here is the disabled resize tests. I've manually verified that this still works and so I think there's a subtle bug in Ghost or our test. The resize tests don't actually resize the browser window but instead resize the container so at best they were just testing a proxy for the behavior we wanted and should probably be further investigated at some point in time.

/cc @kevingrandon


pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

PR #57, opened 2018-03-06
SelectList: Hardcode 40px height for consistency

kennandavison:
Currently, SelectList's total height is 43px which looks a bit funny when side by side with TextField which is 40px. I've hard coded height to 40px.
screen shot 2018-03-06 at 3 48 01 pm
screen shot 2018-03-06 at 3 47 47 pm
screen shot 2018-03-06 at 3 47 29 pm


pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

packages/gestalt/src/SelectList/SelectList.css
```diff
@@ -11,7 +11,7 @@
   border-radius: 4px;
   border-style: solid;
   border-width: 1px;
-  padding: 11px 35px 11px 14px; /* right padding prevents it from running into arrow */
+  padding: 9.5px 35px 9.5px 14px; /* right padding prevents it from running into arrow */
```
chrislloyd: Fractional pixels here are going to cause the text to be blurry - perhaps you can remove the padding and fix the total `line-height` of the select list?

chrislloyd:
Also this is worth a changelong entry :)

packages/gestalt/src/SelectList/SelectList.css
kennandavison: It appears that using anything below line-height 17px results in the value/placeholder being cutoff. See: 
screen shot 2018-03-06 at 3 29 11 pm
screen shot 2018-03-06 at 3 29 04 pm

kennandavison: I've used a mix of padding that isn't fractional and line height to achieve the 40px in the most recent commits.

PR #72, opened 2018-03-10
Docs: Fix directly browsing to components

christianvuerings:
https://deploy-preview-72--gestalt.netlify.com/Avatar => Should work
https://deploy-preview-70--gestalt.netlify.com/Avatar => Results in a 404

See https://www.netlify.com/docs/redirects/#history-pushstate-and-single-page-apps


christianvuerings:
Seems like we'll also need some extra things to make it work with github pages: https://github.com/rafrex/spa-github-pages

chrislloyd:
Would we just add this file so it’s more obvious what’s going on? Or just use the HashRouter…

christianvuerings:
@serranoarevalo Fixed this in #88

PR #75, opened 2018-03-12
Add support for nofollow prop in Link component

istian11:
We'd like to be able to explicitly specify whether to include the rel=nofollow attribute to Gestalt links. This is important for SEO, so that we can avoid needless following of links by search bots.

### DIFF
--- a/packages/gestalt/src/Link/Link.js
+++ b/packages/gestalt/src/Link/Link.js
@@ -8,6 +8,7 @@ type Props = {|
   children?: any,
   href: string,
   inline?: boolean,
+  nofollow?: boolean,
   onClick?: ({ event: SyntheticMouseEvent<> }) => void,
   target?: null | 'self' | 'blank',
 |};
@@ -59,7 +60,7 @@ export default class Link extends React.Component<Props, State> {
       inline = false,
       nofollow = false,
       target = null,
-      href
+      href,
     } = this.props;
     const relAttributes = [];
     if (target === 'blank') {
@@ -68,7 +69,7 @@ export default class Link extends React.Component<Props, State> {
     if (nofollow) {
       relAttributes.push('nofollow');
     }
-    const rel = relAttributes.join(' ');
+    const rel = relAttributes.join(' ') || null;
     const linkTarget = target ? `_${target}` : null;
 
     return (
--- a/packages/gestalt/src/Link/Link.js
+++ b/packages/gestalt/src/Link/Link.js
@@ -23,6 +23,7 @@ export default class Link extends React.Component<Props, State> {
     children: PropTypes.node.isRequired,
     href: PropTypes.string.isRequired,
     inline: PropTypes.bool,
+    nofollow: PropTypes.bool,
     onClick: PropTypes.func,
     target: PropTypes.oneOf([null, 'self', 'blank']),
   };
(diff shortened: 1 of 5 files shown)


pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

packages/gestalt/src/Link/Link.js
```diff
@@ -53,8 +54,21 @@ export default class Link extends React.Component<Props, State> {
   };
 
   render() {
-    const { children, inline = false, target = null, href } = this.props;
-    const rel = target === 'blank' ? 'noopener noreferrer' : null;
+    const {
+      children,
+      inline = false,
```
christianvuerings: Note to self - we should be using `defaultProps` instead of destructuring the props and setting default values.

https://stackoverflow.com/a/42733457/117193

packages/gestalt/src/Link/Link.js
```diff
@@ -53,8 +54,21 @@ export default class Link extends React.Component<Props, State> {
...
+      href
+    } = this.props;
+    const relAttributes = [];
+    if (target === 'blank') {
+      relAttributes.push('noopener', 'noreferrer');
+    }
+    if (nofollow) {
+      relAttributes.push('nofollow');
+    }
+    const rel = relAttributes.join(' ');
```
christianvuerings: @istian11 We want to make sure `rel` is `null` instead of an empty string when nothing needs to be set.

packages/gestalt/src/Link/Link.js
chrislloyd: That link is pretty... light on any details. Claims about perf without any numbers aren't particularly convincing. Claims about correctness without pointing out any actual bugs aren't convincing either :/

chrislloyd:
@istian11 I mentioned this to @cpapazian but I was wondering if there was any generalizable rule for when we should be adding `nofollow` to links? Is there a good default here? As a developer I really have to understand what `nofollow` does to be able to use it. I think we might be missing out on an opportunity to set this transparently for all devs _or_ educating them as to its consequences.

cpapazian:
i don't think there is a simple way to generalize these rules. What makes sense in one context is the opposite in another. it's actually the structure of the site, not something intrinsic to the link that informs whether a it makes sense to allow a crawler to follow a particular link.

Another way of thinking about generalizing this is to allow Tabs and other components that contain links to include a link component as a child, rather than automagically construct the link. Then callers could use any component that renders to a link or button or whatever they want.

cpapazian:
@chrislloyd thoughts? i think the main point is that any attempts to pick out sane defaults for nofollow, etc, need to be in the context of site structure, and so there's no way to do this within gestalt. i suppose i could see rules for defaults passed down via context, but that feels pretty awkward.

cpapazian:
@chrislloyd is the recommendation that this PR be updated to surface `rel` generally? if we make htat change, can we merge it?

chrislloyd:
100% - that would be awesome.

chrislloyd:

PR #77, opened 2018-03-13
Revert Masonry + React 16 changes

bishwei:
- Revert "Promotes ExperimentalMasonry to Masonry (#46)"
- Revert "React 16 support (#53)"


chrislloyd:
For those looking on, one of the experiments that we were running regarding Masonry was setup incorrectly. We're reverting this for the moment to recollect metrics with the intention of "reverting the revert" next week-ish.

PR #79, opened 2018-03-13
Flexible tab backgrounds

pfarejowicz:
The `<Tabs />` component feels a little rigid in its coloring choices given that it sets no background on the container, this means that the intended look will only be correct if the parent container is given a white background.

This change makes the tabs have a transparent background which means they show up as expected across all different background colors. The new active color is now a 6% black overlay which means there is no visual difference if you place the tabs on a component with a white background from before, but this change allows the tabs to also show up correctly is the background is something different, like gray.

Before:
screen shot 2018-03-13 at 3 47 56 pm
screen shot 2018-03-13 at 3 48 58 pm
screen shot 2018-03-13 at 4 12 22 pm
screen shot 2018-03-13 at 4 15 00 pm

Notice how if the background of the parent changes to anything besides white, the tabs look inverted

After:
screen shot 2018-03-13 at 3 40 45 pm
screen shot 2018-03-13 at 5 04 47 pm
screen shot 2018-03-13 at 5 04 19 pm
screen shot 2018-03-13 at 5 05 17 pm


pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

chrislloyd:
@pfarejowicz what do the tabs look like on other colored backgrounds? I agree that it looks weird when the background is the same color as the tab.

jvzingarelli:
@pfarejowicz I would remove the transparency but keep the colors the same no matter what IMO. But that still constrains their use on some background colors.

jvzingarelli:
@pfarejowicz What I mean is keep the background color of the SELECTED tab the same at 100% opacity while removing the background pill of the unselected altogether.

PR #80, opened 2018-03-14
Image: Don't show  text when loading the image in Firefox

christianvuerings:
Before:
screen shot 2018-03-13 at 9 57 14 pm

After:
screen shot 2018-03-13 at 9 56 57 pm

### DIFF
--- a/packages/gestalt/src/Image/Image.css
+++ b/packages/gestalt/src/Image/Image.css
@@ -4,6 +4,15 @@
   composes: absolute from "../Layout.css";
 }
 
+/*
+    Disables alt text showing when loading an image in firefox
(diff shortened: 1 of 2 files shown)


packages/gestalt/src/Image/Image.css
```diff
@@ -4,6 +4,15 @@
   composes: absolute from "../Layout.css";
 }
 
+/*
+    Disables alt text showing when loading an image in firefox
+    https://stackoverflow.com/questions/12667926
+*/
+
+img[alt] {
```
chrislloyd: This will be global for all images - I’m pretty sure that’s not you want here. Can we make this a class?

PR #89, opened 2018-03-20
Docs: Run the docs locally on port 8080 + update README.md

christianvuerings:


README.md
```diff
@@ -31,19 +31,10 @@ yarn
...
-cd packages/gestalt
-yarn watch
-```
-
-Start the docs server:
-
-```bash
-cd docs
 yarn start
-open "http://localhost:8080"
```
christianvuerings: Before the default was `3000`, however this didn't allow us to run it concurrently with the masonry playground (see below)

We also don't need to run the `open` command, create-react-app automatically does that for us.

christianvuerings:

Built with commit c90a1cdd543988a90f8d522c5f2f69edcc63d162

https://deploy-preview-89--gestalt.netlify.com

chrislloyd:
Should we change the Masonry playground instead? CRA has a really awesome port picking tool which would be negated by fixing it in the environment. Ideally we’d run both with CRA and they’d just pick what port they need to run on automatically.

PR #100, opened 2018-03-23
Masonry: Perform measurements "offscreen"

bishwei:
Currently, Masonry mounts all of the items its going to measure in the top left corner of the grid.  This is totally fine and works, but has the downside that unless a user scroll down, all the items being mounted to the grid are immediately considered to be in the viewport.  

This PR moves the measurement items into a separate div that is rendered below the grid.  Since Masonry measures items one row at a time, this means that as newly measured items get rendered, the "measurement" div will naturally be pushed further down.  

Tested this on mobile + desktop and everything looks 👌

### DIFF
--- a/packages/gestalt/src/ExperimentalMasonry/Masonry.js
+++ b/packages/gestalt/src/ExperimentalMasonry/Masonry.js
@@ -506,7 +506,8 @@ export default class ExperimentalMasonry<T> extends React.Component<
             {itemsToRender.map((item, i) =>
               this.renderMasonryComponent(item, i, positions[i])
             )}
-
+          </div>
+          <div className={styles.Masonry} style={{ width }}>
             {itemsToMeasure.map((data, i) => {
               const position = measuringPositions[i];
               return (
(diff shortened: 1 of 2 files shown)


christianvuerings:

Built with commit 4c0e5d7b8ac2a6bb27a3e1ab5c5bd74b4c0b12f1

https://deploy-preview-100--gestalt.netlify.com

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

bishwei:
@chrislloyd so two differences with this approach which should be okay compared with the previous approach (iirc):
1) There is separate relatively positioned container
2) The items being measured are absolutely positioned in this container

This means that the measurement container will never occupy a height and cause the jumpiness when measuring, since all the items being measured are absolutely positioned.  In the approach, the items being measured were initially not absolutely positioned, which cause the container height to change and cause a jumpy effect.  

Happy to chat more about this - I think I know which issue you're referring to

chrislloyd:
Sweet, thanks for the explanation!

PR #107, opened 2018-04-04
Add right to left language Support

lunaruan:
Added right to left language support by swapping certain CSS properties (things that are right are now left and vice versa). To support right to left languages, we append `dir="rtl"` on the `html` header. We also use `html[dir="rtl"]` and `html:not([dir="rtl"])` to specify whether the page is supposed to be right to left or not right to left and use appropriate CSS for each case


christianvuerings:

Built with commit 1ad9b52a4ca27cf1fb15ba10b3443dad287d4638

https://deploy-preview-107--gestalt.netlify.com

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

bishwei:
@lunaruan can you elaborate on the requirements here and how much RTL support are we going for?  

based on the whitespace changes it seems like we're saying swap everything (padding, margin, etc) from right to left - but i echo @chrislloyd 's concerns around bundle size and also, whether or not this is even a safe assumption without additional supporting changes

zackargyle:
@chrislloyd @bishwei what do you think about creating a secondary css file? gestalt-rtl.css? Would avoid the bloat issue.

bishwei:
i'm 👍with having either a separate or additional rtl specific gestalt css for bundle size purposes, but i still think we should think through the whitespace changes more.  just swapping left/right is not the same as RTL imo

one random example: 
this text has a left margin of 3 boints but for RTL, margin-right would only make sense if the display was right aligned and swapped
image

chrislloyd:
In general I'm fine supporting RTL in components (Text, Flyout, Masonry etc.). I'm particularly hesitant with auto changing layout on top of what the browser already does to support RTL because there are plenty of places where asymmetrical padding makes sense and probably shouldn't be RTL compatible (english text next to a logo for instance). Once we do this it would be difficult to go back. I'm curious as to how many occurrences of mis-matched margins show up? Is this a problem we should be fixing in Gestalt?

lunaruan:
So I do think we'll need to do something extra on top of HTML's built in `dir=rtl` feature. For example:

image

In this case, the Search text and the text are all aligned left, and they should be aligned right. However, maybe this isn't the best solution. We can generate a separate right to left gestalt.css in gestalt that users can use for right to left languages. What do you think?

Re. Yen's concern, that would be flipped, so the button would be on the right and the text would be on the left. In general for right to left languages everything is swapped (see [Instagram](https://www.theverge.com/2017/10/31/16584434/instagram-right-to-left-language-support-arabic-farsi-hebrew))

lunaruan:
As per conversation yesterday, going to code inside pinboard to support rtl for whitespace. Only flipping text and borders here

zackargyle:
Summary: I disagree with moving white space hacks to pinboard.

The margin issue is unavoidable, even with using padding in a small number of cases. Even the "Create board" example above wouldn't work with padding because we actually don't want padding on the opposite side. That extra padding will reduce the amount of space that longer translations need. We want the text to be able to flow until the very end. There is good reason why other companies flip the margins in their css. Any one-offs can be adjusted, but we are optimizing for the more general case within the UI library.

chrislloyd:
@zackargyle approving this change for the moment because it's _currently_ uncontroversial. Re: "Create Button" example - I ran through this with @lunaruan and @bishwei yesterday - it's definitely achievable with negative margins.

Generally speaking though, I'd feel waay more comfortable supporting flipping all margins if we knew how we can approach exceptions. It feels like a big change that could be potentially difficult to undo seeing as it relies on global HTML state. Proposals I can think of: 

1. `<Box dir="ltr" />` fixes swapping behavior. Opt in or could be read from context.
2. `<Box marginStart={1} marginEnd={1} />`  - new properties for RTL-aware margins. This mirrors [Android's approach to RTL](https://developer.android.com/about/versions/android-4.2.html#RTL).
3. Drive RTL-ness with context and have some intermediate component to turn it off in places: `<RTLProvider dir="ltr" />`
4. ???

I personally like #2 the most - it's incremental and supporting RTL margins on mweb would be a pretty simple codemod away. We could also do that before upgrading to React 16.3 for new context APIs.

bishwei:
i like the option of driving RTL via context, once we can leverage the next context APIs.  If `Box` and other components were `dir` aware, we wouldn't even need to ship down extra CSS for the browser.  Also, components could still have the option of opting out of any default ltr -> rtl behavior we add.

i think the thing to call out here is having the ability to easily opt out of the default behavior, which we can't really do if we impose the auto flipping (i.e. `borderLeft` -> `borderRight`).

lunaruan:
Hm I also like option 2. However, doesn't that pose the same problems where RTL isn't an opt in experience?

PR #115, opened 2018-04-09
Updating Pulsar component to new ring style

dalvarez:
This will update the Gestalt Pulsar component to the new blue ring version. 
Activation tested. PDS approved.


christianvuerings:

Built with commit 26e4aaa1f9390a219327845aaa5c6a045ddd23a6

https://deploy-preview-115--gestalt.netlify.com

packages/gestalt/src/Pulsar/Pulsar.css
```diff
@@ -1,86 +1,49 @@
...
-.outerCircle {
-  composes: absolute from "../Layout.css";
-  composes: circle from "../Borders.css";
-  composes: darkGrayBg from "../Colors.css";
-  animation-duration: 3s;
-  animation-iteration-count: inherit;
-  animation-name: OuterAnimation;
-  height: 100%;
-  width: 100%;
+.innerCircle:hover {
```
christianvuerings: @dalvarez should we also add a `:focus` here with the same styles? How does this behave when you're navigating with the keyboard only

christianvuerings:
@dalvarez Could we also update `CHANGELOG.md`?

dalvarez:
@chrislloyd good callout! i removed the animation start delay
@christianvuerings going to add the focus and keyboard support, as well as another example in the Pulsar doc

Issue #113, opened 2018-04-09
Docs: can't open components in a new tab


christianvuerings:
From one of our users:

> On https://pinterest.github.io/gestalt, I can't open pages from the menu on the left in a new tab.  Redirects to https://engineering.pinterest.com.  I am new to gestalt and having multiple tabs open is useful for figuring out how different components fit together.

MateusGabi:
Yeah!!

## Use case:
> `Ctrl+click` or `right click -> open new table` in sidebar

pfarejowicz:
Fixed this issue with #120 should go live to the site in the next release

PR #120, opened 2018-04-12
Use createHref from history to make routes and fix navigation in new tabs

pfarejowicz:
In order to fix routing in new tabs we need to use the `history.createHref` function because we chose not to use the `Link` components from `react-router`. The issue was that the href set on the `<a/>` tag did not have the `#` symbol since it was not aware it was inside of a `HashRouter` (which the `history` props knows).

See this for more info: https://github.com/ReactTraining/history/blob/9ff690785f02d5c2554b860ff1a39a6527d18aa1/modules/createHashHistory.js#L176

Solves issue #113

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 
 ### Patch
 * Docs: Add live docs to TextField / TextArea (#116)
+* Internal: Fix navigation to allow opening in new tabs (#120)
 
 </details>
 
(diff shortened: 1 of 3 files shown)


christianvuerings:

Built with commit 55f3bf913b6189ed84e28488e8c7b00ed64aad05

https://deploy-preview-120--gestalt.netlify.com

christianvuerings:

Built with commit 6ce8261319ca125bfb4c918796d2f292eea1adc7

https://deploy-preview-120--gestalt.netlify.com

PR #125, opened 2018-04-16
Made some throttle changes and added debounce

btraut:
This PR makes the following changes:
1. I added comments to throttle.
2. I fixed a tiny bug in throttle that was causing longer than intended delays on throttled calls. If a function is throttled to 5 seconds, the first call should happen immediately and another call at 4 seconds should be executed at the 5 second mark. Our throttle function was executing the first one immediately and then starting a new threshold timer at the 4 second mark, so the second call wasn't being executed until 9 seconds (4 + 5).
3. I fixed a subtle bug in Masonry where `handleResize` could be executed after unmount because its timeout wasn't being cleared. I did this by adding a `clearTimeout` method to the function returned by `throttle`.
4. I created `debounce` and switched a couple Masonry methods to use it. This simplifies Masonry code and removes a couple of class properties.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,10 +7,16 @@
   </summary>
 
 ### Minor
+
 * Link: Remove duplicate CSS declaration (#118)
 
 ### Patch
+
 * Docs: Updated Masonry "comp" definition to be more descriptive
+* Fixed a subtle bug in throttle that would cause longer than intended delays
+* Fixed a timing bug where Masonry's handleResize could be called after unmount
+* Added a debounce method and moved over some Masonry methods to use it
+
   </details>
 
 </details>
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,9 +13,9 @@
 ### Patch
 
 * Docs: Updated Masonry "comp" definition to be more descriptive
-* Fixed a subtle bug in throttle that would cause longer than intended delays
-* Fixed a timing bug where Masonry's handleResize could be called after unmount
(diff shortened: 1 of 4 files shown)


christianvuerings:

Built with commit bee868a2769b26092eb164ef56f20223bad91a8e

https://deploy-preview-125--gestalt.netlify.com

packages/gestalt/src/Masonry/Masonry.js
```diff
@@ -321,6 +317,10 @@ export default class Masonry<T> extends React.Component<Props<T>, State> {
     });
   });
 
+  measureContainerAsync = debounce(() => {
```
bishwei: my only gripe here is that we now have `measureContainerAsync` and `measureContainer` and its not clear when to use one vs the other.  

one option is to make `measureContainer` always async and just pass it a callback or make it return a promise.   then we could update componentDidMount to be:
```
this.measureContainer().then(() => {
   //do other things and set some state
});
```
but that's a separate PR for sure 👍

CHANGELOG.md
```diff
@@ -7,10 +7,16 @@
   </summary>
 
 ### Minor
+
 * Link: Remove duplicate CSS declaration (#118)
 
 ### Patch
+
 * Docs: Updated Masonry "comp" definition to be more descriptive
+* Fixed a subtle bug in throttle that would cause longer than intended delays
```
christianvuerings: @btraut Could we add the PR at the end and always start the comment with a component?

btraut: Oh, sure. Will change.

PR #138, opened 2018-04-25
Flyout: make IE11 compatible

christianvuerings:
IE11 doesn't support `scrollX` and `scrollY` on window which we used when `positionRelativeToAnchor` was set to `false`.

We're using the same patch as React & React Router + made it pass flow:
https://github.com/facebook/react/blob/5d3b12bb3bd6a092cf00ede07b8255a8399c2e58/src/vendor/core/dom/getUnboundedScrollPosition.js#L28
https://github.com/ReactTraining/react-router/pull/607

screen shot 2018-04-25 at 1 54 24 pm

*Before*
screen shot 2018-04-25 at 2 40 15 pm

*After*
screen shot 2018-04-25 at 2 40 29 pm

### DIFF
--- a/packages/gestalt/src/FlyoutUtils/Contents.js
+++ b/packages/gestalt/src/FlyoutUtils/Contents.js
@@ -376,20 +376,24 @@ export default class Contents extends React.Component<Props, State> {
       width,
     } = props;
 
+    // Scroll not needed for relative elements
+    // We can't use window.scrollX / window.scrollY since it's not supported by IE11
+    const scrollX = positionRelativeToAnchor
+      ? 0
+      : window.pageXOffset ||
+        (document.documentElement && document.documentElement.scrollLeft) ||
+        0;
+    const scrollY = positionRelativeToAnchor
+      ? 0
+      : window.pageYOffset ||
+        (document.documentElement && document.documentElement.scrollTop) ||
+        0;
+
     const windowSize = {
       height: window.innerHeight,
       width: window.innerWidth,
-      // scroll not needed for relative elements
-      scrollY: positionRelativeToAnchor
-        ? 0
-        : window.pageYOffset ||
-          (document.documentElement && document.documentElement.scrollTop) ||
-          0,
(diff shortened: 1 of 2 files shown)


christianvuerings:

Built with commit 82350734cb7649d6150c17b53954993826857dd2

https://deploy-preview-138--gestalt.netlify.com

christianvuerings:

Built with commit f2cdc8ce061682af4a0f1e1f2ba7de21b226d5a3

https://deploy-preview-138--gestalt.netlify.com

packages/gestalt/src/FlyoutUtils/Contents.js
```diff
@@ -379,8 +379,17 @@ export default class Contents extends React.Component<Props, State> {
     const windowSize = {
       height: window.innerHeight,
       width: window.innerWidth,
-      scrollY: positionRelativeToAnchor ? 0 : window.scrollY, // scroll not needed for relative elements
-      scrollX: positionRelativeToAnchor ? 0 : window.scrollX,
+      // scroll not needed for relative elements
+      scrollY: positionRelativeToAnchor
+        ? 0
+        : window.pageYOffset ||
+          (document.documentElement && document.documentElement.scrollTop) ||
```
chrislloyd: Can you include a comment here (or extract as a named variable) as to why we have to do this so that it doesn't get accidentally removed in the future?

packages/gestalt/src/FlyoutUtils/Contents.js
```diff
@@ -379,8 +379,17 @@ export default class Contents extends React.Component<Props, State> {
...
+          0,
+      scrollX: positionRelativeToAnchor
+        ? 0
+        : window.pageXOffset ||
+          (document.documentElement && document.documentElement.scrollLeft) ||
```
chrislloyd: Can you include a comment here (or extract as a named variable) as to why we have to do this so that it doesn't get accidentally removed in the future?

PR #142, opened 2018-04-26
[Icon] Add dangerouslySetSvgPath prop to Icon to allow for experiment…

akleiner2:
…ation with non-standardized Icons. 

This would be very beneficial for folks who want to use SVGs that aren't existing icons in PDS. Instead of having rogue SVGs, they can just pass it in through the `dangerouslySetSvgPath` prop. 

This is noted as `dangerous`, since we want to discourage use of this for now -- it's meant strictly for icon experimentation.


pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

christianvuerings:

Built with commit 724014acdbd04acb5777a0dfd85695fc2ffc4785

https://deploy-preview-142--gestalt.netlify.com

chrislloyd:
A couple of things:

* What happens if people don't set either the icon name or the svg path?
* Could you update the docs with some examples of this?
* Tests?

One other thing to consider is that a bunch of other libraries do something like:

```jsx
import { PinIcon } from 'gestalt';
<PinIcon label="Foo" size={32} />;
```

if we did something like that, perhaps it would be a way of "layering" the raw icon (with a path passed to it) and Icons with set paths.

cc @christianvuerings

docs/src/Icon.doc.js
```diff
@@ -51,6 +51,10 @@ card(
         type: 'boolean',
         defaultValue: false,
       },
+      {
+        name: 'dangerouslySetSvgPath',
+        type: `{ __path: string }`,
```
christianvuerings: @akleiner2 could we add a description that the viewbox around the path should be 24x24?

CHANGELOG.md
```diff
@@ -15,10 +15,13 @@
 ## 0.67.0 (April 25, 2018)
 
 ### Minor
+
 * Flyout: make IE11 compatible (#138)
 * Icon: Add new GIF icon (#143)
+* Icon: Add new prop to Icon -- dangerouslySetSvgPath (#142)
```
christianvuerings: @akleiner2 We should move this up into `[Unreleased]`

PR #148, opened 2018-04-30
Button / SearchField / SegmentedControl / SelectList / Tabs / TextField: consistent sizing + improve Windows compatibility

christianvuerings:
Currently some of our components render their medium size at `41px` instead of `40px` which makes them look slightly off next to other components.

image

We also fixed an issue in Windows where the text inside of TextField and SearchField was cut off (we shouldn't set the line-height there)

Before:
screen shot 2018-04-30 at 3 31 15 pm
After:
screen shot 2018-04-30 at 3 31 28 pm

### DIFF
--- a/packages/gestalt/src/Layout.css
+++ b/packages/gestalt/src/Layout.css
@@ -83,6 +83,20 @@
   left: 0;
 }
 
+/* size */
+
+.small {
+  min-height: 36px;
+}
+
+.medium {
+  min-height: 40px;
+}
+
+.large {
+  min-height: 48px;
+}
+
 /* box model */
 
 .borderBox {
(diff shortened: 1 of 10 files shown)


christianvuerings:

Built with commit a3c0458c866ee85823172ea18532f6a71dbc65cf

https://deploy-preview-148--gestalt.netlify.com

packages/gestalt/src/Layout.css
```diff
@@ -83,6 +83,20 @@
   left: 0;
 }
 
+/* size */
+
+.small {
+  min-height: 36px;
```
christianvuerings: Reason for using `min-height` opposed to `height` is multiline:

**min-height**
image

**height**
image

PR #150, opened 2018-05-03
Core Video Component for Gestalt

pfarejowicz:
In terms of media content on web, videos are gaining traction to overtake images as the media of choice, see https://sproutsocial.com/insights/facebook-stats-for-marketers/#videostats

Gestalt has its `Image` component but no support for videos. Our video support in the codebase is also scattered, desktop grid is non-existent, desktop closeup is still largely written as legacy css and elements, and on mobile web we have `rerender = () => this.forceUpdate();` calls which are not advised.

This change is aimed at adding a generics `Video` component that is robust in handling video playback/events, fluid in its UI, and easy to use and integrate into Pinterest in the future.

A point of discussion is how to handle multiple media sources. HTML5 video has support for fallback media sources if a browser is unable to play one (https://developer.mozilla.org/en-US/docs/Web/HTML/Element/video). I chose to only limit to one source for now, but we have a few options here:

1. Treat each sources as a prop (or potentially a list):
```jsx
import { Video } from 'gestalt';

<Video mp4={mp4Url} m3u8={m3u8Url} />

then internally...
<video>
    {m3u8 && <source src={m3u8} type="video/m3u8" />}
    {mp4 && <source src={mp4} type="video/mp4" />}
</video>
```
2. Refactor to accept children and filter for valid elements:
```jsx
import { Video } from 'gestalt';

<Video>
    <source src={m3u8} type="video/m3u8" />
    <source src={mp4} type="video/mp4" />
</Video>
```

Thoughts on addressing the source issue?

Some screenshots, in addition check out the deploy preview
screen shot 2018-05-03 at 4 19 52 pm
screen shot 2018-05-03 at 4 19 44 pm


CHANGELOG.md
```diff
@@ -6,6 +6,9 @@
     Click to see more.
   </summary>
 
+### Major
+* Video: Add new Video component to Gestalt (#150)
```
christianvuerings: @pfarejowicz Even though we're adding a new component, it should be `Minor`, since we're still on `0.x.x` releases

packages/gestalt/src/Icon/icons/index.js
```diff
@@ -44,8 +44,11 @@ import heartBroken from './heart-broken.svg';
 import link from './link.svg';
 import location from './location.svg';
 import lock from './lock.svg';
+import maximize from './maximize.svg';
```
christianvuerings: @pfarejowicz did you confirm that these icons come from PDS spec?

CHANGELOG.md
pfarejowicz: Too excited about adding videos haha

packages/gestalt/src/Icon/icons/index.js
pfarejowicz: Yes, I got all the new icons from the experimental folder that @jaymarsh maintains

packages/gestalt/src/Video/Video.js
```diff
@@ -0,0 +1,273 @@
...
+          preload={preload}
+          ref={this.setVideoRef}
+          className={styles.video}
+          onDurationChange={this.handleDurationChange}
+          onPlaying={this.handlePlay}
+          onPause={this.handlePause}
+          onTimeUpdate={this.handleTimeUpdate}
+          onVolumeChange={this.handleVolumeChange}
+        >
+          {src && <source src={src} type="video/mp4" />}
```
christianvuerings: @pfarejowicz should we be allowing any other types? E.g. `video/webm`

packages/gestalt/src/Video/Video.js
```diff
@@ -0,0 +1,273 @@
...
+      // If the video metadata is missing, default to a standard 16:9 ratio
+      `${9 / 16 * 100}%`;
+
+    return (
+      <div
+        ref={this.setPlayerRef}
+        className={styles.player}
+        style={{ paddingBottom, height: isFullscreen ? '100%' : 0 }}
+      >
+        {/* eslint-disable jsx-a11y/media-has-caption */}
```
christianvuerings: @pfarejowicz Could we build in support for this instead of disabling the a11y linter?
https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/media-has-caption.md

packages/gestalt/src/Video/VideoControls.js
```diff
@@ -0,0 +1,116 @@
...
+    toggleMute,
+    togglePlay,
+  } = props;
+  return (
+    <Box
+      position="absolute"
+      bottom
+      left
+      right
+      width="100%"
```
christianvuerings: @pfarejowicz change this to `column={12}` (1 less inline style)

packages/gestalt/src/Video/VideoControls.js
```diff
@@ -0,0 +1,116 @@
...
+  fullscreen: boolean,
+  muted: boolean,
+  paused: boolean,
+  seek: (time: number) => void,
+  toggleFullscreen: () => void,
+  toggleMute: () => void,
+  togglePlay: () => void,
+|};
+
+const timeToString = (time?: number) => {
```
christianvuerings: @pfarejowicz could we add some tests for this?

packages/gestalt/src/Video/VideoControls.js
```diff
@@ -0,0 +1,116 @@
...
+      right
+      width="100%"
+      padding={2}
+      display="flex"
+      alignItems="center"
+    >
+      <Box padding={2}>
+        <Touchable onTouch={togglePlay} fullWidth={false}>
+          <Icon
+            accessibilityLabel=""
```
christianvuerings: @pfarejowicz for all of these, we should be providing good `accessibilityLabel`s passed in as props. Right now it just reads `button`, `button`, `button`

packages/gestalt/src/Video/VideoPlayhead.js
```diff
@@ -0,0 +1,109 @@
...
+    return (
+      <Box
+        position="relative"
+        display="flex"
+        flex="grow"
+        alignItems="center"
+        height={16}
+      >
+        <div
+          role="presentation"
```
christianvuerings: @pfarejowicz this piece is completely inaccessible - could we use the `progressbar` role or the `<progress>` element?

Also supplying the right aria labels: e.g. `aria-valuenow="20" aria-valuemin="0" aria-valuemax="100"`

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_progressbar_role

packages/gestalt/src/Video/Video.js
```diff
@@ -0,0 +1,273 @@
+// @flow
+
+import * as React from 'react';
+import * as fullscreen from './fullscreen';
```
christianvuerings: @pfarejowicz since we're only importing `fullscreen` and `VideoControls` in this file, why did we extract it out? Would love to hear what @chrislloyd thinks

packages/gestalt/src/Video/Video.js
pfarejowicz: Spoke to @chrislloyd about this (i had some ideas up above in my PR description on how we can handle this) and we're thinking of making `src` and object prop that accepts the different video types as keys with the values as the video url for that type

jackhsu978:
One thing I noticed is that the full screen icon has arrows pointing at different directions than the one we use today, which looks like this:
screenshot 2018-05-03 17 26 31

Is that intentional? Personally I like the existing icon better.

packages/gestalt/src/Video/VideoControls.js
```diff
@@ -0,0 +1,116 @@
...
+            icon={fullscreen ? 'minimize' : 'maximize'}
+            size={20}
+          />
+        </Touchable>
+      </Box>
+    </Box>
+  );
+}
+
+VideoControls.propTypes = {
```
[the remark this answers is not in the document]
pfarejowicz: @chrislloyd `VideoControls` is a stateless functional, this one has to be outside, but I can move the others

packages/gestalt/src/Icon/icons/index.js
pfarejowicz: @jackhsu978 good catch with the maximize icon pointing in the different direction, I didn't even notice. I took that maximize icon straight from the PDS icons folder so my guess is that the flip was intentional, we can talk with @jaymarsh as to why the arrows point in the opposite direction now

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  Bundle size increase
  </td>
    </tr>
  </tbody>
</table>



gestalt.js | :small_red_triangle: 8.3% | :small_red_triangle: 6.9% | 258.66 KB | 280.17 KB | 58.71 KB | 62.73 KB
gestalt.css | :small_red_triangle: 0.1% | :small_red_triangle: 0.0% | 52.63 KB | 52.66 KB | 8.99 KB | 8.99 KB
gestalt.es.js | :small_red_triangle: 8.3% | :small_red_triangle: 6.9% | 256.21 KB | 277.53 KB | 58.32 KB | 62.36 KB
  
<p align="right">
</p>

docs/src/Video.doc.js
```diff
@@ -0,0 +1,345 @@
...
+        name: 'playing',
+        type: 'boolean',
+        description: 'Specifies whether the video should play or not',
+        defaultValue: false,
+      },
+      {
+        name: 'playsInline',
+        type: 'boolean',
+        description:
+          'Indicates that the video is to be played "inline" where supported',
```
chrislloyd: I'm not sure sure what this means from this description - what actually changes with it being inline?

PR #154, opened 2018-05-06
[Docs] Design changes to nav and layout

chrislloyd:


christianvuerings:

Built with commit 74be5579f128f746ca451f57574661c33ee89546

https://deploy-preview-154--gestalt.netlify.com

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

docs/src/components/Header.js
```diff
@@ -0,0 +1,60 @@
...
+                  />
+                </Box>
+                <Box paddingX={1}>Gestalt</Box>
+              </Box>
+            </Box>
+          </Link>
+        </Text>
+      </Box>
+
+      <Box flex="grow" />
```
christianvuerings: @chrislloyd instead of adding an empty `div` / `Box`, can we add `justifyContent="between"` on the parent `<Box /`?

christianvuerings:
😍 Love the design, left 1 small comment

chrislloyd:
@christianvuerings do you mind re-reviewing? I added some more commits last night that tidies it up a bit.

docs/src/components/Example.js
```diff
@@ -10,80 +10,58 @@ type Props = {|
   description: string,
   name: string,
   scope: Object,
-  stacked: boolean,
-  stacked?: boolean,
```
christianvuerings: @chrislloyd this is my least favorite change - previously we were able to see the immediate changes with all examples, now (since they're always stacked) this doesn't work with certain examples: `FlyoutExample` / `TabExample` / `Tooltip`, we have to constantly scroll up / down to see the changes.

docs/src/components/App.js
```diff
@@ -14,15 +15,20 @@ const NavigationWithRouter = withRouter(Navigation);
...
+      <Box smDisplay="flex" direction="row">
+        <Column span={12} smSpan={2}>
           <NavigationWithRouter />
-        </Box>
-      </Column>
-      <Column span={12} mdSpan={10}>
-        <Box padding={4}>{children}</Box>
-      </Column>
+        </Column>
+        <Divider />
```
christianvuerings: @chrislloyd nice trick - actually didn't know we could create a vertical divider like this.

docs/src/components/App.js
```diff
@@ -14,15 +15,20 @@ const NavigationWithRouter = withRouter(Navigation);
...
+        <Column span={12} smSpan={8}>
```
christianvuerings: @chrislloyd How'd you feel if we changed this to `smSpan={10} mdSpan={8}`? The docs are looking pretty weird between the `sm` and `md` breakpoints

Before:
image

After:
image

docs/src/components/NavLink.js
```diff
@@ -0,0 +1,39 @@
...
+  const handleClick = ({ event }) => {
+    if (event.defaultPrevented) return;
+    if (isModifiedEvent(event) || !isLeftClickEvent(event)) return;
+    event.preventDefault();
+    history.push(to);
+  };
+  return (
+    <Route path={to} location={history.location}>
+      {({ match }) => (
+        <Text bold={!!match}>
```
christianvuerings: @chrislloyd love this change!

docs/src/components/Navigation.js
```diff
@@ -13,72 +14,50 @@ const isModifiedEvent = event =>
...
-          onChange={({ value }) => history.push(value)}
-          options={options}
-          value={(m && m[1]) || '#'}
-        />
-      </Box>
-      <Box display="none" mdDisplay="flex" direction="column" flex="grow">
+      <Box
+        display="flex"
+        smDisplay="none"
+        overflow="scrollX"
```
christianvuerings: @chrislloyd Nice! Been thinking it would be great to add something to `SegmentedControl` as well, or should we always let the containing element take care of it?

image

docs/src/components/Navigation.js
```diff
@@ -13,72 +14,50 @@ const isModifiedEvent = event =>
...
+        direction="column"
+        flex="grow"
+        role="list"
+        padding={4}
+        mdPadding={6}
+        lgPadding={8}
+      >
+        <Box marginStart={-2} marginEnd={-2}>
+          {components.map((component, i) => (
+            <Box role="listitem" key={i} flex="none">
```
christianvuerings: @chrislloyd do we need the `flex="none"` even though the container doesn't have `flex` set?

docs/src/components/App.js
chrislloyd: Neither did I until I tried it! I started off adding a vertical option to divider but then I realized that I hadn't actually rebuilt the library. Works well in the mobile case too.

docs/src/components/Navigation.js
chrislloyd: I think we should build out the `Tabs` component. I see SegmentedControl being more of a mis-used form element whereas Tabs are for navigation. I just didn't use them here because they look a bit ridiculous and are _very_ specific to a few pages.

christianvuerings: @chrislloyd makes sense.

Seems like there was some miscommunication, tried to allude to the `overflow=scrollX` specifically. I was suggesting to add something similar to `SegmentedControl` so it works better responsively. Not to use it for the main navigation.

pfarejowicz:
This is amazing!! Only thing I noticed was the main content area feels like it could stretch more to the right edge, right now its not full width and leaves blank space on the right
screen shot 2018-05-07 at 11 02 01 am

PR #158, opened 2018-05-07
Jest 22.4.3

chrislloyd:
Jest 22 includes some really interesting change around multi-project running (eslint etc.). This change updates Jest and gets our tests using the multi-project runner.

~~It also upgrades `react-scripts` to the latest version (which I don't think will be a long term change) but latest fixes an issue that conflicted with the parent Jest version. The latest version of `react-scripts` also requires a strict eslint version and seeing as it's a minor it seems fine to downgrade.~~

We don't need up change `react-scripts` anymore because I peg the version we're on seeing as we don't run tests via. `create-react-app` anyway.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,6 +5,9 @@
     "packages/*",
     "test"
   ],
+  "resolutions": {
+    "jest": "^22.4.3"
+  },
   "devDependencies": {
     "babel-core": "^6.26.0",
     "babel-eslint": "^8.2.1",
@@ -20,7 +23,6 @@
     "concurrently": "^3.5.1",
     "danger": "^3.1.7",
     "enzyme": "^3.3.0",
-    "enzyme-adapter-react-15": "^1.0.5",
     "enzyme-adapter-react-16": "^1.0.5",
     "eslint": "^4.17.0",
     "eslint-config-airbnb": "^16.1.0",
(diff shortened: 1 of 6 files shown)


christianvuerings:

Built with commit c3de9513296a980b9d650c09726e909bd4dc6f56

https://deploy-preview-158--gestalt.netlify.com

christianvuerings:

Built with commit b200fc316579e65e8d577de2c12d890268b71f95

https://deploy-preview-158--gestalt.netlify.com

package.json
```diff
@@ -21,7 +21,7 @@
     "enzyme-adapter-react-16": "^1.0.5",
     "enzyme": "^3.3.0",
     "enzyme-adapter-react-15": "^1.0.5",
-    "eslint": "^4.17.0",
+    "eslint": "4.15.0",
```
bishwei: reason for this?

chrislloyd: `react-scripts@latest` gets really picky about the versions of `eslint` & `jest` (hence also why jest is only updated to `22.1` instead of `22.4`).

chrislloyd: It's not awesome, but seeing as it's a minor version I think it might be OK. Something to keep on the radar though - `react-scripts` is a bit of a leech package at the moment.

pinterest-ios:
gestalt.js | :small_red_triangle_down: 0.0% | :small_red_triangle_down: 0.0% | 260.21 KB | 260.21 KB | 58.34 KB | 58.33 KB
gestalt.es.js | :small_red_triangle_down: 0.0% | :small_red_triangle_down: 0.0% | 257.67 KB | 257.67 KB | 57.97 KB | 57.96 KB
  
<p align="right">
</p>

PR #171, opened 2018-05-10
Adds aspectRatio to Video and hides fullscreen button for unsupported browsers

pfarejowicz:
In the initial implementation, `aspectRatio` was hardcoded to 16:9 in the render because that is the industry standard widescreen size. Instead of doing this, we should open this option up to developers to all them to pass in a custom ratio for the placeholder, and only then default to the 16:9 number. This aspect ratio is really only used as a placeholder, similar to `Image` to hold the height while the video loads and we fully know the real width and height.

Also, the fullscreen toggle button should not be exposed if the browser document cannot support the fullscreen API. Instead of doing that check in the callback, we can move the check to the render of the button and hide if for unsupported browsers. One small not here, is this bad for SSR @chrislloyd since document does not exist on the server, thoughts?

### DIFF
--- a/docs/src/Video.doc.js
+++ b/docs/src/Video.doc.js
@@ -58,6 +58,13 @@ card(
           'Accessibility label for the unmute button if controls are shown',
       },
       {
+        name: 'aspectRatio',
+        type: 'number',
+        description: `Proportional relationship between width and height of the video, calculated as width / height.
+           Used to size the placeholder until the video is loaded. Default is equivalent to 16:9`,
+        defaultValue: 16 / 9,
+      },
+      {
         name: 'captions',
         type: 'string',
         description: 'The URL of the captions track for the video (.vtt file)',
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,8 +10,8 @@
 
 * Sticky: Expand threshold options to take string values (#166)
 * Avatar: Fall back to default letter if image does not load (#156)
-
 * Video: Add new Video component to Gestalt (#150)
+* Video: Add `aspectRatio` prop to Video and hide fullscreen on unsupported browsers (#171)
 
 ### Patch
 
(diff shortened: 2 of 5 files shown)


christianvuerings:

Built with commit 512ab23430391d8ff91561d1c3d5e11e92a8ce76

https://deploy-preview-171--gestalt.netlify.com

christianvuerings:

Built with commit 955cb11392068dbafefc54d8664aef6cae03ce79

https://deploy-preview-171--gestalt.netlify.com

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  Bundle size increase
  </td>
    </tr>
  </tbody>
</table>



gestalt.js | :small_red_triangle: 0.1% | :small_red_triangle: 0.1% | 282.04 KB | 282.19 KB | 63.06 KB | 63.1 KB
gestalt.es.js | :small_red_triangle: 0.1% | :small_red_triangle: 0.1% | 279.39 KB | 279.55 KB | 62.68 KB | 62.72 KB
  
<p align="right">
</p>

docs/src/Video.doc.js
```diff
@@ -57,6 +57,13 @@ card(
         description:
           'Accessibility label for the unmute button if controls are shown',
       },
+      {
+        name: 'aspectRatio',
```
chrislloyd: It might be nice to use the same conventions as Image here - `naturalWidth` & `naturalHeight` instead of asking people to do some math. `aspectRatio` on it's own implies a horizontal ratio.

pfarejowicz: I chose to go the route of `LetterBox` which takes in an `aspectRatio`. Looking at `Image` it takes in the width and height separately but still only uses them together in one spot to actually define the aspect ratio. Its confusing because the formula for aspect ratio is width / height (see https://www.extron.com/company/article.aspx?id=aspectratio) but for the padding trick to work you do height / width (hence the 1 / ratio in my code). I think this math is ok for the client to do as long as the docs describe the correct division which i included. I also like how `aspectRatio` forces the existence into one prop, either defined or defaulted, whereas splitting it into separate width and height would require us to validate that both are present (if one is supplied) otherwise the math gets hairy

chrislloyd: Ok!

PR #174, opened 2018-05-11
Flow-typed support in Gestalt for third party packages

pfarejowicz:
This change adds support for [Flow-typed](https://github.com/flowtype/flow-typed) to declare Flow definitions for our third party dependencies. I manually ran the install command on a few of our most used packages, but this is open to adding more in the future. This allows us to pull Flow definitions from the open-source community and keep up to date with upgrades.

I disabled eslint and prettier on the new `flow-typed` folder because those are auto-generated so we should not bother touching them.

### DIFF
--- a/packages/gestalt/src/Box/Box.js
+++ b/packages/gestalt/src/Box/Box.js
@@ -744,32 +744,32 @@ const PaddingPropType = PropTypes.oneOf([
 
 Box.propTypes = {
   children: PropTypes.node,
-  dangerouslySetInlineStyle: PropTypes.shape({
+  dangerouslySetInlineStyle: PropTypes.exact({
     __style: PropTypes.object,
   }),
 
-  xs: PropTypes.shape({
+  xs: PropTypes.exact({
     display: PropTypes.oneOfType([
       PropTypes.bool,
       PropTypes.oneOf(['flex', 'flexColumn', 'inlineBlock']),
     ]),
     column: PropTypes.number,
   }),
-  sm: PropTypes.shape({
+  sm: PropTypes.exact({
     display: PropTypes.oneOfType([
       PropTypes.bool,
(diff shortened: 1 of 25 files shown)


christianvuerings:

Built with commit 9afa5a84fd5ed31dd0cb93fe658585ec8d39cbaa

https://deploy-preview-174--gestalt.netlify.com

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

packages/gestalt/src/Box/Box.js
```diff
@@ -744,32 +744,32 @@ const PaddingPropType = PropTypes.oneOf([
 
 Box.propTypes = {
   children: PropTypes.node,
-  dangerouslySetInlineStyle: PropTypes.exact({
+  dangerouslySetInlineStyle: PropTypes.shape({
```
chrislloyd: Hey what was your reasoning for this change? Seems to make the interface more loose.

pfarejowicz: @chrislloyd Flow isnt happy with that one (see https://github.com/pfarejowicz/gestalt/blob/8967902e3e96725c1c3113a8caf58acf7d566978/flow-typed/npm/prop-types_v15.x.x.js ). I see `PropTypes.exact` in the source code, but Facebook does not list it on the docs site (not sure why) (see https://github.com/facebook/prop-types ) so I think there is some miscommunication and a bug in flow-typed

chrislloyd: I don't think changing our code here is the right solution to get flow safety here. Are there other approaches?

PR #181, opened 2018-05-15
Add eslint-plugin-jest with recommended settings

pfarejowicz:
Adds the [eslint-plugin-jest](https://github.com/jest-community/eslint-plugin-jest) to gestalt and turns on the recommended settings. Puts the jest globals into the environment and fixes any errors from these settings.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -10,7 +10,7 @@
   ],
   "env": {
     "browser": true,
-    "es6": true,
+    "es6": true
   },
   "parser": "babel-eslint",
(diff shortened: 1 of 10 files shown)


.eslintrc
```diff
@@ -9,10 +10,11 @@
   ],
   "env": {
     "browser": true,
-    "es6": true
+    "es6": true,
+    "jest/globals": true,
```
chrislloyd: This will think that jest globals are valid in all parts of the codebase - can we just configure this for test files? https://eslint.org/docs/user-guide/configuring#configuration-based-on-glob-patterns

PR #182, opened 2018-05-15
Internal: Convert ghostjs to puppeteer

christianvuerings:


pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



gestalt.js | :small_red_triangle_down: 0.0% | :small_red_triangle_down: 0.0% | 282.54 KB | 282.47 KB | 63.21 KB | 63.19 KB
gestalt.es.js | :small_red_triangle_down: 0.0% | :small_red_triangle_down: 0.0% | 279.9 KB | 279.83 KB | 62.83 KB | 62.81 KB
  
<p align="right">
</p>

chrislloyd:
You should check out #158!

christianvuerings:

Built with commit 4a449b99331c03f1ba24e7955eb66df9fdcea34a

https://deploy-preview-182--gestalt.netlify.com

Dockerfile
```diff
@@ -9,20 +9,23 @@ ENV DISPLAY :99
...
 
+RUN  apt-get update \
+     && apt-get install -yq libgconf-2-4 \
+     && apt-get install -y wget --no-install-recommends \
+     && wget -q -O - https://dl-ssl.google.com/linux/linux_signing_key.pub | apt-key add - \
+     && sh -c 'echo "deb [arch=amd64] http://dl.google.com/linux/chrome/deb/ stable main" >> /etc/apt/sources.list.d/google.list' \
+     && apt-get update \
+     && apt-get install -y google-chrome-unstable --no-install-recommends \
+     && rm -rf /var/lib/apt/lists/* \
+     && npm i puppeteer@1.4.0
```
chrislloyd: This is a dependency of the tests - any reason why you're doing this here?

Dockerfile
```diff
@@ -9,20 +9,23 @@ ENV DISPLAY :99
...
-    chmod a+x /etc/init.d/xvfb /usr/bin/xvfb-daemon-run
-
 COPY yarn.lock package.json ./
 
 RUN mkdir -p docs test packages/gestalt
 COPY packages/gestalt/package.json ./packages/gestalt/
 COPY docs/package.json ./docs/
 COPY test/package.json ./test/
 
+RUN  apt-get update \
```
chrislloyd: I think you want this `RUN` command to be high so it's not invalidated by changes to the `package.json` etc.

scripts/jestIntegration.config.json
```diff
@@ -0,0 +1,6 @@
+{
```
chrislloyd: This should probably be baked into the `package.json` - you can always run jest with only a particular project if you want to narrow the scope. `jest --project integration`

scripts/jestIntegration.config.json
```diff
@@ -0,0 +1,6 @@
...
+  "name": "integration",
```
chrislloyd: It should probably be named `puppeteer` to match `node` & `jsdom` (name the tool rather than the type of test).

scripts/jestPuppeteer.config.js
```diff
@@ -0,0 +1,14 @@
+module.exports = {
+  launch: {
```
chrislloyd: This config also takes a `server` option that you can use to have Jest run the dev server. That might mean you can remove `run_integration_tests.sh`

test/package.json
```diff
@@ -12,26 +13,23 @@
...
     "start": "yarn start:webpack-server | yarn start:node-server",
-    "start:webpack-server":
-      "yarn webpack-dev-server --hot --inline --progress --colors",
-    "start:node-server":
-      "BABEL_DISABLE_CACHE=1 NODE_ENV=testing yarn babel-node server.js"
-  }
+    "start:webpack-server": "yarn webpack-dev-server --hot --inline --progress --colors",
+    "start:node-server": "BABEL_DISABLE_CACHE=1 NODE_ENV=testing yarn babel-node server.js"
+  },
+  "devDependencies": {}
```
chrislloyd: necessary?

chrislloyd:
The other reason - we lost parallelism with this. If we can at least consolidate around 1 command to run all the tests then we can add parallelism to that one place and chunk both unit and integration tests across machines.

test/package.json
christianvuerings: @chrislloyd it was not, removed

scripts/jestIntegration.config.json
christianvuerings: @chrislloyd indeed, that should be the case if we make puppeteer a jest project. It seems like we're running into a couple of issues though if we're going for that approach:

image

```
{
  "displayName": "puppeteer",
  "preset": "jest-puppeteer",
  "testMatch": ["<rootDir>**/__integration__/**/*.integration.js"]
}
```

christianvuerings:
@chrislloyd 

> use the @Heredoc at the top of the test?

Could you explain this a bit more? 

> we lost parallelism with this.

Indeed, I removed it within the integration tests since it no longer seemed necessary. Before, the combined integration tests were taking 5+ mins, now they only take 1m 20s

https://buildkite.com/pinterest/gestalt/builds/4241
https://buildkite.com/pinterest/gestalt/builds/4510

scripts/jestPuppeteer.config.js
chrislloyd: `./run_integration_tests` has a timeout for the server to start - perhaps you could just use that script without the bit which actually runs the tests?

chrislloyd:
@christianvuerings https://facebook.github.io/jest/docs/en/configuration.html#testenvironment-string

1:20 is still too long! As we add more we'll still have to parallelize the tests - it would be great if we could get that parallelism for free also with regular Jest unit tests.

packages/gestalt/src/Masonry/__integration__/.eslintrc.json
```diff
@@ -0,0 +1,11 @@
+{
+  "globals": {
```
pfarejowicz: @christianvuerings i think this can be avoided, its baked into eslint now, you just need to add `.integration.js` to the overrides for jest env

chrislloyd: Good catch - I'll do this in a follow up PR. I'd like to get this merged ASAP.

pfarejowicz: @chrislloyd @christianvuerings check out #193 which updates based on the instructions [here](https://github.com/smooth-code/jest-puppeteer#configure-eslint)

PR #192, opened 2018-05-17
Turn on all recommended react linters

pfarejowicz:
We already list [eslint-plugin-react](https://github.com/yannickcr/eslint-plugin-react) as a dependency for our package and subscribe to most of the rules. This change turns on the recommended config from that plugin and fixes the one error that arises from doing so. Our settings already turn off the 3 rules we do not follow so this change causes no other errors.


christianvuerings:

Built with commit f46410bb9c9d62939a124fc12be4afd7976ab944

https://deploy-preview-192--gestalt.netlify.com

PR #196, opened 2018-05-18
Rewrite Column docs to remove scope from Example

pfarejowicz:
`Column.doc.js` is the only doc page currently passing in a custom `scope` to the LiveProvider because we defined a custom React component. In my opinion this is bad because it exposes these custom components to the user in the examples without any context so it can be confusing what this component is or does (see below)
image

We should not allow scope to be passed and limit all examples in Gestalt to be purely written with Gestalt components.


christianvuerings:

Built with commit d8ff402066404af79744b2d3d691bcca7e57b57f

https://deploy-preview-196--gestalt.netlify.com

christianvuerings:

Built with commit 6af01312ccc3692f64c0d1bc2ddca7088d47c2d0

https://deploy-preview-196--gestalt.netlify.com

christianvuerings:
❤️ on all these cleanups!

PR #199, opened 2018-05-18
Update Modal docs to kill StateRecorder

pfarejowicz:
`Modal.doc.js` was the only page still using the old `StateRecorder` component. This change refactors all the examples to use the new live docs which lets us kill the `StateRecorder` component once and for all.


christianvuerings:

Built with commit ca7d81dbcea740a98b4cd9951a0ded2b0eabcf66

https://deploy-preview-199--gestalt.netlify.com

christianvuerings:
@pfarejowicz It looks like some of content for our code formatting shows above our modals:

image

pfarejowicz:
@christianvuerings good call out, the fix is to add
```
color: rgba(197, 200, 198, 0.7);
opacity: 1 !important;
```
to `AtomDark.css` which fixes the issue without changing anything else visually. The `react-live` editor was applying a `0.7` opacity which messes with the Modal overlay. Reseting it to 1 and switching from the opacity CSS attribute to the `rgba()` color attribute fixes the issue without changing the look and feel.

PR #203, opened 2018-05-22
Pass Video events up through to callbacks

pfarejowicz:
When I originally wrote `Video` I did not think the `event` would be needed in the caller but it turns out that it is a good idea to pass these up. One reason these are useful is if the `Video` is wrapped in a `Touchable`, we need to pass all the button click events up so the caller can decide if they need to be stopped or prevented. One caveat is the fullscreen event - since its the actual change event that gets passed up and not the button click, we need to manually stop the propagation on that event.


christianvuerings:

Built with commit b5d88b4ee6d99ba87de19eca4eddaef484400891

https://deploy-preview-203--gestalt.netlify.com

christianvuerings:

Built with commit 69c2d7b685103687e0a9bff8d3e29849e1e59c2d

https://deploy-preview-203--gestalt.netlify.com

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  Bundle size increase
  </td>
    </tr>
  </tbody>
</table>



gestalt.js | :small_red_triangle: 0.8% | :small_red_triangle: 0.2% | 261.9 KB | 263.94 KB | 58.75 KB | 58.89 KB
gestalt.es.js | :small_red_triangle: 0.8% | :small_red_triangle: 0.2% | 259.35 KB | 261.39 KB | 58.39 KB | 58.53 KB
  
<p align="right">
</p>

pfarejowicz:
I will follow up with a PR dedicated to testing as we need to create a way for jsdom only tests to be run

PR #212, opened 2018-05-24
Fix react router dependencies

pfarejowicz:
`App` does not need `withRouter` anymore because `Navigation` does not use it. Likewise, `Link` was the only package using `react-router` instead of `react-router-dom` so we can switch those out since the import is the same.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
   </summary>
 
 ### Minor
+* Internal: Fix react router dependencies (#212)
 
 ### Patch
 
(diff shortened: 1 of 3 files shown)


christianvuerings:

Built with commit cd4ff9554c47ce6b622fe19225f2c3c69e9a23c4

https://deploy-preview-212--gestalt.netlify.com

christianvuerings:

Built with commit 5f0ba35699a179581d7dd8983ef182528d1ac73a

https://deploy-preview-212--gestalt.netlify.com

PR #215, opened 2018-05-25
[Flow] Fix Flow errors in docs directory (part 2)

pfarejowicz:
This change fixes all the remaining Flow relate errors with gestalt:

1. Finishes cleaning up the `docs` Flow errors
2. Removes all the Flow ignore folders as they are good to go now
3. Removes `esproposal.class_instance_fields=enable` and `esproposal.class_static_fields=enable` from the config as Flow already defaults to these [values](https://flow.org/en/docs/config/options/#toc-esproposal-class-instance-fields-enable-ignore-warn)
4. Fixes our `NODE_ENV` check in `PropTable` to the correct string

### DIFF
--- a/docs/src/components/PropTable.js
+++ b/docs/src/components/PropTable.js
@@ -1,6 +1,5 @@
 // @flow
-import type { Node } from 'react';
-import React from 'react';
+import * as React from 'react';
 import { Box, Text, Icon } from 'gestalt';
 
 type Props = {|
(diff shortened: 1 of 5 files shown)


christianvuerings:

Built with commit faa3e45f6aa7ac37d5767d8ed6bd20b3fcee1c20

https://deploy-preview-215--gestalt.netlify.com

christianvuerings:

Built with commit 3cc5b7308947e97508b9140aedacacf68e7cb95a

https://deploy-preview-215--gestalt.netlify.com

docs/src/components/PropTable.js
```diff
@@ -72,7 +71,7 @@ export default function PropTable({ props: properties, Component }: Props) {
     if (missingProps.length > 0) {
       // eslint-disable-next-line no-console
       console.warn(
-        `${displayName} is missing ${
+        `${displayName || ''} is missing ${
```
christianvuerings: @pfarejowicz should we set the default in the destructuring higher up?

`const { displayName = '', propTypes } = Component;`

pfarejowicz: I tried this first too but `displayName` is of type `?string` inside of `React.ComponentType` which means doing `displayName = '',` will catch the `undefined` case but not the `null` case. Flow would still error if we did that so we need to `|| ''` to catch both cases

PR #225, opened 2018-05-30
Fix Video autoplay on SSR on initial render

pfarejowicz:
The `playing` prop covers flipping the video back and forth between being played and being paused, but it breaks on SSR because the browser expects the user to interact with the DOM before the video can play. This means that if for example a user lands on Closeup of a video Pin and `playing` was defaulted to `true`, the video would not actually play until the user starts to click around and clicks play manually (which can lead to inconsistent state issues). By making the `autoPlay` attribute on `video` mirror the `playing` prop on `Video` we can achieve the effect we need of **all** video closeups playing from the get-go regardless of whether they are client navigated from a feed or via a URL.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@
 * TextArea: Makes TextArea React Async compatible (#222)
 * TextField: Makes TextField React Async compatible (#223)
 * ScrollContainer: Makes ScrollContainer React Async compatible (#224)
+* Video: Fix Video playback on SSR if playing is true on first mount (#225)
 
 ### Patch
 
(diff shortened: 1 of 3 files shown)


christianvuerings:

Built with commit 923e16840a48d10621b45bde753b8488cd77bcf8

https://deploy-preview-225--gestalt.netlify.com

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  Bundle size increase
  </td>
    </tr>
  </tbody>
</table>



gestalt.js | :small_red_triangle: 0.0% | :small_red_triangle: 0.0% | 264.62 KB | 264.65 KB | 59.04 KB | 59.04 KB
gestalt.es.js | :small_red_triangle: 0.0% | :small_red_triangle: 0.0% | 262.06 KB | 262.09 KB | 58.67 KB | 58.67 KB
  
<p align="right">
</p>

pfarejowicz:
@chrislloyd we can revisit that in the future, I thought about it but then got stuck thinking about what to do if a user passes `playing=false` and `autoPlay=true` because those seem to conflict and put state out of sync

PR #226, opened 2018-05-30
Cleanup Link component Flow + React prop typing

pfarejowicz:
There is a discrepancy between the Flow typing, the React proptyping, and the gestalt docs for this component. In my opinion, these should both be optional (similar to `Text`) so let's make it all consistent. This also moves the default props to the correct `static defaultProps` and adds and event target to the event handler.

### DIFF
--- a/packages/gestalt/src/Link/Link.js
+++ b/packages/gestalt/src/Link/Link.js
@@ -6,9 +6,9 @@ import styles from './Link.css';
 
 type Props = {|
   children?: React.Node,
-  href: string,
+  href?: string,
   inline?: boolean,
-  onClick?: ({ event: SyntheticMouseEvent<> }) => void,
+  onClick?: ({ event: SyntheticMouseEvent<HTMLAnchorElement> }) => void,
   target?: null | 'self' | 'blank',
 |};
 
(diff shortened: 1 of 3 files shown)


christianvuerings:

Built with commit 1586edbc9f7886157bc0df2f0f9a59c155bcc0a5

https://deploy-preview-226--gestalt.netlify.com

pinterest-ios:
gestalt.js | :small_red_triangle_down: 0.0% | :small_red_triangle_down: 0.0% | 264.62 KB | 264.5 KB | 59.04 KB | 59.02 KB
gestalt.es.js | :small_red_triangle_down: 0.0% | :small_red_triangle_down: 0.0% | 262.06 KB | 261.95 KB | 58.67 KB | 58.64 KB
  
<p align="right">
</p>

packages/gestalt/src/Link/Link.js
```diff
@@ -6,9 +6,9 @@ import styles from './Link.css';
 
 type Props = {|
   children?: React.Node,
-  href: string,
+  href?: string,
```
chrislloyd: What's the case for having a link without `href`?

pfarejowicz: @chrislloyd according to the existing `Link` [docs](https://pinterest.github.io/gestalt/#/Link), it would work like this:
image

Though this is still broken because browsers don't like `<a>` as a child of `<a>`. If we choose to support this we'll have to make the element into a `div` or `span` if `href` is `undefined` (I can make a followup PR). The other option is to continue making it required but I would need to remove that from the docs

chrislloyd: @christianvuerings ^ - I agree that I don't think making it optional is the best way to get React-Router support out of the box.

chrislloyd:

PR #230, opened 2018-06-01
Change flow type of item in SegmentedControl to support React.Node

loserford:
Even with type `string`, we can still put components in the items array, and it just works.  
In this change, I split the string type and component types. 
snip20180531_25

### DIFF
--- a/packages/gestalt/src/SegmentedControl/SegmentedControl.js
+++ b/packages/gestalt/src/SegmentedControl/SegmentedControl.js
@@ -6,7 +6,7 @@ import Text from '../Text/Text';
 import styles from './SegmentedControl.css';
 
 type Props = {|
-  items: Array<string | React.Node>,
+  items: Array<React.Node>,
   onChange: ({ event: SyntheticMouseEvent<>, activeIndex: number }) => void,
   selectedItemIndex: number,
   size?: 'md' | 'lg',
--- a/packages/gestalt/src/SegmentedControl/SegmentedControl.js
+++ b/packages/gestalt/src/SegmentedControl/SegmentedControl.js
@@ -2,6 +2,7 @@
 import * as React from 'react';
 import PropTypes from 'prop-types';
 import classnames from 'classnames';
+import Box from '../Box/Box';
 import Text from '../Text/Text';
 import styles from './SegmentedControl.css';
 
@@ -46,7 +47,7 @@ export default function SegmentedControl(props: Props) {
                 {item}
               </Text>
             ) : (
-              <div>{item}</div>
+              <Box display="flex" justifyContent="center">{item}</Box>
             )}
           </button>
         );
--- a/packages/gestalt/src/SegmentedControl/SegmentedControl.js
+++ b/packages/gestalt/src/SegmentedControl/SegmentedControl.js
@@ -6,7 +6,7 @@ import Text from '../Text/Text';
 import styles from './SegmentedControl.css';
 
 type Props = {|
-  items: Array<string>,
+  items: Array<string | React.Node>,
   onChange: ({ event: SyntheticMouseEvent<>, activeIndex: number }) => void,
   selectedItemIndex: number,
   size?: 'md' | 'lg',
@@ -36,14 +36,18 @@ export default function SegmentedControl(props: Props) {
             onClick={e => onChange({ event: e, activeIndex: i })}
             role="tab"
           >
-            <Text
-              bold
-              color={isSelected ? 'darkGray' : 'gray'}
-              align="center"
-              size={size}
-            >
-              {item}
-            </Text>
+            {typeof item === 'string' ? (
+              <Text
+                bold
+                color={isSelected ? 'darkGray' : 'gray'}
+                align="center"
+                size={size}
+              >
+                {item}
+              </Text>
+            ) : (
+              <div>{item}</div>
+            )}
           </button>
         );
       })}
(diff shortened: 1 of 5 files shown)


packages/gestalt/src/SegmentedControl/SegmentedControl.js
```diff
@@ -6,7 +6,7 @@ import Text from '../Text/Text';
 import styles from './SegmentedControl.css';
 
 type Props = {|
-  items: Array<string>,
+  items: Array<any>,
```
bradencanderson: maybe `string | Node`?

christianvuerings:

Built with commit 2cb310687c0f332f64f8f62f056760342df5a64e

https://deploy-preview-230--gestalt.netlify.com

packages/gestalt/src/SegmentedControl/SegmentedControl.js
akleiner2: +1

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  Bundle size increase
  </td>
    </tr>
  </tbody>
</table>



gestalt.js | :small_red_triangle: 0.0% | :small_red_triangle: 0.0% | 264.81 KB | 264.91 KB | 59.1 KB | 59.12 KB
gestalt.es.js | :small_red_triangle: 0.0% | :small_red_triangle: 0.0% | 262.26 KB | 262.36 KB | 58.73 KB | 58.75 KB
  
<p align="right">
</p>

packages/gestalt/src/SegmentedControl/SegmentedControl.js
loserford: +1 :)

packages/gestalt/src/SegmentedControl/SegmentedControl.js
```diff
@@ -6,7 +6,7 @@ import Text from '../Text/Text';
 import styles from './SegmentedControl.css';
 
 type Props = {|
-  items: Array<string>,
+  items: Array<string | React.Node>,
```
pfarejowicz: `React.Node` already covers `string`, you don't need to duplicate it
https://flow.org/en/docs/react/types/#toc-react-node

christianvuerings: @loserford We should also update the docs

loserford: updated. @pfarejowicz @christianvuerings

christianvuerings:
@loserford Thank you for making all these changes! Could you also update the `changelog.md` file?

loserford:
@christianvuerings How to change the `changelog.md`? just put in `### Minor` section? sorry, it's my first to make a PR for gestalt.

packages/gestalt/src/SegmentedControl/SegmentedControl.js
```diff
@@ -36,14 +36,18 @@ export default function SegmentedControl(props: Props) {
...
+              <Text
+                bold
+                color={isSelected ? 'darkGray' : 'gray'}
+                align="center"
+                size={size}
+              >
+                {item}
+              </Text>
+            ) : (
+              <div>{item}</div>
```
christianvuerings: @loserford could we add use `Box display="flex" justifyContent="center"` instead of this `div`? 
That way
 - We'll make sure that the content of each items is always centered
 - We can remove the wrapping `Box` in the example

image

/cc @chrislloyd

christianvuerings:
@loserford That changelog change was perfect, thank you!

loserford:
@christianvuerings should I end it here or continue to `Squash and merge`?

packages/gestalt/src/SegmentedControl/__tests__/SegmentedControl.test.js
```diff
@@ -13,3 +16,21 @@ test('SegmentedControl renders', () => {
   ).toJSON();
   expect(tree).toMatchSnapshot();
 });
+
+test('SegmentedControl renders component items', () => {
+  const wrapper = shallow(
```
pfarejowicz: This could probably be a snapshot test too in my opinion, what do you think?

loserford: Actually I'm quite unsure about when to use snapshot and when to use dom testing. Always worried that am I over using snapshot?  But for this case, if using snapshot, I would prefer to use shallow snapshot. @christianvuerings any thoughts?

pfarejowicz: I think snapshots work great for UI components like this one. By running a snapshot we isolate the component more to future changes because imagine if one day we added a Text label to SegmentedControl that always appears no matter what the items passed in. In this scenario your current test would fail because the count of Text would bump from 3 to 4. My recommendation is make this a snapshot test but also make the last item a dummy div element instead of an Icon so we can further isolate this test away from other components.

loserford: sounds good. I'll change it.

loserford:
@pfarejowicz @christianvuerings do you know anything about the integration test failure? Checked the error, looks like Masonry related, which I didn't touch.

pfarejowicz:
That one is flaky, I hit retry for you from buildkite

PR #205, opened 2018-05-23
Add jsdom specific tests for VideoControls

pfarejowicz:
Sets up test files named `*.jsdom.test.js` to only be run in the jsdom environment. Regular `.test.js` files will still be run in both jsdom and node environments.

PR #231, opened 2018-06-01
Internal: add better basic test coverage

christianvuerings:
Adds basic jest tests to a couple of components

### DIFF
--- a/packages/gestalt/src/ErrorFlyout/__tests__/ErrorFlyout.test.js
+++ b/packages/gestalt/src/ErrorFlyout/__tests__/ErrorFlyout.test.js
@@ -1,9 +1,25 @@
+/**
+ * @jest-environment jsdom
+ */
 // @flow
 import React from 'react';
 import renderer from 'react-test-renderer';
 import ErrorFlyout from '../ErrorFlyout';
 
 test('ErrorFlyout renders', () => {
+  const element = document.createElement('div');
+  const component = renderer.create(
+    <ErrorFlyout
+      anchor={element}
+      message="Email required"
+      onDismiss={() => {}}
+    />
+  );
+  const tree = component.toJSON();
+  expect(tree).toMatchSnapshot();
+});
+
+test('ErrorFlyout does not render when the anchor is null', () => {
   const component = renderer.create(
     <ErrorFlyout anchor={null} message="Email required" onDismiss={() => {}} />
   );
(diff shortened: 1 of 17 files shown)


christianvuerings:

Built with commit 6d38eb61ae5b5afdbd110297e586c1041e3784fd

https://deploy-preview-231--gestalt.netlify.com

christianvuerings:

Built with commit 0571a5396bfc8196e1da1eb5d3ea9088437c20b6

https://deploy-preview-231--gestalt.netlify.com

packages/gestalt/src/ErrorFlyout/__tests__/__snapshots__/ErrorFlyout.test.js.snap
```diff
@@ -0,0 +1,3 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`ErrorFlyout renders 1`] = `null`;
```
pfarejowicz: Seems like this test isn't doing what we want since its returning `null` which doesnt prove it rendered? Should this set of props be a test for `ErrorFlyout does not render`? Then we actually need to write another test for `ErrorFlyout renders`

christianvuerings: @pfarejowicz Thanks for catching that - should be fixed now.

christianvuerings:
@pfarejowicz Was able to use `@jest-environment jsdom` so it only runs in the jsdom env.

https://facebook.github.io/jest/docs/en/configuration.html#testenvironment-string

packages/gestalt/src/ErrorFlyout/__tests__/ErrorFlyout.test.js
```diff
@@ -0,0 +1,28 @@
+/**
+ * @jest-environment jsdom
```
pfarejowicz: I had a discussion with @chrislloyd about this and I'm not too convinced that in our testing setup this is the best solution. My understanding is that we currently run the same test file for both a `jsdom` env and a `node` env. By setting this flag here in the doc block up top, we are overriding the env to be `jsdom` for both test runs. Now that works fine to pass CI because it means both runs get the correct env, but if you look in the output, itll still say `node PASS: ErrorFlyout.test.js` which to me means we are duplicating the jsdom test twice and not actually running this as a node test ever despite it being output as one. My hope was that we could put tests that are only relevant to jsdom environments somewhere special and not even have them picked up by the node test so we dont duplicate, while keeping the tests that can be run in both as they are. Check out #205 for my proposal on how to run `jsdom` specific tests. What do you and @chrislloyd think?

christianvuerings: @pfarejowicz totally agree and noticed that too (the test running twice in the jsdom env). Doesn't seem a blocker for this PR but would be good to fix moving forward.

PR #237, opened 2018-06-04
Icon/Button 100% code coverage

pfarejowicz:
Add tests or refactor to make the test coverage 100% for these two components. Based on data here https://codecov.io/gh/pinterest/gestalt/tree/master/packages/gestalt/src


packages/gestalt/src/Icon/Icon.js
```diff
@@ -58,15 +58,11 @@ export default function Icon(props: Props) {
...
-  let path;
-  if (icon) {
-    path = icons[icon];
-  } else if (dangerouslySetSvgPath) {
+  const path =
+    (icon && icons[icon]) ||
     /* eslint-disable-next-line no-underscore-dangle */
-    path = dangerouslySetSvgPath.__path;
-  } else {
-    path = '';
```
pfarejowicz: I chose to remove this because Flow would not allow use to reach this line in a test unless we add `$ExpectError` or the likes. More importantly though, the native `<path />` component is ok getting passed `undefined` for the value of `d` because itll function like any other undefined attribute so imo this is actually better than an empty string.

christianvuerings:

Built with commit 0dffceefd72de9057f83060efe5232d103e1961c

https://deploy-preview-237--gestalt.netlify.com

pinterest-ios:
gestalt.js | :small_red_triangle_down: 0.0% | :small_red_triangle: 0.0% | 264.95 KB | 264.89 KB | 59.12 KB | 59.12 KB
gestalt.es.js | :small_red_triangle_down: 0.0% | :small_red_triangle: 0.0% | 262.39 KB | 262.33 KB | 58.75 KB | 58.76 KB
  
<p align="right">
</p>

packages/gestalt/src/Button/__tests__/Button.test.js
```diff
@@ -0,0 +1,19 @@
...
+test('<Button color="transparent" />', () => {
+  const tree = create(
+    <Button color="transparent" text="Hello World" />
+  ).toJSON();
+  expect(tree).toMatchSnapshot();
+});
+
+test('Button handles click', () => {
+  const mockOnClick = jest.fn();
+  const wrapper = shallow(<Button text="Text" onClick={mockOnClick} />);
```
pfarejowicz: Don't need to use a `.jsdom.test.js` file because the `shallow` render works in both and we can still find the native `<button />` element just fine

PR #242, opened 2018-06-05
Merge ErrorFlyout abilities into Flyout

pfarejowicz:
Since `ErrorFlyout` is just a type of `Flyout` it makes sense to merge these as @chrislloyd pointed out.
Migration plan:
1. This PR - merge ErrorFlyout abilities into Flyout. The main differences are the background, the `shouldFocus` prop, and the `children`. I chose to handle this by implementing a `role` prop with options similar to `Modal` (`alertdialog` and `dialog`). @chrislloyd is this ok or should we just switch to `type = 'error' | 'dialog'`?
2. Cut a release with this change and convert all `<ErrorFlyout />` uses to `<Flyout role="alertdialog">`. At this stage we should convert the `message` prop from `ErrorFlyout` to a child that looks like
```
      <Box padding={3}>
        <Text bold color="white">
          <span id={id}>{message}</span>
        </Text>
      </Box>
```
Does this look good, or should we update the Flow typing such that `role = 'alertdialog'` automatically adds this snippet for children and expects a `message` prop as well @chrislloyd ?
3. Cut another release where we remove the exported `ErrorFlyout` once and for all

### DIFF
--- a/docs/src/Flyout.doc.js
+++ b/docs/src/Flyout.doc.js
@@ -47,6 +47,13 @@ card(
           'Depicts if the Flyout shares a relative root with the anchor element',
       },
       {
+        name: 'role',
+        type: `"alertdialog" | "dialog"`,
+        defaultValue: 'dialog',
+        description:
+          'The render style for this Flyout: alertdialog changes the colors to match other errors',
+      },
+      {
         name: 'size',
         type: `'xs' | 'sm' | 'md' | 'lg' | 'xl' | number`,
         description: `xs: 185px, sm: 230px, md: 284px, lg: 320px, xl:375px`,
@@ -93,7 +100,6 @@ class FlyoutExample extends React.Component {
           <Flyout
             anchor={this.anchor}
             idealDirection="up"
-            message="Oops! This item is out of stock."
             onDismiss={this.handleDismiss}
             size="md"
           >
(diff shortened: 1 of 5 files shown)


christianvuerings:

Built with commit 94eb58a1b982ebbb1ca8a5b3a4a90df2943e59f4

https://deploy-preview-242--gestalt.netlify.com

christianvuerings:

Built with commit 3e6e991bca4daf4acbaa0df91d493496d917141c

https://deploy-preview-242--gestalt.netlify.com

pinterest-ios:
gestalt.js | :small_red_triangle_down: 0.1% | :small_red_triangle_down: 0.0% | 266.74 KB | 266.48 KB | 58.92 KB | 58.92 KB
gestalt.es.js | :small_red_triangle_down: 0.1% | :small_red_triangle_down: 0.0% | 264.11 KB | 263.85 KB | 58.56 KB | 58.56 KB
  
<p align="right">
</p>

packages/gestalt/src/Flyout/__tests__/__snapshots__/Flyout.jsdom.test.js.snap
```diff
@@ -9,3 +9,11 @@ exports[`Flyout renders 1`] = `
   <div />
 </div>
 `;
+
+exports[`Flyout renders as error 1`] = `
```
chrislloyd: Hmm this doesn't seem right.

docs/src/Flyout.doc.js
```diff
@@ -46,6 +46,13 @@ card(
         description:
           'Depicts if the Flyout shares a relative root with the anchor element',
       },
+      {
+        name: 'role',
+        type: `"alertdialog" | "dialog"`,
```
chrislloyd: I don't think this is the right way to make this change. It jams in a visual change with a accessibility feature. Perhaps adding `color` (or similar) instead?

packages/gestalt/src/Flyout/__tests__/__snapshots__/Flyout.jsdom.test.js.snap
pfarejowicz: Good catch, the `Flyout` `Controller` child render depends on a ref being existent in order to fully render so the test was treating it as not being set. By using the `createNodeMock` option to some random truthy value we can bypass that and get the full render going.

docs/src/Flyout.doc.js
pfarejowicz: I changed it to `color` let me know what you think. I originally chose role because it matches `Modal` and because it allows a finite set of types with specific purposes. By changing it to `color` at least to me it kind of dilutes the purpose of each color and it can lead to people alter asking why "green" "blue" "red" etc are not supported. In my eyes its harder to map "orange" to being the old `ErrorFlyout` component, than it is to map a role of `alertdialog` to that. Maybe there is some middle ground here with a `type` prop instead? For example look at the `shouldFocus` prop, ErrorFlyout and Flyout set that differently so it feels weird setting that now based on the color. Thoughts @chrislloyd?

PR #248, opened 2018-06-08
[Avatar] Fix error when name is falsey

### LINKED ISSUE #247: Avatar breaks when name=""
Unfortunately this bug is more severe now that errors in React 16 cause the whole tree to get unmounted.

The error happens [here](https://github.com/pinterest/gestalt/blob/master/packages/gestalt/src/Avatar/Avatar.js#L23).

The name in Avatar is usually UGC and it's common for users to, e.g. set their name to `""`.

bradencanderson:
Fixes #247

### DIFF
--- a/packages/gestalt/src/Avatar/Avatar.js
+++ b/packages/gestalt/src/Avatar/Avatar.js
@@ -23,29 +23,31 @@ const DefaultAvatar = ({ name }: { name: string }) => {
   const firstInitial = name ? [...name][0].toUpperCase() : '';
   return (
     <Square color="gray" shape="circle">
-      <svg
-        width="100%"
-        viewBox="-50 -50 100 100"
-        version="1.1"
-        preserveAspectRatio="xMidYMid meet"
-        xmlns="http://www.w3.org/2000/svg"
-      >
-        <title>{name}</title>
-        <text
-          fontSize="50px"
-          fill="#fff"
-          dominantBaseline="central"
-          textAnchor="middle"
-          className={[
-            typography.antialiased,
-            typography.sansSerif,
-            typography.leadingSmall,
-            typography.fontWeightBold,
-          ].join(' ')}
+      {firstInitial && (
+        <svg
+          width="100%"
+          viewBox="-50 -50 100 100"
+          version="1.1"
+          preserveAspectRatio="xMidYMid meet"
+          xmlns="http://www.w3.org/2000/svg"
         >
-          {firstInitial}
-        </text>
-      </svg>
+          <title>{name}</title>
+          <text
+            fontSize="50px"
+            fill="#fff"
+            dominantBaseline="central"
+            textAnchor="middle"
+            className={[
+              typography.antialiased,
+              typography.sansSerif,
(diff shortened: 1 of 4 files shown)


christianvuerings:

Built with commit 62f3b12fae5ce85e7e1f20aad3c3b519403931b9

https://deploy-preview-248--gestalt.netlify.com

packages/gestalt/src/Avatar/Avatar.js
```diff
@@ -20,7 +20,7 @@ const Square = (props: *) => (
 );
 
 const DefaultAvatar = ({ name }: { name: string }) => {
-  const firstInitial = [...name][0].toUpperCase();
+  const firstInitial = name ? [...name][0].toUpperCase() : '';
```
pfarejowicz: Since react renders empty strings into the DOM should that be a null or undefined?

christianvuerings:

Built with commit 6a85e2113512152caaa7bb0f0b09688f71b44ec7

https://deploy-preview-248--gestalt.netlify.com

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  Bundle size increase
  </td>
    </tr>
  </tbody>
</table>



gestalt.js | :small_red_triangle: 0.0% | :small_red_triangle: 0.0% | 268.24 KB | 268.25 KB | 59.53 KB | 59.54 KB
gestalt.es.js | :small_red_triangle: 0.0% | :small_red_triangle: 0.0% | 265.62 KB | 265.63 KB | 59.17 KB | 59.18 KB
  
<p align="right">
</p>

packages/gestalt/src/Avatar/Avatar.js
bradencanderson: Both seem equivalent from an end-user perspective. What's wrong with empty strings in the DOM?

pfarejowicz: Feels dirty, Airbnb has a rule of always explicitly checking against an empty string or a 0 instead of relying on the implicit Boolean conversion so that you don’t get empty strings and 0s in the DOM when you do a conditional render. This sort of feels the same lines. Just thought we should make gestalt as clean as possible for all users

bradencanderson: It sounds like you're talking about 2 different issues - (1) the implicit boolean conversion and (2) putting an empty string in the DOM. No opinion on either in general, but I think in this specific case they're both safe. If the reviewer has an opinion I'm happy to update.

pfarejowicz: Agreed - leaving it up to @christianvuerings

christianvuerings: @bradencanderson @pfarejowicz I'm ok with the empty string in this case. However, we should conditionally render the SVG (within the `Square`) below so we don't render DOM that's not required.

```jsx
<Square color="gray" shape="circle">
  {firstInitial && <svg
```

PR #261, opened 2018-06-25
Internal: Flatten imports and use explicit extensions

chrislloyd:
This is a big looking PR but really on does two small things:

a.) Uses `no-relative-parent-imports` eslint rule to flatten imports and enforce explicit namespaces.
b.) Turns on file extensions by default (something that Ryan Dahl lists as a regret with Nodes module resolution system).

While this may make the `packages/gestalt/src` directory a lot bigger, I think it has serveral positive benefits of showing what stuff is shared in gestalt and allowing us to isolate parts of codebase (behaviors, layouts etc.)

You can see a quick tree-view of the results [on my branch](https://github.com/chrislloyd/gestalt/tree/eslint-warn-about-parent-imports/packages/gestalt/src).


christianvuerings:

Built with commit efc690b8552eda5f58aec8b5eac84a2146c175e1

https://deploy-preview-261--gestalt.netlify.com

christianvuerings:

Built with commit 19c5a3f48a977a4f839902d99f9b3a972eb85431

https://deploy-preview-261--gestalt.netlify.com

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  :memo: CHANGELOG - <i>Please add a CHANGELOG.md entry for your changes</i>
  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

docs/src/components/PageHeader.js
```diff
@@ -1,15 +1,14 @@
...
+import Markdown from './Markdown.js';
 
 type Props = {|
   name: string,
   description?: string,
 |};
 
-const gestaltPath = component =>
-  `/packages/gestalt/src/${component}/${component}.js`;
+const gestaltPath = component => `/packages/gestalt/src/${component}.js`;
```
pfarejowicz: Just noticed this, not a result of your changes, but a "bug" we've had, this is adding 2 `/` characters because of the way we do string templating
`https://github.com/pinterest/gestalt/blob/master//packages/gestalt/src/Checkbox/Checkbox.js`

packages/gestalt/dist/gestalt.js.flow
```diff
@@ -1,2 +1,2 @@
 // @flow
-export * from '../src';
+export * from './src.js';
```
pfarejowicz: Is this one actually a file called `src.js`? I thought this reaches into the `src` folder hence the `..`

christianvuerings:
@chrislloyd seems like this might break all the `Source` links for a while. Might be good to do a release soon after

PR #263, opened 2018-06-25
Flyout: Reuse OutsideEventBehavior for flyout controllers

chrislloyd:


packages/gestalt/src/Controller.js
```diff
@@ -126,30 +122,21 @@ export default class Controller extends React.Component<Props, State> {
...
+          idealDirection={idealDirection}
+          onKeyDown={this.handleKeyDown}
+          onResize={this.handleResize}
+          positionRelativeToAnchor={positionRelativeToAnchor}
+          relativeOffset={this.state.relativeOffset}
+          shouldFocus={shouldFocus}
+          triggerRect={this.state.triggerBoundingRect}
+          width={width}
         >
-          {this.contents ? (
```
pfarejowicz: Does this change anything? `this.contents` does not exist on first render, and setting a ref does not trigger a re-render, so I just recently ran into this in our code 😞
```
  componentDidMount() {
    // This setState causes a re-render.  Flyout is not designed to be added immediately to visible components.
    // Flyout will not render on the first render.  This re-render works around this.
    this.setState({});
  }
```
Looking at `OutsideEventBehavior` looks like it always renders children now which means `Content` will always be rendered.

chrislloyd: `Content` also re-renders itself so the "re-render" on mount behavior will still be there. Also, at worst it'll just mean that your code will render more than it needs to - it shouldn't actually break that code. Also, `this.forceUpdate()` is better in that case than `this.setState({})` because `setState` can be intercepted by `shouldComponentUpdate`.

PR #266, opened 2018-06-27
Fix release script for gh-pages

pfarejowicz:
My change to add pre-commit linting broke the gh-pages release script because it tries to run on the compiled files and blocks completion of the release. This flag bypasses that check on a release commit for gh-pages - something I had to work around in the 0.75.0 release - with this fix that will no longer be an issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@
 
 ### Patch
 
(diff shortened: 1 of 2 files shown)


christianvuerings:

Built with commit 76c46bc963002bb84258e47af5982656100c8d02

https://deploy-preview-266--gestalt.netlify.com

PR #273, opened 2018-07-11
Prepare gestalt for eslint 5 via bump to plugins

pfarejowicz:
[ESLINT 5](https://eslint.org/docs/user-guide/migrating-to-5.0.0) is out, but we cannot upgrade because the [eslint-plugin-airbnb](https://github.com/airbnb/javascript/blob/cae920fc308325b40f49c5e8f76bbeb5d8663fc5/packages/eslint-config-airbnb/package.json) latest version has a peer dependency on eslint 4.19.1 still. Trying to upgrade straight to eslint 5 leads to lots of peer dep errors from yarn. The [fix](https://github.com/airbnb/javascript/commit/a510095acf20e3d96a94e6d0d0b26cfac71d2c7f) has already landed on the airbnb master but it has not been released yet. 

Tracking https://github.com/airbnb/javascript/issues/1834 for when we can finally update to eslint 5!

In the meantime, this change:

1. Upgrades all other eslint related packages to the latest versions to be ready for eslint 5
2. Fixes the errors that came from the upgrades
3. Disables the `react/destructuring-assignment` rule temporarily as that one is a big offender ~80 times so I will handle it on its own
4. Disables the `button-has-type` rule on `Button` because apparently the creators want developers to hard code values and duplicate code instead of using the dynamic type. See https://github.com/yannickcr/eslint-plugin-react/issues/1555 for more info
5. For places that needed a new `button` `type` I used the `submit` value as that is what the default behavior is if the attribute is ignored according to [HTML docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button)


christianvuerings:

Built with commit 64ba012e4a8fad28095ff50f690f5ada1c2ca433

https://deploy-preview-273--gestalt.netlify.com

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  :file_folder: Bundlesize - <i>Bundle size increase detected, please review</i>
  </td>
    </tr>
  </tbody>
</table>


gestalt.js | :small_red_triangle: 0.1% | :small_red_triangle: 0.0% | 254.56 KB | 254.74 KB | 56.73 KB | 56.74 KB
gestalt.es.js | :small_red_triangle: 0.1% | :small_red_triangle: 0.0% | 251.93 KB | 252.11 KB | 56.35 KB | 56.36 KB
<p align="right">
</p>

packages/gestalt/src/Button.js
```diff
@@ -54,6 +54,7 @@ export default function Button(props: Props) {
     [styles.block]: !inline,
   });
 
+  /* eslint-disable react/button-has-type */
```
christianvuerings: @pfarejowicz Any idea why we need to disable this even though we set it below? Is it because `type` is dynamic?

`type={type}`

packages/gestalt/src/Flyout.test.js
```diff
@@ -6,7 +6,11 @@ import Flyout from './Flyout.js';
 it('Flyout renders', () => {
   const wrapper = shallow(
     <Flyout
-      anchor={<button onClick={() => null}> test </button>}
+      anchor={
+        <button onClick={() => null} type="submit">
```
christianvuerings: @pfarejowicz we should probably change this to `type="button"`. A flyout inside of a form should not be submitting the form. Also `button` is the default for `<Button>`

packages/gestalt/src/SegmentedControl.js
```diff
@@ -36,6 +36,7 @@ export default function SegmentedControl(props: Props) {
             key={i}
             onClick={e => onChange({ event: e, activeIndex: i })}
             role="tab"
+            type="submit"
```
christianvuerings: @pfarejowicz could we change this to `type="button"`?

test/containers/ExampleGridItem.js
```diff
@@ -48,6 +48,7 @@ export default class Item extends Component {
             <button
               id={`add-more-${itemIdx}`}
               onClick={() => fetchMore().then(addRelatedItems)}
+              type="submit"
```
christianvuerings: @pfarejowicz same here - change to `type="button"`

packages/gestalt/src/Button.js
pfarejowicz: @christianvuerings exactly! the rule does not like us using the dynamic value `type={type}` even if it is Flow typed and prop typed to match the possible values. There's more discussion in https://github.com/yannickcr/eslint-plugin-react/issues/1555 and the recommendation is to fork the component and render two different JSX components but to me that feels like overkill, what are your thoughts? For reference:

> I'd say that the rule should be forcing a single, hardcoded type value - and that you should conditionally render two different elements if you want two types.

packages/gestalt/src/Flyout.test.js
pfarejowicz: Will change

packages/gestalt/src/SegmentedControl.js
pfarejowicz: I can make the change, I agree `button` is better than `submit`. My goal was to introduce the minimal amount of changes and since this is a `<button />` and not a `<Button />` the default here was actually always setting this to `type="submit"` since we were not specifying it explicitly. See the [Mozilla docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button) for reference. Making this change now, but just making a not here for future readers that the behavior is changing from our released version.

> **type**
> The type of the button. Possible values are:
>
> **submit**: The button submits the form data to the server. This is the default if the attribute is not specified, or if the attribute is dynamically changed to an empty or invalid value.

packages/gestalt-codemods/0.33.0-0.34.0/avatar-non-responsive.js
```diff
@@ -56,7 +56,8 @@ export default function transformer(file, api) {
             if (convertSizes[value]) {
               attribute.value.value = convertSizes[value];
               return attribute;
-            } else if (invalidSizes.includes(value)) {
+            }
+            if (invalidSizes.includes(value)) {
```
chrislloyd: wow huh

test/containers/ExampleGridItem.js
pfarejowicz: Done, same comment as above but this is a test component so Im not too concerned

packages/gestalt/src/Button.js
```diff
@@ -76,8 +77,17 @@ export default function Button(props: Props) {
       </Text>
     </button>
   );
+  /* eslint-enable react/button-has-type */
 }
 
+Button.defaultProps = {
```
chrislloyd: I'm so torn on this one. None of our tools really introspect `defaultProps` and it's not a required part of the React API so this definitely feels like extra typing...

packages/gestalt/src/Button.js
christianvuerings: Disabling it here seems fine - thanks for clarifying

packages/gestalt-codemods/0.33.0-0.34.0/avatar-non-responsive.js
pfarejowicz: @chrislloyd same reaction haha, its this rule https://eslint.org/docs/rules/no-else-return which does not like an `else if` clause after a `return`

packages/gestalt/src/Button.js
pfarejowicz: Agreed - I started off by playing with the typing and defaulting, hoping to fix the `type={type}` error but to no avail since I found the github thread linked above saying it needs to be hardcoded to `"button"` or `"submit"`. Figured I would keep the `defaultProps` that I created anyways since it is a common pattern. Thoughts, I can revert the defaulting since the rule needs to be disabled anyways

PR #286, opened 2018-07-19
Toast: Make icon prop nullable

mminute:
Accommodate code like `const iconName = href && !imageUrl ? 'arrow-circle-forward' : null;` by allowing icon to be null.

### DIFF
--- a/packages/gestalt/src/Toast.js
+++ b/packages/gestalt/src/Toast.js
@@ -8,7 +8,7 @@ import Icon from './Icon.js';
 
 type Props = {|
   color?: 'darkGray' | 'orange',
-  icon?: 'arrow-circle-forward', // leaving open to additional icons in the future
+  icon?: ?'arrow-circle-forward', // leaving open to additional icons in the future
   text: string | Array<string>,
   thumbnail?: React.Element<any>,
 |};
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
   </summary>
 
 ### Minor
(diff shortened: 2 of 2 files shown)


christianvuerings:

Built with commit 713d385557d0d104e20043b27a25dc71484ef81a

https://deploy-preview-286--gestalt.netlify.com

packages/gestalt/src/Toast.js
```diff
@@ -8,7 +8,7 @@ import Icon from './Icon.js';
 
 type Props = {|
   color?: 'darkGray' | 'orange',
-  icon?: 'arrow-circle-forward', // leaving open to additional icons in the future
+  icon?: ?'arrow-circle-forward', // leaving open to additional icons in the future
```
christianvuerings: As a reference https://wietse.loves.engineering/flowtype-understanding-optionals-and-null-2fa7b98210ce

@chrislloyd @pfarejowicz how do you both feel about this?

pfarejowicz: Personally if neither `icon?: 'arrow-circle-forward'` nor `icon: ?'arrow-circle-forward' work for you, id rather do `icon?: 'arrow-circle-forward' | null`, but im curious what the reason for this change is, do we have a conflict in pinboard that requires `null` to be acceptable

pfarejowicz: For the example of `const iconName = href && !imageUrl ? 'arrow-circle-forward' : null;` i think id personally do `const iconName = href && !imageUrl ? 'arrow-circle-forward' : undefined;`

PR #287, opened 2018-07-23
Checkbox: Add error prop and styling

mminute:
image

### DIFF
--- a/docs/src/Checkbox.doc.js
+++ b/docs/src/Checkbox.doc.js
@@ -153,7 +153,7 @@ class CheckboxExample extends React.Component {
     return (
       <Box alignItems="center" direction="row" display="flex">
         <Checkbox
-          error
+          hasError
           id="error"
           name="error"
           onChange={() => {}}
--- a/docs/src/Checkbox.doc.js
+++ b/docs/src/Checkbox.doc.js
@@ -32,6 +32,11 @@ card(
         defaultValue: false,
       },
       {
+        name: 'hasError',
+        type: 'boolean',
+        defaultValue: false,
+      },
+      {
         name: 'id',
         type: 'string',
         required: true,
--- a/docs/src/Checkbox.doc.js
+++ b/docs/src/Checkbox.doc.js
@@ -145,6 +145,33 @@ class CheckboxExample extends React.Component {
 );
 
 card(
+  <Example
+    name="Example: Error state"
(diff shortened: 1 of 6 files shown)


packages/gestalt/src/Checkbox.css
```diff
@@ -20,22 +32,18 @@
 
 .checkDarkGray {
```
pfarejowicz: since these css classes no longer set anything else besides composing the background color from `Colors.css` can we just update the js code to directly import those styles and have the logic directly set those classnames? this will reduce the duplication in my opinion

packages/gestalt/src/Checkbox.js
```diff
@@ -9,6 +9,7 @@ import Icon from './Icon.js';
 type Props = {|
   checked?: boolean,
   disabled?: boolean,
+  error?: boolean,
```
pfarejowicz: does it make more sense to call this `hasError` like `TextField` and `TextArea`? whats the long term plan for these props? i think one consistent name is clearer to users

packages/gestalt/src/Checkbox.js
```diff
@@ -24,6 +25,7 @@ export default class Checkbox extends React.Component<Props, State> {
   static propTypes = {
     checked: PropTypes.bool,
     disabled: PropTypes.bool,
+    error: PropTypes.bool,
```
pfarejowicz: do you want to set a default for this in `defaultProps` to false?

docs/src/Checkbox.doc.js
```diff
@@ -144,6 +144,33 @@ class CheckboxExample extends React.Component {
...
+card(
+  <Example
+    name="Example: Error state"
+    defaultCode={`
+class CheckboxExample extends React.Component {
+  render() {
+    return (
+      <Box alignItems="center" direction="row" display="flex">
+        <Checkbox
+          error
```
pfarejowicz: this should be `hasError`

mminute: thanks for catching

docs/src/Checkbox.doc.js
```diff
@@ -144,6 +144,33 @@ class CheckboxExample extends React.Component {
   />
```
christianvuerings: @mminute could we update the flowtypes in the docs?

image

PR #291, opened 2018-07-25
Collage component

jackhsu978:
Adding a Collage component to Gestalt.

Without gutter
---
screenshot 2018-07-25 01 12 14

With gutter
---
```jsx
<Collage ... gutter={8}  />
```
screenshot 2018-07-25 15 52 32

With gutter and cover image
---
```jsx
<Collage  ...  cover gutter={8} />
```
screenshot 2018-07-25 15 50 45

### DIFF
--- a/docs/src/Collage.doc.js
+++ b/docs/src/Collage.doc.js
@@ -127,7 +127,7 @@ card(
       <Mask wash width={width} height={height}>
         <Image
           alt="collage image"
-          color={image.color || '#EFEFEF'}
+          color={image.color}
           fit="cover"
           naturalHeight={image.naturalHeight}
           naturalWidth={image.naturalWidth}
@@ -199,7 +199,7 @@ card(
               {image ? (
                 <Image
                   alt="collage image"
-                  color={image.color || '#EFEFEF'}
+                  color={image.color}
                   fit="cover"
                   naturalHeight={image.naturalHeight}
                   naturalWidth={image.naturalWidth}
@@ -277,7 +277,7 @@ card(
         <Mask wash width={width} height={height}>
           <Image
             alt="collage image"
-            color={image.color || '#EFEFEF'}
+            color={image.color}
             fit="cover"
             naturalHeight={image.naturalHeight}
             naturalWidth={image.naturalWidth}
@@ -403,7 +403,7 @@ card(
               {image ? (
                 <Image
                   alt="collage image"
-                  color={image.color || '#EFEFEF'}
+                  color={image.color}
                   fit="cover"
                   naturalHeight={image.naturalHeight}
                   naturalWidth={image.naturalWidth}
@@ -489,7 +489,7 @@ card(
             <Mask wash width={width} height={height}>
               <Image
                 alt="collage image"
-                color={image.color || '#EFEFEF'}
+                color={image.color}
                 fit="cover"
                 naturalHeight={image.naturalHeight}
                 naturalWidth={image.naturalWidth}
--- a/docs/src/Collage.doc.js
+++ b/docs/src/Collage.doc.js
@@ -1,0 +1,270 @@
+// @flow
+import * as React from 'react';
+import stock1 from './images/stock1.jpg';
+import stock2 from './images/stock2.jpg';
+import stock3 from './images/stock3.jpg';
+import stock4 from './images/stock4.jpg';
+import stock5 from './images/stock5.jpg';
+import stock6 from './images/stock6.jpg';
+import PropTable from './components/PropTable.js';
+import Example from './components/Example.js';
+import PageHeader from './components/PageHeader.js';
+
+const cards = [];
+const card = c => cards.push(c);
+
+card(
+  <PageHeader
+    name="Collage"
+    description="
+Like Masonry, Collage creates a deterministic grid layout that can absolutely position and virtualize images.
+"
+  />
+);
+
+card(
+  <PropTable
+    props={[
+      {
+        name: 'height',
+        type: 'number',
+        description: 'Height of the collage',
+        required: true,
+      },
+      {
+        name: 'imageAlt',
+        type: 'string',
+        description: 'Accessibility label for every image',
+      },
+      {
+        name: 'images',
+        type:
+          'Array<{| color?: string, naturalHeight: number, naturalWidth: number, src: string |}>',
+        description: 'Images to show in the collage',
+        required: true,
+      },
+      {
+        name: 'imagesAreRounded',
+        type: 'boolean',
+        description: 'Use rounded corner for every image',
+      },
+      {
+        name: 'numCols',
+        type: 'number',
+        description: 'Number of columns',
+        required: true,
+      },
+      {
+        name: 'renderCoverImage',
+        type: '({ width: number, height: number }) => React.Node',
+        description: `Render prop for rendering the first image (a.k.a cover image).`,
+      },
+      {
+        name: 'space',
+        type: '{| type: "around" | "between", value: number |}',
+        description:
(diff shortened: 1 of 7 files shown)


christianvuerings:

Built with commit 4bda4be9709a7f6fec31ea786907fda0c9ad3981

https://deploy-preview-291--gestalt.netlify.com

docs/src/Collage.doc.js
```diff
@@ -0,0 +1,270 @@
...
+      {
+        name: 'space',
+        type: '{| type: "around" | "between", value: number |}',
+        description:
+          'The type of spacing (similar to flexbox space-around and space-between) and the amount of vertical & horizontal padding around/between images',
+      },
+      {
+        name: 'width',
+        type: 'number',
+        description: 'The URL of the captions track for the video (.vtt file)',
```
chrislloyd: Copy 🍝

docs/src/Collage.doc.js
```diff
@@ -0,0 +1,270 @@
...
+        required: true,
+      },
+      {
+        name: 'renderCoverImage',
+        type: '({ width: number, height: number }) => React.Node',
+        description: `Render prop for rendering the first image (a.k.a cover image).`,
+      },
+      {
+        name: 'space',
+        type: '{| type: "around" | "between", value: number |}',
```
chrislloyd: Seeing as both of these properties are required is there any value in having them be an object? If `value` changes we'd need to implement a custom `shouldComponentUpdate` to check the value of the `space` object. How about `gutterStyle: "around" | "between", gutterSize: number`?

packages/gestalt/src/Collection.js
```diff
@@ -45,6 +45,7 @@ import layoutStyles from './Layout.css';
 
 type Props = {|
   Item: ({ idx: number }) => React.Node,
+  itemPadding?: number,
```
chrislloyd: This can be calculated outside of `<Collection/>` - it's what the `paddingAll` functions in `<Collage/>` do.

packages/gestalt/src/Collage.js
```diff
@@ -0,0 +1,270 @@
...
+    />
+  );
+}
+
+Collage.propTypes = {
+  height: PropTypes.number.isRequired,
+  imageAlt: PropTypes.string,
+  images: PropTypes.arrayOf(ImagePropType),
+  imagesAreRounded: PropTypes.bool,
+  numCols: PropTypes.number.isRequired,
```
chrislloyd: This might be nicer as just `columns` - `num` is a type hint and is already expressed elsewhere.

docs/src/Collage.doc.js
jackhsu978: oops... will update

pfarejowicz:
@jackhsu978 I'm curious what the need for the space `around` is for? It looks a little asymmetrical to me as the spacing specified is doubled in between the images because the they sit side by side unlike around the edges
image
If they user wants this effect, is there any reason to not just use the `between` spacing and then wrap the whole thing in a box with some padding? That way the `space` prop essentially just signals how much in between images rather than the type and the spacing

docs/src/Collage.doc.js
jackhsu978: The reason I put it in an object is to make sure that either both `gutterSize` and `gutterStyle` are passed or none are passed. It doesn't make sense to require `gutterStyle` prop if `gutterSize` is 0 or undefined (which is the default).

Maybe we should make `gutterStyle` default to "between" and have both of them optional? Thoughts?

Also, this "space" value here is technically the padding of each image. The actual gutter size is 2 * image padding. That's why I didn't pick the name "gutter".

If we want to use "gutter" (which might be more intuitive), the parent could potentially pass a odd number and we might end up doing float point pixel values for the image padding, which I'm trying to avoid. Thoughts on this?

packages/gestalt/src/Collection.js
jackhsu978: The width and height of the item layer in `<Collection>` are currently max of x-of-right-edge and max of y-of-bottom-edge. It includes the left / top border gutters but excludes the right / bottom border gutters.

For example, in the collage below, the Collection item layer would have width and height the same as what the red box shows:
collection-size

While this doesn't actually cause any problems, I figured that it's probably best if we make sure the Collection size matches the Collage's size.

collection-size-fixed

I understand this additional logic might not be necessary. Happy to change it if you don't think Collection item layer size needs to match the Collage's.

packages/gestalt/src/Collage.js
jackhsu978: I know `columns` is a number hinted by flow, but I usually thought of plural as an array of things instead of a number. I know Masonry uses `minCols` so we're already using plural for non-Array, but I feel that's more intuitive because of the adjective `min`. Thoughts?

packages/gestalt/src/Collection.js
chrislloyd: I think here you could account for the extra padding by passing in a bigger `width` & `height` or `viewportWidth`/`viewportHeight` to collection. It's behavior of maxing the items is only a convenient default.

jackhsu978:
@chrislloyd Passing a render prop for each individual image is a great idea. I will do that. I also replied your comments inline.

@pfarejowicz I have "around" style because that was what we are doing with the board collage, which has 2px padding between items and 1px padding around the border. I guess it wasn't a problem because you can barely tell. I've just changed the board collage to use the "between" style and it actually looks slightly better

Before:
screenshot 2018-07-25 01 59 27

After:
screenshot 2018-07-25 01 59 36


With that, there's currently no use case for the around style. I think we should just drop that case altogether. Thoughts? @pfarejowicz @chrislloyd

docs/src/Collage.doc.js
jackhsu978: On a second thought, padding gutter size makes more sense, especially now that we don't have the around case anymore. Will update to just accepting `gutter` prop

jackhsu978:
@chrislloyd @pfarejowicz I addressed the comments. Can you take another look when you get a chance?

packages/gestalt/src/Collage.js
```diff
@@ -0,0 +1,193 @@
...
+          height: positions[index].height,
+        })
+      }
+      layout={positions}
+    />
+  );
+}
+
+Collage.propTypes = {
+  columns: PropTypes.oneOf([1, 2, 3, 4]).isRequired,
```
pfarejowicz: this is mismatched to the Flow typing above, should be `[2,3,4]`

jackhsu978: changed! good catch!

chrislloyd:
I was wondering what the ‘layoutKey’ prop was? It’s not entirely clear what it is or how to use it from the docs (or if it’s even nessecary).

jackhsu978:
Given the number of columns, a board collage can have different layouts/templates. Each layout controls how the height of each item in each column. In the docs, I have a section that shows all possible layouts for 3 columns.

The design is pre-existing and I don’t really have context of the design decision. I do think I understand the motivation. We want to make board collage layout to look slightly different to make the profile boards more visually appealing. We generate a hash number based for each collage, and use that number to choose the layout from the list of predefined layouts.

jackhsu978:
Here are screenshots from the docs.

Docs for the optional `layoutKey` prop in the prop table:
screenshot 2018-07-26 20 07 35

Live sample for how `layoutKey` is used.
screenshot 2018-07-26 20 07 46

jackhsu978:
@chrislloyd I answered your question. Can you take another look?

chrislloyd:
Looks good, thanks for the update @jackhsu978. Sorry for the slow r+

jackhsu978:
@chrislloyd thanks for the review! merged.

PR #289, opened 2018-07-24
Bump flow version to 0.77.0

pfarejowicz:
No new errors arising from the bump to version 0.77.0

PR #292, opened 2018-07-25
Add strict flow linting rules to codebase

pfarejowicz:
Flow exposes some pretty cool and powerful [linting rules](https://flow.org/en/docs/linting/) that can bring our codebase to the next level! This PR enabled most of them as errors so that we maintain a high quality standard moving forward.

All [rules](https://flow.org/en/docs/linting/rule-reference/) except the `sketchy-null` and `sketchy-number` rules are enabled in this PR. This includes: `untyped-type-import` `untyped-import` `unclear-type` `unsafe-getters-setters` and `nonstrict-import`.


packages/gestalt/src/Box.js
```diff
@@ -32,12 +32,14 @@ import {
   toProps,
 } from './style.js';
 import {
+  // flowlint untyped-import:off
```
pfarejowicz: Turning the rule off for the imports in this file as I need to research @chrislloyd 's "functors" from `transforms.js` more before typing them properly

christianvuerings:

Built with commit 340599a1fe84b20fa0e64c3b665558e0aaa6bec6

https://deploy-preview-292--gestalt.netlify.com

packages/gestalt/src/Flyout.test.js
```diff
@@ -1,25 +0,0 @@
-// @flow
```
pfarejowicz: These tests are being deleted as they are duplicated in the `Flyout.jsdom.test.js` file and actually throw a flow error now for the `anchor` prop

packages/gestalt/src/Video.js
```diff
@@ -377,7 +374,7 @@ export default class Video extends React.PureComponent<Props, State> {
   };
 
   // Sent when the video is switched to/out-of fullscreen mode
-  handleFullscreenChange = (event: SyntheticEvent<HTMLDivElement>) => {
+  handleFullscreenChange = (event: Event) => {
```
pfarejowicz: This is changing because `document.addEventListener` actually shuffles the native `Event` along instead of the React `SyntheticEvent` so this is the correct typing based on the updates above.

pinterest-ios:
gestalt.js | :small_red_triangle_down: 0.0% | :small_red_triangle_down: 0.1% | 256.34 KB | 256.26 KB | 57.12 KB | 57.08 KB
gestalt.es.js | :small_red_triangle_down: 0.0% | :small_red_triangle_down: 0.1% | 253.69 KB | 253.61 KB | 56.75 KB | 56.71 KB
<p align="right">
</p>

packages/gestalt/src/debounce.js
```diff
@@ -5,10 +5,13 @@
...
 
 // @flow
-export default function debounce(fn: () => void, threshhold: number = 100) {
+export default function debounce(
+  fn: (...args: *) => void,
+  threshhold: number = 100
+) {
   let deferTimer: TimeoutID | null = null;
 
-  const debounced = (...args: any) => {
```
pfarejowicz: Had to change this because there is some confusion in the typing here and worth a revisit. The `*` while deprecated at least lets us bypass the rule for now. Originally `fn` was typed as `fn: () => void` but down below we call it as `fn(...args);` which is contradictory. In actual implementation we never debounce or throttle a function that does take args but I did not want to fully remove the args param as we should keep it generic and revisit later.

yarn.lock
```diff
@@ -4000,7 +4000,7 @@ flatten@^1.0.2:
   version "1.0.2"
   resolved "https://registry.yarnpkg.com/flatten/-/flatten-1.0.2.tgz#dae46a9d78fbe25292258cc1e780a41d95c03782"
 
-flow-bin@^0.77.0:
```
pfarejowicz: I think this was my bad from #289 which didnt sync up to Github properly

packages/gestalt/src/style.js
```diff
@@ -16,11 +16,13 @@ order in which you do so doesn't really matter.
 
 */
 
+type InlineStyle = { [key: string]: string | number };
+
 // TODO: This type should be opaque, however the Babel parser doesn't support
 //       the opaque syntax yet.
 type Style = {|
   className: Set<string>,
-  inlineStyle: { [key: string]: any },
```
pfarejowicz: We can be more specific with `string | number` here because those are the only two types that work for a style key https://reactjs.org/docs/dom-elements.html#style

packages/gestalt/src/debounce.js
chrislloyd: Can this just be `debounce<T>(fn: T, threshold: number)`?

pfarejowicz:
@chrislloyd had to disable in that one place in `Box.js` because `transforms.js` has no flow types and I need to spend more time picking your brain on the exact typings that go into that file. My todos are exactly that: flow type `transforms.js` and remove the disable, turn on the `sketchy-number` rule, and turn on the `sketchy-null` rule. All three PRs will follow soon 😛

PR #304, opened 2018-08-09
Made the SearchField placeholder text color to contrast its background better

btraut:
We recently darkened the grays of Gestalt to create better contrast as a win for accessibility. Unfortunately, the background of SearchField is still too close to the placeholder text color to achieve a WCAG AA rating.

This PR darkens the placeholder color just for SearchField, something our designers have already approved.

### DIFF
--- a/packages/gestalt/src/SearchField.css
+++ b/packages/gestalt/src/SearchField.css
@@ -17,7 +17,7 @@
 }
 
 .input::placeholder {
-  color: #8e8e8e;
+  color: #6c6c6c;
 }
 
 .input::-webkit-search-decoration,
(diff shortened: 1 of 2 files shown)


christianvuerings:

Built with commit ade69fc063bbd830bb272441bf6dc5e44688cc8f

https://deploy-preview-304--gestalt.netlify.com

packages/gestalt/src/SearchField.css
```diff
@@ -17,7 +17,7 @@
 }
 
 .input::placeholder {
-  color: #8e8e8e;
+  color: #6c6c6c;
```
pfarejowicz: Is this something we should fo for `<TextArea />` and `<TextField />` as well? The placeholder there is also the old gray
```
.textArea::placeholder {
  color: #8e8e8e;
}
```

btraut: Nope. Keeping the old placeholder on those because they have a white background. It's slightly inconsistent, but accessibility is the north star on this decision.

PR #307, opened 2018-08-14
Masonry: MasonryInfinite for infinite fetching

willyau:
- Pulled fetching out from Masonry
- MasornyInfinite is exported as the old Masonry
  and should function the same
- export MasonryBeta as it can be used with custom
  fetching logic
- This provides a graceful migration path.
- Final migration will involve renaming
  MasonryBeta/MasonryInfinite/Masrony
- Existing Masonry integration will be testing MasonryInfinite to ensure
  compatibility

### DIFF
--- a/packages/gestalt/src/Masonry.js
+++ b/packages/gestalt/src/Masonry.js
@@ -28,6 +28,8 @@ type Layout =
   | LegacyMasonryLayout
   | LegacyUniformRowLayout;
 
+export type MeasurementState = 'idle' | 'measuring';
+
 export type Props<T> = {|
   columnWidth?: number,
   comp: React.ComponentType<{
@@ -45,7 +47,7 @@ export type Props<T> = {|
     content: Position,
     viewport: Position
   ) => void,
-  onPendingMeasurementsUpdate?: (hasPendingMeasurements: boolean) => void,
+  onAutoMeasuringUpdate?: (state: MeasurementState) => void,
   layout?: Layout,
   // Support legacy loadItems usage.
   // TODO: Simplify non falsey flowtype.
@@ -310,7 +312,9 @@ export default class Masonry<T> extends React.Component<Props<T>, State<T>> {
     }));
 
     // need to make sure parent component has the correct pending measurement value
-    this.handleOnPendingMeasurementsUpdate(this.state.hasPendingMeasurements);
+    this.handleOnAutoMeasuringUpdate(
+      this.state.hasPendingMeasurements ? 'measuring' : 'idle'
+    );
   }
 
   componentDidUpdate(prevProps: Props<T>, prevState: State<T>) {
@@ -327,9 +331,9 @@ export default class Masonry<T> extends React.Component<Props<T>, State<T>> {
     );
 
     if (hasPendingMeasurements && !prevState.hasPendingMeasurements) {
-      this.handleOnPendingMeasurementsUpdate(true);
+      this.handleOnAutoMeasuringUpdate('measuring');
     } else if (!hasPendingMeasurements && prevState.hasPendingMeasurements) {
-      this.handleOnPendingMeasurementsUpdate(false);
+      this.handleOnAutoMeasuringUpdate('idle');
     }
     this.handleVirtualizationWindowUpdate();
 
@@ -452,9 +456,9 @@ export default class Masonry<T> extends React.Component<Props<T>, State<T>> {
     }
   };
 
-  handleOnPendingMeasurementsUpdate = (hasPendingMeasurements: boolean) => {
-    if (this.props.onPendingMeasurementsUpdate) {
-      this.props.onPendingMeasurementsUpdate(hasPendingMeasurements);
+  handleOnAutoMeasuringUpdate = (state: MeasurementState) => {
+    if (this.props.onAutoMeasuringUpdate) {
+      this.props.onAutoMeasuringUpdate(state);
     }
   };
 
--- a/packages/gestalt/src/Masonry.js
+++ b/packages/gestalt/src/Masonry.js
@@ -34,6 +34,7 @@ export type Layer = {
   width: number,
   height: number,
 };
+type Position = Layer;
 
 export type Props<T> = {|
   columnWidth?: number,
@@ -68,7 +69,12 @@ export type Props<T> = {|
 
 type State<T> = {|
   hasPendingMeasurements: boolean,
+  height: number,
   items: Array<T>,
+  itemsToMeasure: Array<T>,
+  itemsToRender: Array<T>,
+  measuringPositions: Array<Position>,
+  renderPositions: Array<Position>,
   scrollTop: number,
   width: ?number,
 |};
@@ -204,8 +210,13 @@ export default class Masonry<T> extends React.Component<Props<T>, State<T>> {
       hasPendingMeasurements: props.items.some(
         item => !!item && !props.measurementStore.has(item)
       ),
+      height: 0,
+      itemsToRender: [],
+      itemsToMeasure: [],
       // eslint-disable-next-line react/no-unused-state
       items: props.items,
+      measuringPositions: [],
+      renderPositions: [],
       scrollTop: 0,
       width: undefined,
     };
@@ -262,9 +273,15 @@ export default class Masonry<T> extends React.Component<Props<T>, State<T>> {
       prevState.width == null
     ) {
       this.insertAnimationFrame = requestAnimationFrame(() => {
+        const renderLayout = this.calculateRenderLayout();
         this.setState({
           hasPendingMeasurements,
+          ...renderLayout,
         });
+      });
+    } else if (hasPendingMeasurements || prevState.items !== items) {
+      this.insertAnimationFrame = requestAnimationFrame(() => {
+        this.setState({ ...this.calculateRenderLayout() });
       });
     }
   }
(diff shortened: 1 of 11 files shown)


christianvuerings:

Built with commit ee2766ed15669d5598d435df5b4f08338d758a8e

https://deploy-preview-307--gestalt.netlify.com

packages/gestalt/src/MasonryInfinite.js
```diff
@@ -0,0 +1,142 @@
...
+      });
+    }
+  };
+
+  static getDerivedStateFromProps(props: Props<T>, state: State<T>) {
+    const { items } = props;
+
+    // Shallow compare all items, if any change reflow the grid.
+    // checks below are comparable to Masonry.getDerivedStateFromProps
+    for (let i = 0; i < items.length; i += 1) {
```
bishwei: this will likely cause a perf regression since we're going to be doubling up the number of times we compare items in the array. 

i think for the purposes of fetching items, we can assume immutable items so we can change this to
```
if (props.items === state.items) {
  return null;
}
return {
  items,
  isFetching: false
}
```

what do you think?

packages/gestalt/src/MasonryInfinite.js
```diff
@@ -0,0 +1,142 @@
...
+      this.setState(
+        {
+          isFetching: true,
+        },
+        () => loadItems({ from: this.props.items.length })
+      );
+    }
+  };
+
+  handlePendingMeasurements = () => {
```
bishwei: same comment as above: `handlePendingMeasurementsUpdate`

packages/gestalt/src/Masonry.js
```diff
@@ -72,6 +81,13 @@ const VIRTUAL_BUFFER_FACTOR = 0.7;
...
 
+/**
+ * TODO this should be renamed to MasronyBeta or something else so it is clear
+ * that this is not the exported Masonry.
+ *
+ * The goal is to eventually not have any scroll fetching concerns in this component.
+ * The name is kept for now to have an easier time seeing the diffs.
+ */
 export default class Masonry<T> extends React.Component<Props<T>, State<T>> {
   static createMeasurementStore() {
```
bishwei: i think we can kill this since `MasonryInfinite` has it

packages/gestalt/src/Masonry.js
```diff
@@ -219,6 +237,9 @@ export default class Masonry<T> extends React.Component<Props<T>, State<T>> {
       scrollTop,
       width: this.gridWrapper ? this.gridWrapper.clientWidth : prevState.width,
     }));
+
+    this.handleOnLayerUpdate();
```
bishwei: these methods feel weird to call on mount - what was the reasoning for this?

packages/gestalt/src/Masonry.js
```diff
@@ -41,6 +47,10 @@ type Props<T> = {|
   items: Array<T>,
   measurementStore: Cache<T, *>,
   minCols: number,
+  // Content layer and Viewport layer is as defined in Collection.
+  onLayerUpdate?: (contentLayer: Layer, viewportLayer: Layer) => void,
+  onPendingMeasurements?: () => void,
```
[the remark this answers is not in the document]
willyau: done

packages/gestalt/src/Masonry.js
```diff
@@ -525,6 +571,7 @@ export default class Masonry<T> extends React.Component<Props<T>, State<T>> {
       const height = positions.length
         ? Math.max(...positions.map(pos => pos.top + pos.height))
         : 0;
+      this.contentLayer = { top: this.containerOffset, left: 0, height, width };
```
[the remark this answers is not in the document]
willyau: refactored out

packages/gestalt/src/MasonryInfinite.js
willyau: sounds good!

packages/gestalt/src/MasonryInfinite.js
willyau: done

packages/gestalt/src/Masonry.js
```diff
@@ -29,7 +28,15 @@ type Layout =
   | LegacyMasonryLayout
   | LegacyUniformRowLayout;
 
-type Props<T> = {|
+export type Layer = {
```
christianvuerings: Could we use exact flow types? `{|` https://flow.org/en/docs/types/objects/#toc-exact-object-types

willyau: done!

packages/gestalt/src/Masonry.js
```diff
@@ -72,11 +85,14 @@ const VIRTUAL_BUFFER_FACTOR = 0.7;
 
 const layoutNumberToCssDimension = n => (n !== Infinity ? n : undefined);
 
+/**
+ * TODO this should be renamed to MasronyBeta or something else so it is clear
```
chrislloyd: Lets not leave TODO's in the code - either file an issue or fix it :)

packages/gestalt/src/Masonry.js
```diff
@@ -325,18 +361,110 @@ export default class Masonry<T> extends React.Component<Props<T>, State<T>> {
...
+    return {
+      height,
+      itemsToRender,
+      itemsToMeasure,
+      measuringPositions,
+      renderPositions,
+    };
+  };
+
+  handleOnLayerUpdate = () => {
```
chrislloyd: You don't need the `on` here - just `handleLayerUpdate`.

packages/gestalt/src/Masonry.js
```diff
@@ -325,18 +361,110 @@ export default class Masonry<T> extends React.Component<Props<T>, State<T>> {
...
+        left: 0,
+        height,
+        width: width || 0,
+      };
+
+      this.props.onLayerUpdate(contentLayer, viewportLayer);
+    }
+  };
+
+  handleOnPendingMeasurementsUpdate = (hasPendingMeasurements: boolean) => {
```
chrislloyd: You don't need the `on` here - just `handlePendingMeasurementsUpdate `.

packages/gestalt/src/Masonry.js
```diff
@@ -29,7 +28,14 @@ type Layout =
   | LegacyMasonryLayout
   | LegacyUniformRowLayout;
 
-type Props<T> = {|
+export type Layer = {|
```
chrislloyd: I know this is internal, but the naming is a little confusing (seeing as we also have a Layer component in Gestalt). How about reusing `Position` from the layout? They're the same things.

packages/gestalt/src/Masonry.js
```diff
@@ -41,6 +47,9 @@ type Props<T> = {|
   items: Array<T>,
   measurementStore: Cache<T, *>,
   minCols: number,
+  // Content layer and Viewport layer is as defined in Collection.
+  onLayerUpdate?: (contentLayer: Layer, viewportLayer: Layer) => void,
```
chrislloyd: Again, I think the concept of a "layer" here is internal to Masonry and not very obvious from the outside. How about `onVirtualizationUpdate`, `onBoundsUpdate`, `onWindowUpdate`?

chrislloyd: "Layer" type info is duplicated in arg names too. `content: Layer, viewport: Layer`.

packages/gestalt/src/Masonry.js
```diff
@@ -41,6 +47,9 @@ type Props<T> = {|
...
+  onPendingMeasurementsUpdate?: (hasPendingMeasurements: boolean) => void,
```
chrislloyd: This is almost a double negative. It would be made clearer by not using a boolean for the arg and instead thinking of the state being an enum. `onAutoMeasuringUpdate?: (state: 'idle' | 'measuring')`.

packages/gestalt/src/Masonry.js
```diff
@@ -325,18 +361,110 @@ export default class Masonry<T> extends React.Component<Props<T>, State<T>> {
...
-  fetchMore = () => {
-    const { loadItems } = this.props;
-    if (loadItems && typeof loadItems === 'function') {
-      this.setState(
-        {
-          isFetching: true,
-        },
-        () => loadItems({ from: this.props.items.length })
-      );
+  getLayout = () => {
```
chrislloyd: As a style nit - these functions (`getLayout`, `calculateRenderLayout`) should probably sit outside of the class as helper functions. The words "get" (a function that returns something) and "calculate" (a function that returns something) are smells of this to me. Advantages are that dependencies on props vs. state can be clearly factored away.

test/containers/MasonryExample.js
```diff
@@ -57,7 +57,7 @@ export default class MasonryExample extends React.Component {
       window.RESIZE_MEASUREMENT_DONE = false;
       setTimeout(() => {
         const checkIfMeasuring = () => {
-          if (!this.gridRef.state.hasPendingMeasurements) {
+          if (!this.gridRef.hasPendingMeasurements) {
```
chrislloyd: Can this use the public API rather than reading internal measurements now?

packages/gestalt/src/MasonryInfinite.js
```diff
@@ -0,0 +1,127 @@
...
+  };
+
+  reflow = () => {
+    if (this.gridRef.current) {
+      this.gridRef.current.reflow();
+    }
+  };
+
+  // $FlowFixMe - this is the right definition
+  gridRef: React.ElementRef<typeof Masonry<T>>;
```
chrislloyd: Can we use the function API and null check here instead of the Ref box? Better type safety and lifecycle checking. (ElementRef is only for backwards compat with old usages of string refs)

packages/gestalt/src/Masonry.js
willyau: ok. will file an issue ticket!

packages/gestalt/src/Masonry.js
willyau: done. moved to sit outside the class as helpers and renamed.

packages/gestalt/src/Masonry.js
willyau: done. updated name to onVirtualizationWindowUpdate

packages/gestalt/src/Masonry.js
willyau: done. using Position

packages/gestalt/src/Masonry.js
willyau: done

test/containers/MasonryExample.js
willyau: done

packages/gestalt/src/MasonryInfinite.js
willyau: updated to use ref callback!

PR #327, opened 2018-08-31
Add in greenkeeper-lockfile to auto update yarn lockfile on dep updates

pfarejowicz:
Followed the steps outlined here: https://github.com/greenkeeperio/greenkeeper-lockfile#setup

I think this is correct


christianvuerings:

Built with commit 3bb989fc717613b99ed660344763addc88e05428

https://deploy-preview-327--gestalt.netlify.com

christianvuerings:

Built with commit 60bdcab321fd48a8a3d63853835a27709e4d1bb4

https://deploy-preview-327--gestalt.netlify.com

.buildkite/pipeline.yml
```diff
@@ -17,6 +17,16 @@ steps:
       docker-compose:
         run: gestalt
   - wait
+  - name: ":yarn:"
```
chrislloyd: This will run on every PR - where I think you only want it to run on master. Try looking into branch targeting.

package.json
```diff
@@ -60,6 +61,8 @@
     "build": "cd packages/gestalt && yarn build",
     "lint:js": "eslint .",
     "lint:css": "stylelint \"packages/**/*.css\"",
+    "greenkeeper:update": "greenkeeper-lockfile-update",
```
chrislloyd: This seems like unnecessary indirection - why not just call `yarn greenkeeper-lockfile-update`?

.buildkite/pipeline.yml
pfarejowicz: Actually the greenkeeper bot PRs are done in separate branches, not master, so that won't work. I think this is fine because the program has a way of detecting whether it came from the bot and if not then it short circuits like this
image

package.json
pfarejowicz: I was trying to mimic it so it looks like `lint:js` and `lint:css` but i can update

PR #361, opened 2018-09-17
Tabs: Add optional wrap prop

kennandavison:


packages/gestalt/src/Tabs.js
```diff
@@ -53,10 +55,16 @@ export default class Tabs extends React.Component<Props, State> {
...
   render() {
-    const { tabs, activeTabIndex } = this.props;
+    const { tabs, activeTabIndex, wrap } = this.props;
     const { focusedTabIndex, hoveredTabIndex } = this.state;
     return (
-      <div className={styles.Tabs} role="tablist">
+      <div
+        className={classnames(
+          styles.Tabs,
+          wrap !== undefined && styles.TabsWrap
```
pfarejowicz: This logic is wrong, imagine I set `wrap=false`, in that case `false !== undefined` evaluates to `true` so `true && styles.TabsWrap` would set that CSS class on this component when in fact we didn't want it because we passed `false`.

packages/gestalt/src/Tabs.css
```diff
@@ -3,6 +3,10 @@
   user-select: none;
 }
 
+.TabsWrap {
+  composes: flexWrap from "./Layout.css";
```
pfarejowicz: Since this class only composes from another CSS file, can we not duplicate the class and instead import the `Layout.css` file into the `Tabs.js` component and then use it directly there

packages/gestalt/src/Tabs.js
```diff
@@ -53,10 +56,13 @@ export default class Tabs extends React.Component<Props, State> {
   handleTabMouseLeave = () => this.setState({ hoveredTabIndex: undefined });
 
   render() {
-    const { tabs, activeTabIndex } = this.props;
+    const { tabs, activeTabIndex, wrap } = this.props;
     const { focusedTabIndex, hoveredTabIndex } = this.state;
     return (
-      <div className={styles.Tabs} role="tablist">
+      <div
+        className={classnames(styles.Tabs, wrap === true && layout.flexWrap)}
```
pfarejowicz: why not just `wrap && layout.flexWrap` since only a truthy value will hit the second clause and since its typed to a boolean, we know the only truthy value is `true`

docs/src/Tabs.doc.js
```diff
@@ -34,6 +34,12 @@ card(
         type: `({ event: SyntheticMouseEvent<>, activeTabIndex: number }) => void`,
         required: true,
       },
+      {
+        name: 'wrap',
+        type: 'boolean',
+        defaultValue: false,
```
pfarejowicz: we usually only add a `defaultValue` if its explicitly assigned via a react components default props. i would either remove this line here, or add it as a default to the component

chrislloyd:
@kennandavison also just for context - is this meant to wrap or is it meant to change the orientation of 
the tabs? i.e. can two tabs appear on one line with another beneath it? Also, are there any cases where it _shouldn't_ wrap?

kennandavison:
@chrislloyd - The intention of this PR is to allow wrapping when there isn't enough width (too many tabs for window width). With this in mind, yeah, it'll result in tabs creating new lines as needed which is why making wrapping default might break current usages if it's not expecting extra height. I think wrapping is nicer than having things just cut off though. :X

screen shot 2018-09-19 at 6 16 23 pm

PR #402, opened 2018-10-17
Collage should not create a new item component in every re-render

jackhsu978:
Collage items get unmounted and mounted again on every re-render because a new item component is created on every re-render. This should fix it.

### DIFF
--- a/docs/src/Collage.doc.js
+++ b/docs/src/Collage.doc.js
@@ -50,6 +50,15 @@ card(
         required: true,
       },
       {
+        name: 'Item',
+        type: 'React.ComponentType',
+        description: `
+        A React component (or functional component) that renders the item displayed in the collage.
+        This component is passed props: \`width : number\`, \`height: number\`, \`idx: number\`
+        `,
+        required: true,
+      },
+      {
         name: 'layoutKey',
         type: 'number',
         description: `
@@ -57,13 +66,6 @@ card(
         If there are N layouts available, (layoutKey % N) will determine which layout is used.
         `,
         defaultValue: 0,
-      },
-      {
-        name: 'renderImage',
-        type:
-          '({ width: number, height: number, index: number }) => React.Node',
-        description: 'Render prop for the collage images',
-        required: true,
       },
       {
         name: 'width',
@@ -83,7 +85,7 @@ card(
   columns={3}
   height={300}
   width={300}
-  renderImage={({ index, width, height }) => {
+  Item={({ idx, width, height }) => {
     const images = [
       {
         color: 'rgb(111, 91, 77)',
@@ -122,7 +124,7 @@ card(
         src: '${stock6}',
       },
     ];
-    const image = images[index];
+    const image = images[idx];
     return (
       <Mask wash width={width} height={height}>
         <Image
@@ -154,7 +156,7 @@ card(
         columns={columns}
         height={150}
         width={150}
-        renderImage={({ index, width, height }) => {
+        Item={({ idx, width, height }) => {
           const images = [
             {
               color: 'rgb(111, 91, 77)',
@@ -193,7 +195,7 @@ card(
               src: '${stock6}',
             },
           ];
-          const image = images[index];
+          const image = images[idx];
           return (
             <Mask wash width={width} height={height}>
               {image ? (
@@ -233,7 +235,7 @@ card(
     gutter={8}
     height={300}
     width={300}
-    renderImage={({ index, width, height }) => {
+    Item={({ idx, width, height }) => {
       const images = [
         {
           color: 'rgb(111, 91, 77)',
@@ -272,7 +274,7 @@ card(
           src: '${stock6}',
         },
       ];
-      const image = images[index];
+      const image = images[idx];
       return (
         <Mask wash width={width} height={height}>
           <Image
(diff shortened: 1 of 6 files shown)


christianvuerings:

Built with commit 2927cb3f614f5ad19ef9984b2b233659bafa4520

https://deploy-preview-402--gestalt.netlify.com

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  :memo: CHANGELOG - <i>Please add a CHANGELOG.md entry for your changes</i>
  </td>
    </tr>
  </tbody>
</table>


gestalt.js | :small_red_triangle_down: 0.1% | :small_red_triangle_down: 0.1% | 300.99 KB | 300.84 KB | 61.88 KB | 61.85 KB
gestalt.es.js | :small_red_triangle_down: 0.0% | :small_red_triangle_down: 0.1% | 284.65 KB | 284.52 KB | 60.98 KB | 60.95 KB
<p align="right">
</p>

jackhsu978:
@chrislloyd Unless I'm missing something, not making a breaking changes in `<Collection>`, which hasn't been documented and is currently only used by `<Collage>`, means we have to introduce a breaking change in `<Collage>`, which has been documented. Is that preferred? Please take another look at the PR, which avoids breaking change for `<Collection>` but contains breaking changes in `<Collage>`

packages/gestalt/src/Collection.js
```diff
@@ -109,7 +112,7 @@ export default class Collection extends React.PureComponent<Props, void> {
       <div className={layoutStyles.relative} style={{ width, height }}>
         {items.map(({ idx, ...style }) => (
           <div key={idx} className={layoutStyles.absolute} style={style}>
-            <Item idx={idx} />
+            <Item idx={idx} width={style.width} height={style.height} />
```
chrislloyd: I'm confused by this. Width and height were passed into Collection by Collage so I don't think it's necessary to pass it back to Collage's Item.

docs/src/Collage.doc.js
```diff
@@ -58,13 +67,6 @@ card(
         `,
         defaultValue: 0,
       },
-      {
-        name: 'renderImage',
```
chrislloyd: This prop seemed fine - I'm not sure why we needed to get rid of it?

packages/gestalt/src/Collection.js
jackhsu978: Suppose each collection item is rendered with component `<MyCollectionItem>` which takes `width` and `height` as props:
```jsx
function MyCollectionItem(props: { width: number, height: number, idx: number }) {
  ...
}
```

Without this change, `<MyCollection>`, which renders a list of `<MyCollectionItem>` in `<Collection>`, will look like this:

```jsx
function MyCollection(props) {
  const layout = computeLayout(props);
  return (
    <Collection
     layout={layout}
      Item={({ idx }) => (
        <MyCollectionItem
          width={layout[idx].width}
          height={layout[idx].height}
          idx={idx}
        />
      )}
    />
  );
}
```
where the anonymous function effectively creates a new class on every  re-render, causing the unmount and re-mount.

It's possible to avoid this problem by memoizing layout and pass a instance method:
```jsx
class MyCollection extends React.PureComponent {
  layout = computeLayout(this.props);

  renderItem = () => (
    <MyCollectionItem
      width={this.layout[idx].width}
      height={this.layout[idx].height}
      idx={idx}
    />
  );

  render() {
    return <Collection layout={this.layout} Item={this.renderItem} />;
  }
}
```
But it requires keeping `layout` updated when props are updated and doesn't feel as clean

With this change, using a collection without worrying about its items being unmounted/remounted is simpler:
```jsx
function MyCollection(props) {
  return <Collection layout={computeLayout(props)} Item={MyCollectionItem} />;
}
```

chrislloyd: There's a subtle bug that you're overlooking - `Collection` is only doing a shallow comparison on the `layout` array - it's up to the caller to determine when that has changed and re-render. Only `Collage` can know the inputs to `computeLayout` and when those are invalidated.

jackhsu978: While doing perf profiling, I noticed that collage items are being unmounted and remounted on each re-render, and that's the problem I'm trying to solve.

I think the most elegant solution is to pass the React class component directly to the `Item` prop of Collection.
```jsx
function MyCollection(props) {
  return <Collection layout={computeLayout(props)} Item={MyCollectionItem} />;
}
```

To your point, we _can_ also turn the `<Collage>` or `<MyCollection>` into a class component:
```jsx
function getLayoutFromProps(props) {
  return ...;
}

class MyCollection extends React.PureComponent {
  renderItem = ({ idx }) => {
    const { width, height } = getLayoutFromProps(this.props)[idx];
    return (
      <MyCollectionItem
        width={width}
        height={height}
        idx={idx}
      />
    );
  };

  render() {
    const layout = getLayoutFromProps(this.props);
    return <Collection layout={layout} Item={this.renderItem} />;
  }
}
```
with some memoization of the layout computation. I'm probably missing something here, but I don't really know why this is better than passing the react component directly.

chrislloyd:

PR #421, opened 2018-11-21
Remove default value of short for Text leading property

jennyscript:
## TODO

- [ ] Documentation
- [ ] Tests
- [ ] Accessibility checkup


packages/gestalt/src/Text.js
```diff
@@ -61,7 +61,7 @@ export default function Text({
   smSize,
   mdSize,
   lgSize,
-  leading = 'short',
+  leading,
```
pfarejowicz: Can we update the react proptypes down below as well? Noticed they are missing this property

jennyscript: There is a propType specified here https://github.com/pinterest/gestalt/blob/0ee5c40a3f15a3e4c4fa4704f6f61657286c8671/packages/gestalt/src/Text.js#L150 Was there another place missing it?

pfarejowicz: My mistake, the page formatting in the side by side diff view got me

jennyscript:
The Travis CI Branch step appears to be timing out here. Since the Pull Request Travis CI step is passing, I'm going to go ahead and admin merge. This possibly needs a fix in the Travis settings to turn "Build pushed branches" off, but I don't have admin rights to Travis https://docs.travis-ci.com/user/pull-requests/#double-builds-on-pull-requests. @chrislloyd Could you check the settings?

PR #439, opened 2018-12-15
Removes integration tests

chrislloyd:
We now expect that integration tests lie closer to the consuming
libraries. They're still _really_ important, however a lot of these
tests rely on the environment in which they're run. In particular,
Masonry's tests are highly dependent on the actual things you put
inside Masonry.

The tests themselves still exist internally within Pinterest so no
coverage is actually being lost. The upside is that we can
significantly simply the project setup here.

In addition I've:

* Updated Travis to also test against lts/dubnium seeing as
   `yarn install` was throwing errors with it locally
* Updated CRA and a couple of other deps to make sure
   everything works and installs correctly.


package.json
```diff
@@ -14,12 +13,10 @@
   },
   "devDependencies": {
     "babel-core": "^6.26.0",
-    "babel-eslint": "^10.0.1",
+    "babel-eslint": "9.0.0",
```
btraut: You downgraded the version of eslint? Was that intentional?

package.json
```diff
@@ -14,12 +13,10 @@
...
     "babel-jest": "^23.4.0",
```
btraut: Are there any jest files left? I suspect there's more deps we can remove.

package.json
chrislloyd: Yep - `create-react-app@2` now requires exact versions of eslint. Not awesome but also not terrible (not sure what that version got us). For this and a few other reasons I don't actually think CRA is a great fit anymore for the docs so will address in further PRs.

package.json
chrislloyd: There are tonnes of jest files :) Lots of unit tests and snapshots for components (which are great IMHO).

PR #452, opened 2019-01-08
Bump gestalt to 0.86.2

pfarejowicz:
This patch bump includes the fix for CSS modules that was causing issues noticed with `Button`


btraut:
Peter mentioned there may be some yarn lock issues associated with this diff because our computers are producing different lock files. We may need to go back and fix if it turns out to be on my PR's end.

PR #453, opened 2019-01-09
Add title to IconButton

dangerismycat:
This adds a `title` prop to `IconButton`, displaying the `accessibilityLabel` on hover. This helps clarify intent where the icon alone is not obvious.

icon-button-title

- [x] Documentation
- [x] Tests
- [x] Accessibility checkup


docs/src/IconButton.doc.js
```diff
@@ -36,7 +36,7 @@ card(
         type: 'string',
         required: true,
         description:
-          'String that clients such as VoiceOver will read to describe the element. Always localize the label.',
+          'String that clients such as VoiceOver will read to describe the element. Will also be used for hover text (title). Always localize the label.',
```
chrislloyd: Are there any cases where the a11y label and the hover label would be different?

dangerismycat: I was thinking the same. I could see that being the case if the a11y label is too long and presents UI concerns for the title. One possibility would be to default the title to the a11y label, and allow the developer to provide a different title, or indicate "no title" in some way (pass an empty string, null, etc).

Jenny suggested we start with this small change and iterate from there. I'm fine with either approach.

PR #454, opened 2019-01-09
Handle server rendering in Layer

jennyscript:
Currently, `Layer` will break if it's used in server rendering because it uses `document`. This adds existence checks for `document` and just renders the children in that case. It also adds documentation for this behavior and a `console.warn`

## TODO

- [x] Documentation
- [x] Tests
- [x] Accessibility checkup


packages/gestalt/src/Layer.js
```diff
@@ -9,17 +9,24 @@ type Props = {|
...
+      // eslint-disable-next-line no-console
+      console.warn(
+        'Using Layer without document present. Children will be rendered directly.'
+      );
+    }
   }
 
   componentDidMount() {
-    if (document.body) {
+    if (typeof document !== 'undefined' && document.body) {
```
chrislloyd: I think you can avoid the `document` checks here by just testing for `this.el` (though technically Flow should be able to figure this out).

jennyscript:
The main use case I can think of for wanting it to server render is SEO. What if the content includes important keywords? What if it's rendering something like a sidebar in Layer and there are links to other pages that need to get crawled?

Yes, it's possible there would be hydration issues here. Some more discussion: https://github.com/facebook/react/issues/12615

If we decide not rendering server side is fine, I've seen some examples using `componentDidMount` which would look like this:

```
export default class Layer extends React.Component<Props> {
  componentDidMount() {
    if (typeof document !== 'undefined' && document.body) {
      this.el = document.createElement('div');
      document.body.appendChild(this.el);
      this.forceUpdate();
    }
  }

  componentWillUnmount() {
    if (typeof document !== 'undefined' && document.body) {
      document.body.removeChild(this.el);
    }
  }

  el: HTMLDivElement;

  render() {
    if (!this.el) {
      return null;
    }
    const { children } = this.props;
    return createPortal(children, this.el);
  }
}
```

packages/gestalt/src/Layer.js
jennyscript: Flow can't figure it out, sadly.
screen shot 2019-01-09 at 1 39 23 pm

PR #455, opened 2019-01-09
CSS classname minification is now deterministic

btraut:
There was a lot of churn on our snapshots every time we updated the Gestalt version and one or more classnames had been updated. This change to Gestalt minifies classnames in a deterministic manner.

Because two-character classnames result in so many cache collisions, we needed to determine a way to increment the hash values. To preserve these incremented hash values across edits to Gestalt, an on-disk cache is used. Much like `yarn.lock`, this cache will remain stable across re-builds, but may change when editing files or when rebuilding from scratch.


btraut:
Okay, updated it to get rid of the disk cache grossness. Instead, I removed the underscore and just protected against classnames that started with a number or "-". This should have the same file size for CSS bundle, but have 99.99% deterministic classnames through changes to the library.

christianvuerings:
@btraut Looks like the latest build on netlify has some issues: https://deploy-preview-455--gestalt.netlify.com/#/

image

I preferred the starting with `_` to avoid collisions with external CSS code & make it easier to spot what is gestalt and what's not. Were you able to find any 3 character class names in the Pinterest source code?

packages/gestalt/lib/classnameBuilder.js
```diff
@@ -0,0 +1,84 @@
...
+// solved this by assigning each classname an incrementing number and hashing on
+// that number, but this resulted in lots of name/hash changes whenever
+// classnames were added or removed. To solve this, we now hash based on the
+// class name and only add variation when hashes collide.
+
+import md5 from 'blueimp-md5';
+
+class ClassnameBuilder {
+  constructor() {
+    // Map of classnames to their minified versions.
```
pfarejowicz: No change needed but curious if you considered using a `Map` instead - im interested if there was an internal discussion and the trade-offs https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map

btraut: No reason. Just old habits. I'm pretty sure the way I'm using the objects as a Map makes literally 0 difference in speed, but you're right that actually making it a Map reads more explicitly.

PR #456, opened 2019-01-10
Pass accessibilityLabel down as title through Pog to Icon for hover text

dangerismycat:
Adding on to the IconButton `title` PR, this passes IconButton's `accessibilityLabel` as `title` through Pog to Icon, where it is used as the `svg`s `title`. This yields the hover text when hovering over the icon part of IconButton.

- [ ] Documentation
- [x] Tests
- [ ] Accessibility checkup

### DIFF
--- a/docs/src/Icon.doc.js
+++ b/docs/src/Icon.doc.js
@@ -60,7 +60,6 @@ card(
         type: 'string',
         description:
           'Used for hover text in place of accessibilityLabel if provided. Always localize the label.',
-        defaultValue: 'accessibilityLabel',
       },
     ]}
   />
--- a/docs/src/Icon.doc.js
+++ b/docs/src/Icon.doc.js
@@ -55,6 +55,13 @@ card(
         type: `{ __path: string }`,
         description: `When using this prop, make sure that the viewbox around the SVG path is 24x24`,
       },
+      {
+        name: 'title',
+        type: 'string',
+        description:
+          'Used for hover text in place of accessibilityLabel if provided. Always localize the label.',
+        defaultValue: 'accessibilityLabel',
+      },
     ]}
   />
 );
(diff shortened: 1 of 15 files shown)


packages/gestalt/src/Icon.js
```diff
@@ -76,7 +78,7 @@ export default function Icon(props: Props) {
       aria-label={accessibilityLabel}
       role="img"
     >
-      <title>{accessibilityLabel}</title>
+      <title>{accessibilityLabel || title}</title>
```
pfarejowicz: should this be the opposite now? `title || accessibilityLabel`?

packages/gestalt/src/Icon.js
```diff
@@ -28,6 +28,7 @@ type IconProps = {
     | 'white',
   inline?: boolean,
   size?: number | string,
+  title?: string,
```
pfarejowicz: If we are adding this as a prop to `Icon` it will be exposed to users, can we update the docs and the react prop types as well?

dangerismycat: Sure, can do.

packages/gestalt/src/Icon.js
dangerismycat: If we're exposing `title` to users (as indicated in your other comment), then yep, I think that makes sense.

docs/src/Icon.doc.js
```diff
@@ -55,6 +55,13 @@ card(
         type: `{ __path: string }`,
         description: `When using this prop, make sure that the viewbox around the SVG path is 24x24`,
       },
+      {
+        name: 'title',
+        type: 'string',
+        description:
+          'Used for hover text in place of accessibilityLabel if provided. Always localize the label.',
+        defaultValue: 'accessibilityLabel',
```
pfarejowicz: `defaultValue` is for the default value of the prop if not passed. we should remove this since there is none

PR #459, opened 2019-01-12
Add Badge component

rootvar:
Add Badge component to make an existing usage pattern more consistent. 
PDS team suggested limiting color options to a single gray color.

- [x] Documentation
- [x] Tests
- [x] Accessibility checkup

badge


packages/gestalt/src/Badge.js
```diff
@@ -0,0 +1,22 @@
...
+type Props = {|
+  size?: 'sm' | 'md' | 'lg',
+  text: string,
+|};
+
+export default function Badge(props: Props) {
+  const { size = 'sm', text } = props;
+  const cs = cx(styles.Badge, styles[size]);
+
+  return <span className={cs}>{text}</span>;
```
pfarejowicz: Is there a way to achieve this using `Box` and `Text`? I think as we grow our library we should reuse as much as possible so changes to one attribute system-wide, say font-size, are reflected everywhere

yarn.lock
```diff
@@ -9954,6 +9959,15 @@ postcss-selector-parser@^3.0.0, postcss-selector-parser@^3.1.0:
     indexes-of "^1.0.1"
     uniq "^1.0.1"
 
+postcss-selector-parser@^5.0.0-rc.3:
```
pfarejowicz: @btraut these look like the yarn lock file changes I was talking about when I was trying to update gestalt. It seems others are also experiencing them

packages/gestalt/src/Badge.css
```diff
@@ -0,0 +1,24 @@
+.Badge {
+  composes: antialiased sansSerif letterSpacing from "./Typography.css";
+  composes: fontWeightBold from "./Typography.css";
+  composes: gray from "./Colors.css";
```
pfarejowicz: Nothing wrong with just gray but Im curious why design suggested this since we use other colors in the webapp already. Is the plan to move all those to gray as well? Was this a PDS recommendation?
image

rootvar: My original intention was to add more colors (mimicking <Text /> component) but after talking to PDS (Collin and Scott) we decided to use only gray color. One of the biggest reason is that we (PDS/Gestalt) don't have a clear and consistent guidelines for what meaning colors convey. Orange is being used for errors and warnings, green (e.g. Early access) and blue are used for info, success state is rarely (if ever?) conveyed through color. (This is something I would like to discuss in more detail in the future.)
Another reason is that badges already have treatment (rounded border) and don't need an additional color treatment.
As usual, we can start with just gray and add more colors later?

packages/gestalt/src/Badge.js
rootvar: We can get close but not completely match this design. In most use cases `Badge` font size has to be 1 size smaller that the parent (because of additional visual treatment). So `Badge` `sm` is the same as `Text` `xs`. So we could either have more consistency between `Badge` and `Text` (but not `Heading`)—just not sure how useful that would be.
Also `Box` border-radius is larger than `Badge`, so `Box` is not fully compatible with this component.

packages/gestalt/src/Badge.css
```diff
@@ -0,0 +1,24 @@
...
+  composes: inlineBlock from "./Layout.css";
+  border: 1px solid #8e8e8e;
+  border-radius: 4px;
+  margin-left: 4px;
```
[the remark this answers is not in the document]
rootvar: Yes I agree with that, I matched the design originally but this is not a good practice.

packages/gestalt/src/Badge.css
collino: Thanks! As @rootvar mentioned, we felt like the original color choices weren't intentional. If anything one could argue that our blue used for education would be fitting but for now secondary gray wasn't controversial so we went with that.

yarn.lock
btraut: Gotchya. I have no idea why I don't see the issue, but if this fixes it, I'm happy you updated it.

packages/gestalt/src/Badge.js
rootvar: @btraut that's a good point and it's highlighting a possible problem with this component in general. If we want to match text's line height, then the badge has to be much smaller—and less effective. If we want to keep it bigger, it will be larger than text's line height and will overlap with text above/below.
I don't necessarily think it's a problem though. Depends what we want this component to be. Currently, we don't have any use cases where it's used in a multi-line text block. It's only used in menus and short headers. If moving forward we want it to be used differently, then we might want to use a different visual element. Discussing with PDS!

packages/gestalt/src/Badge.css
```diff
@@ -0,0 +1,23 @@
+.Badge {
+  composes: antialiased sansSerif letterSpacing from "./Typography.css";
+  composes: fontWeightBold from "./Typography.css";
+  composes: gray from "./Colors.css";
+  composes: inlineBlock from "./Layout.css";
+  border: 1px solid #8e8e8e;
```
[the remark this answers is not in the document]
rootvar: @btraut we set a variable in `Colors.css` already with this value (`--gray: #8e8e8e;`). Do you mean adding a style for a border with this color?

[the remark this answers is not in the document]
btraut: Hmm, I guess we don't have any border definitions in there right now, and they seem to be scattered all over. Just seemed strange to pull in a named color (`gray`) and then immediately use a hard-coded value after. Probably fine to keep this way.

PR #460, opened 2019-01-14
Add Pinterest p verified icon to Avatar (#455)

jcpark89:
## TODO

- [ ] Documentation
- [ ] Tests
- [ ] Accessibility checkup


jcpark89:
screen shot 2019-01-13 at 9 58 30 pm

CHANGELOG.md
```diff
@@ -57,6 +58,7 @@
 - Masonry: Add a onFinishedRendering prop which allows for better testing (#423)
 - Tabs: (Update to #368) Use composite of index and href for tab keys (#419)
 - Text: Default value for leading is now undefined to allow browser to determine line-height based on language (#421)
+- Avatar: Add property to use pinterest icon for verified (#422)
```
pfarejowicz: we should remove this line

packages/gestalt/src/Avatar.js
```diff
@@ -60,6 +60,7 @@ type AvatarProps = {|
   size?: 'sm' | 'md' | 'lg',
   src?: string,
   verified?: boolean,
+  icon?: string,
```
pfarejowicz: why is `icon` a string if down below you force it to be either `pinterest` or `check-circle`?

jcpark89: @pfarejowicz  it was for test that was breaking initially but I changed it back to `'check-circle' | 'pinterest'`

packages/gestalt/src/Avatar.js
```diff
@@ -84,10 +85,12 @@ export default class Avatar extends React.PureComponent<AvatarProps, State> {
   handleImageError = () => this.setState({ isImageLoaded: false });
 
   render() {
-    const { name, outline, size, src, verified } = this.props;
+    const { name, outline, size, src, verified, icon } = this.props;
     const { isImageLoaded } = this.state;
     const width = size ? sizes[size] : '100%';
     const height = size ? sizes[size] : '';
+    const verifiedIcon = icon && icon === 'pinterest' ? icon : 'check-circle';
```
pfarejowicz: i dont think you need this logic, you can just have `icon={icon}` down below then set `check-circle` as the default prop for the icon value since that prop is only ever used if verified is true anyways.

another route is just `icon || 'check-circle'` but i prefer the first way of default props

packages/gestalt/src/Avatar.js
```diff
@@ -60,6 +60,7 @@ type AvatarProps = {|
   size?: 'sm' | 'md' | 'lg',
   src?: string,
   verified?: boolean,
+  icon?: 'check-circle' | 'pinterest',
```
pfarejowicz: One more thing - can we update `Avatar.doc.js` to mention this new prop and to provide an example please

docs/src/Avatar.doc.js
```diff
@@ -45,6 +45,11 @@ card(
         type: 'boolean',
         defaultValue: false,
       },
+      {
+        name: 'icon',
+        type: '"check-circle" | "pinterest"',
+        defaultValue: false,
```
pfarejowicz: default value should be `check-circle` here and i would add a `description` that says "This is the icon shown only when `verified=true`"

PR #464, opened 2019-01-23
Video: add children prop

loserford:
This PR will add the `children` prop to `<Video />`, like it in `<Image />`. With this prop, we can add overlays or other UI elements on top of the video.

we can use it like
```
<Video ...>
  <Box position="absolute" top bottom left right 
    display="flex"
    justifyContent="center"
    alignItems="center"
    dangerouslySetInlineStyle={{__style:{backgroundColor:'rgba(0, 0, 0, 0.3)'}}}>
      <IconButton       
        accessibilityLabel="Love"
        bgColor="white"
        icon="trash-can"
        size="lg" />
  </Box> 
</Video>
```


packages/gestalt/src/Video.js
```diff
@@ -477,6 +483,12 @@ export default class Video extends React.PureComponent<Props, State> {
             ))}
           <track kind="captions" src={captions} />
         </video>
+        {overlay && (
```
pfarejowicz: So instead of all this custom HTML and stylings, I think im in favor of having either the `overlay` prop be of type `React.Node` so the user gets to choose what to show in this middle layer, or to utilize the `Video` component `children` for this purpose since that is not used right now.

I think long term this is a good option as it will open up the door for more overlaid UI from other teams such as post-complete video actions like follow buttons, next video thumbnails, or things like overlaid buttons for specific parts of a video clip.

loserford: It's a great idea. PR updated as you suggested.

docs/src/Video.doc.js
```diff
@@ -277,6 +281,42 @@ card(
   />
 );
 
+card(
+  <Example
+    name="Video with children"
+    description={`
```
pfarejowicz: can we add a line here to make it clear that `Video` children are _not_ the same as `video` children https://developer.mozilla.org/en-US/docs/Web/HTML/Element/video which expects only `source` or `track` elements. In this gestalt version the children actually refers to overlays "outside" of the html video element in the code - i think this would make it less error prone to users

loserford: will add it on both prop description and example description.

docs/src/Video.doc.js
```diff
@@ -68,6 +68,12 @@ card(
         description: 'The URL of the captions track for the video (.vtt file)',
         required: true,
       },
+      {
+        name: 'children',
+        type: 'React.Node',
+        description: `This \`children\` prop is not same as children inside the native html \`video\` element.
+          It will add overlays on top of the html video element like overlays, while under controls if they are shown.`,
```
pfarejowicz: small nit: "It will add overlays on top of the html video element like overlays, while under controls if they are shown." doesnt sound right, maybe "Instead it serves to add overlays on top of the html video element, while still being under the video controls."

loserford: thanks! Will update. Please always correct my english writing!

PR #469, opened 2019-01-26
Make <PropTable/> in docs link to <Example/>'s

mminute:
I often find myself looking at the table of props for each component, then scrolling down to find an example usage for that prop.  This adds the ability to make the name of a prop a link which scrolls to the relevant example on the page.

This is a little hacky atm.  I tried just wrapping the name of the prop in a `<Link href="#someId"/>` but I end up getting navigated to `/#/someId` instead of scrolling to the element with associated id.

scrollto

### DIFF
--- a/docs/src/Text.doc.js
+++ b/docs/src/Text.doc.js
@@ -21,11 +21,13 @@ card(
         name: 'align',
         type: `"left" | "right" | "center" | "justify"`,
         defaultValue: 'left',
+        id: 'align',
       },
       {
         name: 'bold',
         type: 'boolean',
         defaultValue: false,
+        id: 'styles',
       },
       {
         name: 'children',
@@ -35,11 +37,13 @@ card(
         name: 'color',
         type: `"blue" | "darkGray" | "eggplant" | "gray" | "green" | "lightGray" | "maroon" | "midnight" | "navy" | "olive" | "orange" | "orchid" | "pine" | "purple" | "red" | "watermelon" | "white"`,
         defaultValue: 'darkGray',
+        id: 'color',
       },
       {
         name: 'inline',
         type: 'boolean',
         defaultValue: false,
+        id: 'inline',
       },
       {
         name: 'leading',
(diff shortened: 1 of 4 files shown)


docs/src/Text.doc.js
```diff
@@ -21,11 +21,13 @@ card(
         name: 'align',
         type: `"left" | "right" | "center" | "justify"`,
         defaultValue: 'left',
+        id: 'align',
       },
       {
         name: 'bold',
         type: 'boolean',
         defaultValue: false,
+        id: 'styles',
```
pfarejowicz: since each prop name has to be unique on the component, maybe we should just reuse the name field for the id as well all around - this `bold` to `styles` naming is a bit confusing at first glance

docs/src/components/Card.js
```diff
@@ -15,11 +16,12 @@ export default function Card({
   children,
   description,
   heading = true,
+  id,
   name,
   stacked = false,
 }: Props) {
   return (
-    <Box>
+    <Box {...(id ? { id } : {})}>
```
pfarejowicz: `<Box id={id}` >` should fine since if its undefined itll just skip it in the DOM?

docs/src/components/PropTable.js
```diff
@@ -126,7 +135,22 @@ export default function PropTable({ props: properties, Component }: Props) {
...
                         <Text overflow="normal" bold leading="tall">
-                          <code>{name}</code>
+                          {id ? (
+                            <Link
+                              href={`#${id}`}
+                              onClick={({ event }) => {
+                                event.preventDefault();
+                                const elem = document.getElementById(id);
+                                if (elem) {
+                                  elem.scrollIntoView({ behavior: 'smooth' });
```
christianvuerings: @mminute Could we also add `block: 'start'`? It was confusing to me that the headings didn't appear on the top.

docs/src/Text.doc.js
mminute: @pfarejowicz I elected a separate prop for two reasons.
1st, it made it easier to only wrap prop that led to examples in a `<Link/>`
2nd, in the case of the `<Text/>` docs both bold and italic lead to the same example so using name didnt seem quite right.  Watcha think?

PR #471, opened 2019-01-28
Light Masonry cleaup

btraut:
This PR has a couple small parts, mostly cosmetic:

1. Cleaned up a few places in Masonry that mixed destructuring and direct access of props.
2. Used React.createRef over ref functions.
3. Fixed a bug where the `scrollContainer` prop was being evaluated as a bool, although it was possible that it was actually a function that returned `undefined` and thus should be considered falsy in that case.


CHANGELOG.md
```diff
@@ -10,6 +10,8 @@
 
 ### Patch
 
+- Masonry: Small cosmetic refactoring and fixed a tiny edge-case bug with `scrollContainer` prop (#471)
```
pfarejowicz: should this go under `minor`?

packages/gestalt/src/Masonry.js
```diff
@@ -322,22 +328,14 @@ export default class Masonry<T> extends React.Component<Props<T>, State<T>> {
...
   fetchMore = () => {
-    const { loadItems } = this.props;
+    const { loadItems, items } = this.props;
     if (loadItems && typeof loadItems === 'function') {
       this.setState(
         {
           isFetching: true,
         },
-        () => loadItems({ from: this.props.items.length })
+        () => loadItems({ from: items.length })
```
pfarejowicz: is this change ok? the callback here is executed asynchronously after the state update but by pulling out `items` into a variable in this closure we are setting a pointer to it before the function is executed. tl;dr would `this.props.items` ever change reference by the time its used inside this set state callback?

packages/gestalt/src/Masonry.js
```diff
@@ -186,6 +186,10 @@ export default class Masonry<T> extends React.Component<Props<T>, State<T>> {
     virtualize: false,
   };
 
+  gridWrapper: { current: HTMLElement | null } = React.createRef();
```
pfarejowicz: should we clean this up to be a `HTMLDivElement` since we are here?

CHANGELOG.md
btraut: I guess the bug fix portion pushes it over. The cosmetic definitely wouldn't.

packages/gestalt/src/Masonry.js
btraut: We don't actually care if it's a `div` or `span` or whatever in this case, as long as it exposes a `clientWidth`. I'd rather pick the most ambiguous type that still accomplishes 100% type safety.

packages/gestalt/src/Masonry.js
btraut: I waffled on this a bit when I wrote it, but I'm pretty sure it's right. Not only is it cosmetically better, but it actually conveys the intent better as well, and I'm pretty confident that the edge case would never be exposed in our codebase, or if it was, it would have no adverse effects.

pfarejowicz: 👍

packages/gestalt/src/Masonry.js
pfarejowicz: Hmm, Id probably pick the most specific type that still covers all cases but non-blocker

PR #473, opened 2019-01-31
<SegmentedControl /> should have equal item widths by default

jackhsu978:
Per the [design spec](https://design.pinadmin.com/pds/#segmented-controls), the width of each item in a segmented control should be the same, rather than based on what's inside of the segmented control item.

screenshot 2019-01-30 22 45 25

~I added an optional prop `equalWidths`. If it's true, the segmented control items would have equal widths.~
**UPDATE**: Before this PR, the items in `<SegmentedControl>` have width responsive to their content. After this PR is merged, the items in `<SegmentedControl>` will have equal width by default and will be adhere the design spec. To use the old behavior, pass the optional prop `responsive`.

~I am including an example in the doc that shows the side-by-side comparison between responsive width (default) and the equal/fixed widths.~
**UPDATE**: I am including an example in the doc that shows the side-by-side comparison between the equal width (now the default) and the responsive width (the old default)
screenshot 2019-01-30 22 32 19

## TODO
- [x] Documentation
- [x] Tests
- [x] Codemode

### DIFF
--- a/packages/gestalt/src/SegmentedControl.js
+++ b/packages/gestalt/src/SegmentedControl.js
@@ -7,6 +7,7 @@ import Text from './Text.js';
 import styles from './SegmentedControl.css';
 
 type Props = {|
+  equalWidths?: boolean,
   items: Array<React.Node>,
   onChange: ({ event: SyntheticMouseEvent<>, activeIndex: number }) => void,
   selectedItemIndex: number,
@@ -14,7 +15,16 @@ type Props = {|
 |};
 
 export default function SegmentedControl(props: Props) {
-  const { items, onChange, selectedItemIndex, size = 'md' } = props;
+  const {
+    equalWidths,
+    items,
+    onChange,
+    selectedItemIndex,
+    size = 'md',
+  } = props;
+  const buttonStyle = equalWidths
+    ? { width: `${Math.floor(100 / Math.max(1, items.length))}%` }
+    : undefined;
   return (
     <div
       className={classnames(styles.SegmentedControl, {
@@ -37,6 +47,7 @@ export default function SegmentedControl(props: Props) {
             onClick={e => onChange({ event: e, activeIndex: i })}
             role="tab"
             type="button"
+            style={buttonStyle}
           >
             {typeof item === 'string' ? (
               <Text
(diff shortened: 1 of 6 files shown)


jackhsu978:
> Hmm if its spec from PDS I am actually for a more disruptive option that is to make the equal widths the default. We can either 1) let the change propogate through our codebase and make it adhere to pds or 2) we can codemod so that all current uses of `SegmentedControl` get `equalWidths=false` and moving forward new uses get the correct stylings

@pfarejowicz IIUC, for the 2nd option, `equalWidths` prop should have default `true`. Is that right?

pfarejowicz:
> > Hmm if its spec from PDS I am actually for a more disruptive option that is to make the equal widths the default. We can either 1) let the change propogate through our codebase and make it adhere to pds or 2) we can codemod so that all current uses of `SegmentedControl` get `equalWidths=false` and moving forward new uses get the correct stylings
> 
> @pfarejowicz IIUC, for the 2nd option, `equalWidths` prop should have default `true`. Is that right?

@jackhsu978 yes - based on that PDS description you posted that should be the default (we can either codemod webapp to set the field to false or just let the change take effect since it should be following PDS anyways)

packages/gestalt/src/SegmentedControl.js
```diff
@@ -9,12 +9,16 @@ import styles from './SegmentedControl.css';
...
   selectedItemIndex: number,
   size?: 'md' | 'lg',
 |};
 
 export default function SegmentedControl(props: Props) {
-  const { items, onChange, selectedItemIndex, size = 'md' } = props;
+  const { items, onChange, responsive, selectedItemIndex, size = 'md' } = props;
+  const buttonStyle = responsive
+    ? undefined
+    : { width: `${Math.floor(100 / Math.max(1, items.length))}%` };
```
pfarejowicz: i would make `style` always present on the div and instead move the ternary to `width: responsive ? undefined : '...'`

jackhsu978: Good catch. FWIW, the reason I did this is because I wanted to keep the default the same and all the snapshots unchanged, but this is no longer relevant. Will update.

PR #476, opened 2019-02-14
Update <Button/> border radius

mminute:
Design has updated their standard regarding button border radius
    
When a button is inline render it with a border radius of 8px
When a button is full width render it with a border radius of 12px

image

- [N/A] Documentation
- [✔️] Tests
- [N/A] Experimental evidence (required for Masonry changes)
- [ ] Accessibility checkup

### DIFF
--- a/packages/gestalt/src/Button.css
+++ b/packages/gestalt/src/Button.css
@@ -8,7 +8,6 @@
 
 .button {
   composes: borderBox from "./Layout.css";
-  border-radius: 4px;
 }
 
 .solid {
--- a/packages/gestalt/src/Button.js
+++ b/packages/gestalt/src/Button.js
@@ -2,6 +2,7 @@
 import * as React from 'react';
 import PropTypes from 'prop-types';
 import classnames from 'classnames';
+import borderStyles from './Borders.css';
 import styles from './Button.css';
 import Text from './Text.js';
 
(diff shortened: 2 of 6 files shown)


packages/gestalt/src/Button.css
```diff
@@ -8,7 +8,14 @@
 
 .button {
   composes: borderBox from "./Layout.css";
-  border-radius: 4px;
+}
+
+.borderRadiusLarge {
```
pfarejowicz: Lets add these to the Borders.css file innstead
https://github.com/pinterest/gestalt/blob/a3b08fa9bd4d0eef413d9671301a22594e440cd5/packages/gestalt/src/Borders.css

packages/gestalt/src/Button.js
```diff
@@ -42,7 +42,9 @@ export default function Button(props: Props) {
     white: 'darkGray',
   };
 
-  const classes = classnames(styles.button, {
+  const borderRadius = `borderRadius${inline ? 'Small' : 'Large'}`;
```
pfarejowicz: This feels hacky, can we instead import the Borders.css file and add both the small and large class names to the object down below insde of `classname`. Then for one you use `inline` and for the other use `!inline`

mminute: I did that as a first pass, not all that crazy about it.  Having one thing set to `inline` and the other set to `!inline` felt a little weird to me.  Also wondering what would happen if PDS added another border-radius requirement, something like borderRadiusMedium

pfarejowicz: If pds adds a third option then both of our methods fail. The string templating just feels very off for creating class names. If you dont like the `inline` `!inline` thing then there is another option of adding a param to `classnames` that looks like `inline ? borders.small : borders.large` since it can take infinite params

mminute: right, but in one version there is 3 places to update, while in the other the logic is all in one place.  I'm fine adding inline/!inline for now though

PR #479, opened 2019-02-16
[dep] flow 0.84.0

jackhsu978:
Upgrade flow to 0.84.0 with flow error fixes


jackhsu978:
Please ignore the misleading origin branch name. I want to rename it, but it seems like renaming it would mean I have to close this PR and create a new one.

PR #486, opened 2019-02-23
Display errors inline for form components

dangerismycat:
We've run into many issues with the current usage of Flyout to display error messages from design and engineering perspective. This PR inlines those errors for TextField, TextArea, and SelectList. While present, the error message will be displayed regardless of focus. Also included are minor styling tweaks related to error states.

- [x] Documentation
- [x] Tests
- [ ] Accessibility checkup


packages/gestalt/src/ErrorMessage.js
```diff
@@ -0,0 +1,27 @@
...
+|};
+
+export default function ErrorMessage({ errorMessage, id }: Props) {
+  return errorMessage ? (
+    <Box marginTop={1}>
+      <Text color="orange">
+        <span id={formatErrorId(id)}>{errorMessage}</span>
+      </Text>
+    </Box>
+  ) : null;
```
pfarejowicz: Is this what design wanted? rendering `null` when there is no error will cause jitters in the UI when the error appears because itll occupy new space, that feels off to me

packages/gestalt/src/formatErrorId.js
```diff
@@ -0,0 +1,5 @@
+// @flow
+
+export default function formatErrorId(id: string): string {
```
pfarejowicz: the overhead and the indirection of this helper method in my opinion is a bit confusing, I would prefer to just keep the string inline

packages/gestalt/src/ErrorMessage.js
dangerismycat: I think I was being overly defensive, and I don't feel strongly about it. I'll get rid of the ternary.

packages/gestalt/src/formatErrorId.js
dangerismycat: True. I abstracted it out so that if we decide to change it, we only have to change it in one place instead of many.

chrislloyd: It's probably also not to expose the word "gestalt" here - it'll show up in anchors etc. It's probably better served localized to a component or explicitly passed in by the user.

docs/src/TextArea.doc.js
```diff
@@ -32,12 +32,6 @@ card(
         type: 'string',
         required: true,
       },
-      {
-        name: 'idealErrorDirection',
```
chrislloyd: What's the migration strategy for removing these props?

packages/gestalt/src/SelectList.js
```diff
@@ -115,77 +92,63 @@ export default class SelectList extends React.Component<Props, State> {
...
-                  <span id={`${id}-gestalt-error`}>{errorMessage}</span>
-                </Text>
-              </Box>
-            </Flyout>
-          )}
+            ))}
+          </select>
+        </Box>
+
+        {errorMessage && <ErrorMessage errorMessage={errorMessage} id={id} />}
```
chrislloyd: This seems a little funny to me. `<ErrorMessage>` isn't composable because it has the `marginTop` (that assumes the usage next to the calling component. If we removed that out to the SelectList then ErrorMessage really just becomes text colored orange. The value of the abstraction breaks down a bit then.

An alternative design is to have ErrorMessage take a prop which is the element to wrap:

```jsx
<ErrorMessage field={<SelectList /> message="asdf" />
```

I like this because a.) it removes an unnessecary `<Box>` on elements when they're not erroring b.) consumers can choose _how_ to format their error messages without them always being below. This approach is similar to experimental "Field" approaches we've seen around the place and helps clarify the naming (`ErrorMessage` is vague yet the component is specific).

packages/gestalt/src/ErrorMessage.js
```diff
@@ -0,0 +1,27 @@
...
+}
+
+ErrorMessage.propTypes = {
+  errorMessage: PropTypes.string,
```
chrislloyd: No need to duplicate the name here - `text`, `message` would be good. Also just using `children` is a good way of not duplicating the name.

docs/src/TextArea.doc.js
dangerismycat: Not sure. What would you recommend? The removed prop refers to a feature we no longer offer (direction of error), so they seem superfluous but not harmful?

packages/gestalt/src/SelectList.js
dangerismycat: Does changing to a more specific `FormErrorMessage` component resolve this? If the component is indicating that's it's pretty specifically to be used as an error message for forms, that makes the `marginTop` seem less awkward…and this is intended to be a purely "internal" component, not one exposed to users. I get the flexibility of an error component that wraps its `children` and can be arbitrarily styled, but that seems beyond the scope of this change.

chrislloyd: IMHO one of the benefits of the previous implementation was that the error message didn't change the field's layout. It was always the same. Now, if you want to color the field yellow, the layout will change and a message is *required*. A good example of where this mightn't work is in top banners where it might be more convenient to display an error message side by side with a field. To reduce the scope of this change the "errorMessage" part of the component could be deprecated and the yellow text added to experimental fields. Another stage would to be to take those fields and make them less experimental :)

dangerismycat: Good point, and probably something we should / may need to address in the future. For the sake of expediency, I'm going to modify the existing paradigm rather than introduce a new one.

You're definitely right about moving the Box with marginTop out of FormErrorMessage.

docs/src/TextArea.doc.js
pfarejowicz: @dangerismycat in reference to @chrislloyd's point can we include a codemod with this PR (or a quick followup) that will remove the prop from our codebase? if we dont then this bump will cause flow errors everywhere the component prop is used

dangerismycat: 👍

PR #494, opened 2019-03-12
Introduce willChangeTransform as a prop for Mask.js

aerozen3:
Introduce new boolean prop willChangeTransform for Mask.js which allows for toggling on and off `will-change: transform` css

## TODO

- [x] Documentation
- [x] Tests
- [ ] Accessibility checkup

### DIFF
--- a/packages/gestalt/src/Mask.css
+++ b/packages/gestalt/src/Mask.css
@@ -1,7 +1,9 @@
 .Mask {
   composes: overflowHidden from "./Layout.css";
   composes: relative from "./Layout.css";
+}
 
+.hintBrowser {
   /* ⬇ https://github.com/pinterest/gestalt/pull/965 */
   will-change: transform;
 }
--- a/packages/gestalt/src/Mask.css
+++ b/packages/gestalt/src/Mask.css
@@ -3,7 +3,7 @@
   composes: relative from "./Layout.css";
 }
 
-.hintBrowser {
+.willChangeTransform {
   /* ⬇ https://github.com/pinterest/gestalt/pull/965 */
   will-change: transform;
 }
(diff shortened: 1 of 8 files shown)


aerozen3:
This PR attempts to fix a Safari full screen video only issue.  See the screenshots here

with `will-change: transform`:with will change transform

without `will-change: transform`:no will change transform

packages/gestalt/src/__snapshots__/Avatar.test.js.snap
```diff
@@ -110,7 +110,7 @@ exports[`Avatar renders the Pinterest P on verified 1`] = `
   }
 >
   <div
-    className="Mask circle"
+    className="Mask circle hintBrowser"
```
aerozen3: hmm not crazy that we're introducing this classname in all the tests, maybe there's another way?  Ultimate I need to turn off `will-change: transform` for Safari full screen, but there's not an easy way to use css selectors for Safari: https://stackoverflow.com/questions/16348489/is-there-a-css-hack-for-safari-only-not-chrome.  And the element that is full screen is a child of `Mask`

btraut: It's totally fine that it's added to all the snapshots.

packages/gestalt/src/Mask.css
```diff
@@ -1,7 +1,9 @@
 .Mask {
   composes: overflowHidden from "./Layout.css";
   composes: relative from "./Layout.css";
+}
 
+.hintBrowser {
```
pfarejowicz: +1 to @btraut comment, this naming feels off

aerozen3:
I got the name looking at the example in MDN docs here: https://developer.mozilla.org/en-US/docs/Web/CSS/will-change, but thinking about it some more, I agree `willChangeTransform` is probably clearer, updating it

aerozen3:
added to the change log and docs:
Screen Shot 2019-03-12 at 4 50 43 PM

PR #496, opened 2019-03-13
Add some more text related icons

loserford:
PD-673, add below text related icons

- text-all-caps
- text-line-height
- text-sentence-case
- text-size
- text-spacing


pfarejowicz:
image

good to go!

PR #500, opened 2019-03-18
Make Modal heading an overloaded prop to accept react elements

pfarejowicz:
This opens the Modal `heading` prop up for more extensibility similar to the `footer` prop

### DIFF
--- a/docs/src/Modal.doc.js
+++ b/docs/src/Modal.doc.js
@@ -45,9 +45,9 @@ card(
       },
       {
         name: 'heading',
-        type: `string`,
+        type: `React.Node`,
         required: true,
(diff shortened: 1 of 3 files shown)


docs/src/Modal.doc.js
```diff
@@ -45,9 +45,9 @@ card(
       },
       {
         name: 'heading',
-        type: `string`,
+        type: `React.Node`,
```
btraut: I get that `React.Node` includes `string`, but because we treat strings as a special type, I'd prefer if the type here was `React.Node | string`

pfarejowicz: Makes sense and I agree - updated!

PR #501, opened 2019-03-20
Tooltip reincarnation within Gestalt

pfarejowicz:
Following the design docs from PDS to add these back into our library but this time in a more opinionated way.

- [x] Documentation
pinch/tooltips
- [x] Tests
Tooltip.jsdom.test.js
- [x] Accessibility checkup
Tested with ChromeVox to make sure the screenreader is capturing the title text in the tooltip
Followed the guide here https://inclusive-components.design/tooltips-toggletips/


packages/gestalt/src/Contents.js
```diff
@@ -52,8 +52,10 @@ type Shift = { x: number, y: number };
 
 type EdgeShift = { caret: Shift, flyout: Shift };
 
+/* eslint-disable react/no-unused-prop-types */
```
btraut: What's up with this? It looks like all of these props are used.

packages/gestalt/src/Controller.js
```diff
@@ -5,14 +5,15 @@ import Contents from './Contents.js';
...
+  anchor: HTMLElement,
   bgColor: 'blue' | 'darkGray' | 'orange' | 'white',
+  caret?: boolean,
   children?: React.Node,
   idealDirection?: 'up' | 'right' | 'down' | 'left',
   onDismiss: () => void,
   positionRelativeToAnchor: boolean,
   shouldFocus?: boolean,
-  size?: 'xs' | 'sm' | 'md' | 'lg' | 'xl' | number,
+  size?: 'xs' | 'sm' | 'md' | 'lg' | 'xl' | number | null,
```
btraut: Size should be non-optional if you're specifying a `defaultProps`.

packages/gestalt/src/Contents.js
pfarejowicz: Not sure, Flow/eslnt are having a moment, it was previously on line 487 so I just moved it up, it thinks idealDirection is not used because of the way we pass `this.props` I think

packages/gestalt/src/Tooltip.js
```diff
@@ -0,0 +1,81 @@
...
+          onBlur={this.handleBlur}
+          onFocus={this.handleFocus}
+          onMouseEnter={this.handleMouseEnter}
+          onMouseLeave={this.handleMouseLeave}
+          ref={this.childRef}
+        >
+          {children}
+        </Box>
+        {(hovered || focused) && (
+          <Popout anchor={this.childRef.current} text={text} />
```
btraut: The first time this Tooltip is rendered, `this.childRef.current` is null and even though it's immediately set, it won't trigger another re-render. I get that it's a super rare edge case because we really only need the ref after focus/hover, but it's something that you should consider when designing these sorts of interactions.

packages/gestalt/src/Controller.js
pfarejowicz: Good catch, changed

packages/gestalt/src/Tooltip.js
pfarejowicz: I agree and I followed current implementation practices with Flyout within Tooltip. The popup mini-component here checks if its null before further rendering. The only way around this would be to trigger a re-render by saving the ref to state for example but since its only appearing on focus/hover I think its safe to follow the practice of Flyout and keep as is like this

PR #508, opened 2019-04-01
Update TextField to take new-password as a prop value for autocomplete

ericadoh:
Add `new-password` prop value for `autocomplete` on TextFields to make it possible for browsers to suggest strong passwords.

### DIFF
--- a/packages/gestalt/src/TextField.js
+++ b/packages/gestalt/src/TextField.js
@@ -8,7 +8,12 @@ import FormErrorMessage from './FormErrorMessage.js';
 import styles from './TextField.css';
 
 type Props = {|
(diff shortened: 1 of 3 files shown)


packages/gestalt/src/TextField.js
```diff
@@ -8,7 +8,12 @@ import FormErrorMessage from './FormErrorMessage.js';
 import styles from './TextField.css';
 
 type Props = {|
-  autoComplete?: 'current-password' | 'on' | 'off' | 'username',
+  autoComplete?:
```
pfarejowicz: can we update the TextField.doc.js file as well for this new option and add a CHANGELOG.md entry for this please

PR #509, opened 2019-04-10
Upgrade to flow v0.69.0

christianvuerings:
https://github.com/facebook/flow/releases/tag/v0.96.0

### DIFF
--- a/packages/gestalt/src/Contents.js
+++ b/packages/gestalt/src/Contents.js
@@ -371,7 +371,7 @@ export default class Contents extends React.Component<Props, State> {
     mainDir: null,
   };
 
-  flyout: {| current: null | HTMLDivElement |} = React.createRef();
+  flyout: {| current: null | React.ElementRef<'div'> |} = React.createRef();
 
   componentDidMount() {
     this.setFlyoutPosition(this.props);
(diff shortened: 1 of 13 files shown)


packages/gestalt/src/Contents.js
```diff
@@ -371,7 +371,7 @@ export default class Contents extends React.Component<Props, State> {
     mainDir: null,
   };
 
-  flyout = React.createRef();
+  flyout: {| current: null | HTMLDivElement |} = React.createRef();
```
chrislloyd: This is such a weird type - it would feel way more natural to do something like `flyout = React.createRef<HTMLDivElement>();` but obviously what do I know.

pfarejowicz: agree we dont have a nice polymorphic type, but can we use `React.ElementRef` maybe?
https://github.com/facebook/flow/commit/d74ab13a8dfc5395ec5883aa2fa56be0a03a3a3f
https://flow.org/en/docs/react/types/#toc-react-elementref

packages/gestalt/src/Tooltip.js
```diff
@@ -22,7 +22,7 @@ export default class Tooltip extends React.Component<Props, State> {
     hovered: false,
   };
 
-  childRef = React.createRef();
+  childRef: {| current: null | HTMLDivElement |} = React.createRef();
```
pfarejowicz: same here, `React.ElementRef`?

PR #513, opened 2019-04-17
Add lightning bolt icon (#513)

noragrossman:


noragrossman:
@pfarejowicz Thanks for the input, I'll chat with my designer and make sure this is the asset they intended.

noragrossman:
@pfarejowicz I talked with my designer, and this asset did come from PDS, but apparently something got a little messed up in the compression process. Hopefully this version should be better.

pfarejowicz:
> @pfarejowicz I talked with my designer, and this asset did come from PDS, but apparently something got a little messed up in the compression process. Hopefully this version should be better.

@noragrossman can we double check again that the actual SVG file came from the PDS team? It looks better now and renders correctly 
image
but it looks like its not optimized which would have been done by them
image

PR #516, opened 2019-04-25
New <List /> component for gestalt

pfarejowicz:
## A new presentational component for gestalt

This change:
image

Inspiration (real world example):
image

Example use case we could replace in product:
image

Preview: https://deploy-preview-516--gestalt.netlify.com

- [x] Documentation
- [x] Tests
- [x] Accessibility checkup (verified tabbing works)

### DIFF
--- a/packages/gestalt/src/ListItem.js
+++ b/packages/gestalt/src/ListItem.js
@@ -1,0 +1,51 @@
+// @flow
+
+import * as React from 'react';
+import Box from './Box.js';
+import Text from './Text.js';
+import Touchable from './Touchable.js';
+
+type Props = {|
+  children: React.Node,
+  onClick: ({
+    event:
+      | SyntheticMouseEvent<HTMLDivElement>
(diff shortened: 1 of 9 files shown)


packages/gestalt/src/ListItem.js
```diff
@@ -0,0 +1,51 @@
...
+
+  handleMouseLeave = () => {
+    this.setState({ hovered: false });
+  };
+
+  render() {
+    const { children, onClick } = this.props;
+    const { hovered } = this.state;
+    return (
+      <Box role="listitem" padding={1} color={hovered ? 'lightGray' : 'white'}>
```
christianvuerings: @pfarejowicz Should this be hovered or focussed? Usually (e.g. Button) there should be no difference between hover & focus apart from the extra outline on focus.

christianvuerings:
Closing for now since this hasn't gotten a lot of traction

PR #518, opened 2019-04-27
Fix 514

kimuraz:
- [ ] Documentation
- [X] Tests
- [ ] Accessibility checkup

### DIFF
--- a/packages/gestalt/src/Avatar.js
+++ b/packages/gestalt/src/Avatar.js
@@ -52,7 +52,7 @@ const DefaultAvatar = ({ name }: { name: string }) => {
   );
 };
 
-type State = {| isImageLoaded: boolean |};
+type State = {| isImageLoaded: boolean, localSrc: string |};
 
 type AvatarProps = {|
   name: string,
@@ -80,7 +80,17 @@ export default class Avatar extends React.PureComponent<AvatarProps, State> {
 
   state = {
     isImageLoaded: true,
+    localSrc: '',
   };
+
+  static getDerivedStateFromProps = (props: AvatarProps, state: State) => {
+    if (props.src !== state.localSrc) {
+      return { localSrc: props.src, isImageLoaded: true };
+    }
+    return null;
+  };
+
+  componentDidMount = () => this.setState({ localSrc: this.props.src });
 
   handleImageError = () => this.setState({ isImageLoaded: false });
 
@@ -93,7 +103,7 @@ export default class Avatar extends React.PureComponent<AvatarProps, State> {
       verified,
       icon = 'check-circle',
     } = this.props;
-    const { isImageLoaded } = this.state;
+    const { isImageLoaded, localSrc } = this.state;
     const width = size ? sizes[size] : '100%';
     const height = size ? sizes[size] : '';
 
@@ -121,7 +131,7 @@ export default class Avatar extends React.PureComponent<AvatarProps, State> {
               color="#EFEFEF"
               naturalHeight={1}
               naturalWidth={1}
-              src={src}
+              src={localSrc}
               onError={this.handleImageError}
             />
           </Mask>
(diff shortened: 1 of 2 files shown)


packages/gestalt/src/Avatar.js
```diff
@@ -84,6 +84,12 @@ export default class Avatar extends React.PureComponent<AvatarProps, State> {
 
   handleImageError = () => this.setState({ isImageLoaded: false });
 
+  componentWillReceiveProps = (nextProps: *) => {
```
pfarejowicz: This lifecycle method is deprecated, we shouldnt be using it to reset state https://reactjs.org/docs/react-component.html#unsafe_componentwillreceiveprops

kimuraz: Replaced by `componentDidUpdate` method

packages/gestalt/src/Avatar.js
```diff
@@ -84,6 +84,12 @@ export default class Avatar extends React.PureComponent<AvatarProps, State> {
 
   handleImageError = () => this.setState({ isImageLoaded: false });
 
+  componentDidUpdate = (prevProps: *) => {
```
pfarejowicz: This is better but it will cause an extra render because React will handle it as follows:
New src comes in -> React renders Avatar (1) -> ComponentDidUpdate is called here -> State is set -> New rerender because prop change caused state change (2)

image

One option to avoid this double render is `getDerivedStateFromProps` which will update state from the props change but make sure to read up on the docs if that is truly the route to take

https://reactjs.org/docs/react-component.html?utm_source=caibaojian.com#static-getderivedstatefromprops

btraut: I'm not a huge fan of `getDerivedStateFromProps`, but this seems to be a reasonable use case for it. The loading state is mostly controlled by render callbacks, but a change in props may also cause it to reset.

kimuraz: I can do it, but maybe there's another solution. Since the child component `Image` handles the error and also calls the parent method when it occurs, maybe the responsibility to handle the reset on `Avatar` state should be on `Image`, because it already "knows" when there's an error or not.
Implementing it this way, the state on Avatar will only be changed when there's no error on `Image` and not when the src changes.
What do you think?

.gitignore
```diff
@@ -8,3 +8,6 @@ screenshots
 packages/gestalt/dist/*
 # the flow export file as it's maintained by hand
 !packages/gestalt/dist/gestalt.js.flow
+
+# exclude vim files
+*.sw*
```
chrislloyd: This is local config that you should probably keep in your machine's global gitignore. Gestalt doesn't support local editor setups.

kimuraz: I'll remove this commit 👍

packages/gestalt/src/Avatar.js
kimuraz: I've noticed that there is an `onLoad` on `Image`, would it be ok to control the `Image` and the `DefaultAvatar` components `opacity` according to the variable?
If so, the component wouldn't need anything but to listen to both `onLoad` and `onError`.

kimuraz: Did it using `getDerivedStateFromProps`, hope it fits the need 👍

kimuraz:
Please, let me know if there's anything missing here :+1:

christianvuerings:
@kimuraz `Avatar` has been converted to use `hooks` under the hood: https://github.com/pinterest/gestalt/pull/598

Would you be able to see if this is still an issue?

kimuraz:
> @kimuraz `Avatar` has been converted to use `hooks` under the hood: #598
> 
> Would you be able to see if this is still an issue?

I'll check, thanks for the feedback :)

PR #521, opened 2019-05-09
Disabled IconButton states

pfarejowicz:
Stop gap solution to introduce `disabled` prop to `<IconButton />` until we get a more cohesive unified solution to buttons.

image

- [x] Documentation - added new examples
- [x] Tests - added new tests

### DIFF
--- a/packages/gestalt/src/IconButton.css
+++ b/packages/gestalt/src/IconButton.css
@@ -6,8 +6,16 @@
   composes: block from "./Layout.css";
   composes: noBorder from "./Borders.css";
   composes: p0 from "./Whitespace.css";
-  composes: pointer from "./Cursor.css";
   background: transparent;
+}
+
+.enabled {
+  composes: pointer from "./Cursor.css";
+}
+
+.disabled {
+  cursor: default;
+  opacity: 0.5;
 }
 
 .button:focus {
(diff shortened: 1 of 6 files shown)


packages/gestalt/src/IconButton.css
```diff
@@ -6,10 +6,18 @@
   composes: block from "./Layout.css";
   composes: noBorder from "./Borders.css";
   composes: p0 from "./Whitespace.css";
-  composes: pointer from "./Cursor.css";
   background: transparent;
 }
 
+.enabled {
```
btraut: Tiny nit: I'm not sure how the rest of Gestalt does it, but in past projects, I tended to include the default case (enabled, in this case) along with the base styles, and then use the alternate (disabled) state as overrides. Otherwise the component isn't truly "styled" until you add either an enabled or disabled style, and the code doesn't communicate this well.

packages/gestalt/src/IconButton.js
```diff
@@ -93,10 +101,10 @@ export default class IconButton extends React.Component<Props, State> {
         type="button"
       >
         <Pog
-          active={active}
+          active={disabled ? false : active}
```
btraut: `active && !disabled`?

packages/gestalt/src/IconButton.css
pfarejowicz: I tried pulling the pointer back into the base styles and only adding the disabled one when the prop is set but it doesnt work. The base styles always override this one. Im going to keep it as is since thats what Button.js does

PR #527, opened 2019-05-23
update flashlight.svg to use new design

allollama:
PDS has a new asset for the flashlight icon, so this just switches them out


btraut:
Please add an entry to CHANGELOG.md to describe your change.

PR #532, opened 2019-05-31
Fixed bugs with Box's margin:auto

btraut:
I missed a PropType when I added margin:auto to Box. This adds it.


btraut:
Added another piece to this. I found the bug that was preventing minification of the classname. This now minifies, and as such, properly applies the style.

PR #534, opened 2019-06-05
Button- Make the border radius 8px regardless of button width

mminute:
Design has abandoned their spec where inline and full width buttons have different border radii and height.  Removing logic for changing button border radius depending on the inline prop.  All buttons will now have a border radius of 8px

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@
 - SegmentedControl: Update outer border radius to 8px from new design spec (#530)
 - Masonry: remove `MasonryBeta` and `MasonryInfiniteBeta` from source code (#531)
 - Spinner: add `delay` prop to optionally remove 300ms delay to appear (#533)
-- Undo Button border radii changes for full width buttons.  Conform all to 8px (#534)
+- Button: Undo Button border radii changes for full width buttons.  Conform all to 8px (#534)
 
 ### Patch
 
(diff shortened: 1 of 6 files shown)


packages/gestalt/src/Button.css
```diff
@@ -8,6 +8,7 @@
 
 .button {
   composes: borderBox from "./Layout.css";
+  border-radius: 8px;
```
pfarejowicz: can we compose rounded from Borders.css here instead of adding more

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@
 - SegmentedControl: Update outer border radius to 8px from new design spec (#530)
 - Masonry: remove `MasonryBeta` and `MasonryInfiniteBeta` from source code (#531)
 - Spinner: add `delay` prop to optionally remove 300ms delay to appear (#533)
+- Undo Button border radii changes for full width buttons.  Conform all to 8px (#534)
```
pfarejowicz: prefix this with `Button: ` please

packages/gestalt/src/Button.css
mminute: I was undoing this as closely as possible: https://github.com/pinterest/gestalt/pull/476/files#diff-0ae4272be4122c77370759def80050a3L11

pfarejowicz: that one had a custom border radius of 4px, i think its best to reuse as much as we can now that its 8px and we have a class for that

PR #535, opened 2019-06-05
margin top override

jouhyoung:
- Add margin override to provide for additional flexibility.
Use case:
Currently, unauth pin pages at Pinterest uses the Letterbox to render images.
Example:
Screen Shot 2019-06-05 at 11 35 15 AM
Notice how the center of the image is centered(effects of using Letterbox with aspect ratio), which is undesired in this case.
I propose that by making this change, we will have the ability to shift images based on contents to provide better UX, providing the image below instead. 
Proposed result:
Screen Shot 2019-06-05 at 11 35 22 AM

### DIFF
--- a/packages/gestalt/src/Letterbox.js
+++ b/packages/gestalt/src/Letterbox.js
@@ -19,6 +19,7 @@ type Props = {|
   contentAspectRatio: number,
   height: number,
   width: number,
+  marginTop?:number,
 |};
 
 export default function Letterbox({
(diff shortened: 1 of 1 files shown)


chrislloyd:
Hey @jouhyoung - what is the usecase here? It's a little unclear why you're adding this.

chrislloyd:
Ah, great! Thanks for updating the description. It should be possible to compose Letterbox with other components to achieve the same result:

```jsx
<Letterbox>
  <div style={{marginTop: -100}}>
    <img ... />
  </div>
</Letterbox>
```

Does that solve your usecase?

jouhyoung:
closing this as the effects can be achieved without change the gestalt library

PR #540, opened 2019-06-21
Convert all error states and components from orange to red

pfarejowicz:
Based on recent decision from design leadership
TextField (error)
image
TextArea (with focus and error)
image
SelectList
image
Checkbox
image
Flyout (error)
image
Toast (error)
image


packages/gestalt/src/SelectList.css
```diff
@@ -22,11 +22,11 @@
 }
 
 select.errored {
-  border-color: #e3780c;
+  border-color: #e60023;
 }
 
 select.errored:focus {
-  box-shadow: 0 0 0 4px rgba(227, 120, 12, 0.75);
```
pfarejowicz: Talked to Scott A about this, the error state focus outline should have the same 50% opacity that the normal focus  state outline has on it.

PR #542, opened 2019-06-21
carat sizing tweaks

NEsfandiari:
## TODO

- Documentation: elongated the sizing of the carets so that they no longer conflict with the modal border and look like one cohesive whole when clicked on.
- [ ] Tests
- [ ] Accessibility checkup


pfarejowicz:
@NEsfandiari can you provide more context on this change? Before after pictures, the spec from the PDS team, etc? It would be helpful in reviewing this change -  I'm not aware of the request for this change

NEsfandiari:
Oh, sorry. Yes this was as it pertained to this Jira bug https://jira.pinadmin.com/browse/BUG-94660?filter=-1
Before:
Screen Shot 2019-06-24 at 9 32 44 AM

After photos:
Screen Shot 2019-06-24 at 9 19 38 AM
Screen Shot 2019-06-24 at 9 19 59 AM
Screen Shot 2019-06-24 at 9 20 31 AM
Screen Shot 2019-06-24 at 9 20 48 AM

Also, there was no spec for how a caret should look on PDS, but I cleared the change with Scott Andreae from design.

btraut:
@NEsfandiari: Can you check if this commit is still necessary? I think we might not need it anymore.

christianvuerings:
@NEsfandiari This should have been fixed as part of #576 - can you confirm?

PR #547, opened 2019-07-15
Add lightning bolt icon

jesselumarie:
Add lightning bolt icon ⚡️

image


CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
   </summary>
 
 ### Minor
+  - Icon: Add new lightning icon
```
pfarejowicz: Can you add this after the button changes and attach the PR number like the rest of these please

packages/gestalt/src/icons/index.js
```diff
@@ -206,6 +207,7 @@ export default {
   'fill-transparent': fillTransparent,
   filter,
   flag,
+  flash,
```
pfarejowicz: Down below we have `'lightning-bolt-circle': lightningBoltCircle,` shouldn't this be called `lightning-bolt` to match the similarly looking icon?

packages/gestalt/src/icons/flash.svg
```diff
@@ -0,0 +1 @@
+<svg height="24" viewBox="0 0 24 24" width="24"><path d="M17.45 0L3.5 14h6.586L6.55 24 20.5 10h-6.586z" fill="#8e8e8e" fill-rule="evenodd"></path></svg>
```
pfarejowicz: This SVG markup is invalid (see the file). You need to self close the `path` tag, did this file come from PDS? They should have given you the correct file so that you dont need to worry about this stuff
image

packages/gestalt/src/icons/index.js
jesselumarie: i can do that — i didn't want to confuse folks who looked in the icon generator (where it's called flash).  but i agree this makes more sense.

jesselumarie: Talked offline a bit — we use flash internally so it makes sense to keep it as flash here

packages/gestalt/src/icons/flash.svg
```diff
@@ -0,0 +1 @@
+<svg height="24" width="24" xmlns="http://www.w3.org/2000/svg"><path d="M17.45 0L3.5 14h6.586L6.55 24 20.5 10h-6.586z" fill="#8e8e8e"/></svg>
```
pfarejowicz: Looks good 
image

PR #548, opened 2019-07-16
 Add tooltipText to IconButton

powerlanguage:
Optional prop to allow better hints for icons:

image

- [x] Documentation
- [x] Tests
- [ ] ~Experimental evidence (required for Masonry changes)~
- [ ] ~Accessibility checkup~


dangerismycat:
I like this approach. Seems better than the HOC version, as the dev doesn't need to have knowledge about how the components fit together; dev passes a string as a prop, gets what they expect.

powerlanguage:
After some discussion, I've removed the implementation for `Icon` as the intent is to be able to provide a hint for an action that is about to take place.

christianvuerings:
Closing due to inactivity. Also, Josh, I miss you.

PR #550, opened 2019-07-23
Fix tooltips when used inside a segmented control

pfarejowicz:
Previously when using `Tooltip` inside of a `SegmentedControl` clicking on the buttons would cause the flyout to go bonkers because of this little bit of CSS that changes the position attribute.

Removing it seems to fix the issue and it looks like nothing needed the position relative when focused

Before:
image
After:
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,12 +7,13 @@
   </summary>
 
 ### Minor
-  - Icon: Update send icon (#549)
-  - Icon: Add new lightning icon (#547)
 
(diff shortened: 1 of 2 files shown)


chrislloyd:
This was originally there to fix the focus outline tabbing through options - does that still work?

chrislloyd:
And shouldn't the tooltip introduce it's own stacking context?

pfarejowicz:
> This was originally there to fix the focus outline tabbing through options - does that still work?

Tested in the sandbox and it works to tab through https://deploy-preview-550--gestalt.netlify.com/#/SegmentedControl
test

PR #552, opened 2019-07-24
Remove the white background of the outer box of SearchField

Neet-Nestor:
## Remove the white background color of SeachField

- Remove the white background color of SeachField to make it can be used on background which has color other than white without the weird white sharp corner.

Before:
Screen Shot 2019-07-29 at 11 50 57 AM
Screen Shot 2019-07-29 at 11 50 50 AM
After:
Screen Shot 2019-07-29 at 11 51 14 AM
Screen Shot 2019-07-29 at 11 51 18 AM

### DIFF
--- a/packages/gestalt/src/SearchField.js
+++ b/packages/gestalt/src/SearchField.js
@@ -107,7 +107,6 @@ export default class SearchField extends React.Component<Props, State> {
         onMouseLeave={this.handleMouseLeave}
         onFocus={this.handleFocus}
         onBlur={this.handleBlur}
-        color="white"
       >
         <Box
           dangerouslySetInlineStyle={{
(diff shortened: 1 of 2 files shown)


packages/gestalt/src/SearchField.js
```diff
@@ -107,7 +107,6 @@ export default class SearchField extends React.Component<Props, State> {
         onMouseLeave={this.handleMouseLeave}
         onFocus={this.handleFocus}
         onBlur={this.handleBlur}
-        color="white"
```
pfarejowicz: This should be fine but can you show before and after screenshots of this change in the PR description? Also, do you mind adding a line entry to the CHANGELOG.md file @Neet-Nestor

Neet-Nestor: Done. Thanks!

PR #553, opened 2019-07-26
Add the size prop to Spinner

cvuong:
Add the `size` prop for the Spinner. The value of `size` can be the following: `"xs" | "sm" | "md" | "lg" | "xl"`.

`size="sm"`
Screen Shot 2019-07-26 at 1 10 33 PM

No `size` prop present (size defaults to `"md"`)
Screen Shot 2019-07-26 at 1 10 57 PM

### DIFF
--- a/docs/src/Spinner.doc.js
+++ b/docs/src/Spinner.doc.js
@@ -35,8 +35,8 @@ card(
       },
       {
         name: 'size',
-        type: `"xs" | "sm" | "md" | "lg" | "xl"`,
-        description: `xs: 24px, sm: 32px, md: 40px, lg: 48px, xl: 56px`,
+        type: `"sm" | "md"`,
+        description: `sm: 32px, md: 40px`,
         defaultValue: 'md',
       },
     ]}
--- a/docs/src/Spinner.doc.js
+++ b/docs/src/Spinner.doc.js
@@ -33,6 +33,12 @@ card(
         description:
           'Whether or not to render with a 300ms delay. The delay is for perceived performance so you should rarely need to remove it.',
       },
+      {
+        name: 'size',
+        type: `"xs" | "sm" | "md" | "lg" | "xl"`,
+        description: `xs: 24px, sm: 32px, md: 40px, lg: 48px, xl: 56px`,
+        defaultValue: 'md',
+      },
     ]}
   />
 );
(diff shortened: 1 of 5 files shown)


packages/gestalt/src/Spinner.js
```diff
@@ -6,26 +6,35 @@ import Box from './Box.js';
...
 }: Props) {
+  const iconSize = SIZE_NAME_TO_PIXEL[size];
   return show ? (
     <Box display="flex" justifyContent="around" overflow="hidden">
       <div className={classnames(styles.icon, { [styles.delay]: delay })}>
         <Icon
           icon="knoop"
           accessibilityLabel={accessibilityLabel}
-          size={SIZE}
+          size={iconSize}
```
pfarejowicz: Can we add a test for a non-default size please

docs/src/Spinner.doc.js
```diff
@@ -33,6 +33,12 @@ card(
         description:
           'Whether or not to render with a 300ms delay. The delay is for perceived performance so you should rarely need to remove it.',
       },
+      {
+        name: 'size',
+        type: `"xs" | "sm" | "md" | "lg" | "xl"`,
+        description: `xs: 24px, sm: 32px, md: 40px, lg: 48px, xl: 56px`,
```
pfarejowicz: Just curious how these were chosen? Did PDS help with the spec?

packages/gestalt/src/Spinner.js
cvuong: Will do.

docs/src/Spinner.doc.js
cvuong: Our designer said that the sizes should match the same size specs as PDS.

packages/gestalt/src/Spinner.js
```diff
@@ -6,26 +6,35 @@ import Box from './Box.js';
...
+  size?: 'xs' | 'sm' | 'md' | 'lg' | 'xl',
 |};
 
 export default function Spinner({
   accessibilityLabel,
   delay = true,
   show,
+  size = 'md',
 }: Props) {
+  const iconSize = SIZE_NAME_TO_PIXEL[size];
```
pfarejowicz: Can we inline this directly for the prop that way if `show` is false then the browser doesnt need to compute this extraneously

PR #555, opened 2019-07-31
[Breaking] Update the gestalt icons to keep parity with PDS

pfarejowicz:
PDS has been working on a better system for icons, and has provided a new tool and set of icons to use on web.

This change ports all those icons over and abides by the PDS naming for consistency. A script was added to generate the `index.js` file for the icons folder and the `.eslintrc` file was also moved out of the icons directory in favor of using the `overrides` at the root level.

Note: don't focus too much on the script, its not 100% perfect but it helps get the job done. Some manual changes were needed such as fixing the names of icons that start with a number and running prettier on the file to remove superfluous quotes.

The list of icons that are changing names:
TODO


christianvuerings:
Closing this PR for now - we'll also need a codemod to address the breaking changes

PR #566, opened 2019-09-12
Update componentWillReceiveProps to UNSAFE_componentWillReceiveProps

janephilipps:
In order to prepare for an upgrade to React 17, we need to rename `componentWillReceiveProps` to `UNSAFE_componentWillRecieveProps`.

https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html#gradual-migration-path

## TODO

- [x] Documentation
- [ ] Tests

### DIFF
--- a/packages/gestalt/src/Contents.js
+++ b/packages/gestalt/src/Contents.js
@@ -386,7 +386,7 @@ export default class Contents extends React.Component<Props, State> {
     window.addEventListener('keydown', this.props.onKeyDown);
   }
 
-  componentWillReceiveProps(nextProps: Props) {
+  UNSAFE_componentWillReceiveProps(nextProps: Props) {
     this.setFlyoutPosition(nextProps);
   }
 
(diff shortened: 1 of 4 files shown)


packages/gestalt/src/Contents.js
```diff
@@ -474,6 +470,11 @@ export default class Contents extends React.Component<Props, State> {
     });
   };
 
+  // eslint-disable-next-line camelcase
+  UNSAFE_componentWillReceiveProps(nextProps: Props) {
```
christianvuerings: @janephilipps Could we instead use `allow` for this rule in our eslint config? https://eslint.org/docs/rules/camelcase#allow
```
/*eslint camelcase: ["error", {allow: ["UNSAFE_componentWillMount"]}]*/

function UNSAFE_componentWillMount() {
    // ...
}
```

janephilipps:
The update to the eslint config failed CI, so in lieu of making this change more complex than it needs to be, I'm reverting back to disabling the rule.

PR #567, opened 2019-09-24
Add new replace/scale icons

loserford:
Add two new icons: replace and scale


CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 
 ### Minor
 
+- Icon: Add new replace/scale icons (#567)
```
christianvuerings: @loserford Could we move this to the bottom of this list? We always try to keep these in order of the PR #

loserford: updated.

Issue #568, opened 2019-09-27
Documentation for responsive properties


durchanek:
I may be wrong but it looks like there is no documentation on responsive variants of properties, like what are the actual breakpoints or what is the desired behaviour. This makes it very hard to see if I am doing something wrong or there is a bug in the component. E.g. I would expect that setting `smDisplay="none"` hides the component on small screens but in fact, it outputs:
```
@media (min-width: 576px)
._As {
    display: none;
}
```
This feels counter-intuitive because it forces `display="none" smDisplay="block"` but it may be by design. Having a clear documentation would totally help here.

christianvuerings:
@durchanek that's indeed by design - everything we do on gestalt is mobile first so `display="none" smDisplay="block"` is correct.

You can find the list of breakpoints at https://github.com/pinterest/gestalt/blob/master/packages/gestalt/src/breakpoints.json

Feel free to open a PR to add this information to our docs

PR #569, opened 2019-10-01
remove React defaultProps and move them to ES6 static defaultProps. u…

AlbertCarreras:
…pdated Changelog

## TODO

- [ ] Documentation
- [ ] Tests
- [ ] Accessibility checkup


AlbertCarreras:
On Changelog, I referred my PR with #568 which was the next number on the list. I see now that's actually #569.

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@
 - Icon: Replace existing filter icon (#565)
 - Contents/Controller: Replace componentWillReceiveProps with UNSAFE_componentWillReceiveProps (#566)
 - Icon: Add new replace/scale icons (#567)
+- defaultProps: Remove the last React defaultProps and transfer it to the already existing but empty ES6 static defaultProps on the same file (#568)
```
btraut: Even though the motivation here was to modify `defaultProps`, you specifically edited the `ScrollFetch` component. If I were writing the comment, I think I would've called that out as the header of this item.

packages/gestalt/src/ScrollFetch.js
```diff
@@ -32,7 +32,9 @@ export default class ScrollFetch extends React.PureComponent<Props, State> {
     this.setState(this.getScrollState());
   });
 
-  static defaultProps: {};
+  static defaultProps = {
```
btraut: As it turns out, this code was actually compliant already. `defaultProps` was only deprecated on function components, and this component is a class component. That said, I prefer moving it into a static declaration rather than putting it at the bottom of the file, so I like this change.

PR #570, opened 2019-10-02
Removed unsafe lifecycle methods from Controller/Contents

btraut:
As part of the React 17 readiness effort, we're doing our best to remove/refactor the deprecated lifecycle methods. In PR #566, we renamed the only remaining occurrences of these methods to use the `UNSAFE_` prefix. This diff follows this up by killing those methods outright, and instead migrating to `getDerivedStateFromProps`.

To test, I used the docs page to verify the Flyouts and Tooltips (both of these use Controller/Contents) still worked as intended. I also triggered props changes and validated that things work. Lastly, I validated that the snapshot test did not change.

### DIFF
--- a/packages/gestalt/src/Controller.js
+++ b/packages/gestalt/src/Controller.js
@@ -43,12 +43,39 @@ type State = {|
   triggerBoundingRect: ClientRect,
 |};
 
+function getTriggerRect(
+  anchor: HTMLElement,
+  positionRelativeToAnchor: boolean
+) {
+  let triggerBoundingRect;
+  let relativeOffset;
+  if (anchor) {
+    triggerBoundingRect = anchor.getBoundingClientRect();
+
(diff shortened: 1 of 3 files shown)


packages/gestalt/src/Controller.js
```diff
@@ -87,32 +114,14 @@ export default class Controller extends React.Component<Props, State> {
     this.updateTriggerRect(this.props);
```
bishwei: question here - could we actually just do a `this.forceUpdate()` here since that would ultimately result in `this.updateTriggerRect` being called from gDSFP?

btraut: `forceUpdate` seems a little smelly to me but I don't think I can defend the position. Seems the same in this case. Do you feel like that's a cleaner solution?

PR #573, opened 2019-10-16
Fixed memory leak in Masonry’s measurementStore

btraut:
When comparing Gestalt's Masonry (Masonry8) with Masonry9, I stumbled on a bug in Masonry8. The bug is that a single `MeasurementCache` gets created when `defaultProps` are defined, and then this same `MeasurementCache` is used for _all_ grids that don't specify their own `measurementCache` via props. This means, since the `defaultProps` are static and live outside any instance of a component, the cache is shared across grids and it's never cleared, even when grids are unmounted.

This PR fixes the issue by turning `measurementCache` into an unmanaged prop and pulling it into `state` on component construction. While I tend not to like unmanaged props (especially when they're not named something like `initialMeasurementCache` or `defaultMeasurementCache`), I think this is the simplest solution to fix the bug. AFAIK, we never change up the `measurementCache` values within a component in Pinterest anyway.

Note: Masonry9 has the exact same bug and it'll need to fixed separately.

### DIFF
--- a/packages/gestalt/src/Masonry.js
+++ b/packages/gestalt/src/Masonry.js
@@ -58,6 +58,7 @@ type Props<T> = {|
 |};
 
 type State<T> = {|
+  measurementStore: Cache<T, *>,
   hasPendingMeasurements: boolean,
   isFetching: boolean,
   items: Array<T>,
@@ -176,7 +177,6 @@ export default class Masonry<T> extends React.Component<Props<T>, State<T>> {
   static defaultProps = {
     columnWidth: 236,
     // $FlowFixMe: new errors found from flow 0.96 upgrade
-    measurementStore: new MeasurementStore(),
     minCols: 3,
     layout: DefaultLayoutSymbol,
     loadItems: () => {},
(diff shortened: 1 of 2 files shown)


packages/gestalt/src/Masonry.js
```diff
@@ -176,7 +177,6 @@ export default class Masonry<T> extends React.Component<Props<T>, State<T>> {
   static defaultProps = {
     columnWidth: 236,
     // $FlowFixMe: new errors found from flow 0.96 upgrade
-    measurementStore: new MeasurementStore(),
```
jackhsu978: Since the prop `measurementStore` no longer has a default, flow type for the prop should be changed from required to optional

packages/gestalt/src/Masonry.js
```diff
@@ -374,6 +381,7 @@ export default class Masonry<T> extends React.Component<Props<T>, State<T>> {
    */
   reflow() {
     this.props.measurementStore.reset();
```
jackhsu978: need to check if `this.props.measurementStore` is defined first

packages/gestalt/src/Masonry.js
btraut: Good call! Done.

packages/gestalt/src/Masonry.js
btraut: So smart! Fixed.

PR #576, opened 2019-10-18
[Flyout] Fix lightgray border separating caret from contents on white…

remingtonholt:
… flyout

Before:
スクリーンショット 2019-10-17 午後5 29 24

After:
スクリーンショット 2019-10-17 午後5 31 35

No other flyout separate its content from the caret with a white flyout border; so, this PR removes that separation to make the white flyout design behavior match those of all the other color flyouts.

- [x] Documentation
- [x] Tests

### DIFF
--- a/packages/gestalt/src/Contents.js
+++ b/packages/gestalt/src/Contents.js
@@ -483,7 +483,6 @@ export default class Contents extends React.Component<Props, State> {
     const visibility = mainDir === null ? 'hidden' : 'visible';
     const background = `${bgColor}Bg`;
     const stroke = bgColor === 'white' ? '#efefef' : null;
-    const borderColor = bgColor === 'white' ? 'lightGray' : bgColor;
 
     return (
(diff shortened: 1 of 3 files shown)


packages/gestalt/src/Contents.js
```diff
@@ -483,7 +483,7 @@ export default class Contents extends React.Component<Props, State> {
     const visibility = mainDir === null ? 'hidden' : 'visible';
     const background = `${bgColor}Bg`;
     const stroke = bgColor === 'white' ? '#efefef' : null;
-    const borderColor = bgColor === 'white' ? 'lightGray' : bgColor;
+    const borderColor = bgColor;
```
pfarejowicz: You can probably clean this up more by getting rid of the variable all together :)

remingtonholt: Removed the variable altogether. :)

PR #587, opened 2019-11-14
Fix flow types on ScrollFetch

btraut:
With #569, we updated the way ScrollFetch specifies default props to use `static` instead of setting an object after the class definition. This was functionally the same, but flow interpreted them differently. Specifically: flow is now covering these defaults whereas it wasn't before. Unfortunately, this didn't get exposed until we tried to import Gestalt v106 into pinboard.

This diff attempts to fix the default prop type. Unfortunately, I don't know how to test it other than to land it and pull it down from there.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
 
 ### Minor
 
+- ScrollFetch: Fixed a tiny flow type bug on the default props (#587)
+
 ### Patch
 
 </details>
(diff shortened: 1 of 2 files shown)


btraut:
Figured it out! I struggled testing this thing with `yarn link` and with `package.json` entries that pointed to branches/hashes, but I managed to make it work by copying the Gestalt dist package directly into pinboard's `node_modules`. This allowed me to run `yarn flow` and there were no errors.

PR #591, opened 2019-11-24
Text / Heading: Add semibold weight & introduce codemod unit tests

christianvuerings:
Adds a semi bold option to `<Text />` and `<Heading />`

image

- [*] Documentation
- [*] Tests
- [*] Accessibility checkup

We also introduce codemod unit tests to more easily write codemods, check edge cases & avoid regressions.

More info at
- https://github.com/facebook/jscodeshift#unit-testing
- https://github.com/benjamn/recast
- https://github.com/facebook/jscodeshift
- https://github.com/cpojer/js-codemod
- https://github.com/reactjs/react-codemod

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -5,6 +5,7 @@
 # Tracking https://github.com/facebook/flow/issues/4015
 <PROJECT_ROOT>/docs
 <PROJECT_ROOT>/test
+<PROJECT_ROOT>/packages/gestalt-codemods/.*/__testfixtures__
 
 [options]
 suppress_comment= \\(.\\|\n\\)*\\$FlowFixMe
--- a/README.md
+++ b/README.md
@@ -61,20 +61,15 @@ Running Masonry's integration tests. This will leave lots of Firefox processes h
 
 When a release will cause breaking changes — in usage or in typing — we provide a codemod to ease the upgrade process. Codemods are organized by release in `/packages/gestalt-codemods`. We recommend using [jscodeshift](https://github.com/facebook/jscodeshift) to upgrade.
 
-### Prerequisite:
-Install `jscodeshift` globally if you haven't already.
-```bash
-yarn global add jscodeshift
-```
-
 ### Usage:
+
 Clone the Gestalt repo locally if you haven't already. Run the relevant codemod(s) in the relevant directory of your repo (not the Gestalt repo): anywhere the component to be updated is used. Example usage for a codebase using Flow:
+
 ```bash
-jscodeshift --parser=flow -t={relative/path/to/codemod} relative/path/to/your/code
+yarn codemod --parser=flow -t={relative/path/to/codemod} relative/path/to/your/code
 ```
+
 For a dry run to see what the changes will be, add the `-d` (dry run) and `-p` (print output) flags (pipe stdout to a file for easier inspection if you like).
-
-
 
 ## Releasing
 
--- a/packages/gestalt-codemods/0.108.0-0.109.0/__testfixtures__/.eslintrc
+++ b/packages/gestalt-codemods/0.108.0-0.109.0/__testfixtures__/.eslintrc
@@ -1,0 +1,6 @@
+{
+  "rules": {
+    "prettier/prettier": "off",
+    "no-undef": "off"
+  }
+}
--- a/packages/gestalt-codemods/0.108.0-0.109.0/__testfixtures__/convert-font-weight-dynamic.input.js
+++ b/packages/gestalt-codemods/0.108.0-0.109.0/__testfixtures__/convert-font-weight-dynamic.input.js
@@ -1,0 +1,15 @@
+// @flow
+import React from 'react';
+import { Text } from 'gestalt';
+
+export default function BoldText({
+  isCurrentPage,
+}: {
+  isCurrentPage?: boolean,
+}) {
+  return (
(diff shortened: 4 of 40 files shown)


.flowconfig
```diff
@@ -5,6 +5,7 @@
 # Tracking https://github.com/facebook/flow/issues/4015
 <PROJECT_ROOT>/docs
 <PROJECT_ROOT>/test
+<PROJECT_ROOT>/packages/gestalt-codemods/.*/__testfixtures__
```
christianvuerings: Flow fails on `<Text bold />` referenced in test fixtures, we could add something like `$FlowExpectedError`, however that breaks the `input` / `output` behavior `defineTest` expects

packages/gestalt-codemods/0.108.0-0.109.0/__testfixtures__/convert-font-weight-dynamic.input.js
```diff
@@ -0,0 +1,15 @@
+// @flow
```
christianvuerings: More info on naming requirements are at https://github.com/facebook/jscodeshift#unit-testing

> jscodeshift comes with a simple utility to allow easy unit testing with [Jest](https://facebook.github.io/jest/), without having to write a lot of boilerplate code. This utility makes some assumptions in order to reduce the amount of configuration required:
> - The test is located in a subdirectory under the directory the transform itself is located in (eg. `__tests__`)
> - Test fixtures are located in a `__testfixtures__` directory

packages/gestalt-codemods/0.108.0-0.109.0/__testfixtures__/.eslintrc
```diff
@@ -0,0 +1,8 @@
+{
+  "rules": {
+    "prettier/prettier": 0,
```
christianvuerings: Had to disable prettier in the test fixture because of
```
<Text size="xl" weight={isCurrentPage ? "bold" : "normal"}>
```
should instead be
```
<Text size="xl" weight={isCurrentPage ? 'bold' : 'normal'}>
```

However, when calling jscodeshift with `.toSource({quote: 'single'})`, it converts all quotes instead of only the modified ones.

packages/gestalt-codemods/0.108.0-0.109.0/__tests__/convert-font-weight.test.js
```diff
@@ -0,0 +1,17 @@
+import { defineTest } from 'jscodeshift/dist/testUtils.js';
+
+jest.mock('../convert-font-weight', () => {
+  return Object.assign(require.requireActual('../convert-font-weight'), {
```
christianvuerings: Found this fix in `react/react-codemods`: https://git.io/JePID

README.md
```diff
@@ -61,20 +61,15 @@ Running Masonry's integration tests. This will leave lots of Firefox processes h
 
 When a release will cause breaking changes — in usage or in typing — we provide a codemod to ease the upgrade process. Codemods are organized by release in `/packages/gestalt-codemods`. We recommend using [jscodeshift](https://github.com/facebook/jscodeshift) to upgrade.
 
-### Prerequisite:
-Install `jscodeshift` globally if you haven't already.
-```bash
-yarn global add jscodeshift
```
christianvuerings: `jscodeshift` is now installed as a dev dependency and we wrap it with `yarn codemod` to ensure the correct `jscodeshift` version.

packages/gestalt-codemods/0.108.0-0.109.0/__tests__/convert-font-weight.test.js
```diff
@@ -0,0 +1,17 @@
...
+    parser: 'flow',
+  });
+});
+
+describe('convert-font-weight', () => {
+  [
+    'convert-font-weight-dynamic',
+    'convert-font-weight-spread',
+    'convert-font-weight-static',
```
btraut: Do we also need to worry about any cases with `<Text bold={false} />`?

packages/gestalt/src/Heading.js
```diff
@@ -74,15 +76,16 @@ export default function Heading(props: Props) {
     lgSize && styles[`lgFontSize${SIZE_SCALE[lgSize]}`],
     colors[color],
     overflow === 'breakWord' && typography.breakWord,
-    truncate && typography.truncate
+    truncate && typography.truncate,
+    weight === 'semibold' && typography.fontWeightSemibold
```
btraut: Do you tend to like to make secondary properties like `fontWeightSemibold` as just overrides to the base class? I will often create two separate classes each that extend the base, then choose which one of those I should use. I'm certain your approach works, but I'm curious what you think, and if you think I'm being too pedantic.

packages/gestalt-codemods/0.108.0-0.109.0/__tests__/convert-font-weight.test.js
christianvuerings: Thanks - also added a test case for that (didn't find any in pinboard)

packages/gestalt/src/Heading.js
christianvuerings: @btraut this approach is a bit more atomic, just one property per CSS class and also allows for better re-use across components.

PR #597, opened 2019-12-02
Card: Convert examples & component to use hooks

christianvuerings:
Convert examples & component to use hooks

### DIFF
--- a/docs/src/Card.doc.js
+++ b/docs/src/Card.doc.js
@@ -52,47 +52,25 @@ card(
   `}
     name="Example"
     defaultCode={`
-class CardExample extends React.Component {
-  constructor(props) {
-    super(props);
-    this.state = { hovered: false };
-    this.handleMouseEnter = this._handleMouseEnter.bind(this);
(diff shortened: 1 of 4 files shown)


docs/src/Card.doc.js
```diff
@@ -52,47 +52,25 @@ card(
...
-            color="red"
-            text="Follow"
-          />
-        </Card>
-      </Box>
-    );
-  }
+function CardExample() {
+  return (
+    <Box maxWidth={236} padding={2} column={12}>
```
christianvuerings: We do not need to use `state` in this example

docs/src/components/Card.js
```diff
@@ -29,7 +29,7 @@ export default function Card({
         display="flex"
         direction={stacked ? 'column' : 'row'}
       >
-        <Box paddingX={2} column={12}>
+        <Box marginTop={1} paddingX={2} column={12}>
```
christianvuerings: Adds a bit of extra spacing

PR #598, opened 2019-12-03
Avatar: Convert component to use hooks

christianvuerings:
Also introduces the following:
- Internal: Add `react-testing-library`
- Internal: Enable `react-hooks/exhaustive-deps` lint rule

### DIFF
--- a/packages/gestalt/src/Avatar.js
+++ b/packages/gestalt/src/Avatar.js
@@ -52,9 +52,7 @@ const DefaultAvatar = ({ name }: { name: string }) => {
   );
 };
 
-type State = {| isImageLoaded: boolean |};
-
-type AvatarProps = {|
+type Props = {|
   name: string,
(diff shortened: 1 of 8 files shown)


packages/gestalt/src/Avatar.js
```diff
@@ -69,95 +67,82 @@ const sizes = {
...
-    src: PropTypes.string,
-    size: PropTypes.oneOf(['sm', 'md', 'lg']),
-    verified: PropTypes.bool,
-  };
-
-  state = {
-    isImageLoaded: true,
-  };
+export default function Avatar(props: Props) {
+  const [isImageLoaded, setIsImageLoaded] = React.useState(true);
```
btraut: This is a bad var name. We should follow up with another diff that renames to `imageHasFailedToLoad = false`.

PR #600, opened 2019-12-04
Checkbox: Convert examples & component to use hooks

christianvuerings:


.eslintrc.json
```diff
@@ -34,6 +34,15 @@
       "error",
       { "components": ["label"], "allowChildren": true }
     ],
+    "jsx-a11y/label-has-associated-control": [
```
christianvuerings: https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/label-has-associated-control.md#case-my-label-and-input-components-are-custom-components

docs/src/Checkbox.doc.js
```diff
@@ -122,37 +115,35 @@ card(
...
-        <Box paddingY={1}>
-          <CheckboxWithLabel label="Email" id="email" />
-        </Box>
-        <Box paddingY={1}>
-          <CheckboxWithLabel label="Mobile push" id="push" />
-        </Box>
-        <Box paddingY={1}>
-          <CheckboxWithLabel label="Carrier pidgeon" id="pidgeon" />
+function CheckboxExample() {
+  const CheckboxWithLabel = ({ id, label }) => (
```
christianvuerings: This is a bit odd but `react-live` breaks if we extract `CheckboxWithLabel` out of `CheckboxExample` since it does an automatic `return` on whatever is passed into `defaultCode`

packages/gestalt/src/Checkbox.jsdom.test.js
```diff
@@ -0,0 +1,23 @@
+// @flow
```
christianvuerings: Extracted out of `src/Checkbox.test.js` since clicking is a browser only event (doesn't happen on the server)

PR #603, opened 2019-12-06
Internal: Convert tests to React Testing Library

christianvuerings:
- Add [@testing-library/jest-dom](https://github.com/testing-library/jest-dom): adds Custom jest matchers to test the state of the DOM (e.g. `toBeVisible`/ `toHaveFocus`)
- Update flow types
- Remove enzyme
- Ensure snapshot tests span multiple lines: previously some of these were very long single line strings which made them hard to compare in diffs. `.html()` to `toJSON()`. See https://reactjs.org/docs/test-renderer.html#testrenderertojson
> Return an object representing the rendered tree. This tree only contains the platform-specific nodes like <div> or <View> and their props, but doesn’t contain any user-written components. This is handy for snapshot testing.


packages/gestalt/src/SelectList.jsdom.test.js
```diff
@@ -0,0 +1,73 @@
...
+    />
+  );
+  expect(getByText('Error message')).toBeVisible();
+});
+
+it('SelectList with disabled', () => {
+  const { container } = render(
+    <SelectList disabled id="test" onChange={jest.fn()} options={options} />
+  );
+  expect(container.querySelector('select[disabled]')).toBeVisible();
```
btraut: I wonder if we should just invest in building a few more selectors into RTL. This one seems legit, but annoying that we need to use the querySelector hack.

packages/gestalt/src/TextArea.jsdom.test.js
```diff
@@ -0,0 +1,87 @@
...
+      <TextArea
+        id="test"
+        onChange={jest.fn()}
+        onFocus={jest.fn()}
+        onBlur={jest.fn()}
+      />
+    );
+    expect(() => {
+      getByText('Error message');
+    }).toThrow('Unable to find an element with the text: Error message');
```
btraut: Is catching an exception really the right pattern for this? Without reading the docs, it feels like there should be a selector that just returns something falsy on no results.

packages/gestalt/src/Video.jsdom.test.js
```diff
@@ -0,0 +1,186 @@
...
+  it('Does not load when string src does not change', () => {
+    const props = {
+      ...A11Y_LABELS,
+      aspectRatio: 1,
+      captions: 'https://media.w3.org/2010/05/sintel/captions.vtt',
+      src: 'https://media.w3.org/2010/05/sintel/trailer_hd.mp4',
+    };
+
+    const { container, rerender } = render(<Video {...props} />);
+    const spy = jest.spyOn(container.querySelector('video'), 'load');
```
btraut: Do we automatically clear spies on teardown?

packages/gestalt/src/SelectList.jsdom.test.js
christianvuerings: Agreed - seems like a good follow up. This initial PR is to remove enzyme and we can improve our tests later.

packages/gestalt/src/TextArea.jsdom.test.js
christianvuerings: That's the way React Testing Library does it - haven't found a better way yet

packages/gestalt/src/Video.jsdom.test.js
christianvuerings: The spy is scoped to the actual `test` so we it should not leak across others.

PR #604, opened 2019-12-06
Adds info-circle and workflow SVG icons

Randyjp:
## TODO

- [x] Documentation
- [x] Tests

iconsFixed

infocircle


packages/gestalt/src/icons/workflow-status-halted.svg
```diff
@@ -0,0 +1 @@
+<svg width="24" height="24" fill="none" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24"><path fill-rule="evenodd" clip-rule="evenodd" d="M20 12c0 4.4-3.6 8-8 8s-8-3.6-8-8 3.6-8 8-8 8 3.6 8 8zM6 12c0 3.3 2.7 6 6 6s6-2.7 6-6-2.7-6-6-6-6 2.7-6 6zm4-3c.6 0 1 .4 1 1v4c0 .6-.4 1-1 1s-1-.4-1-1v-4c0-.6.4-1 1-1zm4 0c.6 0 1 .4 1 1v4c0 .6-.4 1-1 1s-1-.4-1-1v-4c0-.6.4-1 1-1z" fill="#333"/></svg>
```
christianvuerings: @Randyjp could we remove the `fill="none"` from the `svg` element? Also, remove `clip-rule="evenodd"` and set the `fill ` to be `#8e8e8e` like all our other icons.

That way the SVG becomes a bit smaller and we can set the color in gestalt

CHANGELOG.md
```diff
@@ -6,6 +6,13 @@
     Click to see more.
   </summary>
 
+### Minor
```
christianvuerings: @Randyjp Can we move these into the other `### Minor section below` ?

packages/gestalt/src/icons/workflow-status-halted.svg
ponori: One interesting thing about these particular workflow icons is that we don't want them to be any other color than what's directed. There's no reason to change what we're doing in the short-term, but there may be a case to move these icons into a separate component in the future that enforces the specific color treatment.

Randyjp: SVG icons have been updated with suggestions

PR #609, opened 2019-12-10
Add IconWithTooltip component and tests

denamwangi:
Background: We have an initiative to push for in product education within the sterling codebase in H1 that requires a tooltip on an info-icon hover with support for a "Learn More" link.

Proposed: We don't want to support this for all our Tooltips so this new IconWithTooltip component will allow us to be very specific in the implementation. PDS approved it for the info-icon specifically- i'm leaving the icon assignment flexible for others' use cases.

~Open question: accessibility standards. I had one suggestion to make the tooltip visible on focus in addition to on hover. Is this sufficient?~

*Updated behavior with focus and blur*

iconwithtooltip3

## TODO

- [x] Documentation
- [x] Tests
- [x] Accessibility checkup

### DIFF
--- a/packages/gestalt/src/IconWithTooltip.jsdom.test.js
+++ b/packages/gestalt/src/IconWithTooltip.jsdom.test.js
@@ -1,0 +1,96 @@
+// @flow
+import React from 'react';
+import '@babel/polyfill';
+import {
+  render,
+  fireEvent,
+  wait,
+  waitForElement,
+} from '@testing-library/react';
+import IconWithTooltip from './IconWithTooltip.js';
+
+test('Icon with Tooltip and no href renders', () => {
+  const { getByRole, queryByRole } = render(
+    <IconWithTooltip
+      accessibilityLabel="informational-tooltip"
+      icon="circle-outline"
+      inline
+      tooltipText="Icon with Tooltip and Learn More link"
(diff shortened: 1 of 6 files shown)


packages/gestalt/src/IconWithTooltip.jsdom.test.js
```diff
@@ -0,0 +1,55 @@
...
+      tooltipText="Icon with Tooltip and Learn More link"
+      href="https://www.pinterest.com"
+    />
+  );
+
+  expect(getByRole('img')).toBeVisible();
+  fireEvent.mouseEnter(getByRole('img'));
+  expect(getByRole('tooltip')).toBeVisible();
+  expect(queryByRole('link')).toEqual(null);
+});
```
jennyscript: Please add test coverage for the other events handled by the component. Be sure to fire all of these and check for correct behavior: `fireEvent.mouseLeave(getByRole('img'))`, `fireEvent.mouseEnter(getByRole('tooltip'))`, `fireEvent.mouseLeave(getByRole('tooltip'))`

denamwangi: Thanks @jennyscript - i've added a test for the full cycle.

PR #628, opened 2020-01-16
Switch: Add flip style support for right-to-left language locales like arabic

selena-zeng:
## TODO

- [x] Documentation
- [x] Tests
~~Experimental evidence (required for Masonry changes)~~
- [x] Accessibility checkup

Added Switch RTL support description and example
switch

Updated snapshot and tested on webapp for RTL and LTR, no server side errors
image

### DIFF
--- a/docs/src/Switch.doc.js
+++ b/docs/src/Switch.doc.js
@@ -1,6 +1,6 @@
 // @flow
 import * as React from 'react';
-import { Switch, Button, Box } from 'gestalt';
+import { Box, Icon, Text, Switch } from 'gestalt';
 import PropTable from './components/PropTable.js';
 import Example from './components/Example.js';
 import Combination from './components/Combination.js';
@@ -8,30 +8,6 @@ import PageHeader from './components/PageHeader.js';
 
 const cards = [];
 const card = c => cards.push(c);
-
-const toggleRTL = () => {
-  if (document && document.documentElement) {
-    const isRTL = document.documentElement.dir === 'rtl';
-    document.documentElement.dir = isRTL ? 'ltr' : 'rtl';
-  }
-};
-
-const changeLocaleDirection = (
-  <Box
-    position="relative"
-    justifyContent="center"
-    display="flex"
-    marginTop={10}
-  >
-    <Box maxWidth={600}>
-      <Button
-        size="sm"
-        onClick={toggleRTL}
-        text="Toggle Page and Switch Direction: right-to-left/left-to-right"
-      />
-    </Box>
-  </Box>
-);
 
 card(
   <PageHeader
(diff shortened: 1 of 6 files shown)


docs/src/Switch.doc.js
```diff
@@ -9,11 +9,37 @@ import PageHeader from './components/PageHeader.js';
...
 const card = c => cards.push(c);
 
+const toggleRTL = () => {
+  if (document && document.documentElement) {
+    const isRTL = document.documentElement.dir === 'rtl';
+    document.documentElement.dir = isRTL ? 'ltr' : 'rtl';
+  }
+};
+
+const changeLocaleDirection = (
```
christianvuerings: @selena-zeng Could we add this to the navbar instead of making it specific to `Switch`?

packages/gestalt/src/Switch.css
```diff
@@ -4,6 +4,14 @@
   --border: 1px;
 }
 
+html[dir="rtl"] .rtlSupport {
```
christianvuerings: @selena-zeng Could we just use the `.switch` class? `html[dir="rtl"] .switch {` That way we don't need to add a new class

docs/src/Switch.doc.js
selena-zeng: changed to nav bar

docs/src/Switch.doc.js
```diff
@@ -12,11 +12,30 @@ const card = c => cards.push(c);
...
-    description="Use switches for single cell options that can be turned on and off only.
-If you have a cell with multiple options that can activated, consider using check marks."
+    description={`Use switches for single cell options that can be turned on and off only.
+If you have a cell with multiple options that can activated, consider using check marks.
+\`Switch\` component supports right-to-left(RTL) language locales layout
+(auto flip on RTL locales like Arabic).`}
   />
 );
 
+card(
```
christianvuerings: @selena-zeng Can we remove this `card`? It should be enough to have the control in the nav bar.

docs/src/components/Header.js
```diff
@@ -1,68 +1,102 @@
...
-                    icon="pinterest"
-                    color="white"
-                    size={24}
-                    accessibilityLabel="Pinterest Logo"
-                  />
+type State = {|
+  isRTL: boolean,
+|};
+
+export default class Header extends Component<{}, State> {
```
christianvuerings: @selena-zeng please use a function with `useState` instead. We want to avoid adding new class components in Gestalt

packages/gestalt/src/Switch.css
```diff
@@ -18,6 +18,14 @@
   width: var(--width);
 }
 
+html[dir="rtl"] .switch {
+  /*
```
christianvuerings: @selena-zeng let's remove this comment

PR #631, opened 2020-01-17
Icon: Add right-to-left locales flip style support and update doc

selena-zeng:
## TODO

- [x] Documentation
- [x] Tests
~~Experimental evidence (required for Masonry changes)~~
- [x] Accessibility checkup

image

### DIFF
--- a/docs/src/Icon.doc.js
+++ b/docs/src/Icon.doc.js
@@ -1,6 +1,6 @@
 // @flow
 import * as React from 'react';
-import { Icon } from 'gestalt';
+import { Box, Icon, Text } from 'gestalt';
 import Example from './components/Example.js';
 import PropTable from './components/PropTable.js';
 import Combination from './components/Combination.js';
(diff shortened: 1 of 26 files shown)


docs/src/Icon.doc.js
```diff
@@ -18,6 +18,30 @@ Show icons with different colors and sizes in an accessible way.
...
 );
 
+card(
+  <Box display="flex" direction="column">
+    <Text size="lg">
+      Icon component supports right-to-left(RTL) language locales layout (auto
+      flip on RTL locales like Arabic). Consider whether you want to keep the
+      original layout on rtl if you are adding new icons.
+    </Text>
+    <Box display="flex" direction="row">
```
christianvuerings: @selena-zeng Let's remove this `Box` since the rtl will be a feature across pages

packages/gestalt/src/Icon.js
```diff
@@ -34,6 +34,20 @@ type Props = {|
 
 const IconNames = Object.keys(icons);
 
+const noFlipOnRtlIconNames = [
+  'facebook',
```
christianvuerings: @selena-zeng this is different from the approach we talked about - is this in agreement with PDS, that only these icons should not be flipped and all the other ones should?

packages/gestalt/src/Icon.css
```diff
@@ -7,3 +7,8 @@
 .iconBlock {
   composes: block from "./Layout.css";
 }
+
+html[dir="rtl"] .rtlSupport {
+  /* Flip the icon for right-to-left language locales */
+  transform: rotateY(180deg);
+}
```
christianvuerings: @selena-zeng could we add an extra breaking line at the end of this file?

image

christianvuerings:
@selena-zeng Thank you for working with PDS on this - merged!

PR #635, opened 2020-01-21
Upgrade flow to 0.116.1

jennyscript:
Updates the version of flow we're using to latest

- [x] Documentation
- [x] Tests

### DIFF
--- a/packages/gestalt/src/Contents.js
+++ b/packages/gestalt/src/Contents.js
@@ -497,6 +497,7 @@ export default class Contents extends React.Component<Props, State> {
     return (
       <div
         className={styles.container}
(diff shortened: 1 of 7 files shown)


packages/gestalt/src/Contents.js
```diff
@@ -497,6 +497,7 @@ export default class Contents extends React.Component<Props, State> {
     return (
       <div
         className={styles.container}
+        // $FlowFixMe: new errors found from flow 0.116.1 upgrade
```
christianvuerings: @jennyscript instead of adding these `$FlowFixMe`s, could we fix the errors instead?

Issue #630, opened 2020-01-16
Code sample UI obscures UI examples


ponori:
The z-index of code samples seem to be "above" certain floating UI such as toasts and modals. 

<img width="1440" alt="Screenshot 2020-01-16 15 11 51" src="https://user-images.githubusercontent.com/50343812/72570737-8a8ac980-3872-11ea-8e35-bad0722678cd.png">

<img width="1438" alt="Screenshot 2020-01-16 14 49 39" src="https://user-images.githubusercontent.com/50343812/72570519-02a4bf80-3872-11ea-851e-c9a9bc92689d.png">

I've tested only toasts and modals, but there could very likely be additional cases where this issue arises.

cc @christianvuerings

jennyscript:
Fixed by #639

PR #639, opened 2020-01-25
Fix documentation overlap issues

jennyscript:
Fixes ticket #630 by adding `Layer` tags around `Modal`, `Toast`, and `Flyout` in the examples. I also refactored some of the docs to hooks while I was in there

- [x] Documentation
- [x] Tests
- [x] Accessibility checkup

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@
 
 ### Patch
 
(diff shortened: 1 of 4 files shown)


christianvuerings:
@jennyscript Thank you!

PR #641, opened 2020-01-28
Add value for throttling scroll

tuanderful:
We wrap updateScrollPosition in `throttle` but don't specify a value for the `wait` param, so it defaults to 0, so we're not throttling anything. 16ms is an approximation for 60fps.

## TODO

- [ ] Documentation
- [ ] Tests
- [ ] Accessibility checkup


btraut:
@tuanderful Can you update the changelog to outline this change? I'm actually pretty comfortable pushing this out without an experiment both because I think it's too subtle to see and because we're likely going to phase out this version of Masonry soon.

PR #651, opened 2020-02-07
Ship "fixFetchMoreBug”

btraut:
Following the results of internal experimentation (see web_fix_masonry_fetchmore_bug), I decided it's fine to ship this fix. This will cause Masonry to fetch less aggressively on pages where there's lots of content inside the scroll container above Masonry. An example of this is the Pinterest Pin Closeup page.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@
 ### Minor
 
 - Color: update dark gray color to #111 (#648)
-- Masonry: Shipped "fixFetchMoreBug" behavior and removed flag. This makes Masonry fetch less aggressively in some cases. (#649)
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 ### Minor
 
 - Color: update dark gray color to #111 (#648)
+- Masonry: Shipped "fixFetchMoreBug" behavior and removed flag. This makes Masonry fetch less aggressively in some cases. (#649)
```
christianvuerings: @btraut looks like this should be #651

btraut: Good catch! Updating.

PR #653, opened 2020-02-10
Heading: removed unused weight prop from docs

AlbertCarreras:
`weight` is no longer a prop on `Heading`


jennyscript:
This is a tiny change, but also add a note in the Changelog for completeness

PR #654, opened 2020-02-10
Add opacity prop for Box

jennyscript:
## TODO

- [x] Documentation
- [x] Tests
- [ ] Accessibility checkup

### DIFF
--- a/packages/gestalt/src/Box.js
+++ b/packages/gestalt/src/Box.js
@@ -987,6 +987,8 @@ Box.propTypes = {
   minHeight: PropTypes.oneOfType([PropTypes.number, PropTypes.string]),
   minWidth: PropTypes.oneOfType([PropTypes.number, PropTypes.string]),
 
+  opacity: PropTypes.oneOf([0, 0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9, 1]),
+
   overflow: PropTypes.oneOf([
     'visible',
     'hidden',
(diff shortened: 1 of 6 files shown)


packages/gestalt/src/Box.js
```diff
@@ -190,6 +190,8 @@ type PropType = {
   minHeight?: number | string,
   minWidth?: number | string,
 
+  opacity?: 0 | 0.1 | 0.2 | 0.3 | 0.4 | 0.5 | 0.6 | 0.7 | 0.8 | 0.9 | 1,
```
christianvuerings: @jennyscript let's not forget about our propTypes

packages/gestalt/src/Box.js
```diff
@@ -307,6 +309,14 @@ const lgPaddingX = bind(rangeWithoutZero('lgPaddingX'), whitespace);
 const lgPaddingY = bind(rangeWithoutZero('lgPaddingY'), whitespace);
 const lgPadding = union(lgPaddingX, lgPaddingY);
 
+const opacityMap = mapClassName(name => styles[name]);
+const opacity = val => {
+  if (val > 0 && val < 1) {
```
christianvuerings: @jennyscript Can we avoid this fork?

`String(val).replace('.', '')`

```
[0, 0.3, 1].map(val => String(val).replace('.', ''))
// ["0", "03", "1"]
```

jennyscript: Range only takes number arguments. I'll see if one of the string compatible ones could use that form

jennyscript: I tried a couple hacky workarounds, but this is the most clear option so I'd prefer to keep it

PR #655, opened 2020-02-12
Button: Update border radius / small size + add dark gray option

christianvuerings:
image
combined-border


docs/src/Button.doc.js
```diff
@@ -99,7 +99,7 @@ card(
   <Example
     name="Example"
     defaultCode={`
-<Button text="Medium Sized Button" />
+<Button text="Medium Sized Button" inline />
```
christianvuerings: Changed most of the examples of the buttons to `inline` since we want to get rid of `fullWidth` buttons in gestalt

packages/gestalt/src/Button.css
```diff
@@ -8,7 +8,8 @@
 
 .button {
   composes: borderBox from "./Layout.css";
-  composes: rounded from "./Borders.css";
+  border-radius: 24px;
```
christianvuerings: @jennyscript opted for not yet introducing the new border radius in `"./Borders.css"` since that should be tackled with the `Box` upgrade.

packages/gestalt/src/Button.css
```diff
@@ -22,7 +23,7 @@
 
 .sm {
   composes: small from "./Layout.css";
-  padding: 8px 14px;
+  padding: 6px 14px;
```
christianvuerings: Smaller top/bottom padding since the height is now `32px` instead of `36px`

packages/gestalt/src/Button.js
```diff
@@ -18,7 +19,7 @@ type Props = {|
   accessibilityExpanded?: boolean,
   accessibilityHaspopup?: boolean,
   accessibilityLabel?: string,
-  color?: 'gray' | 'red' | 'blue' | 'transparent' | 'white',
+  color?: 'gray' | 'darkGray' | 'red' | 'blue' | 'transparent' | 'white',
```
christianvuerings: @jennyscript I opted for adding a new `color` instead of going for a new `selected` prop.

jennyscript: That makes sense for the API. If this is a new pattern do you think there's a helpful way to add that info to the doc?

btraut:
Are there any fears that we have incompatible custom buttons on the site along side these buttons? We saw this a bunch when we last adjusted the border radii and colors.

packages/gestalt/src/Button.js
christianvuerings: @jennyscript Updated

christianvuerings:
> Are there any fears that we have incompatible custom buttons on the site along side these buttons? We saw this a bunch when we last adjusted the border radii and colors.

@btraut Most definitely, it will be great to see where we're not yet using gestalt and it might actually drive our adoption rate.

PR #657, opened 2020-02-13
Box and Touchable: Remove roundedX shapes

jennyscript:
These are rarely used in our codebase, so I'm removing them to reduce complexity. There's already a matching PR in the other repo to remove the few places using this I'll merge before updating to this release. I'll be following up with another PR in a few days to add support for more granular control over borderRadius that replaces `shape="rounded"`

- [ ] Documentation
- [ ] Tests
- [ ] Accessibility checkup


CHANGELOG.md
```diff
@@ -6,6 +6,11 @@
     Click to see more.
   </summary>
 
+### Major
+
+- Box: BREAKING CHANGE - Removed support for `shape` prop values `roundedTop`, `roundedRight`, `roundedBottom`, and `roundedLeft` (#657)
```
christianvuerings: @jennyscript could we use less caps? Perhaps use `[Breaking]`? Also, is it truly breaking if we provide a codemod?

jennyscript: I can add a codemod, but as we discussed it won't have exactly the same behavior as before (either rounded or square subsistitute) so I'd still consider that breaking someone's code.

christianvuerings: @jennyscript Yes, it would be ideal to still add the codemod and do a major bump. The codemod will avoid JavaScript / flow and proptype errors.

CHANGELOG.md
```diff
@@ -6,6 +6,11 @@
     Click to see more.
   </summary>
 
+### Major
+
+- Box [Breaking]: Removed support for `shape` prop values `roundedTop`, `roundedRight`, `roundedBottom`, and `roundedLeft` _Has codemod_ (#657)
```
christianvuerings: @jennyscript Instead of `Has codemod`, can we give people the command they need to run for each one of these?

```sh
cd gestalt
yarn run codemod --parser=flow -t=packages/gestalt-codemods/0.108.0-0.109.0/convert-font-weight.js ~/code/repo
```

PR #659, opened 2020-02-13
IconButton/Pog: Add darkGray background option

christianvuerings:
image


packages/gestalt/src/Pog.js
```diff
@@ -33,12 +34,13 @@ type Props = {|
 |};
 
 const defaultIconButtonIconColors = {
-  transparent: 'gray',
+  blue: 'white',
```
christianvuerings: Alphabetized

PR #661, opened 2020-02-16
Button/IconButton/Pog: Convert 'darkGray' color to selected state

christianvuerings:
We decided to go for a selected state instead of a `darkGray` color option:
* It makes the intent of this new color more explicit
* Allows us to easily update the selected state moving forward

## Before
```js
<Button color={selected ? 'darkGray' : 'gray' } text="Show info" />
```
## After
```js
<Button selected={selected} text="Show info" />
```

### DIFF
--- a/docs/src/components/PropTable.js
+++ b/docs/src/components/PropTable.js
@@ -143,7 +143,6 @@ export default function PropTable({ props: properties, Component }: Props) {
                                 const elem = document.getElementById(href);
                                 if (elem) {
                                   elem.scrollIntoView({
-                                    behavior: 'smooth',
                                     block: 'start',
                                   });
                                 }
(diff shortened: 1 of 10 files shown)


packages/gestalt/src/Button.css
```diff
@@ -1,11 +1,3 @@
-:root {
```
christianvuerings: We don't need these since we're importing the colors from `Colors.css`

docs/src/components/PropTable.js
```diff
@@ -143,7 +143,6 @@ export default function PropTable({ props: properties, Component }: Props) {
                                 const elem = document.getElementById(href);
                                 if (elem) {
                                   elem.scrollIntoView({
-                                    behavior: 'smooth',
```
christianvuerings: Removing this since it took too long to scroll down to the appropriate section of the page

PR #663, opened 2020-02-19
Flyout: Remove caret, update shadow

dangerismycat:
## Flyout changes for Lego spec

This PR updates Flyout to the Lego spec:
- Remove caret
- Update shadow
- Update border radius

- [x] Documentation
- [x] Tests
- [ ] Accessibility checkup

### DIFF
--- a/packages/gestalt/src/Contents.css
+++ b/packages/gestalt/src/Contents.css
@@ -5,8 +5,10 @@
 .contents {
   composes: block from "./Layout.css";
   composes: borderBox from "./Layout.css";
-  composes: rounded from "./Borders.css";
+
+  /* composes: rounded from "./Borders.css"; TODO: switch back to "rounded" when that style is updated to 16px */
   border: 1px solid currentColor;
+  border-radius: 16px;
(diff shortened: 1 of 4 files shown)


packages/gestalt/src/Contents.css
```diff
@@ -7,7 +7,7 @@
   composes: borderBox from "./Layout.css";
   composes: rounded from "./Borders.css";
```
christianvuerings: @dangerismycat Should we also update the border radius? Was there a reason for splitting these changes out?

dangerismycat:
The border radius change seems most easily implemented all at once, by
changing the style for `rounded`. It’s unclear how much friction there will
be on that change. These two changes seemed pretty subtle, and shouldn’t
break any current layouts, so I figured I’d put these through here and
isolate the border radius change to a separate PR.

christianvuerings:
@dangerismycat we'll update `rounded` in a separate PR, @jennyscript is working on those. For now, please inline the style instead and no longer compose from `"./Borders.css"`

dangerismycat:
👍

PR #664, opened 2020-02-19
SelectList / TextField / TextArea: Update design

christianvuerings:
- [x] Documentation
- [x] Tests
- [x] Experimental evidence (required for Masonry changes)
- [x] Accessibility checkup
- [x] Design sign-off

### DIFF
--- a/packages/gestalt/src/FormElement.css
+++ b/packages/gestalt/src/FormElement.css
@@ -1,0 +1,42 @@
+.base {
+  appearance: none;
+  border-radius: 16px;
+  border-style: solid;
+  border-width: 2px;
+}
+
+.base:focus {
+  border-color: #111;
+  outline: none;
+}
+
+.normal {
+  border-color: #efefef;
+}
+
+.normal:hover:not(:focus) {
+  border-color: rgba(0, 0, 0, 0.12);
+}
+
+.errored {
+  border-color: #e60023;
+  outline: none;
+}
+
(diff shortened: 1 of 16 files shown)


packages/gestalt/src/FormElement.css
```diff
@@ -0,0 +1,33 @@
+.base {
```
christianvuerings: Since these are common across SelectList / TextField and TextArea, I'm extracting these out.

packages/gestalt/src/TextField.css
```diff
@@ -1,43 +1,10 @@
...
-.enabled {
-  composes: darkGray from "./Colors.css";
-  composes: whiteBg from "./Colors.css";
-  cursor: text;
-}
-
-.disabled {
-  composes: gray from "./Colors.css";
-  composes: lightGrayBg from "./Colors.css";
-}
```
jennyscript: Nice change to unify all these in `FormElement.css` 👍

packages/gestalt/src/FormElement.css
```diff
@@ -0,0 +1,33 @@
...
+  appearance: none;
+  border-radius: 16px;
+  border-style: solid;
+  border-width: 2px;
```
jennyscript: These are all using 2px borders so seems like another argument for adding support for 2px border width when I implement borders for Box

PR #666, opened 2020-02-19
Replace shape props with rounding props

jennyscript:
This adds a support for a larger range of border radii. When looking through the codebase, I saw we frequently used dangerous styling to apply a border radius that wasn't supported by the existing `shape="rounded"`. There are about 125 overrides in our main codebase. Many of these are a boint value between 1 and 8, so that's the range I chose for the new prop.

- [x] Documentation
- [x] Tests
- [x] Accessibility checkup
- [x] Write a codemod
- [x] Add screenshots

Box
Screen Shot 2020-02-19 at 12 55 59 PM

Mask
Screen Shot 2020-02-19 at 12 56 21 PM

Touchable
Screen Shot 2020-02-19 at 12 56 41 PM


CHANGELOG.md
```diff
@@ -10,8 +10,13 @@
 
 - Box [Breaking]: Removed support for `shape` prop values `roundedTop`, `roundedRight`, `roundedBottom`, and `roundedLeft` (#657)
 - Touchable [Breaking]: Removed support for `shape` prop values `roundedTop`, `roundedRight`, `roundedBottom`, and `roundedLeft` (#657)
-- Run codemod for breaking changes: `cd gestalt
+- Box [Breaking]: Replace `shape` prop with `rounding` (#666)
```
christianvuerings: @jennyscript Feel free to combine those: `Box / Mask / Touchable: Replace `shape` prop with `rounding` (#666)`

docs/src/Box.doc.js
```diff
@@ -171,10 +171,9 @@ card(
         href: 'absolutePositioning',
       },
       {
-        name: 'shape',
-        type: `"square" | "rounded" | "pill" | "circle"`,
-        defaultValue: 'square',
-        href: 'shapes',
+        name: 'rounding',
+        type: `"pill" | "circle" | 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8`,
```
jennyscript: Why not use one of these scales?
1) 0-12: This makes sense for `padding` and `margin` but for radius we aren't actually using values higher than 8 boints for radius unless it's a `circle` or `pill` type use case.
2) xs-xl: There's additional mental overhead to convert a pixel value to a tshirt size value. When looking at a mockup for a design it's easier to do the math of dividing by 4 than needing to look at the docs to remember which size maps to which pixel value.

packages/gestalt-codemods/0.125.0-1.0.0/__testfixtures__/convert-touchable-shape-to-rounding-square.input.js
```diff
@@ -0,0 +1,11 @@
+// @flow
+import React from 'react';
+import { Touchable } from 'gestalt';
+
+export default function Square() {
+  return (
+    <Touchable shape="square">
```
christianvuerings: @jennyscript Do we also plan on supporting `shape={rounded ? 'rounded' : 'square'}` with the codemod?

The codemod will get a more verbose but you can find an example here: https://github.com/pinterest/gestalt/blob/master/packages/gestalt-codemods/0.111.0-0.112.0/text-remove-semibold.js#L73

An alternative is to update those instances manually

packages/gestalt-codemods/0.125.0-1.0.0/convert-shape-to-rounding.js
```diff
@@ -0,0 +1,104 @@
+/**
+ * Converts
+ *  <Box shape="roundedTop" /> or <Touchable shape="roundedTop" /> (and other roundedX variants)
```
christianvuerings: @jennyscript can we update the description?

packages/gestalt/src/Borders.css
```diff
@@ -48,8 +43,40 @@ html[dir="rtl"] .borderLeft {
   border-radius: 999px;
 }
 
-.rounded {
-  border-radius: var(--border-radius);
+.rounding0 {
+  border-radius: 0;
```
christianvuerings: @jennyscript instead, could we set these to:

```css
border-radius: calc(var(--bt) * 0);
border-radius: calc(var(--bt) * 1);
```
It makes it more clear that we're using boints for these increases.

Also see `packages/gestalt/src/boxWhitespace.css`

packages/gestalt-codemods/0.125.0-1.0.0/__testfixtures__/convert-touchable-shape-to-rounding-square.input.js
jennyscript: Updated to support ternaries

PR #667, opened 2020-02-19
Masonry: Allowing string enums for Masonry layout

btraut:
Masonry currently allows three ways to specify which layout you'd like to use:

1. Use a `Symbol` as part of `legacyLayoutSymbols.js`, although this is not exported so only relevant inside the package.
2. Use one of class-based layouts in the `/layouts` folder. These _are_ exported but are considered "legacy".
3. Specify nothing and get the default columnar layout.

Because extra exports are annoying and we're already split on what to use, I'd like to update to one true system: strings. I'd like layout to eventually be `type Layout = 'basic' | 'uniformRow' | 'flexible';`. This PR is the first step -- allow the string versions. I already have a pinboard diff that switches all uses over to it, and we can follow up by removing the `Symbol` and class versions once that PR lands.

One other future goal: I'd like to remove the `flexible` bool prop. When true, Masonry currently uses a completely different layout engine. For instance, if the user asks for uniform row layout and specifies `flexible={true}`, the uniform layout flag will be ignored. Once this lands, I intend to convert all instances of `flexible={true}` to `layout="flexible"`.


christianvuerings:
@btraut Looks great to me, let's land this with its own release

PR #671, opened 2020-02-20
Internal: update dependencies

christianvuerings:
Updates all of our internal dependencies except for `flow` and `react`. Those 2 are peer dependencies and we should update the Pinterest codebase at the same time.


.stylelintrc.json
```diff
@@ -9,6 +9,12 @@
         "ignoreProperties": ["composes"]
       }
     ],
+    "value-keyword-case": [
+      "lower",
+      {
+        "ignoreProperties": ["composes"]
```
christianvuerings: More info at: https://stylelint.io/user-guide/rules/value-keyword-case
Since we use CSS modules, we need to ignore `composes`, otherwise we get a bunch of these errors:

image

PR #675, opened 2020-02-21
Internal: Flowtype CSS

christianvuerings:
Be more strict about the CSS we import into JS. We had a couple of instances where we were importing incorrect classnames that did not exist.

It uses [css-modules-flow-types](https://github.com/skovhus/css-modules-flow-types) under the hood and we expose the following commands:

```
# Generate CSS flowtypes
yarn run flow-generate:css

# Generate CSS flowtypes and run flow
"flow:css": "yarn flow-generate:css && yarn flow",
```

image


.flowconfig
```diff
@@ -11,8 +11,6 @@
 include_warnings=true
 suppress_comment=\\(.\\|\n\\)*\\$FlowFixMe
 suppress_comment=\\(.\\|\n\\)*\\$FlowIssue
-# necessary for webpack loader syntax.
-module.name_mapper='^.*\.css$' -> '<PROJECT_ROOT>/flow/stubs/CSSModuleStub.js'
```
christianvuerings: We no longer need this custom loader since webpack is now abstracted from us via Create React App. It also failed to work with `css-modules-flow-types-cli`

packages/gestalt/src/Avatar.js
```diff
@@ -40,7 +40,6 @@ const DefaultAvatar = ({ name }: { name: string }) => {
             className={[
               typography.antialiased,
               typography.sansSerif,
-              typography.leadingSmall,
```
christianvuerings: Not defined in CSS

packages/gestalt/src/Image.js
```diff
@@ -105,7 +105,7 @@ export default class Image extends React.PureComponent<Props> {
     return isScaledImage ? (
       <div
         aria-label={alt}
-        className={styles[fit]}
+        className={fit === 'contain' || fit === 'cover' ? styles[fit] : null}
```
christianvuerings: `fit` could also be `none`, however no `none` existed in the CSS file.

packages/gestalt/src/Text.js
```diff
@@ -102,7 +102,6 @@ export default function Text({
     align === 'right' && typography.alignRight,
     overflow === 'breakWord' && typography.breakWord,
     italic && typography.fontStyleItalic,
-    !italic && typography.fontStyleNormal,
```
christianvuerings: `fontStyleNormal` didn't exist

packages/gestalt/src/Typography.css
```diff
@@ -106,6 +106,6 @@ html[dir="rtl"] .alignRight {
   text-align: center;
 }
 
-.textJustify {
+.alignJustify {
```
christianvuerings: Misspelling

PR #678, opened 2020-02-21
Add borderSize prop for Box

jennyscript:
This adds a new prop to Box to support borders. Boxes may have a 1px or 2px solid light gray border

-`borderWidth`: "sm" (1px) | "lg" (2px) | "none" (not set)

Screen Shot 2020-02-21 at 3 47 27 PM

- [x] Documentation
- [x] Tests
- [x] Accessibility checkup


docs/src/Box.doc.js
```diff
@@ -494,6 +502,48 @@ card(
...
+card(
+  <Card
+    description={`
+    Borders are controlled by the "borderSize" prop. Setting a size enables a solid light gray color in that width.
+  `}
+    id="border"
+    name="Borders"
+  >
+    <Box display="flex" wrap>
+      {['sm', 'lg'].map((size, i) => (
```
jennyscript: I made this custom instead of using `Combination` because the checkered background made the light gray border hard to see.

docs/src/Box.doc.js
```diff
@@ -494,6 +502,48 @@ card(
...
+        >
+          <Box marginBottom={2}>
+            <Text align="center" size="sm">
+              borderSize=&quot;{size}&quot;
+            </Text>
+          </Box>
+          <Box position="relative" padding={4}>
+            <Box position="relative">
+              <Box
+                color="gray"
```
christianvuerings: @jennyscript Should we make this color `white` so it's easier to see the difference between 1 and 2px? (since the border color is lightgray)

docs/src/Box.doc.js
```diff
@@ -494,6 +502,48 @@ card(
...
 
+card(
+  <Card
+    description={`
+    Borders are controlled by the "borderSize" prop. Setting a size enables a solid light gray color in that width.
+  `}
+    id="border"
+    name="Borders"
+  >
+    <Box display="flex" wrap>
```
christianvuerings: @jennyscript should we use the `Combinations` component like we do in other places?

PR #684, opened 2020-02-25
Tooltip: Add ability to hover over tooltip and add a clickable link

anniejiwon:
## TODO

- [x] Documentation
- [x] Tests
- [ ] Accessibility checkup

notice i can highlight the text because the tooltip stays open when i hover on it:
Screen Shot 2020-02-24 at 4 15 38 PM


packages/gestalt/src/Tooltip.js
```diff
@@ -3,62 +3,128 @@
...
+      };
+    default:
+      throw new Error();
+  }
+};
 
-export default class Tooltip extends React.Component<Props, State> {
-  state = {
-    hovered: false,
+export default function Tooltip({ children, href, hrefText, inline, text }) {
```
christianvuerings: @anniejiwon We probably should type these: `{ children, href, hrefText, inline, text }: Props`

packages/gestalt/src/Tooltip.js
```diff
@@ -3,62 +3,128 @@
...
+
+  const handleIconMouseEnter = () => {
+    dispatch({ type: 'hoverInIcon' });
   };
 
-  childRef: {| current: null | React.ElementRef<'div'> |} = React.createRef();
+  const handleIconMouseLeave = () => {
+    setTimeout(() => {
+      dispatch({ type: 'hoverOutIcon' });
+    }, 75);
```
christianvuerings: @anniejiwon Since we use the number `75` twice and they should be the same, could you extract it into a `const`?

christianvuerings:
@anniejiwon Could you check with Annie T whether this is ok from a design perspective? The left/top margin looks very small around `Logout`

image

docs/src/Tooltip.doc.js
```diff
@@ -60,6 +60,22 @@ card(
   />
 );
 
+card(
+  <Example
+    name="Tooltips with link"
+    description="Passing a href prop will render a Learn More link below the provided tooltip text. Be sure to internationalize your hrefText."
+    defaultCode={`
+<Tooltip inline href="pinterest.com/anniejiwon" hrefText="Learn more" text="Logout">
```
christianvuerings: @anniejiwon could we come up with a better example? `Learn more` is not a good descriptive link. See [Failure of Success Criterion 2.4.9 due to using a non-specific link such as "click here" or "more"](https://www.w3.org/TR/WCAG20-TECHS/F84.html)

`Learn more about logout` / `Logout documentation` would be better `hrefText`'s

packages/gestalt/src/Tooltip.js
```diff
@@ -3,62 +3,128 @@
...
-          ref={this.childRef}
+  return (
+    <Box display={inline ? 'inlineBlock' : 'block'}>
+      <Box
+        ref={childRef}
+        onMouseEnter={handleIconMouseEnter}
+        onMouseLeave={handleIconMouseLeave}
+        onFocus={handleIconMouseEnter}
+        onBlur={handleIconMouseLeave}
+        role="tooltip-container"
```
christianvuerings: @anniejiwon this is an invalid role, can we please remove it? https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques#Roles

We also have an `onFocus` and `onBlur` event but no `tabIndex` set to `0` - could we add that `tabIndex`?

packages/gestalt/src/Tooltip.js
```diff
@@ -3,62 +3,128 @@
...
-          </Controller>
-        )}
-      </Box>
-    );
-  }
+            <Text color="white" size="xs">
+              {text}
+            </Text>
+            {allowTooltipHover && (
+              <Box paddingY={1} role="href-link">
```
christianvuerings: @anniejiwon this is an invalid ARIA role, can we please remove it? https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques#Roles

Also see the second rule of ARIA: https://www.w3.org/TR/using-aria/#secondrule `Do not change native semantics, unless you really have to.`

christianvuerings:
@anniejiwon thank you for these changes! Left some feedback and please ping me if you have any questions about them.

docs/src/Tooltip.doc.js
```diff
@@ -39,6 +39,12 @@ card(
         description:
           'String that is shown as addition information in a tooltip bubble to describe the child. Always localize the text.',
       },
+      {
+        name: 'link',
+        type: 'React.Node',
+        required: false,
+        description: 'A link element',
```
christianvuerings: @anniejiwon could we be more descriptive? E.g. `A link which will show on the bottom of a tooltip`

docs/src/Tooltip.doc.js
```diff
@@ -60,6 +66,32 @@ card(
   />
 );
 
+card(
+  <Example
+    name="Tooltips with link"
+    description="Passing a href prop will render a Learn More link below the provided tooltip text. Be sure to internationalize your hrefText."
```
christianvuerings: @anniejiwon we no longer have `href` and `hrefText` - can we update this?

packages/gestalt/src/Tooltip.jsdom.test.js
```diff
@@ -12,3 +15,35 @@ test('Tooltip renders', () => {
...
+      }
+      text="This is a tooltip"
+    >
+      <div>Hi</div>
+    </Tooltip>
+  );
+
+  fireEvent.mouseEnter(getByRole('tooltip'));
+  expect(getByText('Learn more about logout')).toBeVisible();
+  expect(container.firstChild).toMatchSnapshot();
```
christianvuerings: @anniejiwon can we remove these extra snapshot tests + add a test to make sure the link `https://pinterest.com` is rendered?

packages/gestalt/src/Tooltip.jsdom.test.js
```diff
@@ -12,3 +15,35 @@ test('Tooltip renders', () => {
   const tree = component.toJSON();
   expect(tree).toMatchSnapshot();
 });
+
+test('Tooltip renders the hrefText when hovered', () => {
```
christianvuerings: @anniejiwon this should get updated, we no longer have `hrefText`

packages/gestalt/src/Tooltip.jsdom.test.js
```diff
@@ -12,3 +15,35 @@ test('Tooltip renders', () => {
...
+
+test('Tooltip should render as expected when hovered', () => {
+  const { container, getByRole } = render(
+    <Tooltip text="This is a tooltip">
+      <div>Hi</div>
+    </Tooltip>
+  );
+
+  fireEvent.mouseEnter(getByRole('tooltip'));
+  expect(container.firstChild).toMatchSnapshot();
```
christianvuerings: @anniejiwon instead, can we check for `This is a tooltip`

PR #689, opened 2020-02-25
Button: update horizontal padding to 12px

christianvuerings:
Before the spacing `left/right` was `14px`, now it's `12px` which is the same as our Figma files. Design signed off on these changes.

Before: https://deploy-preview-688--gestalt.netlify.com/#/Button
After: https://deploy-preview-689--gestalt.netlify.com/#/Button

### DIFF
--- a/packages/gestalt/src/Button.css
+++ b/packages/gestalt/src/Button.css
@@ -10,17 +10,20 @@
 
 .sm {
   composes: small from "./Layout.css";
-  padding: 6px 14px;
+  composes: paddingX3 from "./boxWhitespace.css";
+  composes: paddingY1 from "./boxWhitespace.css";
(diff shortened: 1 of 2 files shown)


packages/gestalt/src/Button.css
```diff
@@ -10,17 +10,20 @@
 
 .sm {
   composes: small from "./Layout.css";
-  padding: 6px 14px;
+  composes: paddingX3 from "./boxWhitespace.css";
+  composes: paddingY1 from "./boxWhitespace.css";
```
christianvuerings: Since we have a `min-height` set on each button, the height doesn't change when we update the vertical padding. This approach ultimately results in less CSS being put out.

PR #693, opened 2020-02-26
Typography updates

jennyscript:
This change reduces the sizes available on `Heading` and `Text` from 5 options to 3 options. Sizes map as follows:

## Heading

Before | After
-- | --
-- | sm: 20px
xs: 24px | md: 28 px
sm: 36px | lg: 36 px (default)
md: 48px (default) | (lg: 36 px)
lg: 64px | (lg: 36 px)
xl: 96px | (lg: 36 px)

## Text

Before | After
-- | --
xs: 12px | sm: 12px
sm: 14px | md: 14px
md: 16px (default) | lg: 16px (default)
lg: 18px | (lg: 16px)
xl: 21px | (lg: 16px)

This PR also includes codemods to make the updates 
1. `cd gestalt; yarn run codemod --parser=flow -t=packages/gestalt-codemods/1.7.0-1.8.0/heading-size-replace.js ~/code/repo`
2. `yarn run codemod --parser=flow -t=packages/gestalt-codemods/1.7.0-1.8.0/text-size-replace.js ~/code/repo`

- [x] Documentation
- [x] Tests
- [x] Accessibility checkup

PR #696, opened 2020-02-27
Added codemod to remove responsive sizes and support breaking changes…

AlbertCarreras:
… from #693

This change provides support for breaking changes on #693, where responsive size attributes (`smSize`, `mdSize`, `lgSize`) on `Text` and `Heading` are deprecated. 

The `/1.8.0-1.9.0/remove-responsive-text-sizing` codemod executes the following changes on the `Text` and `Heading` component (the component/s):

- Removes responsive size attributes (`smSize`, `mdSize`, `lgSize`) attributes from the component except the `size` attribute.
- If the component has no `size` attribute, it creates a new `size` attribute with the value from the largest responsive size existing (`lgSize` < `smSize` < `mdSize`)
- If the resulting `size` has the default value ('md'), it's also removed . 

The codemod does not change the component if:
- it has the attribute `size` but none of the responsive size attributes.

The codemod throws an error if:
- it has spread props; spread props are opaque to scripts.
- it any of the existing responsive size attributes (`lgSize`, `smSize`, `mdSize`) is not a string (for instance, size={__}). 

cd gestalt; yarn run codemod --parser=flow -t=packages/gestalt-codemods/1.8.0-1.9.0/remove-responsive-text-sizing.js  ~/code/repo

## TODO

- [x] Documentation
- [x] Tests
- [x] ~~Experimental evidence (required for Masonry changes)~~
- [x] ~~Accessibility checkup~~


christianvuerings:
@AlbertCarreras Could you address the linter errors / warnings? 
image

package-lock.json
```diff
@@ -0,0 +1,12541 @@
+{
```
christianvuerings: @AlbertCarreras Can we remove this `package-lock.json`?

PR #705, opened 2020-02-28
SelectList / TextField / TextArea: Add label and helperText props

christianvuerings:
Adds a `label` and `helperText` prop to SelectList / TextField and TextArea.

That way we can enforce consistent label styling moving forward and add helper text for fields that require it.

image


docs/src/SelectList.doc.js
```diff
@@ -24,18 +24,28 @@ card(
       },
       {
         name: 'errorMessage',
-        type: '?string',
```
christianvuerings: Removed these, since we don't do this anywhere else. We should use the `required` option instead

packages/gestalt/src/FormElement.css
```diff
@@ -14,7 +14,7 @@
   border-color: #efefef;
 }
 
-.normal:hover:not(:focus) {
+.normal:hover:not(:focus):not(.disabled) {
```
christianvuerings: Fixes a bug where hovering a disabled field still shows the hover border

packages/gestalt/src/TextArea.css.flow
```diff
@@ -1,9 +1,5 @@
 // @flow strict
 /* This file is automatically generated by css-modules-flow-types */
 declare module.exports: {|
-  +'disabled': string,
```
christianvuerings: All of these classes have moved to `FormElement.css`

packages/gestalt/src/TextArea.js
```diff
@@ -3,16 +3,19 @@
 import * as React from 'react';
```
christianvuerings: One interesting bug I found in Chrome is that they always add an extra 4px of spacing to the bottom of each `<textarea />` https://bugs.chromium.org/p/chromium/issues/detail?id=89530

christianvuerings: Not going to address that in this PR since it'll break in other browsers.

packages/gestalt/src/FormLabel.css
```diff
@@ -0,0 +1,4 @@
+.formLabel {
+  composes: paddingX2 from "./boxWhitespace.css";
+  padding-bottom: 8px;
```
christianvuerings: Not using `Box` in this case since we can't set asymmetrical padding. We want only `padding-bottom` for the following reasons:

* Components in gestalt should not have extra spacing around them (reason why we don't make it symmetrical)
* Margin bottom is not good enough - it results in "flickering" when you hover from the label to the form field

christianvuerings: Example of the "flickering" behavior
flickering-margin-bottom

jennyscript: That flickering is subtle! Good catch

docs/src/TextArea.doc.js
```diff
@@ -119,17 +123,36 @@ card(
...
+card(
+  <Example
+    id="helperText"
+    name="Example: Helper Text"
+    description={`Whenever you want to provide more information about a form field, you should use \`helperText\`.`}
+    defaultCode={`
+function Example(props) {
+  const [value, setValue] = React.useState('')
+  return (
+    <Box padding={2} color="white">
```
jennyscript: The white Box really improves readability here and would be nice on all the other examples as well. What about adding it to the others OR making the change we discussed before about making the checkerboard optional?

christianvuerings: @jennyscript added in https://github.com/pinterest/gestalt/pull/714 - plan is to merge this PR first and then update the examples after

PR #707, opened 2020-02-29
Make typography changes more backwards-compatible

jennyscript:
After running the codemods on the repo using gestalt, we found many places where the `Text size="xl"` change was too big. I made two changes here to address this:
1. I updated the codemod to change any XL `Text` that is also `weight="bold"` and has props that are valid for `Heading` update to `Heading size="sm"` since that's only a 1px difference
2. I left `Text size="xl"` in as an undocumented deprecated value so we can clean up places not safe to codemod manually before removing

- [x] Documentation
- [x] Tests
- [x] Accessibility checkup


christianvuerings:
@jennyscript an alternative to re-adding the `xl` font size to `Text` would be to add an inline `<span />` with `font-size`.

## Pros / Cons

### Pros
* No longer have to include the non-supported `21px` in Gestalt
* Not have to do a follow up where we remove it from Gestalt

### Cons
* Gestalt usage takes a slight hit

## Before / After

### Before
Align right/font-weight normal
```jsx
<Text size="xl" align="right">
  Wunderbar
</Text>
```

### After
Default on `Text` changes to `lg`
```jsx
<Text align="right">
  <span style={{ fontSize: 21 }}>
    Wunderbar
  </span>
</Text>
```

I believe it would be a better solution than re-adding `xl` but still r+'ing this PR to unblock you and let you drive the decision.

---

We could also extract the style declaration into a `dangerouslyUnsupportedXLSize` to make the exception grep-able so teams can update it themselves.

```jsx
const dangerouslyUnsupportedXLSize = { fontSize: 21 };
// ...
<span style={dangerouslyUnsupportedXLSize}>
```

jennyscript:
I like the `dangerouslyUnsupportedXLSize` idea. I'm going to merge this as-is since I've already made a 1.9.1 release with the changes in this commit then I'll do a follow up to add a codemod for places still using `size="xl"`

PR #713, opened 2020-03-03
Form fields: Add "lg" size option

christianvuerings:
| Size | Min Height |
| ------- | ----- |
| `md` |`40px` |
| `lg` | `48px` |

Touches the following components

* `<SearchField />`
* `<SelectList />`
* `<Tabs />`
* `<TextField />`

form-fields-updated-heights

### DIFF
--- a/packages/gestalt/src/SegmentedControl.css
+++ b/packages/gestalt/src/SegmentedControl.css
@@ -8,14 +8,6 @@
   user-select: none;
 }
 
-.md {
-  composes: medium from "./Layout.css";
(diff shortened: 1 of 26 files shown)


packages/gestalt/src/SegmentedControl.css
```diff
@@ -8,14 +8,6 @@
   user-select: none;
 }
 
-.md {
```
christianvuerings: Imported `Layout.css` in JavaScript instead

PR #731, opened 2020-03-06
Tabs: Add white background, background colors for active/focus/hover states

dangerismycat:
This PR adds "overlays" (not actual overlays, represented as hex colors) for active and focus/hover states for tabs in the Tabs component.. Additionally, to ensure text is always readable regardless of what color background Tabs are used on, this adds a white background to each tab. Finally, this adds a tab min width of 60px — since this matches Button's min width, I moved things around a bit so both are composing from the same style.

## TODO

- [ ] Documentation
- [ ] Tests
- [ ] Accessibility checkup

### DIFF
--- a/packages/gestalt/src/Tabs.css
+++ b/packages/gestalt/src/Tabs.css
@@ -19,7 +19,7 @@
 }
 
 .tabIsNotActive {
-  composes: transparentBg gray from "./Colors.css";
+  composes: whiteBg from "./Colors.css";
 }
 
 .tab:focus,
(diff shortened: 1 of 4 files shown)


dangerismycat:
don't review this yet, forgot to add min width as well

packages/gestalt/src/Tabs.css
```diff
@@ -29,9 +29,14 @@
 
 .tabIsNotActive:focus,
 .tabIsNotActive:hover {
+  background-color: #f1f1f1;
   color: #111;
 }
 
+.tabIsNotActive:active {
+  background-color: #e8e8e8;
```
jennyscript: I see this and #f1f1f1 are also used in `Button.css`. Maybe they should be named colors in `Colors.scss` for sharing?

PR #735, opened 2020-03-07
Docs: Open in CodeSandbox & remove checkerbox from example

christianvuerings:
## Open examples in CodeSandbox

Makes it way easier for developers to:
* Prototype
* Share their code with others

gestalt-code-sandbox

## Other notes

* Removed checkerbox from examples since it's pretty hard to read content on top of it (we still use it for `<Combinantion />`)
* Had to use a 3rd party service to host the stock images. With CodeSandbox we no longer allow dynamic properties in our code.


btraut:
Great job! I love to see the improvements roll in.

PR #736, opened 2020-03-09
Modal: lower bottom heading spacing for alert dialogs

christianvuerings:
Design asked us to update the spacing on `heading` for alert dialogs.

## Before
Screen Shot 2020-03-09 at 9 17 01 AM

## After
Screen Shot 2020-03-09 at 9 16 46 AM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
 
 ### Minor
 
+- Modal: lower bottom heading spacing for alert dialogs (#736)
+
 ### Patch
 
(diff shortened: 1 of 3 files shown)


chrislloyd:
Yeah, if anything the buttons look odd to me.

chrislloyd:
And dang, this example gets me every time.

christianvuerings:
@btraut @chrislloyd thank you both for raising the feedback. Did a double check with design and we're going to leave it as is

PR #751, opened 2020-03-16
Internal: upgrade packages

christianvuerings:
Upgrade internal packages


docs/package.json
```diff
@@ -16,8 +16,8 @@
     "history": "^4.10.1",
     "lz-string": "^1.4.4",
     "marked": "^0.8.0",
-    "react": "^16.8.5",
-    "react-dom": "^16.12.0",
+    "react": "^16.13.0",
```
christianvuerings: Only update the docs since these include react as a dependency, not a peer dependency. That way we check whether everything still works in the latest version of React.

PR #759, opened 2020-03-20
SelectList: Refactor placeholder for better React support

rootvar:
React warns against using "selected" prop directly on an "option". We currently only use "selected" for the placeholder feature.
This change removes "selected" and instead uses "value" prop when a placeholder is used.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 ### Minor
 
 ### Patch
+- SelectList: Remove selected prop from the placeholder option tag for better React support (#759)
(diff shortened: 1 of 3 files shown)


rootvar:
@christianvuerings do I need to bump Gestalt version as well? Does every PR need to update the changelog and bump the version?

christianvuerings:
@rootvar No need to bump the Gestalt version. Every PR should update the changelog.md file
image

rootvar:
@christianvuerings got it! thank you

PR #766, opened 2020-03-24
Button: Add iconEnd prop

dangerismycat:
This adds an `iconEnd` prop to Button, allowing for an icon to be displayed after the button text. This is intentionally left wide open here; we will use linter rules in Pinboard to restrict usage to PDS guidelines (only size `md`, only color white, only `arrow-down` icon).

- [x] Documentation
- [ ] Tests
- [ ] Accessibility checkup


packages/gestalt/src/Button.js
```diff
@@ -72,19 +95,22 @@ export default function Button(props: Props) {
...
-          (selected && 'white') ||
-          textColor ||
-          DEFAULT_TEXT_COLORS[color]
-        }
-        overflow="normal"
-        weight="bold"
-      >
-        {text}
-      </Text>
+      {iconEnd ? (
```
christianvuerings: @dangerismycat could you also add a simple (snapshot) test to verify the icon shows up when we add `iconEnd`?

PR #767, opened 2020-03-24
Heading: Add 'align' prop

aiweithesushigirl:
## TODO

- [x] Documentation
- [x] Tests
- [x] Experimental evidence (required for Masonry changes)
- [x] Accessibility checkup

Screen Shot 2020-03-24 at 10 48 33 AM


packages/gestalt/src/Heading.js
```diff
@@ -7,6 +7,7 @@ import styles from './Heading.css';
 import typography from './Typography.css';
 
 type Props = {|
+  align?: 'left' | 'right' | 'center' | 'justify',
```
christianvuerings: @aiweithesushigirl could we also update the `Heading.propTypes` below? As a UI library, we need to support not just Flow but also other usages.

aiweithesushigirl: Just updated!

PR #772, opened 2020-03-25
RadioButton/Checkbox: Moved shared classes from RadioButton.css to FormElement.css

AlbertCarreras:
- RadioButton/Checkbox: Moved shared classes from RadioButton.css to FormElement.css

## TODO

- [x] Documentation
~~- [ ] Tests~~
~~- [ ] Experimental evidence (required for Masonry changes)~~
~~- [ ] Accessibility checkup~~

### DIFF
--- a/packages/gestalt/src/FormElement.css
+++ b/packages/gestalt/src/FormElement.css
@@ -1,9 +1,22 @@
+:root {
+  --size-sm: 16px;
+  --size-md: 24px;
+}
+
 .base {
   composes: accessibilityOutline from "./Focus.css";
   appearance: none;
   border-radius: 16px;
   border-style: solid;
   border-width: 2px;
+}
+
+.input {
+  composes: absolute from "./Layout.css";
+  composes: m0 from "./Whitespace.css";
+  appearance: none;
+  opacity: 0;
+  outline: 0;
 }
 
 .normal {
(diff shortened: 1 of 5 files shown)


packages/gestalt/src/FormElement.css
```diff
@@ -36,3 +49,14 @@
   composes: gray from "./Colors.css";
   composes: lightGrayBg from "./Colors.css";
 }
+
```
christianvuerings: @AlbertCarreras can you ensure you have `stylelint` installed as a vscode extension? Looks like we have double line breaks here

packages/gestalt/src/FormElement.css
```diff
@@ -1,3 +1,8 @@
+:root {
```
christianvuerings: Previously `FormElement` was for all the elements that included text in them - e.g. `TextField` / `TextArea` / `SelectList` - I'm not sure whether adding `RadioButton` and `CheckBox` in here makes sense.

Instead, can we create a `RadioButtonCheckbox.css` file which lists the shared styles? It should also be fine to directly import this new CSS file into `RadioButton.js` and `CheckBox.js`. That way we have even less code we need to maintain.

AlbertCarreras:
new PR

PR #776, opened 2020-03-26
Borders: Update lightgray border color to #ddd

christianvuerings:
Request from design - this improves accessibility and visibility of our borders


packages/gestalt/src/Box.js
```diff
@@ -493,7 +493,7 @@ const propToFn = {
         // default: none
       })(value),
       fromClassName(borders.solid),
-      fromClassName(colors.lightGrayBorder),
+      fromClassName(borders.borderColorLightGray),
```
christianvuerings: Looks like we were defining the same CSS class twice, once in `Colors.css` and the other time in `Borders.css`

PR #777, opened 2020-03-26
Modal: Update OutsideEventBehavior to work well with Portals

emilyalwayss:
Only trigger outside event handler if click is outside in the react component tree

Before:
outside-event-behavior-before

After: 
outside-event-behavior-after

### DIFF
--- a/docs/src/Modal.doc.js
+++ b/docs/src/Modal.doc.js
@@ -536,7 +536,6 @@ card(
 function Example(props) {
   const [showModal, setShowModal] = React.useState(false);
   const [showFlyout, setShowFlyout] = React.useState(false);
-
   const anchorRef = React.useRef();
 
   return (
--- a/docs/src/Modal.doc.js
+++ b/docs/src/Modal.doc.js
@@ -529,6 +529,72 @@ function Example(props) {
 );
 
 card(
+  <Example
+    name="Outside Event Exception Test"
+    description={`
+      Testing that outside event is working as expected with portals
+    `}
+    defaultCode={`
+function Example(props) {
+  const [showModal, setShowModal] = React.useState(false);
+  const [hasLoaded, setHasLoaded] = React.useState(false);
+  const [showFlyout, setShowFlyout] = React.useState(false);
+
+  const anchorRef = React.useRef();
+
+  const handleToggleModal = () => {
+    setShowModal(!showModal);
+    setHasLoaded(false);
+  }
+
+  return (
+    <Box marginLeft={-1} marginRight={-1}>
+      <Box padding={1}>
+        <Button
+          text="View Modal"
+          onClick={handleToggleModal}
+        />
+        {showModal && (
+          <Modal
+            accessibilityModalLabel="View flyout modal"
+            heading="Modal"
+            onDismiss={handleToggleModal}
+          >
+            <Box margin={4} ref={anchorRef}>
+              <Button text="Open Flyout" onClick={() => setShowFlyout(true)} />
+            </Box>
+            {showFlyout && (
+              <Layer>
+                <Flyout
+                  anchor={anchorRef.current}
+                  color="blue"
+                  idealDirection="up"
+                  onDismiss={() => setShowFlyout(false)}
+                  positionRelativeToAnchor={false}
+                  showCaret
+                  shouldFocus={false}
+                  size="md"
+                >
+                  <Box padding={3}>
+                    <Text color="white" weight="bold">
+                      This flyout is in a React portal.
+                    </Text>
+                    <Button text="Click me and the modal should not close" />
+                  </Box>
+                </Flyout>
+              </Layer>
+            )}
+          </Modal>
+        )}
+      </Box>
+    </Box>
+  );
+}
+`}
+  />
+);
+
+card(
   <Card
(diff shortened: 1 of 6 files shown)


docs/src/Modal.doc.js
```diff
@@ -528,6 +528,72 @@ function Example(props) {
   />
 );
 
+card(
+  <Example
+    name="Outside Event Exception Test"
```
christianvuerings: @emilyalwayss Would be wonderful if we could write an actual jest test for this use case (with https://github.com/testing-library/react-testing-library)

Would you be able to update the name (title) and description to something like the following:
- name: `Flyout inside of Modal Example`
-  description: `Opens a Flyout inside of a Modal and verifies that outside events work as expected with portals`

"Tests" should live in our tests and not our docs.

docs/src/Modal.doc.js
```diff
@@ -528,6 +528,72 @@ function Example(props) {
...
+    description={`
+      Testing that outside event is working as expected with portals
+    `}
+    defaultCode={`
+function Example(props) {
+  const [showModal, setShowModal] = React.useState(false);
+  const [hasLoaded, setHasLoaded] = React.useState(false);
```
christianvuerings: @emilyalwayss can we remove `hasLoaded` - doesn't look like it's used in this example.

docs/src/Modal.doc.js
```diff
@@ -528,6 +528,72 @@ function Example(props) {
...
+
+  return (
+    <Box marginLeft={-1} marginRight={-1}>
+      <Box padding={1}>
+        <Button
+          text="View Modal"
+          onClick={handleToggleModal}
+        />
+        {showModal && (
+          <Modal
```
christianvuerings: @emilyalwayss can we wrap `Modal` in a `<Layer />`? Otherwise, it'll show behind the code example in the docs:
image

docs/src/Modal.doc.js
```diff
@@ -528,6 +528,72 @@ function Example(props) {
...
+            accessibilityModalLabel="View flyout modal"
+            heading="Modal"
+            onDismiss={handleToggleModal}
+          >
+            <Box margin={4} ref={anchorRef}>
+              <Button text="Open Flyout" onClick={() => setShowFlyout(true)} />
```
christianvuerings: @emilyalwayss there seems to be a bug when I take the following steps:
1. Open the modal
2. Click the "Show flyout" button => Flyout shows correctly
3. Click outside the modal (closes the modal)
4. Re-open the modal
5. Click the "Show flyout" button => Flyout does not show up

Could you take a look and fix this bug?

docs/src/Modal.doc.js
```diff
@@ -528,6 +528,72 @@ function Example(props) {
...
+                  color="blue"
+                  idealDirection="up"
+                  onDismiss={() => setShowFlyout(false)}
+                  positionRelativeToAnchor={false}
+                  showCaret
+                  shouldFocus={false}
+                  size="md"
+                >
+                  <Box padding={3}>
+                    <Text color="white" weight="bold">
```
christianvuerings: @emilyalwayss could we clean the spacing up a bit? 

```
<Box padding={3}>
  <Text color="white" weight="bold">
    This flyout is in a React portal.
  </Text>
  <Box marginTop={3}>
    <Button text="Click me and the modal should not close" />
  </Box>
</Box>
```

Before:
image

After:
image

packages/gestalt/src/behaviors/OutsideEventBehavior.js
```diff
@@ -1,39 +1,35 @@
...
-  handleClickEvent = (event: MouseEvent) => {
-    // eslint-disable-next-line react/no-find-dom-node
-    const el = findDOMNode(this);
-    if (
-      !this.props.onClick ||
-      !el ||
-      (event.target instanceof Node && el.contains(event.target))
-    ) {
+const OutsideEventBehavior = ({ children, onClick }: Props) => {
+  let isClickedInside = false;
```
christianvuerings: @emilyalwayss I'm curious to see if we can use React `useState` for this variable instead. Could you give that a try?

packages/gestalt/src/behaviors/OutsideEventBehavior.js
```diff
@@ -1,39 +1,35 @@
...
+  const handleClickEvent = (event: MouseEvent) => {
```
christianvuerings: @emilyalwayss can we rename this to `handleDocumentClickEvent`? Differentiates with the other `handleClick` we now have.

packages/gestalt/src/behaviors/OutsideEventBehavior.js
emilyalwayss: @christianvuerings What are some of the benefits of storing it as a state variable? I don't think we want to trigger a re-render when setting this variable.

Storing as a state var also introduces some unexpected behavior (ie. clicking outside of the modal only clears the flyout). However, if you feel we should do it this way, I will look into it!

docs/src/Modal.doc.js
emilyalwayss: fixed! good catch

PR #778, opened 2020-03-26
IconButton/Pog: Add "red" backgroundColor + update icon sizes

dangerismycat:
About a year ago, I forked a handful of Gestalt components into Pinboard to make changes necessary for Lego PinRep and Closeup. This PR will allow the deletion of the last of those forks, by making two changes:
- Add the background color "red" to both IconButton and Pog, including active/focus/hover states
- Update icon sizes app-wide

- [x] Documentation
- [ ] Tests
- [ ] Accessibility checkup

## Before
image

## After
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@
 
 ### Minor
 
+- IconButton: Add bgColor "red", add iconSize prop (#778)
+- Pog: Add bgColor "red" (#778)
+
 ### Patch
 
 </details>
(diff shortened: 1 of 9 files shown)


christianvuerings:
@dangerismycat thank you! Could you share a couple of screenshots for the use cases where we use a smaller `iconSize`? I'm wondering if we can reach agreement with design and update pinboard to use the regular / default size instead.

dangerismycat:
> @dangerismycat thank you! Could you share a couple of screenshots for the use cases where we use a smaller `iconSize`? I'm wondering if we can reach agreement with design and update pinboard to use the regular / default size instead.

Sure, though I doubt Design would be into that — the usages are all on the Lego PinRep, the reason why I forked these components in the first place:
Screen Shot 2020-03-27 at 10 56 41 AM

PR #785, opened 2020-03-30
[Fix] OutsideEventBehavior

emilyalwayss:
- only set document event listener once on mount
- use `useCallback` to prevent new handle function to be created (and un/registered) on every render

flyout inside of modal 2

### Components using `OutsideEventBehavior` being used in pinboard:
**Flyout -**
pinboard-flyout

**Modal (with code change in pinboard) -** 
pinboard-modal

**Flyout in Modal**
pinboard-flyout-in-modal


christianvuerings:
@emilyalwayss Thank you! Closing this in favor of #793

PR #786, opened 2020-03-30
Icon: add code icon

eliakaris:
Add new code icon.

This is used currently in the Tag Install page:
Pinterest


packages/gestalt/src/icons/code.svg
```diff
@@ -0,0 +1 @@
+<svg width="74" height="58" xmlns="http://www.w3.org/2000/svg"><g transform="translate(2 2)" fill="none" fill-rule="evenodd"><text font-family="SFProText-Semibold, SF Pro Text" font-size="23" font-weight="500" letter-spacing="-.533" fill="#979797"><tspan x="11" y="38">&lt; / &gt;</tspan></text><rect stroke="#979797" stroke-width="3" width="70" height="54" rx="5"/><path d="M5 0h60a5 5 0 015 5v4H0V5a5 5 0 015-5z" fill="#979797"/><circle stroke="#979797" fill="#D8D8D8" cx="7" cy="4" r="2"/><circle stroke="#979797" fill="#D8D8D8" cx="13" cy="4" r="2"/><circle stroke="#979797" fill="#D8D8D8" cx="19" cy="4" r="2"/></g></svg>
```
christianvuerings: @eliakaris could you request PDS to export this icon? Looks like it has transforms and other properties which we don't support.

eliakaris: sounds good, yeah it is causing the build to break

PR #788, opened 2020-03-31
Revert "Modal: Update OutsideEventBehavior to work well with Portals"

christianvuerings:
This reverts commit 5482612230a6a7075489e276446de977d5218b75.

Revert the changes to `OutsideEventBehavior` for now since we need better testing before putting it in pinboard. We already had a fix in #785 but let's revert instead of fix forward.


christianvuerings:
cc @emilyalwayss we are reverting the original PR for now, could you update #785 with both fixes combined? After that, let's set up a devapp where we can test the new behavior before landing it again.

PR #793, opened 2020-04-01
Modal: support Flyout inside of a Modal

christianvuerings:
Alternative approach to #785 with the following changes:

* Limit the amount of click listeners - we only require 1 for the document (Flyout) and 1 for the backdrop
* Modal uses similar approach to reach ui https://github.com/reach/reach-ui/blob/master/packages/dialog/src/index.tsx#L110 and material ui https://github.com/mui-org/material-ui/blob/master/packages/material-ui/src/Modal/Modal.js#L171
* Avoid `isClickedInside` - `useRef` boolean variable (used to avoid re-renders)

/cc @emilyalwayss


packages/gestalt/src/Modal.js
```diff
@@ -89,32 +105,30 @@ export default function Modal({
           className={styles.container}
           role={role}
         >
-          <Backdrop>
-            <OutsideEventBehavior onClick={handleOutsideClick}>
```
christianvuerings, line 93: main change is to no longer use `OutsideEventBehavior` here - we add the `onClick` on the `Backdrop` instead

PR #799, opened 2020-04-10
[Container] Allow user to set max width

dangerismycat:
Centering content is essential in many designs, but not all use a max-width of 800px. This keeps the existing behavior as the default, but allows the user to override if needed.

- [x] Documentation
- [x] Tests
- [ ] Accessibility checkup

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
 
 ### Minor
 
+- Container: Allow user to set max width (#799)
+
 ### Patch
 
 </details>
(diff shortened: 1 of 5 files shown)


christianvuerings:
@dangerismycat would it be possible to figure out a better default with design?

dangerismycat:
Context: I'm creating the Lego Profile header right now, which has a 488px max width for pinners and 656px max width for partners/creators. I suppose we could make maxWidth an enum of those two values and 800px, but I'm not sure I see the value that would add. It seems like it'd be better to encourage the use of this component over Box by allowing the dev to set the max width, which at least encapsulates behavior better than the wide-open Box.

dangerismycat:
@chrislloyd's good points: we don't want to complicate Container or allow devs to set a larger max width than the rest of the app. Column works for this sort of layout centering.

PR #803, opened 2020-04-13
[Box, Flexbox, Row, Stack] Create Flexbox, Row, Stack components, DRY out Box types

dangerismycat:
This work begins with DRYing out and exporting Box types for use in Box-derived components (also removed a couple of types that were no longer used).

I then created a FlexBox component, which is a Box with `display="flex"` and all non-layout props removed. That internal, un-exported component is used to compose two new components, Row and Stack. These components are used for horizontal and vertical flex layouts, with a much smaller, more-focused API than Box. They also feature a new prop, `gap`, which defines symmetrical padding around children along the main axis. These components will allow developers to easily create flex layouts and prevent overloading Boxes with both layout and other concerns.

- [x] Documentation
- [x] Tests
- [ ] Accessibility checkup


packages/gestalt/src/Row.js
```diff
@@ -0,0 +1,125 @@
...
+  alignContent?: AlignContent,
+  alignItems?: AlignItems,
+  alignSelf?: AlignSelf,
+  children?: React.Node,
+  fit?: boolean,
+  flex?: Flex,
+  height?: Dimension,
+  justifyContent?: JustifyContent,
+
+  maxHeight?: Dimension,
```
christianvuerings, line 36: @dangerismycat should we limit the API of `Row` / `Stack` even more? Just keeping it to the absolute necessary flex props. If people need to set width / height / padding / overflow, they can always wrap it in a `<Box />`
Mainly trying to avoid the issue we have with `Box` where the API is too wide.

```
  alignContent?: AlignContent,
  alignItems?: AlignItems,
  alignSelf?: AlignSelf,
  children?: React.Node,
  direction?: Direction,
  fit?: boolean,
  flex?: Flex,
  justifyContent?: JustifyContent,
  gap?: Margin,
  wrap? boolean,
```

packages/gestalt/src/Row.js
```diff
@@ -0,0 +1,125 @@
...
+  return (
+    <FlexBox
+      alignItems={alignItems ?? 'center'}
+      direction="row"
+      justifyContent={justifyContent ?? 'start'}
+      {...rest}
+    >
+      {React.Children.map(children, (child, index) => (
+        <Box
+          marginStart={index === 0 ? 0 : spacing}
```
christianvuerings, line 79: @dangerismycat this will cause an issue when `wrap` is set to `true` since the spacing of the individual items will not be correct when it flex wraps:

```
item1  |
  item2  |
  item3|
```

Instead, can we use symmetrical padding?

```js
<Box marginStart={-gap} marginEnd={-gap}>
  {React.Children.map(children, (child, index) => (
    <Box paddingX={gap}>{child}</Box>
  ))}
</Box>
```

packages/gestalt/src/Row.js
```diff
@@ -0,0 +1,125 @@
...
+  smPaddingX?: Padding,
+  mdPaddingX?: Padding,
+  lgPaddingX?: Padding,
+
+  paddingY?: Padding,
+  smPaddingY?: Padding,
+  mdPaddingY?: Padding,
+  lgPaddingY?: Padding,
+
+  spacing?: Margin,
```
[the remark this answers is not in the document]
dangerismycat: Absolutely! I didn't realize there was prior art here.

packages/gestalt/src/Row.js
dangerismycat: Sure, limiting the API has been a goal of mine. I wasn't sure where to draw the line. You think the above list of props should be all that is available?

packages/gestalt/src/Row.js
dangerismycat: Yep, good point! I hadn't considering wrapping.

packages/gestalt/src/Row.js
christianvuerings: @dangerismycat Yes, I think that would be a good start - we can always add more of them later (taking away is harder).

dangerismycat: Also removing `direction`, as that's one of the props we're hard-coding as part of Row/Stack

dangerismycat:
@christianvuerings A few remaining questions/notes:
- Should we remove more FlexBox props? I think it could probably be pared down more.
- Any additional test cases I should cover? Or additional examples in the docs I should create?
- Are the defaults I established in Row/Stack for `alignItems` and `justifyContent` good, or should we choose different ones? I picked the ones that feel the most "base case" to me, but very open to adjusting those if needed.

docs/src/FlexBox.doc.js
```diff
@@ -0,0 +1,173 @@
...
+  <Example
+    description={`
+    With a limited set of props that only relate to layout, FlexBox is useful for separating concerns to prevent overloaded Box usage.
+    Compare this example to the first example in the Box docs.
+  `}
+    name="Example: Media object"
+    defaultCode={`
+<FlexBox
+  alignItems="center"
+  marginStart={-1}
```
christianvuerings, line 135: @dangerismycat is this a prop supported by `FlexBox`? If so, could we add it to the docs.

docs/src/FlexBox.doc.js
```diff
@@ -0,0 +1,173 @@
...
+);
+
+card(
+  <Combination
+    description={`
+    FlexBox is strictly for flex layouts. If you're new to flex layout, please read the excellent [CSS-Tricks guide to flexbox](https://css-tricks.com/snippets/css/a-guide-to-flexbox/).
+  `}
+    id="layout"
+    name="Layout"
+    // alignContent={['start', 'end', 'center', 'between', 'around', 'stretch']}
```
christianvuerings, line 157: @dangerismycat let's remove this comment

docs/src/Row.doc.js
```diff
@@ -0,0 +1,131 @@
...
+
+card(
+  <Example
+    description={`
+    With a very limited set of props that only relate to horizontal layout, Row is useful for separating concerns to prevent overloaded Box usage.
+    Compare this example to the first example in the Box docs.
+  `}
+    name="Example: Media object"
+    defaultCode={`
+<Box marginStart={-1} marginEnd={-1}>
```
christianvuerings, line 93: @dangerismycat could we showcase `gap` instead?

```js
<Row gap={1}>
  <Avatar name="chrislloyd" size="md" />
  <Box>
      <Text weight="bold">Chris Lloyd</Text>
      <Text>joined 2 years ago</Text>
  </Box>
</Row>
```

docs/src/Row.doc.js
```diff
@@ -0,0 +1,131 @@
...
+);
+
+card(
+  <Combination
+    description={`
+    Row is strictly for horizontal flex layouts. If you're new to flex layout, please read the excellent [CSS-Tricks guide to flexbox](https://css-tricks.com/snippets/css/a-guide-to-flexbox/).
+  `}
+    id="layout"
+    name="Layout"
+    // alignContent={['start', 'end', 'center', 'between', 'around', 'stretch']}
```
christianvuerings, line 115: Let's remove this comment

docs/src/Stack.doc.js
```diff
@@ -0,0 +1,126 @@
...
+      {
+        name: 'justifyContent',
+        type: `"start" | "end" | "center" | "between" | "around" | "evenly"`,
+        defaultValue: 'center',
+        description:
+          'Defines the alignment along the vertical axis. It helps distribute extra free space left over when either all the flex items on a line are inflexible, or are flexible but have reached their maximum size. It also exerts some control over the alignment of items when they overflow the line.',
+        href: 'layout',
+      },
+      {
+        name: 'wrap',
```
christianvuerings, line 76: @dangerismycat does this make sense for `<Stack />`? Aren't the items always stacked on top of each other / wrapped?

docs/src/Row.doc.js
```diff
@@ -0,0 +1,128 @@
...
+
+card(
+  <Example
+    description={`
+    With a very limited set of props that only relate to horizontal layout, Row is useful for separating concerns to prevent overloaded Box usage.
+    Compare this example to the first example in the Box docs.
+  `}
+    name="Example: Media object"
+    defaultCode={`
+<Box marginStart={-1} marginEnd={-1}>
```
christianvuerings, line 93: @dangerismycat Can we remove this line? We should no longer need the `<Box marginStart={-1} marginEnd={-1}>` surrounding a row with `gap`

dangerismycat: yep! good catch!

PR #815, opened 2020-04-17
Minor version updates for several dependencies

jennyscript:
This updates a few dependencies to the latest versions:

- react, react-dom, and react-test-renderer
- @testing-library/react and @testing-library/jest-dom
- jest and babel-jest
- stylelint

- [x] Tests

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,8 +5,8 @@
     "packages/*"
   ],
   "dependencies": {
-    "react": "^16.8.5",
-    "react-dom": "^16.8.5"
+    "react": "^16.13.1",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -5,19 +5,19 @@
     "packages/*"
   ],
   "dependencies": {
-    "react": "^16.8.5",
-    "react-dom": "^16.8.5"
+    "react": "^16.13.1",
```
christianvuerings, line 8: @jennyscript could we keep the other `package.json` files also in sync with these updates?

image

PR #818, opened 2020-04-18
Add android share icon to icon type

emilyalwayss:
image


packages/gestalt/src/icons/android-share.svg
```diff
@@ -0,0 +1,19 @@
+<svg
+    width="24"
+    height="24"
+    viewBox="0 0 24 24"
+    version="1.1"
+    xmlns="http://www.w3.org/2000/svg"
+>
+    <path
```
christianvuerings, line 8: @emilyalwayss could you work with PDS and get a streamlined + optimized version of the SVG?

SVG's in gestalt should not contain strokes / transforms / ...

emilyalwayss: updated!

PR #828, opened 2020-04-23
Masonry: Add RTL support for Masonry Grid to have 1st element at top …

vincenttian:
…right.

## TODO

- [X] Documentation
- [ ] Tests
- [ ] Accessibility checkup

### DIFF
--- a/docs/src/Masonry.doc.js
+++ b/docs/src/Masonry.doc.js
@@ -69,6 +69,13 @@ card(
           'A callback when the user scrolls and you need to load more items into the grid. Note that `scrollContainer` must be specified.',
       },
       {
+        name: 'layoutDirection',
+        type: 'string',
+        defaultValue: 'ltr',
+        description:
+          'Support ltr or rtl. Specifies whether Masonry should load first element from right to left instead of left to right for RTL support.',
+      },
+      {
         name: 'scrollContainer',
         type: '() => HTMLElement',
         description:
--- a/packages/gestalt/src/Masonry.js
+++ b/packages/gestalt/src/Masonry.js
@@ -54,6 +54,7 @@ type Props<T> = {|
           from: number,
         }
       ) => void | boolean | {}),
+  layoutDirection?: string,
   scrollContainer?: () => HTMLElement,
   virtualBoundsTop?: number,
   virtualBoundsBottom?: number,
@@ -198,6 +199,7 @@ export default class Masonry<T: {}> extends React.Component<
     columnWidth: 236,
     minCols: 3,
     layout: 'basic',
+    layoutDirection: 'ltr',
     loadItems: () => {},
     virtualize: false,
   };
@@ -401,6 +403,7 @@ export default class Masonry<T: {}> extends React.Component<
   renderMasonryComponent = (itemData: T, idx: number, position: *) => {
     const {
       comp: Component,
+      layoutDirection,
       scrollContainer,
       virtualize,
       virtualBoundsTop,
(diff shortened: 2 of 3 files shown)


vincenttian:
Side x Side comparison of localhost vs prod https://pinterest.github.io/gestalt/#/Masonry

Used React extension to modify and add `rtlLayout: true` to verify correct behavior, will work on adding tests to further verify correctness.

image

vincenttian:
Testing within my devapp

Regular Pin Grid:
image

Users Grid Layout:
image

MWeb Pin Grid:
image

vincenttian:
User Homefeed with `web_masonry9_homepage=control` in ModHeader for experiment override:
image

**Caveat in testing**
there is currently a `web_masonry9_homepage` experiment that uses a different Masonry that forces employees into https://helium.pinadmin.com/helium/web_masonry9_homepage/config/

This masonry does not support RTL and is unaffected by the gestalt Masonry change in this PR.

Not sure what the plan is to ship that/update RTL support into that Masonry if it ship, just wanted to call this out in testing because HF is probably most important surface to support RTL since it has all the slot-specific logic

btraut:
Everything here looks good to me. It's possible at some point that we'll want to carve out a separate "layoutOptions" that contains things like "rtlLayout", but that's a matter for another day. @bishwei, @chrislloyd, or @christianvuerings -- can you all take a look as well? I'd love a second set of eyes on this. As for corresponding pinboard changes, I've reached out to @vincenttian in our #gestalt channel.

packages/gestalt/src/Masonry.js
```diff
@@ -54,6 +54,7 @@ type Props<T> = {|
           from: number,
         }
       ) => void | boolean | {}),
+  rtlLayout?: boolean,
```
btraut, line 57: nit: I'm always suspicious of whether booleans communicate their impact properly. How do you feel about naming this something along the lines of `layoutDirection: 'rtl' | 'ltr'` and then setting a `defaultProps` to `ltr`?

bishwei:
thanks @vincenttian for the PR!

my initial concern with this change is that it introduces a _different_ way for components in gestalt to respond to `rtl` vs `ltr` layout. previous changes, such as #107 and #628, toggle direction based on the html `dir` attribute - it totally makes sense why we can't use the css approach here, but imo it would be great if we could have consistency.    

One option we've talked about previously was using some sort of context provider here that gestalt components could be aware of - wonder if that's worth looking into more now @christianvuerings @btraut 

re: the change itself - i'm okay with this change if we're just using it to experiment on a few surfaces.  i think we should probably reconsider the approach when we decide to roll out further - setting a direction prop on Masonry _feels_ wrong to me long term.

packages/gestalt/src/Masonry.js
bishwei: +1 on this

vincenttian: Sounds good, will rename as such

docs/src/Masonry.doc.js
```diff
@@ -68,6 +68,13 @@ card(
         description:
           'A callback when the user scrolls and you need to load more items into the grid. Note that `scrollContainer` must be specified.',
       },
+      {
+        name: 'layoutDirection',
+        type: 'string',
```
AlbertCarreras, line 73: I would treat this one as `dealDirection = 'up' | 'right' | 'down' | 'left'` in Tooltip, accepting  `ltr` or `rtl` rather than just `string`. https://pinterest.github.io/gestalt/#/Tooltip

christianvuerings:
@vincenttian going to close this for now. As discussed, we should have a `layoutDirectionProvider` that uses React context which can be read by individual components (like Masonry)

PR #819, opened 2020-04-20
SearchField: Update design

nachofassini:
## Before
Screen Shot 2020-04-20 at 12 06 14 PM
Screen Shot 2020-04-20 at 12 06 19 PM
Screen Shot 2020-04-20 at 12 14 19 PM

## After
Screen Shot 2020-04-20 at 11 47 47 AM
Screen Shot 2020-04-20 at 11 47 58 AM
Screen Shot 2020-04-20 at 12 14 04 PM

### New `lg (48px)` variant
Screen Shot 2020-04-20 at 12 06 52 PM

PR #830, opened 2020-04-23
SearchField: convert back to class component

christianvuerings:
#819 converted `SearchField` from a class to a function component. That is great, since it makes the code more readable but it introduced issues in Pinterest specific code which uses `ref` on the components.

An alternative is to use `forwardRef` but that would encourage abstraction leaking, which is not something we'd like to do.

### DIFF
--- a/docs/src/SearchField.doc.js
+++ b/docs/src/SearchField.doc.js
@@ -83,37 +83,37 @@ card(
   function SearchFieldExample() {
     const [value, setValue] = React.useState('');
 
-      return (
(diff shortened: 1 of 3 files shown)


docs/src/SearchField.doc.js
```diff
@@ -83,37 +83,37 @@ card(
   function SearchFieldExample() {
     const [value, setValue] = React.useState('');
 
-      return (
```
christianvuerings, line 86: Docs were not indented correctly.

PR #832, opened 2020-04-23
Touchable: add optional onBlur / onFocus props

christianvuerings:
Use case: in order to make a more accessible `Tooltip`, we need to fire onBlur / onFocus events from `Touchable` so the design syncs up.

In this example you can see the issue, hovering the Icon changes the color but focussing it does not.

touchable-focus-blur


AlbertCarreras:
we could add a simple example to the docs

PR #834, opened 2020-04-27
Support CloseupShop v2 Flyout UseCase

NEsfandiari:
To support the sizing requirements of CloseupShop v2's useCase of a flyout-like component without having to reWrite flyout logic, I added custom logic for if the developer inputs a size below the "xs" size standard.

Examples within the context of the current docs: 
Screen Shot 2020-04-27 at 2 35 06 PM
Screen Shot 2020-04-27 at 2 35 44 PM

### DIFF
--- a/packages/gestalt/src/Contents.css
+++ b/packages/gestalt/src/Contents.css
@@ -24,7 +24,7 @@
 }
 
 .minDimensionsCustom {
-  padding: 3px
+  padding: 2px
 }
 
 .innerContents {
--- a/packages/gestalt/src/Contents.css
+++ b/packages/gestalt/src/Contents.css
@@ -23,6 +23,10 @@
   min-width: 180px;
 }
 
+.minDimensionsCustom {
+  padding: 3px
+}
+
 .innerContents {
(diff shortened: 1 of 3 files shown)


packages/gestalt/src/Contents.css
```diff
@@ -23,6 +23,10 @@
   min-width: 180px;
 }
 
+.minDimensionsCustom {
+  padding: 2px
```
christianvuerings, line 27: @NEsfandiari we should not add this padding when `< 180` - it feels too much like magic.

packages/gestalt/src/Contents.js
```diff
@@ -498,6 +498,10 @@ export default class Contents extends React.Component<Props, State> {
     const visibility = mainDir === null ? 'hidden' : 'visible';
     const background = `${bgColor}Bg`;
     const stroke = bgColor === 'white' ? '#ddd' : null;
+    const minDims =
```
christianvuerings, line 501: @NEsfandiari an alternative would be to remove the `minDimensions` when the width is a number

PR #835, opened 2020-04-28
Readme: Remove greenkeeper reference

christianvuerings:
We no longer use greenkeeper so let's remove it


rviscuso:
Nice change! :-)

PR #838, opened 2020-05-01
Table: add basic table components

PayalModi:
Adds the first round of components for the Table: Table, Table.Body, Table.Cell, Table.Header, Table.HeaderCell, and Table.Row.

- [x] Documentation
- [x] Tests
- [x] Accessibility checkup - I added scope for the header cells and used the Chrome audits tab

Screen Shot 2020-05-01 at 9 52 29 AM


PayalModi:
Screenshots of the docs:
Screen Shot 2020-05-01 at 1 26 54 PM
Screen Shot 2020-05-01 at 1 26 59 PM

docs/src/Table.doc.js
```diff
@@ -0,0 +1,173 @@
...
+
+card(<Card name="Table.Header" />);
+
+card(
+  <PropTable
+    Component={Table.Header}
+    props={[
+      {
+        name: 'children',
+        type: 'React.Node',
```
btraut, line 124: Is it allowed for `Table.Header` to have components other than `Table.HeaderCell` as children? If not, I would type this as `Table.HeaderCell | Array<Table.HeaderCell>` and add a corresponding check in the code.

Same goes for regular cells (outside header).

packages/gestalt/src/Table.js
```diff
@@ -0,0 +1,81 @@
...
+    props = { ...props, colSpan };
+  }
+  if (rowSpan) {
+    props = { ...props, rowSpan };
+  }
+
+  return <td {...props}>{children}</td>;
+};
+
+const TableHeader = ({ children }: { children: React.Node }) => (
```
christianvuerings, line 29: @PayalModi Even though these components will be quite small, can we put them in their own files? Makes it easier to expand / test there individual components.

packages/gestalt/src/Table.js
```diff
@@ -0,0 +1,81 @@
...
+}) => {
+  let props = { className: styles.td };
+  if (colSpan) {
+    props = { ...props, colSpan };
+  }
+  if (rowSpan) {
+    props = { ...props, rowSpan };
+  }
+
+  return <td {...props}>{children}</td>;
```
christianvuerings, line 26: @PayalModi instead, could we write it as the following:

```
return <td className={styles.td} colSpan={colSpan} rowSpan={rowSpan} />
```

We try to avoid prop spreading as much as possible to make the code more explicit and easier to reason about. When `colSpan` is `null` or `undefined`, the prop should not get rendered in the DOM.

image

packages/gestalt/src/Table.js
```diff
@@ -0,0 +1,81 @@
...
+  colSpan,
+  rowSpan,
+  scope = 'col',
+}: {
+  children: React.Node,
+  colSpan?: number,
+  rowSpan?: number,
+  scope?: 'col' | 'colgroup' | 'row' | 'rowgroup',
+}) => {
+  let props = { className: styles.th, scope };
```
christianvuerings, line 44: Same comment as regular `Cell`

packages/gestalt/src/Table.js
```diff
@@ -0,0 +1,81 @@
...
+
+const TableRow = ({ children }: { children: React.Node }) => (
+  <tr>{children}</tr>
+);
+
+type TableProps = {|
+  children: React.Node,
+|};
+
+export default class Table extends React.PureComponent<TableProps> {
```
christianvuerings, line 63: @PayalModi for new components, we're trying to avoid `class` components (since they can have implicit `ref`'s on them). Could you see if we can change this to a function component instead?

Not sure but something like `Table.Body = TableBody` might work

packages/gestalt/src/Table.js
```diff
@@ -0,0 +1,81 @@
...
+
+  static Header = TableHeader;
+
+  static HeaderCell = TableHeaderCell;
+
+  static Row = TableRow;
+
+  render() {
+    return (
+      <div style={{ overflowX: 'scroll' }}>
```
christianvuerings, line 76: @PayalModi could we use a `Box` instead with `overflow` set to `scrollX`?

packages/gestalt/src/Table.test.js
```diff
@@ -0,0 +1,96 @@
...
+        <Table.HeaderCell rowSpan={2} colSpan={3} scope="row">
+          row name
+        </Table.HeaderCell>
+      )
+      .toJSON();
+    expect(tree).toMatchSnapshot();
+  });
+});
+
+describe('<Table.Row />', () => {
```
christianvuerings, line 85: Would be great if we can also split these tests up into multiple files.

docs/src/Table.doc.js
PayalModi: In current tables, there are custom table cell wrappers like: https://phabricator.pinadmin.com/diffusion/P/browse/master/webapp/app/sterling/react/components/BillingPage/BillingHistoryTableCells.js. I intentionally left the flow types for children open so that current tables can be adapted to this one easily, and to allow for experimentation/flexibility. Let me know what you think!

packages/gestalt/src/Table.js
PayalModi: This is much easier to read, thanks for the suggestion!

PR #839, opened 2020-05-01
Add support for Badge with blue background and white text

denamwangi:
This change updates the existing Badge component to match updated specs featuring:
- solid background (blue)
- white text
- remove enforcement of upper casing (consensus is sentence casing is better)
- update docs examples to include sentence casing, remove sentence suggesting we encourage use of top with headings.

I anticipate extending the implementation to include additional sizing and color props so i've tried to write this with that in mind while still being restrictive to match our current spec. For example, I allow for an optional color prop but I don't list this in the documentation and I restrict the Proptypes. Let me know if this is confusing in any way but I figured we'll likely be expanding this in the near future.

After
Screen Shot 2020-05-05 at 5 38 07 PM

Before
Screen Shot 2020-05-05 at 5 37 59 PM

## TODO

- [x ] Documentation
- [x ] Tests
- [ ] Accessibility checkup

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
   </summary>
 
 ### Minor
+- Badge: Update to solid background and white text (#839)
 
 ### Patch
 
--- a/docs/src/Badge.doc.js
+++ b/docs/src/Badge.doc.js
@@ -22,6 +22,8 @@ card(
         name: 'text',
         type: `string`,
         required: true,
+        description:
+          'Text displayed inside of the Badge. Sentence case is best.',
       },
       {
         name: 'position',
@@ -39,7 +41,7 @@ card(
     The `Badge` component is rendered inline within parent element."
     name="Example"
     defaultCode={`
-<Text>Some text that uses Badge component as its child <Badge text="new" /></Text>
+<Text>Some text that uses Badge component as its child <Badge text="New" /></Text>
 `}
   />
 );
@@ -47,10 +49,10 @@ card(
 card(
   <Example
     description="
-    Larger text looks better with a superscript `Badge`."
-    name="Example"
+    Larger text example rendered with a top positioned `Badge`."
+    name="Example: positioning"
     defaultCode={`
-  <Heading>Heading <Badge text="new" position="top" /></Heading>
+  <Heading>Heading <Badge text="Beta" position="top"/></Heading>
 `}
   />
 );
--- a/packages/gestalt/src/Badge.css
+++ b/packages/gestalt/src/Badge.css
@@ -1,16 +1,12 @@
 .Badge {
   composes: antialiased sansSerif from "./Typography.css";
-  composes: fontWeightBold from "./Typography.css";
-  composes: blue from "./Colors.css";
   composes: inlineBlock from "./Layout.css";
-  border-radius: 4px;
-  border-style: solid;
-  border-width: 1px;
+  composes: rounding1 from "./Borders.css";
+  composes: white from "./Colors.css";
+  composes: fontWeightBold from './Typography.css';
   font-size: 10px;
-  line-height: 1;
   margin-top: -4px;
   padding: 2px;
-  text-transform: uppercase;
   white-space: nowrap;
 }
 
--- a/packages/gestalt/src/Badge.js
+++ b/packages/gestalt/src/Badge.js
@@ -3,6 +3,7 @@ import * as React from 'react';
 import PropTypes from 'prop-types';
 import cx from 'classnames';
 import styles from './Badge.css';
+import colors from './Colors.css';
 
 type Props = {|
   position?: 'middle' | 'top',
@@ -11,7 +12,8 @@ type Props = {|
 
 export default function Badge(props: Props) {
   const { position = 'middle', text } = props;
-  const cs = cx(styles.Badge, styles[position]);
+
+  const cs = cx(styles.Badge, styles[position], colors.blueBg);
 
   return <span className={cs}>{text}</span>;
 }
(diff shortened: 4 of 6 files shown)


docs/src/Badge.doc.js
```diff
@@ -23,6 +23,13 @@ card(
         type: `string`,
         required: true,
       },
+      {
+        name: 'textColor',
+        type: `string`,
+        defaultValue: 'blue',
+        description:
+          'Changes the text color and predefined matching background',
```
christianvuerings, line 31: @denamwangi instead of adding a new prop, could we update the current treatment as specced by Annie?

packages/gestalt/src/Badge.css
```diff
@@ -1,19 +1,27 @@
...
+  composes: fontWeightNormal from './Typography.css';
+  line-height: 2;
   margin-top: -4px;
-  padding: 2px;
-  text-transform: uppercase;
+  text-align: center;
   white-space: nowrap;
 }
 
+.md {
```
christianvuerings, line 19: @denamwangi Can we remove this class? We want to avoid setting specific widths and heights on elements - for i18n reasons. Also let's keep the `font-size` in `.Badge` since we only have 1 size.

image
image

packages/gestalt/src/Badge.css
```diff
@@ -1,19 +1,27 @@
...
   composes: inlineBlock from "./Layout.css";
-  border-radius: 4px;
-  border-style: solid;
-  border-width: 1px;
-  font-size: 10px;
-  line-height: 1;
+  composes: rounding2 from "./Borders.css";
+  composes: white from "./Colors.css";
+  composes: fontWeightNormal from './Typography.css';
+  line-height: 2;
```
christianvuerings, line 13: @denamwangi instead of adding `line-height`, can we use padding - especially `Thai` has issues with using line height? Been was playing around with it and this seems like a better approach:

```
display: inline-flex;
align-items: center;
white-space: nowrap;
font-size: 12px;
padding: 4px 8px;
min-height: 24px;
box-sizing: border-box;
```

packages/gestalt/src/Badge.js
```diff
@@ -3,20 +3,29 @@ import * as React from 'react';
 import PropTypes from 'prop-types';
 import cx from 'classnames';
 import styles from './Badge.css';
+import colors from './Colors.css';
 
 type Props = {|
   position?: 'middle' | 'top',
   text: string,
+  color?: 'blue',
```
christianvuerings, line 11: @denamwangi since we're changing the default treatment, can we remove this extra prop?

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
   </summary>
 
 ### Minor
+- Update Badge component to match new specs with solid background and white text (#839)
```
christianvuerings, line 10: Can we use the standard way of adding entries into CHANGELOG? `ComponentName: Description (#839)`

```
Badge: Update to solid background and white text (#839)
```

packages/gestalt/src/Badge.css
denamwangi: That's good to know!

packages/gestalt/src/Badge.css
denamwangi: Ah yes- didn't realize this affected i18n. I'll take it out. For the changes you've added above with Annie's suggested new font size, padding and height- i'm going to hold off on that until there is consensus on the new standard and put that in a follow up diff but this approach looks great and i'll use for the next iteration.

packages/gestalt/src/Badge.js
denamwangi: Sure!

CHANGELOG.md
denamwangi: done

packages/gestalt/src/Badge.css
```diff
@@ -1,16 +1,14 @@
...
+  composes: rounding1 from "./Borders.css";
+  composes: white from "./Colors.css";
+  composes: fontWeightBold from './Typography.css';
   font-size: 10px;
-  line-height: 1;
   margin-top: -4px;
+  min-height: 12px;
   padding: 2px;
-  text-transform: uppercase;
+  text-align: center;
```
christianvuerings, line 11: @denamwangi Can we remove this line? Since the context always expands - I don't think this will have any effect.

packages/gestalt/src/Badge.css
```diff
@@ -1,16 +1,14 @@
...
-  border-radius: 4px;
-  border-style: solid;
-  border-width: 1px;
+  composes: rounding1 from "./Borders.css";
+  composes: white from "./Colors.css";
+  composes: fontWeightBold from './Typography.css';
   font-size: 10px;
-  line-height: 1;
   margin-top: -4px;
+  min-height: 12px;
```
christianvuerings, line 9: @denamwangi Do we need `min-height`? The `text` prop is required so it should already have `font-size` (10px) + padding (2px top/bottom)

packages/gestalt/src/Badge.test.js
```diff
@@ -1,10 +1,18 @@
 // @flow strict
 import React from 'react';
-import renderer from 'react-test-renderer';
+// import renderer from 'react-test-renderer';
```
christianvuerings, line 3: @denamwangi can you remove this comment?

denamwangi: 🤦‍♀️ thank you

packages/gestalt/src/Badge.css
denamwangi: you're so right- thanks

denamwangi:
Thanks for taking the time to go through this and for all the great comments!

PR #844, opened 2020-05-06
Flyout: shift caret left/right towards flyout to align with rounded corners

ldubberstein:
Bug: There is a gap between flyout and caret when flyout has one line of text.
Fixed by shifting the caret inwards towards the flyout 2 pixels when caret is on the left/right.  

~~2 callouts about this:~~
- ~~One downside to this solution: When the flyout is white, the stroke cuts into the flyout (fig 3)~~
- ~~The caret is less visible by 2 pixels, if we stick with this solution and this is undesirable, I can make the caret larger.~~
- ~~This issue does not exist if the contained box has paddingY > 3.  However there are many cases on web where paddingY = 3~~

~~**alternative solutions**~~
- ~~Increasing the minimum height of the flyout.  However this would cause any existing one line flyouts on web containing a Box element with a margin of 3 or less to be misaligned vertically.~~
- ~~Make the corners less round.  Corners are more rounded for Lego~~
- ~~Make the caret much smaller or less obtuse~~

Moved the background color to the element in front of the caret.  The shadow is still layered behind the caret and the flyout is layered over the caret.

Please comment any suggestions on these solutions/drawbacks.

- [x] Documentation
- [x] Tests
- [x] Accessibility checkup

Updated 5/13: 

**Before:**

Screen Shot 2020-05-05 at 9 16 52 PM

**After (Updated 5/13):**

Screen Shot 2020-05-13 at 7 57 22 PM
Screen Shot 2020-05-13 at 7 57 39 PM
Screen Shot 2020-05-13 at 11 29 13 AM

"up" position calculation needed to be changed after moving `caret` underneath `innerContent` layer
Screen Shot 2020-05-13 at 7 56 46 PM


christianvuerings:
@ldubberstein Thank you for this PR and mentioning the trade-offs! The `stroke` on the caret keeps causing issues (e.g. #837). Could we remove the `stroke` and use `drop-shadow` instead of `box-shadow`?

With these changes, please ensure Annie Teng signs off from the design side.

#### packages/gestalt/src/Contents.css
(uses `@support` for to make sure we still have [a good fallback in IE11](https://caniuse.com/#feat=css-filters))
```diff
+@supports (filter: drop-shadow(0 0 8px rgba(0, 0, 0, 0.1))) {
+  .contents {
+    box-shadow: none;
+    filter: drop-shadow(0 0 8px rgba(0, 0, 0, 0.1));
+  }
+}
```

#### packages/gestalt/src/Contents.js
```diff
     // Needed to prevent UI thrashing
     const visibility = mainDir === null ? 'hidden' : 'visible';
     const background = `${bgColor}Bg`;
-    const stroke = bgColor === 'white' ? '#efefef' : null;

     return (
-      <div
-        className={styles.container}
-        style={{ stroke, visibility, ...flyoutOffset }}
-      >
+      <div className={styles.container} style={{ visibility, ...flyoutOffset }}>
         <div
           className={classnames(
             border && styles.border,
```

#### `filter: drop-shadow`
image

#### `box-shadow`
image

ldubberstein:
@christianvuerings I tried removing the `stroke` as mentioned, but the shadow would still be layered above the flyout.  Since the `shadow` would need to be applied on the svg, it does not seem possible to control the caret `shadow` with css.

Instead I found a fix by bringing the background color to the inner-contents div and layering it over the caret.  The design now looks the same as originally intended without the bug.

packages/gestalt/src/Contents.css
```diff
@@ -26,7 +26,8 @@
 .innerContents {
   composes: flex from "./Layout.css";
   composes: overflowAuto from "./Layout.css";
-  composes: rounding2 from "./Borders.css";
+  position: relative;
+  z-index: 1;
```
christianvuerings, line 30: @ldubberstein thank you for working through this. In Gestalt we want to avoid using `z-index` as much as possible. Instead of applying the `z-index`, could we re-order the `<Caret />` to come before the content? That way the content will overlap the Caret

```diff
diff --git a/packages/gestalt/src/Contents.js b/packages/gestalt/src/Contents.js
index 7e9c191..f41f824 100644
--- a/packages/gestalt/src/Contents.js
+++ b/packages/gestalt/src/Contents.js
@@ -518,6 +518,14 @@ export default class Contents extends React.Component<Props, State> {
           ref={this.setFlyoutRef}
           tabIndex={-1}
         >
+          {caret && (
+            <div
+              className={classnames(colors[bgColor], styles.caret)}
+              style={{ ...caretOffset }}
+            >
+              <Caret direction={mainDir} />
+            </div>
+          )}
           <div
             className={classnames(
               styles.innerContents,
@@ -528,14 +536,6 @@ export default class Contents extends React.Component<Props, State> {
           >
             {children}
           </div>
-          {caret && (
-            <div
-              className={classnames(colors[bgColor], styles.caret)}
-              style={{ ...caretOffset }}
-            >
-              <Caret direction={mainDir} />
-            </div>
-          )}
         </div>
       </div>
     );
```

Also created this video explaining what it looks like in the DOM: https://youtu.be/_4eu_diNopc

ldubberstein: Good to know!  Yes that works.  I made this change, it required recalculating where the `caretBottom` when the flyout is up.  Screenshots are updated above.  Let me know if anything else seems amiss.

PR #851, opened 2020-05-11
Table: Add table border prop

PayalModi:
## TODO

- [x] Documentation
- [x] Tests
- [ ] Accessibility checkup

With border: 
Screen Shot 2020-05-11 at 2 56 05 PM

### DIFF
--- a/docs/src/Table.doc.js
+++ b/docs/src/Table.doc.js
@@ -24,6 +24,11 @@ card(
         name: 'children',
         type: 'React.Node',
       },
+      {
+        name: 'hasBorder',
+        type: 'boolean',
+        defaultValue: false,
+      },
     ]}
   />
 );
--- a/docs/src/Table.doc.js
+++ b/docs/src/Table.doc.js
@@ -25,9 +25,10 @@ card(
         type: 'React.Node',
       },
       {
-        name: 'hasBorder',
-        type: 'boolean',
-        defaultValue: false,
+        name: 'borderSize',
+        type: `"sm" | "none"`,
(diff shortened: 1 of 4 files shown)


packages/gestalt/src/Table.js
```diff
@@ -10,12 +10,18 @@ import TableRow from './TableRow.js';
...
 
 export default function Table(props: Props) {
+  const { hasBorder, children } = props;
+
   return (
-    <Box overflow="scrollX">
-      <table className={styles.table}>{props.children}</table>
+    <Box
+      overflow="scrollX"
+      {...(hasBorder ? { borderSize: 'sm', rounding: 1 } : {})}
```
PayalModi, line 22: I used a conditional here with spread props because the border should only show if hasBorder is true, but Box will always show a solid light gray border if anything (null/'none'/undefined) is passed in for borderSize.

docs/src/Table.doc.js
```diff
@@ -24,6 +24,11 @@ card(
         name: 'children',
         type: 'React.Node',
       },
+      {
+        name: 'hasBorder',
```
christianvuerings, line 28: @PayalModi instead of a boolean prop, could we pass in `borderSize` and limit it to `sm` for now? Gives us more flexibility moving forward + maintains a consistent API with the rest of Gestalt (e.g. `Box`)

PayalModi: Yes, I'll update it to be borderSize that allows 'sm' | 'none' to keep it the same as Box

PayalModi:
Did not update CHANGELOG.md since #849 has landed

PR #853, opened 2020-05-12
Internal: Require release type on every PR

christianvuerings:
Suggestion by @bishwei in #848 to make the release type required:

> thinking out loud - is defaulting to `patch` here the right behavior or should we throw if a release type label wasn't properly specified?

We also require for just one release type to be defined, if you define multiple, the build will fail. Uses [zwaldowski/match-label-action](https://github.com/zwaldowski/match-label-action) under the hood.

### No release type label defined => Build fails
image

### Add release label
image

### Release label defined => Build passes
image

### DIFF
--- a/.github/workflows/requireReleaseType.yml
+++ b/.github/workflows/requireReleaseType.yml
@@ -6,7 +6,6 @@ on:
       - opened
       - labeled
       - unlabeled
-      - edited
       - synchronize
 
 jobs:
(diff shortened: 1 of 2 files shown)


christianvuerings:
> cool! just curious, what made you go with `zwaldowski/match-label-action` vs the other ones?

@bishwei it had the best API, clean code & ensured you can not define multiple release type labels.

PR #856, opened 2020-05-12
Icon: add star-half icon

ouryuken:
Add star-half icon to gestalt.


christianvuerings:
@ouryuken Thanks for your contribution! We already have a `plus` icon in Gestalt but it's named `add`. Sorry for the confusion.

I removed the `plus` icon from this PR and will land it as soon as the build is green.

image

PR #859, opened 2020-05-14
Modal: Fix shadows on scroll

christianvuerings:
We have these bugs with shadows that should show when you scroll:
1. On Safari, the shadow is too thick and shows up a grey border
2. Components like `<TextField />` show on top of the shadow

This PR addresses both by moving away from the CSS only approach and add event listeners + use box shadow instead.

**Chrome before / after**
chrome-modal-shadows

**Safari before / after**
safari-modal-shadows

### DIFF
--- a/packages/gestalt/src/Modal.js
+++ b/packages/gestalt/src/Modal.js
@@ -1,6 +1,7 @@
 // @flow strict
 import * as React from 'react';
 import PropTypes from 'prop-types';
+import classnames from 'classnames';
 import Box from './Box.js';
 import Heading from './Heading.js';
 import StopScrollBehavior from './behaviors/StopScrollBehavior.js';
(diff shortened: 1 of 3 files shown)


packages/gestalt/src/Modal.js
```diff
@@ -95,6 +100,30 @@ export default function Modal({
...
+    window.addEventListener('resize', updateShadows);
+    return () => {
+      window.removeEventListener('resize', updateShadows);
+    };
+  }, []);
+
+  React.useEffect(() => {
+    updateShadows();
+  }, []);
+
```
btraut, line 126: This code makes me realize we don't have the hooks exhaustive deps linter plugged in properly. In this case, the `useEffect` calls should have `updateShadows` set as deps, and `updateShadows` should be memoized. I'm fairly certain there's no bugs in the way you've written it, but it'd be better to make it future-proof.

christianvuerings: @btraut interesting - we should have enabled it here:
https://github.com/pinterest/gestalt/blob/455e6d3bdc7caad0ca0991d692fb65219eea8353/.eslintrc.json#L61

adding `updateShadows` to the dependency list does show the follow expected lint error: 
image

For this particular line we only want to call `updateShadows` on initial render, so the the empty array should do the trick.

PR #871, opened 2020-05-18
Table: Add clickable row

PayalModi:
Adds support for clickable rows in the Table by adding an optional prop onClick for Table.Row. If onClick is passed in, the row will be clickable. Similar to Touchable, the row can be clicked by mouse click, space key, or enter key and can be tabbed through.

This also includes the related CSS changes:
- adds padding within the border to match design
- adds spacing between rows
- adds default cursor style for table
- adds hover style changes for clickable rows (background color, rounded borders, pointer cursor)

- [x] Accessibility checkup
- [x] Update documentation
- [x] Add/update Tests

With table border:
Screen Shot 2020-05-18 at 12 01 27 PM

Without table border:
Screen Shot 2020-05-18 at 12 04 30 PM

### DIFF
--- a/packages/gestalt/src/Table.css
+++ b/packages/gestalt/src/Table.css
@@ -1,10 +1,12 @@
 .table {
-  border-collapse: collapse;
-  border-spacing: 0;
+  border-collapse: separate;
+  border-spacing: 0 4px;
+  cursor: default;
   width: 100%;
 }
 
 .th {
+  composes: borderBottom from "./Borders.css";
   composes: paddingY3 from "./boxWhitespace.css";
   composes: paddingX3 from "./boxWhitespace.css";
 }
@@ -14,6 +16,15 @@
   composes: paddingX3 from "./boxWhitespace.css";
 }
 
-.thead {
-  composes: borderBottom from "./Borders.css";
+.trClickable:hover {
+  background-color: #efefef;
+  cursor: pointer;
+}
+
+.trClickable:hover td:first-child {
+  border-radius: 8px 0 0 8px;
+}
+
+.trClickable:hover td:last-child {
+  border-radius: 0 8px 8px 0;
 }
--- a/packages/gestalt/src/Table.js
+++ b/packages/gestalt/src/Table.js
@@ -20,6 +20,7 @@ export default function Table(props: Props) {
   return (
     <Box
       overflow="scrollX"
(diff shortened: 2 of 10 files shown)


packages/gestalt/src/Table.css
```diff
@@ -14,6 +16,15 @@
...
 }
 
-.thead {
-  composes: borderBottom from "./Borders.css";
+.trClickable:hover {
+  background-color: #efefef;
+  cursor: pointer;
+}
+
+.trClickable:hover td:first-child {
```
christianvuerings, line 24: @PayalModi could we split the border radius properties up? I know it's a bit of an edge case but this would break if the table only had 1 column:
image

```css
border-top-left-radius: 8px;
border-bottom-left-radius: 8px;
```

packages/gestalt/src/Table.js
```diff
@@ -20,6 +20,7 @@ export default function Table(props: Props) {
   return (
     <Box
       overflow="scrollX"
+      paddingX={1}
```
christianvuerings, line 23: @PayalModi could we remove this? In Gestalt, every container element should we 100% width by default without any extra padding.

packages/gestalt/src/TableRow.js
```diff
@@ -1,10 +1,52 @@
...
 |};
 
+const SPACE_CHAR_CODE = 32;
+const ENTER_CHAR_CODE = 13;
+
 export default function TableRow(props: Props) {
-  return <tr>{props.children}</tr>;
+  const { children, onClick } = props;
+
+  const handleKeyPress = (event: SyntheticKeyboardEvent<HTMLDivElement>) => {
```
christianvuerings, line 20: @PayalModi could you update with `HTMLDivElement` here and below with `HTMLTableRowElement` since we add these handlers to a `tr`

packages/gestalt/src/Table.css
```diff
@@ -14,6 +16,15 @@
   composes: paddingX3 from "./boxWhitespace.css";
 }
 
-.thead {
-  composes: borderBottom from "./Borders.css";
+.trClickable:hover {
+  background-color: #efefef;
```
christianvuerings, line 20: @PayalModi could you also add `composes: accessibilityOutline from "./Focus.css";` so we use the same focus style here as all of our components?

PR #877, opened 2020-05-22
Internal: fix yarn watch keeps appending CSS

christianvuerings:
Use `sane` instead of rollup's `watch method which kept appending CSS (instead of replacing it) with every change.

### DIFF
--- a/packages/gestalt/package.json
+++ b/packages/gestalt/package.json
@@ -25,7 +25,7 @@
     "postpack": "rm README.md",
     "prepack": "cp ../../README.md .",
     "prepublish": "rollup -c rollup.config.js",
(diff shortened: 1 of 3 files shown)


packages/gestalt/package.json
```diff
@@ -42,7 +42,6 @@
     "rollup-plugin-node-resolve": "^5.2.0",
     "rollup-plugin-replace": "^2.2.0",
     "rollup-plugin-visualizer": "^3.3.2",
-    "rollup-watch": "^4.3.1",
```
christianvuerings, line 45: Rollup added its own `watch` method in `0.46` - so we haven't used this package in a while.

PR #878, opened 2020-05-22
Table: Add sortable header cells

PayalModi:
Adds Table.SortableHeaderCell

- [x] Accessibility checkup - added same behavior for onFocus as onHover. Can tab through header cells, and click with space or enter key
- [x] Update documentation - updated docs and added new example
- [x] Add/update Tests - added snapshot and jsdom test

hover state:
Screen Shot 2020-05-19 at 3 45 02 PM

active state:
Screen Shot 2020-05-19 at 3 45 28 PM

docs:
Screen Shot 2020-05-22 at 9 39 25 AM


packages/gestalt/src/TableSortableHeaderCell.js
```diff
@@ -0,0 +1,93 @@
...
+    rowSpan,
+    status,
+    sortOrder,
+    onSortChange,
+  } = props;
+
+  const [isFocused, setFocused] = React.useState(false);
+  const [isHovered, setHovered] = React.useState(false);
+
+  const handleKeyPress = (
```
christianvuerings, line 39: @PayalModi Could we wrap the header `<Box />` in a `<Touchable />` instead? That way:

- Focus styles would work better 
image
- Only the title + icon would be clickable (similar to material UI table and other tables)
- We don't have to copy/paste the clickable / focus code from `Touchable` in this file.

```
<th
      className={classes}
      scope={scope || 'col'}
      colSpan={colSpan}
      rowSpan={rowSpan}
    >
      <Touchable {...}>
        {children}
       <Touchable />
```

We might even be able to use `<Table.HeaderCell />` and pass all of this as children:

```
<Table.HeaderCell
      scope={scope}
      colSpan={colSpan}
      rowSpan={rowSpan}
    >
      <Touchable {...}>
        {children}
       <Touchable />
```

packages/gestalt/src/TableSortableHeaderCell.js
```diff
@@ -0,0 +1,93 @@
...
+      onBlur={() => setFocused(false)}
+      tabIndex="0"
+      role="button"
+    >
+      <Box display="flex">
+        {children}
+        <Box
+          marginLeft={2}
+          dangerouslySetInlineStyle={{
+            __style: { visibility },
```
christianvuerings, line 77: @PayalModi can we just conditionally render this `Box` instead of changing the `visibility`?

PayalModi: I used visibility because the header cell changes size (looks like the whole cells jumps to expand in size) on hover/click if I only render this conditionally

christianvuerings: @PayalModi that works - thanks for explaining

packages/gestalt/src/TableSortableHeaderCell.js
PayalModi: Thanks for suggesting - I update to use TableHeaderCell wrapped around a Touchable!

PayalModi: Screen Shot 2020-05-26 at 3 00 45 PM
this is the new focus state

christianvuerings: @PayalModi Thank you - might be worthwhile to follow up (we can do it later) to set `fullWidth` to `false` on `Touchable` - that way it would not span the whole 100% width.

PayalModi: Ok! If I create a box with fit-content around the Touchable with fullWidth=False, the clickable area will only surround the title + arrow - I can put up a PR for that

PR #890, opened 2020-05-29
TableSortableHeaderCell: Change clickable area

PayalModi:
Changed the clickable area on the sortable header cell to only be around the children and the sorting icon, so that the clickable area doesn't span the entire width of the header cell. I added a Box with width as fit-content around the Touchable, and added fullWidth={false} to Touchable.

Focus state:
Screen Shot 2020-05-29 at 2 05 35 PM

After clicking:
Screen Shot 2020-05-29 at 2 06 06 PM

- [x] Accessibility checkup - no change required
- [x] Update documentation - no change required
- [x] Add/update Tests

### DIFF
--- a/packages/gestalt/src/TableSortableHeaderCell.js
+++ b/packages/gestalt/src/TableSortableHeaderCell.js
@@ -38,33 +38,36 @@ export default function TableSortableHeaderCell(props: Props) {
 
   return (
     <TableHeaderCell colSpan={colSpan} rowSpan={rowSpan} scope={scope}>
-      <Touchable
-        onTouch={onSortChange}
-        onMouseEnter={() => setHovered(true)}
-        onMouseLeave={() => setHovered(false)}
-        onFocus={() => setFocused(true)}
-        onBlur={() => setFocused(false)}
-      >
-        <Box display="flex">
-          {children}
-          <Box
-            marginLeft={2}
-            dangerouslySetInlineStyle={{
-              __style: { visibility },
-            }}
-          >
-            <Icon
-              accessibilityLabel=""
-              icon={
-                status === 'active' && sortOrder === 'asc'
-                  ? 'sort-ascending'
-                  : 'sort-descending'
-              }
-              color={status === 'active' ? 'darkGray' : 'gray'}
-            />
+      <Box width="fit-content">
+        <Touchable
+          fullWidth={false}
+          onTouch={onSortChange}
+          onMouseEnter={() => setHovered(true)}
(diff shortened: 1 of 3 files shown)


packages/gestalt/src/TableSortableHeaderCell.js
```diff
@@ -38,33 +38,36 @@ export default function TableSortableHeaderCell(props: Props) {
...
-            <Icon
-              accessibilityLabel=""
-              icon={
-                status === 'active' && sortOrder === 'asc'
-                  ? 'sort-ascending'
-                  : 'sort-descending'
-              }
-              color={status === 'active' ? 'darkGray' : 'gray'}
-            />
+      <Box width="fit-content">
```
christianvuerings, line 41: @PayalModi `fit-content` is not supported on IE 11. Instead let's try to use `display="inlineBlock"`.

https://caniuse.com/#search=fit-content

image

packages/gestalt/src/TableSortableHeaderCell.js
```diff
@@ -38,33 +38,36 @@ export default function TableSortableHeaderCell(props: Props) {
...
+          onTouch={onSortChange}
+          onMouseEnter={() => setHovered(true)}
+          onMouseLeave={() => setHovered(false)}
+          onFocus={() => setFocused(true)}
+          onBlur={() => setFocused(false)}
+        >
+          <Box display="flex">
+            {children}
+            <Box
+              marginLeft={2}
```
christianvuerings, line 53: @PayalModi can we use `marginStart` instead of `marginLeft`? That way we support RTL out of the box:

marginLeft
Screen Shot 2020-06-04 at 3 04 11 PM

marginStart (correct margin, even with RTL - Right to Left)

Screen Shot 2020-06-04 at 3 04 03 PM

packages/gestalt/src/TableSortableHeaderCell.js
```diff
@@ -38,33 +38,36 @@ export default function TableSortableHeaderCell(props: Props) {
...
+      <Box width="fit-content">
+        <Touchable
+          fullWidth={false}
+          onTouch={onSortChange}
+          onMouseEnter={() => setHovered(true)}
+          onMouseLeave={() => setHovered(false)}
+          onFocus={() => setFocused(true)}
+          onBlur={() => setFocused(false)}
+        >
+          <Box display="flex">
```
christianvuerings, line 50: @PayalModi can we center align the items? Looks like design also wants this:

Figma:
image

Before:
image

After:
image

PR #895, opened 2020-06-02
Docs: Add search

christianvuerings:
Add search to the Gestalt docs

gestalt-docs-search

# Notes
* Under the hood this uses Algolia DocSearch: see the docs at https://docsearch.algolia.com/docs/what-is-docsearch/
* The demo will trigger a reload to another page (so navigation will be slow)
* Design is ok, we should add a search icon link and perhaps a reset button in a follow up PR
* We could not use the Gestalt `<SearchField />` since Algolia makes DOM mutations outside of the React tree

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -5,3 +5,4 @@ dist
 flow-typed
 node_modules
 package.json
+*.css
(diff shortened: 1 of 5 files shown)


btraut:
Duuuuuude! This is so cool!!!

christianvuerings:
Depends on #897

PR #896, opened 2020-06-02
Docs: Add missing parameter for Toast

Neet-Nestor:
* Docs: The parameter `color` exists in the code but not in the docs. Add the parameter and a corresponding example for it in Toast.

image
image

### DIFF
--- a/docs/src/Toast.doc.js
+++ b/docs/src/Toast.doc.js
@@ -27,6 +27,11 @@ card(
         href: 'imageTextButtonExample',
       },
       {
+        name: 'color',
+        type: `'darkGray' | 'red'`,
+        href: 'redBackgroundColorExample',
+      },
+      {
         name: 'thumbnail',
         type: 'React.Node',
         href: 'imageTextExample',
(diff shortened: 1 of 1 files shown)


docs/src/Toast.doc.js
```diff
@@ -97,6 +102,56 @@ function ToastExample() {
...
+          }}
+          paddingX={1}
+          position='fixed'
+        >
+          {showToast && (
+            <Toast
+              color="red"
+              text={
+                <>
+                  Saved to{' '}
```
christianvuerings, line 137: @Neet-Nestor Could we update this to be an error message instead? E.g. `Oops, we couldn't find that!`. It makes the intent more clear

christianvuerings:
See https://github.com/pinterest/gestalt/pull/760  for more information on why it's no longer documented. I'm ok adding it back in since it's confusing for developers.

Neet-Nestor:
Thanks for the advise! Updated:
image

Neet-Nestor:
> See #760 for more information on why it's no longer documented. I'm ok adding it back in since it's confusing for developers.

Didn't know it. I'll also check with our designer about this. Thank you!

christianvuerings:
@Neet-Nestor merged - thanks for the fix!

PR #900, opened 2020-06-04
TableHeader: Add optional sticky header

PayalModi:
Adds the sticky header feature. Added an optional prop height for the Table component, and added an optional prop sticky for the TableHeader which makes the header sticky and the body scrollable

- [x] Accessibility checkup
- [x] Update documentation
- [x] Add/update Tests

Screen Shot 2020-06-04 at 2 11 08 PM


packages/gestalt/src/Table.css
```diff
@@ -14,6 +14,12 @@
...
-.thead {
-  composes: borderBottom from "./Borders.css";
+.sticky tr th {
+  background-color: #fff;
+  position: sticky;
+  top: 0;
+}
+
+.thead tr:last-child th {
+  border-bottom: 1px solid #efefef;
```
christianvuerings, line 24: @PayalModi borders should now be `#ddd` - see #776 /cc @ashleyseto

We also had this with the previous `composes: borderBottom from "./Borders.css";`

packages/gestalt/src/Table.js
```diff
@@ -13,15 +13,17 @@ import TableSortableHeaderCell from './TableSortableHeaderCell.js';
...
 
 export default function Table(props: Props) {
-  const { borderSize, children } = props;
+  const { borderSize, children, height } = props;
 
   return (
     <Box
       overflow="auto"
       {...(borderSize === 'sm' ? { borderSize: 'sm', rounding: 1 } : {})}
+      height={height}
```
christianvuerings, line 26: @PayalModi should we change this to `maxHeight` instead? Otherwise it might look off when we set a height higher than the number of items

# Height
image

# Max Height
image

packages/gestalt/src/Table.css
PayalModi: I get this error when trying to use compose for this class: `Error: composition is only allowed when selector is single :local class name not in ":local(.thead) tr:last-child th"`. But I updated to use #ddd for this border - thanks for catching that!

packages/gestalt/src/Table.css
```diff
@@ -1,5 +1,5 @@
 .table {
-  border-collapse: collapse;
+  border-collapse: separate;
```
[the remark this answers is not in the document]
PayalModi: I think this should be ok as a general change for the table for now, because it should work with follow up changes

packages/gestalt/src/Table.js
PayalModi: Good call, I updated to have maxHeight instead of height

PR #901, opened 2020-06-04
TextField: Add forward ref (BREAKING CHANGE)

christianvuerings:
Adds ref forwarding to `TextField`

We have 13 places in the Pinterest codebase where we already set a `ref` on the current `<TextField />`. However, that ref is set on the component's instance, not the actual input. So in each of those cases, people drill down to get the `<input />`. That is pretty hacky, so let's add a way for them to get it in a cleaner way.

## Why is this a breaking change?

https://reactjs.org/docs/forwarding-refs.html#note-for-component-library-maintainers

## Can I get more documentation around this feature?

## How can we find in which location we already have these legacy refs set?

Run the codemod:

```
cd ~/code/gestalt
yarn run codemod --parser=flow -t=packages/gestalt-codemods/1.61.0-2.0.0/textfield-find-deprecated-refs.js ~/code/pinboard/webapp
```

Which will output the following:

```
Update legacy ref on TextField: /home/christian/code/pinterest/PasswordStep.js:125:12
```
We supply the line number and column name so you brings you to the right location in the editor.

### DIFF
--- a/flow-typed/npm/jest_v26.x.x.js
+++ b/flow-typed/npm/jest_v26.x.x.js
@@ -259,6 +259,7 @@ type DomTestingLibraryType = {
   // 5.x
   toHaveDisplayValue(value: string | string[]): void,
   toBeChecked(): void,
+  toHaveDescription(value: string | string[]): Result,
   ...
 };
 
--- a/.flowconfig
+++ b/.flowconfig
@@ -8,6 +8,8 @@
 
 [options]
 include_warnings=true
+esproposal.optional_chaining=enable
+
 suppress_comment=\\(.\\|\n\\)*\\$FlowFixMe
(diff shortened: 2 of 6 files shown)


flow-typed/npm/jest_v26.x.x.js
```diff
@@ -259,6 +259,7 @@ type DomTestingLibraryType = {
   // 5.x
   toHaveDisplayValue(value: string | string[]): void,
   toBeChecked(): void,
+  toHaveDescription(value: string | string[]): Result,
```
christianvuerings, line 262: It's part of https://github.com/testing-library/jest-dom/releases/tag/v5.6.0 and we use `"@testing-library/jest-dom": "^5.7.0",`

PR #905, opened 2020-06-06
WIP[Masonry]- Add option to center justify grid content

mminute:
Add a layout `basicCentered` that uses the basic layout but centers the contents of the grid.

ex) https://www.pinterest.com/pinterestcreators/story-pins/

Existing:
Screen Shot 2020-06-08 at 2 03 51 PM

Proposed (in red):
Screen Shot 2020-06-05 at 2 53 58 PM

## TODO

- [ ] Accessibility checkup
- [ ] Update documentation
- [ ] Add/update Tests

### DIFF
--- a/packages/gestalt/src/Masonry.js
+++ b/packages/gestalt/src/Masonry.js
@@ -29,6 +29,7 @@ type Layout =
   | LegacyMasonryLayout
   | LegacyUniformRowLayout
   | 'basic'
+  | 'basicCentered'
   | 'flexible'
   | 'uniformRow';
 
@@ -490,6 +491,7 @@ export default class Masonry<T: {}> extends React.Component<
         cache: measurementStore,
         columnWidth,
         gutter,
+        justify: layout === 'basicCentered' ? 'center' : 'start',
         minCols,
         width,
       });
--- a/packages/gestalt/src/defaultLayout.js
+++ b/packages/gestalt/src/defaultLayout.js
@@ -29,11 +29,13 @@ export default <T>({
   cache,
   columnWidth = 236,
   gutter = 14,
+  justify = 'start',
   minCols = 2,
   width,
 }: {|
   columnWidth?: number,
   gutter?: number,
+  justify?: 'center' | 'start',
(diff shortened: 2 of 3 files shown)


btraut:
@chrislloyd @bishwei I wanted your thoughts on this as well. I talked with Mason and he's aware that we'll need to change our `gridCentered` styles as well, but this was the starting point. We discussed two options:

1. Creating a `basicCentered` layout, or
2. Adding an option into the existing `basic` layout that chooses to either center or left-align.

There's tradeoffs to both, and it's possible I'm missing other options, so I'd love your feedback.

chrislloyd:
@mminute what's the goal of this diff? It's a bit difficult to review because I'm not sure _why_ you're making the change.

mminute:
@chrislloyd there is a design request to center the contents of the grid.  Take this board for example: https://www.pinterest.com/pinterestcreators/story-pins/

The board sections are centered, but the grid items are left aligned.  The desired output is that the pins are also centered

bishwei:
@mminute fyi adding a screenshot in the future would be 👍 

yeah, i feel like ideally this is something that we should be able to achieve by composing something else with Masonry, but I do see the challenge atm since Masonry assumes 100% of its container. 

instead of making this a new layout, would it make more sense to just add a `justifyContent` prop on Masonry and pass that into the layout functions?

mminute:
I considered adding a `justifyContent` prop, but felt that for that to make sense it should be an option to apply to all layouts.  Seems reasonable though a broader change.  Limiting the change to just one layout felt like a good first step.  Would y'all be happier with the broader change of adding this option to all layouts @bishwei @btraut @chrislloyd ?

btraut:
fwiw, with Masonry9, I actually carved out a generic property bag of "layout specific settings" that Masonry passes to the layouts. `justifyContent` would be a perfect example of these given that it may not apply to all custom layout types. I had originally suggested to Mason that he just creates a separate layout type for the time being given that it's a happy medium between amount of new code and as little disruption as possible to pinboard.

PR #906, opened 2020-06-06
Touchable: forward ref, touch feedback, and more (BREAKING CHANGE)

jackhsu978:
### Major (breaking) changes:
- `onContextMenu` is now disabled.
- `onFocus` event no longer bubbles up.
- refactor `Touchable` from a class component as a function component with [`React.forwardRef`](https://reactjs.org/docs/forwarding-refs.html)

Note that covert a class component to a function component with forward ref is a [breaking change](https://reactjs.org/docs/forwarding-refs.html#note-for-component-library-maintainers).

### Minor changes:
- add `tapStyle` support, which is "none" by default. If `tapStyle="compress"`, the Touchable would compress (scale down) by 1% when clicked / touched.

### Revisions:
- `onTouch` is optional but the doc says it's required. fixed.
- rename `touchable.xxx` test files to `Touchable.xxx`.

## Codemod

To find which location we already have these legacy refs set:
```
cd ~/code/gestalt
yarn run codemod --parser=flow -t=packages/gestalt-codemods/2.0.1-3.0.0/touchable-find-deprecated-refs.js /path/to/your/project
```

For Pinboard webapp, there is none.

## TODO

- [x] Accessibility checkup
- [x] Update documentation
- [x] Add/update Tests

### Props docs
Screenshot 2020-06-11 17 16 11

### Tap style prop example
Screenshot 2020-06-11 17 18 39

tapStyle

### Ref example
Screenshot 2020-06-09 14 58 42

### Why disable default behavior of `onContextMenu`?

This is what Duplo uses and I assume there was a reason for that (although there was no documentation on that).

PS: here is the doc on context menu for reference:

> The `contextmenu` event fires when the user attempts to open a context menu. This event is typically triggered by clicking the right mouse button, or by pressing the context menu key. (https://developer.mozilla.org/en-US/docs/Web/API/Element/contextmenu_event)

### Why stops `onFocus` event from bubbling up?

See diff D304752

> "This makes it so that with the STL ... on a pin is focused the whole pin is not focused."

### Why are we not handling Link?

We plan to have another component like `TapLink` that is a link (anchor tag) with the same look & feel as Touchable. That is out of the scope of this PR.

### Why call it "tapStyle" instead of "pressStyle" or "touchStyle"?

Consider the following common associations:

- "Click" <-> mouse events
- "Press" <-> keyboard events
- "Touch" <-> touch events

The word "tap" is not directly associated with any of the events (although one might argue it's associated with touch), and is a good option for describing something that's common for both mouse events and touch events.

We also have plans for replacing Touchable with TapArea. When we do that, `tapStyle` would make a lot of sense and it would also be a good idea to rename `onTouch` to `onTap`

### Why isn't tapStyle="compress" the default?
Ideally we do want that as the default. However, making that the default requires additional codemod that safely handles the existing use cases where compress-on-tap is not desired. This can be addressed later and is out of the scope of this PR.

### Why support mouse down/up events when Duplo does not?
Based on Annie's vision, we want non-touch devices to behave the same way. Since "compress" is not a default and requires opt-in, this shouldn't break anything.

### Why don't we have tap style that shows light background or shows light gray border/outline/shadow?

Annie believes that the compress styling would be a better treatment and we should stay consistent. Given that we want compress styling as the tap feedback, those styles would be redundant. Also, those styles don't work on dark mode.

### Why are we still showing the blue border/outline/shadow on tap? Why blue?

The blue border is a way to indicate that the item is in focused state and is the only accessible way for users to effectively keyboard-navigate between items. Ideally, we do _not_ want the blue outline to show up when user taps on the items (we _only_ want to show it when user keyboard-navigates to it). However, this is not a quick thing to do, and @christianvuerings will be looking into that. Making it only show up on keyboard-focus is out of the scope of this PR.

As to the choice of color, Annie said that we only picked that color because it was a common color that many sites use. We didn't really consider branding and dark mode, and we may revisit that in the future. This is also out of the scope of this PR.


packages/gestalt/src/Touchable.js
```diff
@@ -73,126 +77,167 @@ const getRoundingStyle = (r: Rounding): Style => {
...
+      }}
+      onFocus={event => {
+        if (!disabled && onFocus) {
+          onFocus({ event });
+        }
+        if (!isPressed && !ignoreNextFocusEventRef.current) {
+          setFocused(true);
+        }
+        ignoreNextFocusEventRef.current = false;
+        event.stopPropagation();
```
jackhsu978, line 149: calling `event.stopPropagation();` in `onFocus` is a breaking change here. See https://phabricator.pinadmin.com/differential/changeset/?ref=7777595

packages/gestalt/src/Touchable.css
```diff
@@ -1,5 +1,7 @@
 .touchable {
   composes: accessibilityOutline from "./Focus.css";
+  transition: transform .2s ease-out;
```
jackhsu978, line 3: this shouldn't break anything because there was no transition before

packages/gestalt/src/Touchable.css
```diff
@@ -1,5 +1,7 @@
...
+  outline: 0;
```
jackhsu978, line 3: Adding `outline` here for consistency with what we do on mweb today. Not sure if it is okay. Open to suggestions.

packages/gestalt/src/Touchable.js
```diff
@@ -73,126 +82,167 @@ const getRoundingStyle = (r: Rounding): Style => {
...
+  pressScaleEnabled = false,
+  rounding = 0,
+}: Props) {
+  const [isFocused, setFocused] = React.useState<boolean>(false);
+  const [isPressed, setPressed] = React.useState<boolean>(false);
+  const [coordinate, setCoordinate] = React.useState<Coordinate>({
+    x: 0,
+    y: 0,
+  });
+  const ignoreNextFocusEventRef = React.useRef<boolean>(false);
```
jackhsu978, line 110: IIUC, this is to handle undesired focus event fired immediately after touch ends. See https://phabricator.pinadmin.com/D213915 for details.

packages/gestalt/src/Touchable.js
```diff
@@ -73,126 +82,167 @@ const getRoundingStyle = (r: Rounding): Style => {
...
+
+  return (
+    <div
+      aria-controls={accessibilityControls}
+      aria-disabled={disabled}
+      aria-expanded={accessibilityExpanded}
+      aria-haspopup={accessibilityHaspopup}
+      aria-label={accessibilityLabel}
+      className={classes}
+      onContextMenu={event => event.preventDefault()}
```
jackhsu978, line 133: See https://developer.mozilla.org/en-US/docs/Web/API/Element/contextmenu_event

This has always been set in Duplo. See https://phabricator.pinadmin.com/D192884

packages/gestalt/src/Touchable.js
```diff
@@ -73,126 +85,187 @@ const getRoundingStyle = (r: Rounding): Style => {
...
-        onKeyPress={this.handleKeyPress}
-        role="button"
-        tabIndex={disabled ? null : '0'}
-      >
-        {children}
-      </div>
-    );
-  }
-}
+export default React.forwardRef<Props, HTMLDivElement>(forwardRef);
```
jackhsu978, line 271: `React.forwardRef` here is to allow manually setting focus on the Touchable.

This is needed to support focusing on the next item in an item list when user uses the tab key:

preview-Jun-25-2019_18-53-47

See https://phabricator.pinadmin.com/D417565 for more details.

jackhsu978:
@christianvuerings thanks for your feedback offline. I've added the codemod for completeness.

packages/gestalt/src/Touchable.css
christianvuerings: @jackhsu978 take a look at https://github.com/pinterest/gestalt/blob/bcc420b7e3e2086b165757b7e7ab35f0517fe9a7/packages/gestalt/src/Focus.css#L3 - we already set `outline: 0` when we compose from `./Focus.css`

docs/src/Touchable.doc.js
```diff
@@ -60,14 +60,23 @@ card(
       {
         name: 'fullHeight',
         type: 'boolean',
+        description: 'Expands to the full height of the parent.',
```
christianvuerings, line 63: Thanks for adding these descriptions

docs/src/Touchable.doc.js
```diff
@@ -94,9 +103,30 @@ card(
         name: 'onTouch',
         type:
           '({ event: SyntheticMouseEvent<HTMLDivElement> | SyntheticKeyboardEvent<HTMLDivElement> }) => void',
-        required: true,
         href: 'basicExample',
       },
+      {
+        name: 'pressBackgroundEnabled',
```
christianvuerings, line 109: @jackhsu978 we are trying to avoid boolean props as much as possible. Should we use `PressState` similar to what we use in Pinterest's codebase?

```
type PressState = 'border' | 'background' | 'compress' | 'none';
```

packages/gestalt/src/Touchable.css
```diff
@@ -1,5 +1,8 @@
 .touchable {
   composes: accessibilityOutline from "./Focus.css";
+  display: block;
```
christianvuerings, line 3: @jackhsu978 do we need `block` here? Since it is a `div`, won't the default `display` be `block`?

packages/gestalt/src/Touchable.css
```diff
@@ -1,5 +1,8 @@
...
+  outline: 0;
+  transition: transform .2s ease-out;
```
christianvuerings, line 5: This is fine for now - we should just check in with Ashley / AJ to see what exact animation we want here. I know in the past it was something like a cubic-bezier of `0.25, 0.375, 0.1, 0.975`

The only reference I found was: https://design.pinadmin.com/pds/index.html#container-tapstate which provides more details

packages/gestalt/src/Touchable.css
```diff
@@ -41,3 +44,20 @@
...
+  transform: scale(0.99);
+  transform-origin: center;
+}
+
+.pressBackground {
+  composes: lightGrayBg from "./Colors.css";
+}
+
+.pressShadow {
+  box-shadow: 0 0 0 1px #767676;
```
christianvuerings, line 58: @jackhsu978 could you check with AJ / Ashley to see what the defaults should be here?

packages/gestalt/src/Touchable.js
```diff
@@ -73,126 +85,187 @@ const getRoundingStyle = (r: Rounding): Style => {
...
+  );
 
-  handleFocus = (event: SyntheticFocusEvent<HTMLDivElement>) => {
-    const { disabled, onFocus } = this.props;
-    if (!disabled && onFocus) {
-      onFocus({ event });
-    }
-  };
+  return (
+    <div
```
bishwei, line 135: one difference here is that `TapArea` in pinboard supports an `href` prop, which will change it to an anchor.  options here for supporting that could be:
- update gestalt's `Link` to support touch
- update this to support an href prop (feels gross)

@christianvuerings @jackhsu978 wdyt?

packages/gestalt/src/Touchable.js
```diff
@@ -25,8 +31,10 @@ type Props = {|
   accessibilityLabel?: string,
   children?: React.Node,
   disabled?: boolean,
+  forwardedRef?: React.Ref<'div'>,
   fullHeight?: boolean,
   fullWidth?: boolean,
+  inline?: boolean,
```
bishwei, line 37: i wonder if we actually need this as a prop? this seems like something we should be able to achieve via a container instead of setting it explicitly on Touchable

docs/src/Touchable.doc.js
bishwei: +1 on this

jackhsu978: The reason I went with independent boolean props rather than just one value is that there are 31 places in Pinboard where we use both "background" and "compress":
```
pressState={['compress', 'background']}
```
there are also places where we only use "compress" and only use "background".

Alternatively, I could also do this instead:
```
type PressState = 'border' | 'background' | 'compress' | 'compressAndBackground'

pressState?: PressState,
```
Would that be better?

packages/gestalt/src/Touchable.css
jackhsu978: I don't think we need to. will remove.

packages/gestalt/src/Touchable.css
jackhsu978: this is what Duplo uses today. will check with them.

packages/gestalt/src/Touchable.css
jackhsu978: this is what Duplo uses today. will check with them.

packages/gestalt/src/Touchable.js
jackhsu978: I thought about supporting `href` prop, but if we do that, we are going to a lot more props like `nofollow` and `target`. It's gonna be gross.

I think we should make components single responsibility. Touchable should just support the look & feel of a touchable UI element, but it should not care what's going to happen after touch.

For the link use case, this is what I have in mind (I tested in Pinboard and it seems to work fine):

```
<Touchable ...>
  <a href={href} rel={nofollow ? 'nofollow' : ''}>
    {children}
  </a>
</Touchable>
```

PS: Note that I am not using Gestalt's `<Link>` in the code snippet above because: (1) `Link` always underlines text on hover and (2) it doesn't support `nofollow` prop. We can make Gestalt's `<Link>` component support those, but that's a separate discussion.

packages/gestalt/src/Touchable.js
jackhsu978: good catch! I can just wrap it with `<Box display="inlineBlock">`. will remove.

packages/gestalt/src/Touchable.css
jackhsu978: Here is one of the existing use cases: mweb searchbox shows black outline on focus
Screenshot 2020-06-09 12 43 35

packages/gestalt/src/Touchable.css
jackhsu978: Comparison:

`200ms ease-out`
200-ease-out

`250ms cubic-bezier(0.25, 0.375, 0.1, 0.975)`
250ms-cubic-bezier

docs/src/Touchable.doc.js
bishwei: what about: `pressState: PressState | Array<PressState>,`?

packages/gestalt/src/Touchable.js
bishwei: ^yeah the problem with that (and why we originally added this to `TapArea`) is accessibility (having a button that wraps a link)

we could address this by adding touch support to `Link` - would definitely make the intent clearer (imo)

docs/src/Touchable.doc.js
jackhsu978: The (very minor) issues with passing an array are:

- we might be inviting more unnecessary re-render of Touchable because the array reference is new each time parent renders
- an array can have duplicated elements

Thoughts?

packages/gestalt/src/Touchable.js
jackhsu978: `Touchable` uses a div element with role "button". What if we add a prop "role" which can be "button" or null, and make it "button" by default? (also, `tab-index` attribute would be 0 if role is `null`)

Would that work?

docs/src/Touchable.doc.js
btraut: > we might be inviting more unnecessary re-render of Touchable because the array reference is new each time parent renders

As long as we don't have side-effects based on these re-renders, I highly doubt this will be a problem. It's technically okay to set a contract with the parent component to maintain memoization of params if we really want stability. I'm not saying we should or shouldn't go with an array, but just adding my 2 cents that re-rendering probably shouldn't be a big concern.

jackhsu978: thanks for the suggestions! updated the PR!

packages/gestalt/src/Touchable.css
ashleyseto: Acknowledging that I've seen this but I don't have any context on this work. AJ is on PTO for the week. Is there any designer working through touch specifically?

jackhsu978: @bishwei do you remember why the Duplo TapArea press animation was built the way it is today?

bishwei: we based as much as we could off the ios designs.  iirc, the touch animation was done according to pds spec at that time (~2 years ago).  its definitely possible that, since then, there have been changes to the design spec or changes to TapArea.

imo we should go with the animation based on pds spec here

jackhsu978:
I talked to Annie and have a much better understanding of Touchable. Here is a summary of Annie's vision:
- touch and non-touch devices should have the same treatment in focus state and pressed state
- focus state: always show blue border and should persist after touch up for accessibility
- press state: always show compressed with animation defined in the "Lego visual language" paper doc (i.e. `85ms ease-out`).
- we don't want to show the gray background color in neither focus state nor press state
- make sure one can focus by hitting tab for accessibility
- for the instances where we don't follow the pattern above, we can just change them to follow this pattern.

@christianvuerings @bishwei what do you think? does this look good to you?

bishwei:
> I talked to Annie and have a much better understanding of Touchable. Here is a summary of Annie's vision:
> 
> * touch and non-touch devices should have the same treatment in focus state and press state, and all instances of Touchable should have the same logic.
> * focus state: always show blue border and should persist after touch up for accessibility
> * press state: always show compressed with animation defined in the "Lego visual language" paper doc (i.e. `85ms ease-out`). having blue border in the press state is acceptable, but not ideal.
> * we don't want to show the gray background color in neither focus state nor press state. it would be redundant to the compress style in the press state and blue border in the focus state
> * make sure one can focus by hitting tab for accessibility
> * for the instances where we don't follow the pattern above, we can just change them to follow this pattern.
> 
> @christianvuerings @bishwei what do you think? does this look good to you?

that all sounds good to me 👍  it seems like it might even simplify things a bit
is there any concern on rolling out these changes on mweb before lego changes have been made to things like nav?

jackhsu978:
Thank you all for the great feedback! I've just updated the PR accordingly. It should be ready for review again. Thank you 🙏

docs/src/Touchable.doc.js
```diff
@@ -338,7 +348,70 @@ function DisabledEx() {
...
+card(
+  <Combination
+    id="tapStyleCombinations"
+    name="Combinations: tapStyle"
+    description={`\`tapStyle\` "compress" scales down by 1% on click / touch.`}
+    tapStyle={['none', 'compress']}
+  >
+    {props => (
+      <Box padding={2} width={150}>
+        <Touchable {...props} rounding={2}>
```
christianvuerings, line 369: @jackhsu978 could we make this `rounding="circle"`?

Before
image

After
image

packages/gestalt/src/Touchable.css.flow
```diff
@@ -6,9 +6,11 @@ declare module.exports: {|
   +'fullWidth': string,
   +'grab': string,
   +'grabbing': string,
+  +'inline': string,
```
bishwei, line 9: are we adding support for the `inline` prop?  if not - we should remove this (okay with keeping support for the prop btw to make the transition easier)

packages/gestalt/src/Touchable.js
```diff
@@ -73,126 +84,162 @@ const getRoundingStyle = (r: Rounding): Style => {
...
+  disabled = false,
+  forwardedRef,
+  fullHeight,
+  fullWidth = true,
+  mouseCursor = 'pointer',
+  onBlur,
+  onFocus,
+  onMouseEnter,
+  onMouseLeave,
+  onTouch,
```
bishwei, line 102: okay for a followup - does this make sense to leave as `onTouch` if we're going to rename this to `TapArea` eventually?  maybe `onTap` instead?

packages/gestalt/src/Touchable.css.flow
jackhsu978: Oops. I forgot to remove. No, we aren't adding support for `inline` prop since one can just wrap Touchable with `<Box display="inlineBlock">`. Thanks for catching this. I will also update the PR description.

packages/gestalt/src/Touchable.js
jackhsu978: I think we can leave as is and rename when we are ready to codemod everything to `TapArea`. This way we can just codemod both component name and this prop in one diff instead of two.

docs/src/Touchable.doc.js
jackhsu978: good catch! I purposely wanted the rounding to _not_ match (which is generally a bad thing to do) just so that I can show the light gray background tapStyle, but that is no longer relevant. Will update.

packages/gestalt/src/Touchable.js
jackhsu978: Will address it with a new component called `TapLink`

PR #907, opened 2020-06-06
Typeahead: Add New Component

simkessy:
image

image

DarkMode: 

image

AustralianMode: 

image

## TODO

- [x] Accessibility checkup
- [x] Update documentation
- [x] Add/update Tests
- [x] Keyboard navigation of options


packages/gestalt/src/AutoComplete.js
```diff
@@ -0,0 +1,211 @@
...
+      {/* RESULTS CONTAINER */}
+      {focused && (
+        <Layer>
+          <Flyout
+            showCaret
+            anchor={anchorRef.current}
+            idealDirection="down"
+            onDismiss={() => {}}
+            positionRelativeToAnchor={false}
+            size="flexible"
```
simkessy, line 145: Getting error on size
image

ashleyseto:
I now have access! Copying over my slack message for the record.

I will work to get you some better designs by end of week. If you have time in the meantime, there are some functional changes that need to be done. I think ant has the best performing example:
https://codesandbox.io/s/9rzvg

- The field should clear invalid values when it loses focus as to not need to throw an error state
- Caret on field should not be optional as that signals that this is a dropdown
- Bug: Sometimes when I select a value in the list, it does not hold in the field.

As far as how this is documented, is this a separate concept or are there ways to document this with the selectlist?

---
Adding also that we have no need for additional color options as well as we'd discussed earlier.

Final thing I'll also be thinking about how we name this new addition based on where we can document it. I'm not sure if it should be called Autocomplete. Perhaps Typeahead or just Selectlist Search.

packages/gestalt/src/AutoComplete.js
simkessy: Flyout issue fixed

ashleyseto:
Hey. Let me know when this is ready for another review. I see some of the changes still need to be implemented.

As a note this is "Typeahead" not "Typehead".

simkessy:
> Hey. Let me know when this is ready for another review. I see some of the changes still need to be implemented.
> 

@ashleyseto hey, yea I think it's good to review.

ashleyseto:
Thanks for all the updates! Seems like the visuals are correct. There are a few UX things and I think that bug is still happening. I'm looking at this on Chrome.

*On load of the page, the typeahead dropdown won't allow an item to be selected in the dropdown. A user should always be able to select an item vs typing something.
typeahead

*Clicking the x to clear the field causes the dropdown to pop open briefly. Is there any way to prevent that?

*When an invalid value is in the field, clicking anywhere else on the page causes the dropdown to briefly pop open when the field clears then close. The dropdown shouldn't pop open when the user clicks elsewhere on the page.

There may be additional comments from @christianvuerings as I can't evaluate the code quality :)

packages/gestalt/src/Typeahead.js
```diff
@@ -0,0 +1,351 @@
...
+
+type OptionObject = {
+  label: string,
+  value: string,
+};
+
+type Props = {|
+  accessibilityLabel: string,
+  id: string,
+  onChange: ({ event: SyntheticInputEvent<>, value: string }) => void,
```
christianvuerings, line 38: @simkessy could we

* Alphabetize these? Looks like `onChange` appears twice in this list.
* Make the events more specific? Same as what we do with `TextField`: `onFocus | ({ event: SyntheticFocusEvent<HTMLInputElement>, value: string }) => void`
image

packages/gestalt/src/Typeahead.js
```diff
@@ -0,0 +1,351 @@
...
+  | 'maroon'
+  | 'watermelon'
+  | 'orange'
+  | 'darkGray'
+  | 'gray'
+  | 'lightGray'
+  | 'red'
+  | 'white';
+
+type OptionObject = {
```
christianvuerings, line 30: @simkessy could we make these [exact](https://flow.org/en/docs/types/objects/#toc-exact-object-types)? Just to maintain the same syntax in this file `{| |}`

packages/gestalt/src/Typeahead.js
```diff
@@ -0,0 +1,351 @@
...
+    onSelect,
+    value = '',
+    data,
+    placeholder,
+    size,
+    resultHeight = '50vh',
+    searchField = 'label',
+    defaultItem = null,
+    caret = false,
+    noResultText = 'No Results',
```
christianvuerings, line 81: @simkessy this should be a mandatory prop and we should not have a default. That way we enforce `i18n` for this label.

docs/src/Typeahead.doc.js
```diff
@@ -0,0 +1,147 @@
...
+    defaultCode={`
+      function Example(props) {
+        const [item, setItem] = React.useState('')
+
+        const options = Array.from(Array(100).keys()).map(item => ({value: "value-" + item, label: "Label-" + item}))
+
+        const handleOnChange = ({value}) => {
+          setItem(value);
+
+          console.log("onchange", value)
```
christianvuerings, line 121: Could we update:

* The indentation of the example
* Remove all the `console.log` messages.

Might be good to show the selected item below the `<Typeahead />`:
`<Text>Selected Item: ${item}</Text>`

image

packages/gestalt/src/Typeahead.js
```diff
@@ -0,0 +1,351 @@
...
+import React, { useState, useRef } from 'react';
+import PropTypes from 'prop-types';
+import InputField from './TypeaheadInputField.js';
+import Option from './TypeaheadOption.js';
+import Box from './Box.js';
+import Text from './Text.js';
+import Flyout from './Flyout.js';
+import Layer from './Layer.js';
+
+type TextColors =
```
christianvuerings, line 11: @simkessy can we remove all these props and just keep the ones specified in the docs? We don't want

* noResultTextColor
* hoverColor
* textColor
* backgroundColor

packages/gestalt/src/Typeahead.js
```diff
@@ -0,0 +1,351 @@
...
+  searchField?: string,
+  onBlur?: () => void,
+  onChange?: ({
+    event: SyntheticEvent<HTMLInputElement>,
+    value: string,
+  }) => void => void,
+  onFocus?: () => void,
+  onSelect?: OptionObject => void,
+  value?: string,
+  caret?: boolean,
```
christianvuerings, line 55: @simkessy can we remove the caret option?

packages/gestalt/src/Typeahead.js
```diff
@@ -0,0 +1,351 @@
...
+  value?: ?string,
+  searchField?: string,
+  onBlur?: () => void,
+  onChange?: ({
+    event: SyntheticEvent<HTMLInputElement>,
+    value: string,
+  }) => void => void,
+  onFocus?: () => void,
+  onSelect?: OptionObject => void,
+  value?: string,
```
christianvuerings, line 54: @simkessy `value` seems duplicated in this list, I also don't see it in the docs.

Could we make sure the docs + these flow types are in sync?

packages/gestalt/src/TypeaheadOption.js
```diff
@@ -0,0 +1,133 @@
...
+
+type OptionProps = {|
+  option: OptionObject,
+  selected?: OptionObject | null,
+  searchField: string,
+  hoverColor: string,
+  textColor: TextColors,
+  backgroundColor: string,
+  handleSelect: OptionObject => void,
+  // getOptionRef: (HTMLElement | null) => void,
```
christianvuerings, line 41: @simkessy can we remove this comment?

packages/gestalt/src/TypeaheadOption.js
```diff
@@ -0,0 +1,133 @@
...
+type OptionObject = {
+  label: string,
+  value: string,
+};
+
+type OptionProps = {|
+  option: OptionObject,
+  selected?: OptionObject | null,
+  searchField: string,
+  hoverColor: string,
```
christianvuerings, line 37: @simkessy please remove `hoverColor` / `textColor` / `backgroundColor`

packages/gestalt/src/TypeaheadOption.js
```diff
@@ -0,0 +1,133 @@
...
+|};
+
+const Option = ({
+  option,
+  selected,
+  searchField,
+  handleSelect,
+  hoverColor,
+  textColor,
+  // getOptionRef,
```
christianvuerings, line 51: @simkessy could we remove this comment?

packages/gestalt/src/TypeaheadOption.js
```diff
@@ -0,0 +1,133 @@
...
+        onMouseEnter={() => setHover(true)}
+        onMouseLeave={() => setHover(false)}
+      >
+        <Box
+          margin={1}
+          padding={2}
+          color={isSelectedItem || hover ? hoverColor : backgroundColor}
+          rounding={2}
+        >
+          {/* TODO: It'd be cool to render whatever here */}
```
christianvuerings, line 88: @simkessy let's remove this comment

packages/gestalt/src/TypeaheadOption.js
```diff
@@ -0,0 +1,133 @@
...
+  return (
+    <Box
+      // ref={ref => {
+      //   // Only send ref of selected item
+      //   if (selected) getOptionRef(ref);
+      // }}
+      width="100%"
+      display="flex"
+      role="option"
+      aria-selected={isSelectedItem}
```
christianvuerings, line 73: @simkessy could we
* Set the `margin` and `padding` on this surrounding `Box`
* Also set the `rounding` on `TapArea`?
That should improve the focus outline on each option

Current behavior
typeahead-focus-outline

packages/gestalt/src/TypeaheadOption.js
```diff
@@ -0,0 +1,133 @@
...
+  | 'maroon'
+  | 'watermelon'
+  | 'orange'
+  | 'darkGray'
+  | 'gray'
+  | 'lightGray'
+  | 'red'
+  | 'white';
+
+type OptionObject = {
```
christianvuerings, line 28: @simkessy could we make this exact?

packages/gestalt/src/index.js
```diff
@@ -47,6 +47,8 @@ import TextField from './TextField.js';
 import Toast from './Toast.js';
 import Tooltip from './Tooltip.js';
 import Video from './Video.js';
+// eslint-disable-next-line import/extensions
```
christianvuerings, line 50: @simkessy please remove this `eslint-disable-next-line` and add `.js` the the end:

```js
import Typeahead from './Typeahead.js';
```

packages/gestalt/src/Typeahead.js
```diff
@@ -0,0 +1,351 @@
...
+  noResultTextColor: PropTypes.oneOf(TEXT_COLORS),
+  resultHeight: PropTypes.string,
+  hoverColor: PropTypes.oneOf(TEXT_COLORS),
+  textColor: PropTypes.oneOf(TEXT_COLORS),
+  backgroundColor: PropTypes.oneOf(TEXT_COLORS),
+};
+
+export default Typeahead;
+
+// ------------------------------
```
christianvuerings, line 324: @simkessy can we remove these comments?

packages/gestalt/src/Typeahead.js
```diff
@@ -0,0 +1,351 @@
...
+              maxHeight={resultHeight}
+              width={`${inputRef?.current?.offsetWidth || 300 - 10}px`}
+              overflow="auto"
+              color={backgroundColor}
+            >
+              <Box
+                alignItems="center"
+                direction="column"
+                display="flex"
+                marginStart={-1}
```
christianvuerings: @simkessy negative `marginStart` / `marginEnd` should only be used for symmetrical padding. Could we remove it?

packages/gestalt/src/Typeahead.js
```diff
@@ -0,0 +1,351 @@
...
+          >
+            <Box
+              // ref={getContainerRef}
+              position="relative"
+              display="block"
+              padding={1}
+              marginTop={2}
+              marginBottom={2}
+              maxHeight={resultHeight}
+              width={`${inputRef?.current?.offsetWidth || 300 - 10}px`}
```
christianvuerings, line 237: @simkessy could you add a comment to explain these numbers `300 - 10`?

packages/gestalt/src/Typeahead.js
```diff
@@ -0,0 +1,351 @@
...
+              overflow="auto"
+              color={backgroundColor}
```
christianvuerings, line 239: @simkessy as mentioned before - let's remove these color options for the first release

packages/gestalt/src/Typeahead.js
```diff
@@ -0,0 +1,351 @@
...
+            anchor={inputRef.current}
+            idealDirection="down"
+            onDismiss={() => {}}
+            positionRelativeToAnchor={false}
+            size="flexible"
+          >
+            <Box
+              // ref={getContainerRef}
+              position="relative"
+              display="block"
```
christianvuerings, line 232: @simkessy `block` should be the default, can we remove it?

packages/gestalt/src/Typeahead.js
```diff
@@ -0,0 +1,351 @@
...
+    setSelected(item);
+
+    setSearch(item[searchField]);
+    setFocused(false);
+    if (onSelect) onSelect(item);
+  };
+
+  return (
+    <Box ref={componentRef}>
+      {/* INPUT FIELD */}
```
christianvuerings, line 204: @simkessy let's remove these uppercase comments.

packages/gestalt/src/Typeahead.js
```diff
@@ -0,0 +1,351 @@
...
+    setSearch('');
+    setFocused(false);
+  };
+
+  // Handler for when an item is clicked
+  const handleSelect = item => {
+    setSelected(item);
+
+    setSearch(item[searchField]);
+    setFocused(false);
```
christianvuerings, line 198: @simkessy this is not what we want. Once a user selects an item, we do want to have the focus on the input element but not show the options list.

https://material-ui.com/components/autocomplete/ is a good example

packages/gestalt/src/Typeahead.js
```diff
@@ -0,0 +1,351 @@
...
+    setSearch(value);
+
+    // Run onChange callback
+    if (onChange) onChange({ event, value });
+  };
+
+  const handleClear = () => {
+    setSelected(null);
+    setSearch('');
+    setFocused(false);
```
christianvuerings, line 190: @simkessy clicking "Clear" should still focus the input field

packages/gestalt/src/Typeahead.js
```diff
@@ -0,0 +1,351 @@
...
+      setTimeout(() => {
+        setFocused(document.activeElement === componentRef.current);
+      }, 200);
+    }
+    // Run blur callback
+    if (onBlur) onBlur();
+  };
+
+  // Handler for when text is typed
+  // This rule is stupid
```
christianvuerings, line 171: @simkessy let's remove these comments + the eslint-disable

packages/gestalt/src/Typeahead.js
```diff
@@ -0,0 +1,351 @@
...
+  const handleBlur = () => {
+    // Clear input and reset options
+    if (options.length === 0) {
+      setSearch('');
+      setOptions(dataRef.current);
+    }
+
+    // TODO: Is this the best way to hide the results on blur
+    if (document.activeElement && componentRef.current) {
+      setTimeout(() => {
```
christianvuerings, line 162: @simkessy Using `setTimeout` will break if we unmount the component without clearing the timeout.

Let's figure out a better way to do this.

packages/gestalt/src/Typeahead.js
```diff
@@ -0,0 +1,351 @@
...
+  // let containerRef;
+  // const getContainerRef = ref => {
+  //   containerRef = ref;
+  // };
+
+  // Handle when input is in and out of focus
+  const componentRef = useRef();
+  const [focused, setFocused] = useState<boolean>(false);
+
+  // When the menu item opens, scroll to selected item
```
christianvuerings, line 126: @simkessy let's remove these comments.

packages/gestalt/src/Typeahead.js
```diff
@@ -0,0 +1,351 @@
...
+  const [selected, setSelected] = useState<OptionObject | null>(defaultItem);
+
+  const [options, setOptions] = useState<OptionObject[]>(
+    filterOriginalData(search)
+  );
+
+  // Ref to the input
+  const inputRef = useRef();
+
+  // Reference to selected option
```
christianvuerings, line 110: @simkessy lets remove these comments

packages/gestalt/src/Typeahead.js
```diff
@@ -0,0 +1,351 @@
...
+  onBlur?: () => void,
+  onChange?: ({
+    event: SyntheticEvent<HTMLInputElement>,
+    value: string,
+  }) => void => void,
+  onFocus?: () => void,
+  onSelect?: OptionObject => void,
+  value?: string,
+  caret?: boolean,
+  defaultItem?: OptionObject,
```
christianvuerings, line 56: @simkessy could we document this property + show it in an example?

packages/gestalt/src/TypeaheadInputField.css
```diff
@@ -0,0 +1,80 @@
+:root {
+  --bt: 4px;
+  --gray: #767676;
+  --lightGray: #efefef;
+  --white: #fff;
+}
+
+.input {
+  composes: accessibilityOutline from "./Focus.css";
```
christianvuerings, line 9: @simkessy let's share some of this CSS with `SearchField`. Looks like most of it is a copy/paste with a few differences.

packages/gestalt/src/TypeaheadInputField.js
```diff
@@ -0,0 +1,169 @@
...
+    [layout.large]: size === 'lg',
+    [styles.inputActive]: true,
+    [styles.inputHovered]: hovered,
+  });
+
+  const clearButtonSize = size === 'lg' ? 24 : 20;
+  const clearIconSize = size === 'lg' ? 12 : 10;
+
+  return (
+    <span>
```
christianvuerings, line 91: @simkessy can we remove this `<span />` and empty string below `{' '}`

packages/gestalt/src/TypeaheadInputField.js
```diff
@@ -0,0 +1,169 @@
...
+          aria-label={accessibilityLabel}
+          className={className}
+          id={id}
+          onChange={handleChange}
+          placeholder={placeholder}
+          type="text"
+          value={value}
+        />
+
+        {/* INPUT ICON AND CLEAR BUTTON */}
```
christianvuerings, line 114: @simkessy let's remove all these uppercase comments

packages/gestalt/src/TypeaheadInputField.js
```diff
@@ -0,0 +1,169 @@
...
+import PropTypes from 'prop-types';
+import classnames from 'classnames';
+import layout from './Layout.css';
+import styles from './TypeaheadInputField.css';
+import Box from './Box.js';
+import Icon from './Icon.js';
+import FormLabel from './FormLabel.js';
+
+type Props = {|
+  accessibilityLabel: string,
```
christianvuerings, line 13: @simkessy let's remove `accessibilityLabel` and keep `label` - they currently have overlapping behavior

packages/gestalt/src/TypeaheadInputField.js
```diff
@@ -0,0 +1,169 @@
...
+        alignItems="center"
+        display="flex"
+        onMouseEnter={handleMouseEnter}
+        onMouseLeave={handleMouseLeave}
+        onFocus={handleFocus}
+        onBlur={handleBlur}
+        position="relative"
+      >
+        <input
+          ref={forwardedRef}
```
christianvuerings, line 104: @simkessy we probably want to add `autocomplete="off"` so it doesn't show the native autocomplete

image

docs/src/Typeahead.doc.js
```diff
@@ -0,0 +1,147 @@
...
+
+card(
+  <PropTable
+    props={[
+      {
+        name: 'resultHeight',
+        type: 'string',
+        description: 'Set the height of results box',
+        href: 'basicExample',
+        defaultValue: '50vh',
```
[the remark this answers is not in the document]
ashleyseto: I think we'd want to not make it changeable? It should behaving like a normal select list where it scrolls when hitting the limits of the page instead of needing to scroll the page to see more of the dropdown.

packages/gestalt/src/Typeahead.js
simkessy: Sure but can I get a better understanding into why we force a certain object type into the component? If a user's data doesn't return in the format of `label/value` why are we forcing them into it and what benefit does that provide the user?

packages/gestalt/src/Typeahead.js
simkessy: Added

ashleyseto:
Reviewed initial feedback, and it has been resolved in the latest changes. 

Outstanding is the dropdown height and any of Christian's comments. 

Also in the examples, we may want to use the label component. https://gestalt.netlify.app/#/Label

simkessy:
Hey @christianvuerings, should have all the UI issues from @ashleyseto  addressed and I've gone through and resolved the ones you commented on. I started getting these last set of Flow issues 
image

If you have any ideas that would be helpful. Thanks

christianvuerings:
> Hey @christianvuerings, should have all the UI issues from @ashleyseto addressed and I've gone through and resolved the ones you commented on. I started getting these last set of Flow issues
> image
> 
> If you have any ideas that would be helpful. Thanks

@simkessy That's because we enables flow types-first. See https://medium.com/flow-type/types-first-a-scalable-new-architecture-for-flow-3d8c7ba1d4eb for more info. #940 contains a couple of examples of how to flowtype the proptypes.

docs/src/Typeahead.doc.js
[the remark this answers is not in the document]
ashleyseto: I am not sure if this is relevant anymore I think this should be not changeable. @simkessy

simkessy:
> Seems to be working correctly from what I see. The only change I could see that might be needed is the docs using the label component rather than text.

Hey @ashleyseto 

> Seems to be working correctly from what I see. The only change I could see that might be needed is the docs using the label component rather than text.

Hey @ashleyseto , I checked with the TextField component and it is using the FormLabel component. They should probably match? If not let me know and I'll make the update.

docs/src/Typeahead.doc.js
```diff
@@ -0,0 +1,193 @@
...
+    `}
+  />
+);
+
+card(
+  <Example
+    id="defaultItemExample"
+    name="Default Item Example"
+    defaultCode={`
+      function Example(props) {
```
christianvuerings, line 153: @simkessy thanks for adding the up/down arrow key functionality. Could we only update the `selected` option when the user hits enter? Similar as to https://material-ui.com/components/autocomplete/ does it?

Right now, the selected option changes as you go up and down with the arrow keys, which is unexpected
typeahead-arrow-change

packages/gestalt/src/Typeahead.js
```diff
@@ -0,0 +1,301 @@
...
+  }) => void,
+  onFocus?: ({
+    event: SyntheticFocusEvent<HTMLInputElement>,
+    value: string,
+  }) => void,
+  onSelect?: (?OptionObject) => void,
+  placeholder?: string,
+  searchField?: string,
+  size?: 'md' | 'lg',
+  forwardedRef?: React.Ref<'input'>,
```
christianvuerings, line 43: @simkessy thanks for doing the cleanup on all these props!

packages/gestalt/src/Typeahead.js
```diff
@@ -0,0 +1,301 @@
...
+  }) => void,
+  onChange?: ({
+    event: SyntheticInputEvent<HTMLInputElement>,
+    value: string,
+  }) => void,
+  onFocus?: ({
+    event: SyntheticFocusEvent<HTMLInputElement>,
+    value: string,
+  }) => void,
+  onSelect?: (?OptionObject) => void,
```
christianvuerings, line 39: @simkessy could we update the flow type so it reflects the docs? In Gestalt we always return an object in the callback.

`({ event: SyntheticInputEvent<>, value: string }) => void`?

Perhaps it should also include the label?
`({ event: SyntheticInputEvent<>, value: string, label: string }) => void`?

packages/gestalt/src/TypeaheadInputField.js
```diff
@@ -0,0 +1,207 @@
...
+import Box from './Box.js';
+import Icon from './Icon.js';
+import FormLabel from './FormLabel.js';
+
+type Props = {|
+  id: string,
+  onBlur?: ({
+    event:
+      | SyntheticFocusEvent<HTMLInputElement>
+      | SyntheticKeyboardEvent<HTMLInputElement>,
```
christianvuerings, line 19: @simkessy should we remove `| SyntheticKeyboardEvent<HTMLInputElement>,` here since it is only a `SyntheticFocusEvent<HTMLInputElement>`

Similar as to `onFocus`

packages/gestalt/src/TypeaheadInputField.js
```diff
@@ -0,0 +1,207 @@
...
+    else if (event.keyCode === 40) {
+      onKeyNavigation(1);
+    }
+    // Enter Key
+    else if (event.keyCode === 13) {
+      handleBlur(event);
+    }
+  };
+
+  const hasValue = value && value?.length > 0;
```
christianvuerings, line 113: @simkessy should we write this as `const hasValue = value?.length > 0;`?

packages/gestalt/src/TypeaheadInputField.js
```diff
@@ -0,0 +1,207 @@
...
+            alignItems="center"
+            color="transparent"
+            display="flex"
+            height={clearButtonSize}
+            justifyContent="center"
+            rounding="circle"
+            width={clearButtonSize}
+          >
+            <Icon
+              accessibilityLabel="typeahead clear button"
```
christianvuerings, line 168: @simkessy since this is not selectable by keyboard, could you change it to an empty accessibility label? `accessibilityLabel=""`

packages/gestalt/src/Typeahead.js
```diff
@@ -0,0 +1,301 @@
...
+  label: string,
+  value: string,
+|};
+
+type Props = {|
+  data: Array<{
+    label: string,
+    value: string,
+  }>,
+  defaultItem?: OptionObject,
```
christianvuerings, line 22: @simkessy here `defaultItem` is optional but in the docs it is required. Could you make sure the docs and the code align?

packages/gestalt/src/Typeahead.js
```diff
@@ -0,0 +1,301 @@
...
+
+    setSearch(item[searchField]);
+    setContainerOpen(false);
+    if (inputRef.current) inputRef.current.focus();
+    if (onSelect) onSelect(item);
+  };
+
+  const handleScrolling = item => {
+    const selectedNode = [
+      ...document.querySelectorAll('div.typeahead-option'),
```
christianvuerings, line 142: @simkessy this leaks abstraction outside of this component. Instead, could we set a `ref` on the container of options?

packages/gestalt/src/Typeahead.js
```diff
@@ -0,0 +1,301 @@
...
+            onDismiss={() => {}}
+            positionRelativeToAnchor={false}
+            size="flexible"
+          >
+            <Box
+              position="relative"
+              padding={1}
+              marginTop={2}
+              marginBottom={2}
+              maxHeight="50vh"
```
christianvuerings, line 232: @simkessy @ashleyseto what was the resolution on the maxHeight? Should we use an absolute value instead of 50% of the viewport height? E.g. `400px`

packages/gestalt/src/Typeahead.js
```diff
@@ -0,0 +1,301 @@
...
+              width={`${inputRef?.current?.offsetWidth}px`}
```
christianvuerings, line 233: @simkessy could you see if `width={inputRef?.current?.offsetWidth}` also works?

packages/gestalt/src/TypeaheadInputField.css
```diff
@@ -0,0 +1,23 @@
...
+  composes: right0 from "./Layout.css";
+  composes: transparentBg from "./Colors.css";
+  outline: none;
+  position: absolute;
+  top: 50%;
+  transform: translateY(-50%);
+}
+
+.inputRadius {
+  border-radius: 16px;
```
christianvuerings, line 17: @simkessy could we compose from `rounding4` in `packages/gestalt/src/Borders.css`?

packages/gestalt/src/TypeaheadInputField.css
```diff
@@ -0,0 +1,23 @@
+
+.icon {
+  composes: marginEnd2 from "./boxWhitespace.css";
```
christianvuerings, line 3: @simkessy we can combine some of these:

```
composes: marginEnd2 paddingX1 paddingY1 from "./boxWhitespace.css";
```

packages/gestalt/src/TypeaheadOption.js
```diff
@@ -0,0 +1,92 @@
...
+  };
+
+  const className = classnames(styles.touchable, getRoundingClassName(2), {
+    [styles.fullWidth]: true,
+    [styles.pointer]: true,
+  });
+
+  return (
+    <div
+      className={`${className} typeahead-option ${
```
christianvuerings, line 49: @simkessy could we
* remove `typeahead-option` - we do not allow static classnames
* make `isSelectedItem ? focusStyle.typeaheadOptionFocus : ''` part of the classnames function above

docs/src/Typeahead.doc.js
```diff
@@ -0,0 +1,193 @@
...
+        );
+      }
+    `}
+  />
+);
+
+card(
+  <Example
+    id="defaultItemExample"
+    name="Default Item Example"
```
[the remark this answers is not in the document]
simkessy: AntD uses different colors to indicate what's selected vs what's  hovered on. 

They also disable the mouse hovered item when key nav is triggered (I'll how to look into how they're doing that.)


image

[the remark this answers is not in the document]
simkessy: I went with this. @ashleyseto let me know what you think about the colors. 

'gray' : selected item 
'lightGray' : currently hovered item 

image

packages/gestalt/src/Typeahead.js
simkessy: Last statement on this was that users shouldn't have control over the default. No specific number / height was provided.

docs/src/Typeahead.doc.js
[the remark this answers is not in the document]
ashleyseto: Screen Shot 2020-06-29 at 9 05 07 AM

If we want to add a selected state, I'd refer back to the Figma file where we're using a checkmark to indicate selected across other areas.

packages/gestalt/src/Typeahead.js
ashleyseto: If we use an absolute value, it can go off the screen right? The goal is to keep it on the screen. The 50% seems to be working but I was curious how other design systems handle this component.

docs/src/Typeahead.doc.js
```diff
@@ -0,0 +1,196 @@
...
+        setSelected(item);
+      };
+
+      return (
+        <>
+          <Box marginBottom={2}>
+            <Text>Selected Item: {(selected && selected.label) || ""}</Text>
+          </Box>
+
+          <Typeahead
```
christianvuerings, line 136: @simkessy could we use `label` in this case to make it more accessible? That way, when someone clicks on the label, they focus into the typeahead + the label gets read out by screenreaders.

```
      const label = `Selected Item: ${(selected && selected.label) || ""}`;

      return (
        <Typeahead
          label={label}
          id="Typeahead-example"
          noResultText="No Results"
          data={options}
          placeholder="Select a Label"
          onChange={handleOnChange}
          value={item}
          onSelect={handleSelect}
        />
      );
```

christianvuerings: same with the example below

docs/src/Typeahead.doc.js
```diff
@@ -0,0 +1,196 @@
...
+    ]}
+  />
+);
+
+card(
+  <Example
+    id="basicExample"
+    name="Example"
+    defaultCode={`
+    function Example(props) {
```
christianvuerings, line 113: @simkessy could we fix the indentation  of the examples?

`function Example(props) {` instead of `    function Example(props) {`.

packages/gestalt/src/Typeahead.js
```diff
@@ -0,0 +1,341 @@
...
+  size: PropTypes.oneOf(['md', 'lg']),
+  searchField: PropTypes.string,
+  onBlur: PropTypes.func,
+  onFocus: PropTypes.func,
+  onSelect: PropTypes.func,
+  defaultItem: PropTypes.exact({
+    label: PropTypes.string.isRequired,
+    value: PropTypes.string.isRequired,
+  }),
+  noResultText: PropTypes.string,
```
christianvuerings, line 330: @simkessy 
* `label` seems to be missing from this list
* `data` / `id` and `noResultText` should all be required

packages/gestalt/src/Typeahead.js
```diff
@@ -0,0 +1,341 @@
...
+  const containerRef = useRef();
+
+  const handleScrolling = (direction: number) => {
+    const container = containerRef.current;
+
+    // Based on keyboard navigation we get the next or previuos option
+    // When we reach the start or end of the list, move to the start or end of the list based on the direction
+    const nextOption =
+      direction > 0
+        ? // $FlowFixMe[incompatible-cast]
```
christianvuerings, line 160: @simkessy could you explain why we need casting here + can we get rid of the `// $FlowFixMe`?

packages/gestalt/src/TypeaheadInputField.js
```diff
@@ -0,0 +1,206 @@
...
+          onBlur={handleBlur}
+          onClick={() => setContainer(true)}
+          onChange={handleChange}
+          placeholder={placeholder}
+          type="text"
+          value={value}
+          onKeyDown={handleKeyNavigation}
+        />
+
+        <button
```
christianvuerings, line 151: @simkessy please update the `x` button so it also works with RTL. See #970 how we did it for `SearchField`

SearchField (correct)
image

Typeahead
image

packages/gestalt/src/Typeahead.js
```diff
@@ -0,0 +1,341 @@
...
+      selectedOption.clientHeight -
+      containerHeight +
+      overScroll;
+
+    container.scrollTop = scrollPos;
+  };
+
+  const handleKeyNavigation = (
+    event: SyntheticKeyboardEvent<HTMLInputElement>,
+    direction: number
```
christianvuerings, line 190: @simkessy can we make this more specific? `direction: -1 | 0 | 1`

packages/gestalt/src/TypeaheadOption.js
```diff
@@ -0,0 +1,98 @@
...
+  option: PropTypes.exact({
+    label: PropTypes.string.isRequired,
+    value: PropTypes.string.isRequired,
+  }),
+  selected: PropTypes.exact({
+    label: PropTypes.string.isRequired,
+    value: PropTypes.string.isRequired,
+  }),
+  searchField: PropTypes.string,
+  handleSelect: PropTypes.func,
```
christianvuerings, line 97: @simkessy could we update these proptypes with the flow types?

```
  index: number,
  option: OptionObject,
  selected?: OptionObject | null,
  searchField: string,
  handleSelect: ({|
    item: OptionObject,
    event: SyntheticFocusEvent<HTMLInputElement>,
  |}) => void,
  hoveredItem: ?number,
  setHoveredItem: (?React.Node, number) => void,
  setOptionRef: (?HTMLElement) => void,
```

packages/gestalt/src/TypeaheadInputField.css
```diff
@@ -0,0 +1,16 @@
...
+  composes: transparentBg from "./Colors.css";
+  outline: none;
+  position: absolute;
+  top: 50%;
+  transform: translateY(-50%);
+}
+
+.inputRadius {
+  composes: rounding4 from './Borders.css'
+}
```
christianvuerings, line 16: @simkessy NIT could we
* Remove the first empty line in this file
* Add an extra empty line add the end of the file

image

docs/src/Typeahead.doc.js
simkessy: image

ashleyseto:
Not sure where we ended up with the select state since I was out last week but the Figma file does contain the latest. I noticed in the current build the select state is gray, and it changes gray shades depending on where you hover.

Screen Shot 2020-07-15 at 12 58 43 PM

https://www.figma.com/file/rJPa4obG6uWXU4gAI61wLS/List-component-Gestalt?node-id=298%3A81

packages/gestalt/src/Typeahead.js
simkessy: the nextSibling callback returns a type of Node however nextOption is should be htmlElement

PR #911, opened 2020-06-09
Pog: Add padding prop

jackhsu978:
### Minor changes:
- add `padding` prop where possible value are 1 ~ 5 boints. if omitted, default padding (which may or may not be an integer boint) would be used.

### Revisions:
- removed unnecessary use of `Box` in the component

## TODO
- [x] Accessibility checkup
- [x] Update documentation
- [x] Add/update Tests

### docs

Screenshot 2020-06-11 19 23 18

Screenshot 2020-06-11 19 24 04

### DIFF
--- a/docs/src/Pog.doc.js
+++ b/docs/src/Pog.doc.js
@@ -66,8 +66,8 @@ card(
       },
       {
         name: 'padding',
-        type: `"xs" | "sm" | "md" | "lg" | "xl" | 4 | 12 | 20`,
-        description: `Available padding are "xs" (6px), "sm" (8px), "md" (11px), "lg" (14px), "xl" (16px), 4 (px), 12 (px), and 20 (px). By default, this value is the same as the \`size\` prop.`,
+        type: `1 | 2 | 3 | 4 | 5`,
+        description: `Padding in boints. If omitted, padding is derived from the \`size\` prop.`,
         href: 'paddingCombinations',
       },
       {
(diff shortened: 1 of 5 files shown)


christianvuerings:
@jackhsu978 could you share a few screenshots / instances where we need a different `padding` prop in the design?

jackhsu978:
> @jackhsu978 could you share a few screenshots / instances where we need a different `padding` prop in the design?

Here are some examples

- Example where size is "xl" but padding is 2 boints (8px, or "sm"):
Navigation footer button
Screenshot 2020-06-09 16 57 10

- Example where size is "xl" but padding is 3 boints (12px):
Close button on board edit page
Screenshot 2020-06-09 17 00 49

- Example where size is "md" but padding is 1 boint (4px):
Search bar cancel button
Screenshot 2020-06-09 17 02 27

jackhsu978:
I talked to Annie and here is a summary of what we concluded:
- having a fix icon-to-container-size ratio is ideal, but it isn't practical because we have a lot of cases that don't fall into a fixed ratio. in fact, we may have 20+ different ratios at the current state and there are cases that are totally valid.
- The default Pog/IconButton "md" is problematic. The button size is 40x40 (icon size = 18x18 and padding = 11), but that is actually not part of PDS. Also, we don't want the padding size to be an odd number (11). We should change it.
- We want to avoid resizing the icon button container size without manually testing because doing so can break layout. I did some jscodeshift analysis and found that there are 98 instances in Pinboard where we use the default "md" size. It's too time consuming to manually check each instance, so we will try to fix weird 11px padding without changing the container size.
- We will make "md" icon size slightly smaller while keeping the container size the same: icon size = 16x16 and padding = 12px, which is the same as the default padding in the Duplo-forked IconButton. Note that the icon size will be the same as "sm", so it's not great, but we agreed that it's ok.

jackhsu978:
On a second thought, we can simplify the API significantly.

The `padding` is auto-derived by default. This stays the same. The `padding` prop allows custom padding in boints, and that's the Duplo's IconButton uses today. I originally thought that we would also allow one to pass the padding that Gestalt IconButton (or Pog) supports, but that doesn't actually add any values.

PR #915, opened 2020-06-11
Badge: Update design and deprecate positioning

denamwangi:
This adds the second piece of the revamped design on the Badge which increases the font size to 12px, increases the border radius to 8px, and updates the padding to match. It also deprecates the positioning prop and no longer allows for top positioning.

After
Screen Shot 2020-06-11 at 9 13 20 AM

Before
Screen Shot 2020-06-11 at 9 14 50 AM

## TODO

- [ ] Accessibility checkup
- [ ] Update documentation
- [x] Add/update Tests

* also left to do before this lands: update all Badge instances in pinboard using top.


packages/gestalt/src/Badge.js
```diff
@@ -1,24 +1,17 @@
 // @flow strict
 import * as React from 'react';
 import PropTypes from 'prop-types';
-import cx from 'classnames';
 import styles from './Badge.css';
-import colors from './Colors.css';
 
 type Props = {|
-  position?: 'middle' | 'top',
```
christianvuerings, line 9: @denamwangi love the cleanup here! Since we are removing a prop, could you write a codemod for it? That way doing the update in the Pinterest codebase becomes easier.

https://github.com/pinterest/gestalt/tree/master/packages/gestalt-codemods contains a lot of examples and https://astexplorer.net/ is what you can use to test.

https://github.com/pinterest/gestalt/blob/master/packages/gestalt-codemods/1.35.0-1.36.0/leading-text-remove.js might be a good start, which converted
`<Text leading='short' />` to `<Text/>`
just like we will have to convert
`<Badge position="middle" />` to `<Badge/>`

ashleyseto:
From Annie slack:
The text seems 1px low
the text should always be regular weight

Waiting to learn of any edge cases of multi-line text that we'd have to reevaluate. Seems not to be an issue from what we have gathered so far.

denamwangi:
I'll circle back to this in the next week or two!

PR #916, opened 2020-06-11
Icon: Add sparkle icon

denamwangi:
This adds a new icon to our icon library: the sparkle icon.

Screen Shot 2020-06-11 at 9 24 54 AM

## TODO

- [ ] Accessibility checkup
- [ ] Update documentation
- [ ] Add/update Tests


ashleyseto:
Scott on PDS says this is approved.

packages/gestalt/src/icons/sparkle.svg
```diff
@@ -0,0 +1 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="24" height="24"><path d="M23.84 4.21a5.34 5.34 0 00-3.63 3.63c-.06.21-.36.21-.42 0a5.34 5.34 0 00-3.63-3.63c-.21-.06-.21-.36 0-.42A5.34 5.34 0 0019.79.16c.06-.21.36-.21.42 0a5.34 5.34 0 003.63 3.63c.21.06.21.36 0 .42zm-3.25 9.84a14.02 14.02 0 00-9.54 9.54c-.16.55-.94.55-1.1 0A14.02 14.02 0 00.4 14.05c-.55-.16-.55-.94 0-1.1A14.02 14.02 0 009.95 3.4c.16-.55.94-.55 1.1 0a14.02 14.02 0 009.54 9.54c.55.16.55.94 0 1.1z"/></svg>
```
AlbertCarreras, line 1: Hey, not an expert here on icons but I see most icons have the property fill. Do we need that as well?

<path fill="#767676" d="M

AlbertCarreras:
Thanks! looks good!

packages/gestalt/src/icons/sparkle.svg
denamwangi: I'm also not too familiar with when we need it and when we don't but I saw here that the updated icon in this case didn't have a fill either https://github.com/pinterest/gestalt/pull/881/files? @christianvuerings might be able to shed some light for us. 

I think PDS has a standardized way to export these now and I wonder if that optimization changed things up.

christianvuerings: @AlbertCarreras @denamwangi the fill doesn't matter that much since we only use the `path` in from the SVG. 
https://github.com/pinterest/gestalt/blob/a6f7152ed194c68d93d4e8ef626755517d132865/packages/gestalt/rollup.config.js#L36

PR #917, opened 2020-06-11
[WIP] Pog: Add blue, green, and pine background colors

denamwangi:
This adds three additional colors to the Pog component to support designs on a new surface. 
Screen Shot 2020-06-11 at 10 16 38 AM
Screen Shot 2020-06-11 at 10 16 27 AM

## TODO

- [ ] Accessibility checkup
- [ ] Update documentation
- [ ] Add/update Tests


ashleyseto:
Solution from Christian:
In this case, on the technical side, I would recommend a round <Box /> with a background color + <Icon /> inside of it. Pog should be only used if we also have hover / focus states. And if a picture is available, use <Avatar />

No need to add colors to the Pog

denamwangi:
Going with an alternate solution!

PR #919, opened 2020-06-12
Touchable: Add inline prop

dangerismycat:
This is needed for clickable elements that are part of a `inline` layout. The specific use case I'm trying to solve right now is clickable text as part of an inline layout, which currently isn't possible. A flex layout allows for inline behavior if all the elements fit on one line, but that breaks if that wraps to the next line.

Consider this:
Screen Shot 2020-06-11 at 6 06 51 PM

Vs this:
Screen Shot 2020-06-11 at 6 08 04 PM

- [ ] Accessibility checkup
- [x] Update documentation
- [ ] Add/update Tests

### DIFF
--- a/docs/src/Touchable.doc.js
+++ b/docs/src/Touchable.doc.js
@@ -69,6 +69,10 @@ card(
         href: 'fullHeightWidthExample',
       },
       {
+        name: 'inline',
+        type: 'boolean',
+      },
+      {
         name: 'mouseCursor',
         type: `"copy" | "grab" | "grabbing" | "move" | "noDrop" | "pointer" | "zoomIn" | "zoomOut"`,
         defaultValue: 'pointer',
@@ -212,12 +216,12 @@ function A11yExPopup() {
           accessibilityLabel="see more"
           onTouch={() => setOpen(!isOpen)}
         >
-        <Box 
-          alignItems="center" 
+        <Box
+          alignItems="center"
           borderSize="sm"
-          display="flex" 
-          padding={2} 
-          rounding="pill" 
+          display="flex"
+          padding={2}
+          rounding="pill"
         >
             <Box paddingX={1}>
               <Text weight="bold">See more</Text>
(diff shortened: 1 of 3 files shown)


docs/src/Touchable.doc.js
```diff
@@ -68,6 +68,10 @@ card(
         defaultValue: true,
         href: 'fullHeightWidthExample',
       },
+      {
+        name: 'inline',
+        type: 'boolean',
```
btraut, line 73: boolean props :(

Perhaps this could have `display: 'block' | 'boolean'` with default `block`?

dangerismycat: Agreed in general, but I was thinking of this more like the `inline` prop on Text. If we may that an enum, is that implying that we'd like to allow users to specify something like `flex` or `inlineBlock` in the future?

christianvuerings: In this case we should remain consistent with other components. We don't want to add more/new boolean props moving forward but consistency trumps that.

Also, this PR overlaps with #906 where @jackhsu978 also added this `inline` prop + other changes.

dangerismycat: Ah, I didn't realize Jack was implementing the same thing. 😅 This is no longer urgent for me, so I'm fine with abandoning this PR and waiting for Jack's, unless there's reason to continue with this one.

dangerismycat: If we do want to continue with this PR: it looks like there's a failing test related to a new CSS class I added. Is there a script that needs to be run after adding a class?

jackhsu978: I did try to add `inline` support in #906, but @bishwei correctly pointed out that it is not necessary. One can just wrap Touchable with `<Box display="inlineBox">`:

```
<Box display="inlineBox">
  <Touchable>
    ...
  </Touchable>
</Box>
```

Note that `fullWidth` prop in Touchable needs to be true for inline to work, but we don't have to explicitly set it because it is true by default.

dangerismycat: I'm pretty sure it is still necessary for inline layouts. I was facing this problem yesterday, and any non-inline element child in an inline layout breaks to the next line.

jackhsu978: Can you give some code example? Here is an live example in 906 PR: https://deploy-preview-906--gestalt.netlify.app/#/Touchable

dangerismycat:
This isn't necessary — wrapping Touchable with `<Box display="inlineBlock>` does the trick

PR #927, opened 2020-06-16
Table: Move to /Table folder

simkessy:
There are too many Table related files. Adding them to /src/Table will make it clearer where to work on the Table component as more functionality is added. 

## TODO

- [x] Accessibility checkup
- [x] Update documentation
- [x] Add/update Tests


christianvuerings:
This will break the `Source` links we should on every component: 
image

See https://github.com/pinterest/gestalt/blob/2de1d88964b299be9ffb41c8f117464e90a89e8e/docs/src/components/PageHeader.js#L13 for more info

simkessy:
> This will break the `Source` links we should on every component:
> image
> 
> See
> 
> https://github.com/pinterest/gestalt/blob/2de1d88964b299be9ffb41c8f117464e90a89e8e/docs/src/components/PageHeader.js#L13
> 
> for more info

should be resolved

packages/gestalt/src/Table/Table.test.js
```diff
@@ -1,7 +1,7 @@
 // @flow strict
 import React from 'react';
 import renderer from 'react-test-renderer';
-import Table from './Table.js';
+import Table from './index.js';
```
christianvuerings, line 4: @simkessy we should avoid `index.js` files when possible. It's hard to know what's in those files + they don't autocomplete well

packages/gestalt/src/Table/TableSortableHeaderCell.js
```diff
@@ -1,9 +1,11 @@
+/* eslint-disable eslint-comments/disable-enable-pair */
+/* eslint-disable import/no-relative-parent-imports */
```
christianvuerings, line 2: @simkessy We have this rule to avoid `../` in our codebase. So I would recommend to keep the flat structure as we currently have it and not nest components inside of `Table`

simkessy: Why do we want to keep a flat structure? Especially as we're adding more complicated components like Tables, DatePickers, Typeaheads. To have all those pieces flat seems like it would make organization challenging.  

Comparing with other UI libraries:
https://github.com/mui-org/material-ui/tree/next/packages/material-ui/src
https://github.com/ant-design/ant-design/tree/master/components

packages/gestalt/src/Table/Table.test.js
simkessy: I think my comment below also applies regarding the use of Barrels. It's a practice I've seen quite often, i'm not entirely clear why we'd deviate. Regarding the issue of confusion, I think if you're looking at the `/Table/index.js`, I don't find it confusing that this file is the entry point to the Table component. Same as how the `gestalt/src/index.js` file is the entry point to Gestalt. 

If someone is not familiar with Barrels, it would be a good learning opportunity.

christianvuerings:
Going to close this PR for now but we should revisit our folder structure once we start segmenting components into multiple categories (Form / Layout / Media / ...). I don't think it makes sense to just do this for Table, we should have a more generic approach for every component.

PR #929, opened 2020-06-17
Masonry/defaultLayout: Add a layout `basicCentered` to center justify grid content

mminute:
The Masonry grid renders items from left to right with the first item left justified.  This change adds an option to the Masonry `layout` prop called `basicCentered`.  When passed this value `defaultLayout` modifies its calculation of position to center justify the contents of the grid.

Example board: https://www.pinterest.com/pinterestcreators/story-pins/

Masonry.props.layout === 'basic'
Screen Shot 2020-06-17 at 3 10 50 PM

Masonry.props.layout === 'basicCentered'
Screen Shot 2020-06-17 at 3 10 40 PM

## TODO

- [ ] Accessibility checkup
- [ ] Update documentation
- [✓] Add/update Tests
- [✓] Pinterest Designer (for design changes) (@annieteng)
- [✓] [Format PR title](https://github.com/pinterest/gestalt/#releasing): `ComponentName: Description`.

### DIFF
--- a/packages/gestalt/src/Masonry.js
+++ b/packages/gestalt/src/Masonry.js
@@ -29,6 +29,7 @@ type Layout =
   | LegacyMasonryLayout
   | LegacyUniformRowLayout
   | 'basic'
+  | 'basicCentered'
   | 'flexible'
   | 'uniformRow';
 
@@ -490,7 +491,9 @@ export default class Masonry<T: {}> extends React.Component<
         cache: measurementStore,
         columnWidth,
         gutter,
+        justify: layout === 'basicCentered' ? 'center' : 'start',
         minCols,
+        rawItemCount: items.length,
         width,
       });
     }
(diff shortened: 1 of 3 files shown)


btraut:
Adding more context here: this PR is an extension of the former https://github.com/pinterest/gestalt/pull/905. It fixes a bug that existed in that PR, but it's essentially the same approach as far as the API is concerned. Again, I'm looking to @bishwei and @chrislloyd for any additional feedback/concerns.

The addition of `rawItemCount` to the layout function is not my favorite, but the layout really needs that information to do its job properly. FWIW: Masonry9 always passes the full list of items to layout functions even if we don't have layout info yet. This prevents the need to pass that additional information separately.

As far as I'm concerned, this change is fine and I'm going to approve. Would just love to see more eyes from others.

chrislloyd:
(I don't think this should block the PR but...)

If you wanted to avoid the `rawItemsCount` you could create a layout that just composes `basicLayout` and just calculate the full width returned by that. Would avoid all extra args needed on `basicLayout`.

PR #933, opened 2020-06-19
Link: add TapArea-like feedback

jackhsu978:
### Minor changes
- Add `<TapArea>`-like touch feedback to `<Link>`

### Revisions
- minor fix for a test name in `TapArea`

## Details

Here is a screenshot of the API documentation:

Screenshot 2020-06-19 18 09 58

It supports some of props that `<TapArea>` supports, but not all since not every option is relevant to `<TapLink>`.

Here is a screenshot of examples in the documentation:

Screenshot 2020-06-19 18 10 58

Note that the "Prevent default on tap" switch is turn on by default because it would be easier to test the TapArea-like touch feedback without actually opening the link. One can turn it off and test the link navigation.

## TODO

- [x] Accessibility checkup
- [x] Update documentation
- [x] Add/update Tests


packages/gestalt/src/TapLink.js
```diff
@@ -0,0 +1,112 @@
...
+          event.preventDefault();
+          onTap({ event });
+        }
+      }}
+      onTouchStart={handleTouchStart}
+      onTouchMove={handleTouchMove}
+      onTouchCancel={handleTouchCancel}
+      onTouchEnd={handleTouchEnd}
+      rel={[
+        ...(target === 'blank' ? ['noopener', 'noreferrer'] : []),
```
christianvuerings, line 88: @jackhsu978 I like this a lot - we should follow the same logic in `<Link />` (we should do that in a follow up PR)

docs/src/TapLink.doc.js
```diff
@@ -0,0 +1,132 @@
...
+  return (
+    <Box>
+      <Box display="flex" alignItems="center" padding={1} marginStart={-2} marginEnd={-2}>
+        <Box paddingX={2}>
+          <Label htmlFor="preventDefault">
+            <Text>Prevent default on tap</Text>
+          </Label>
+        </Box>
+        <Switch
+          preventDefault
```
christianvuerings, line 85: @jackhsu978 should this have been `id="preventDefault"` to link the label to the switch?

docs/src/TapLink.doc.js
```diff
@@ -0,0 +1,132 @@
...
+        >
+          <Text>TapLink</Text>
+        </TapLink>
+      </Box>
+      <Box padding={1}>
+        <TapLink
+          href="https://pinterest.com"
+          onTap={onTap}
+          rounding="pill"
+          tapStyle="compress"
```
christianvuerings, line 105: @jackhsu978 could we make this example also `inline` - `compress` with full width feels pretty odd

docs/src/TapLink.doc.js
```diff
@@ -0,0 +1,132 @@
...
+          '({ event: SyntheticMouseEvent<HTMLAnchorElement> | SyntheticKeyboardEvent<HTMLAnchorElement> }) => void',
+        href: 'basicExample',
+      },
+      {
+        name: 'rel',
+        type: `"none" | "nofollow"`,
+      },
+      {
+        name: 'rounding',
+        type: `"pill" | "circle" | 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8`,
```
christianvuerings, line 51: @jackhsu978 could we mention that the default value is `0`?

docs/src/TapLink.doc.js
```diff
@@ -0,0 +1,132 @@
...
+        href: 'basicExample',
+      },
+      {
+        name: 'target',
+        type: `"null" | "self" | "blank"`,
+        defaultValue: 'null',
+      },
+    ]}
```
christianvuerings, line 59: @jackhsu978 could we also add `tapStyle` to the docs?

docs/src/TapLink.doc.js
```diff
@@ -0,0 +1,132 @@
...
+      },
+      {
+        name: 'onTap',
+        type:
+          '({ event: SyntheticMouseEvent<HTMLAnchorElement> | SyntheticKeyboardEvent<HTMLAnchorElement> }) => void',
+        href: 'basicExample',
+      },
+      {
+        name: 'rel',
+        type: `"none" | "nofollow"`,
```
christianvuerings, line 47: @jackhsu978 could we mention that the default is `none`?

docs/src/TapLink.doc.js
```diff
@@ -0,0 +1,132 @@
...
+  const [preventDefault, setPreventDefault] = React.useState(true);
+  const onTap = ({ event }) => {
+    if (preventDefault) {
+      event.preventDefault();
+    }
+  };
+
+  return (
+    <Box>
+      <Box display="flex" alignItems="center" padding={1} marginStart={-2} marginEnd={-2}>
```
christianvuerings, line 78: @jackhsu978 seems like a good use case for `Row`

```jsx
<Row gap={2} padding={1}>
  <Label htmlFor="preventDefault">
    <Text>Prevent default on tap</Text>
  </Label>
  <Switch
    id="preventDefault"
    onChange={() => setPreventDefault(!preventDefault)}
    switched={preventDefault}
  />
</Row>
```

ashleyseto:
@ajmoo will be reviewing this from PDS side and all other tap related PRs

ajmoo:
For context… is the purpose of this component to support mWeb? Would love to understand a bit more around why we have both Link and TapLink components. Why not add this functionality to the Link component? Forgive me if these are dumb questions.

jackhsu978:
@christianvuerings @bishwei do you think it makes sense to add the features to Link rather than having a new component?

docs/src/TapLink.doc.js
jackhsu978: Sure, updated. See screenshot below:

Screenshot 2020-06-19 13 47 48

packages/gestalt/src/TapLink.js
jackhsu978: sure

docs/src/TapLink.doc.js
jackhsu978: good catch! done!

jackhsu978:
@ajmoo this is a great question! I talked to @christianvuerings offline and we think that we should just merge these to `<Link>` component rather than creating a new component. We just need to make sure that the `<Link>` today works the same way.

PS: Please feel free to ask any questions here. It can also help others who have the same questions. You can always DM me if you feel more comfortable doing that.

jackhsu978:
@ajmoo @ashleyseto @christianvuerings Thank you for all the great feedback. I've changed the PR to add additional feature to `<Link>` rather than creating a new component. Please take another look when you get a chance ([test link](https://deploy-preview-933--gestalt.netlify.app/#/Link)). Thank you!

docs/src/Link.doc.js
```diff
@@ -111,4 +138,68 @@ card(
...
+  const onClick = ({ event }) => {
+    if (preventDefault) {
+      event.preventDefault();
+    }
+  };
+
+  return (
+    <Box>
+      <Row gap={2} padding={1}>
+        <Box paddingX={2}>
```
christianvuerings, line 157: @jackhsu978 now that we use `<Row gap={2} />`, we no longer need this `<Box paddingX={2}>`.

I would change it to the following:
```
      <Row gap={2} padding={2}>
        <Label htmlFor="preventDefault">
          <Text>Prevent default on tap</Text>
        </Label>
        <Switch
          id="preventDefault"
          onChange={() => setPreventDefault(!preventDefault)}
          switched={preventDefault}
        />
      </Row>
      <Divider />
```

To fix alignment + make it obvious that the toggle changes the examples below.

jackhsu978: done. thank you for suggestions!

PR #934, opened 2020-06-19
Docs: provide more layouts for <Combination />

christianvuerings:
Before we had to use `fullWidth` for `DatePicker` which resulted in each date picker showing below the other one.

# Before
Screen Shot 2020-06-19 at 8 48 02 AM

# After
Screen Shot 2020-06-19 at 8 48 10 AM


christianvuerings:
@AlbertCarreras could you take a look at this?

docs/src/components/Markdown.css
```diff
@@ -7,3 +7,7 @@
   margin-bottom: 8px;
   margin-top: 8px;
 }
+
+.Markdown pre {
```
AlbertCarreras, line 11: what's this?

christianvuerings: @AlbertCarreras on smaller screens, we were seeing a horizontal scrollbar on the whole page because of the code we added to showcase the gestalt datepicker.

```
![image](https://user-images.githubusercontent.com/127199/85160149-4893f400-b213-11ea-8835-cab9195d7853.png)
```

PR #937, opened 2020-06-22
Taparea: Add event.PreventDefault() to onMouseDown event

simkessy:
Was running into issues where onBlur event is being called before the onClick event is registered. The second answer from [this ](https://stackoverflow.com/questions/17769005/onclick-and-onblur-ordering-issue)stackoverflow question explains that adding the `event.preventDefault()` to the onMouseDown event changes the order in which these events are being called in. 

## TODO

- [x] Accessibility checkup
- [x] Update documentation
- [x] Add/update Tests


simkessy:
Bad commits included. Will make a new PR

simkessy:
Bad commits included. Will make a new PR

PR #940, opened 2020-06-22
Internal: Flow: Enable types-first

christianvuerings:
https://medium.com/flow-type/types-first-a-scalable-new-architecture-for-flow-3d8c7ba1d4eb

> The types-first architecture unlocks Flow’s potential at scale by leveraging fully typed module boundaries. We plan to migrate to the new architecture over the next months.

> We are introducing a new architecture known as “types-first”, which allows us to avoid checking dependencies. It is critical to making Flow faster and allowing us to continue to scale. The benefits extend to other problematic areas, such as error messages and debuggability.

image

### DIFF
--- a/packages/gestalt/src/Badge.js
+++ b/packages/gestalt/src/Badge.js
@@ -10,7 +10,7 @@ type Props = {|
   text: string,
 |};
 
-export default function Badge(props: Props) {
+export default function Badge(props: Props): React.Node {
   const { position = 'middle', text } = props;
 
   const cs = cx(styles.Badge, styles[position], colors.blueBg);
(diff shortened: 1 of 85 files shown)


packages/gestalt/src/Badge.js
```diff
@@ -10,7 +10,7 @@ type Props = {|
   text: string,
 |};
 
-export default function Badge(props: Props) {
+export default function Badge(props: Props): React.Element<'span'> {
```
btraut, line 13: Does it make sense to be this specific? I typically type all component returns as `React.Node` since it shouldn't matter externally what it is.

christianvuerings: @btraut those were automatically added since flow inferred the type as `React.Element<'span'>`. Still changed them all to `React.Node` since I agree externally we should not expose that information.

PR #955, opened 2020-06-27
Tabs (and Box, Link, Text): force long tab label in one line with major refactor

jackhsu978:
### Tabs
- No wrapping on long tab text (per Annie's proposal). This could significantly increase the width of the Tabs component.
- The tabs always have pills rounding, rather than weird shapes.

### Box
- Add optional `userSelect`, which sets [`user-select`](https://developer.mozilla.org/en-US/docs/Web/CSS/user-select) CSS property and is "auto" by default. It can be set to "none" to make the containing text in the subtree not selectable.

### Link
- Add optional `id` prop
- Add optional `onBlur` and `onFocus` prop

### Text
- Allow the optional `overview` prop to be "no-wrap", which sets `white-space: nowrap;` (see [`white-space`](https://developer.mozilla.org/en-US/docs/Web/CSS/white-space)) without allowing truncation.

## Revision
- Prefer using Gestalt components over raw HTML elements to implement the component.
- Fixed a bug where expected height for size=md is 40px but the actual height is 42px.
- Updated Tabs documentation and tests

## Before and After comparison

The snapshots below are wrapped inside of:
```
<Box maxWidth={500} overflow="auto" borderSize="sm" padding={1}>
  ...
</Box>
```

Before:
Screenshot 2020-06-30 13 57 45
note that the md size height is supposed to be 40px, but 42px is rendered instead.

After:
Screenshot 2020-06-30 18 10 01

## TODO

- [x] Accessibility checkup
- [x] Update documentation
- [x] Add/update Tests


docs/src/Link.doc.js
```diff
@@ -138,6 +148,51 @@ card(
...
+        <Box
+          color={index === activeIndex ? "darkGray" : undefined}
+          display="inlineBlock"
+          key={text}
+          rounding="pill"
+        >
+          <Link
+            accessibilitySelected={index === activeIndex}
+            hoverStyle="none"
+            href="#"
```
christianvuerings, line 173: @jackhsu978 could we change this to an actual link? It allows for opening a link in a new tab with Right click, open in new tab (a11y)

docs/src/Tabs.doc.js
```diff
@@ -31,7 +31,7 @@ card(
       },
       {
         name: 'onChange',
-        type: `({ event: SyntheticMouseEvent<>, activeTabIndex: number }) => void`,
+        type: `({ +event: SyntheticMouseEvent<> | SyntheticKeyboardEvent<>, +activeTabIndex: number }) => void`,
```
christianvuerings, line 34: @jackhsu978 love the read only change, we should make that the default for other components as well. (out of scope for this PR)

packages/gestalt/src/Tabs.js
```diff
@@ -25,54 +52,68 @@ export default function Tabs({
...
-            href={href}
-            {...(id ? { id } : {})}
-            key={`${i}${href}`}
-            onClick={(e: SyntheticMouseEvent<>) => handleTabClick(i, e)}
-            role="tab"
+          <Box
+            key={id || `${href}_${index}`}
+            id={id}
+            dangerouslySetInlineStyle={{
+              __style: { padding: '0px 2px' },
```
christianvuerings, line 84: @jackhsu978 let's not add the padding here, that should be added by the container which uses `Tabs`. Main reason is that <Tabs /> no longer looks visually in sync with other components.

It also makes the assumption that the outline will always stay 2px, which is not the case.

```jsx
<Row padding={2}>
  <Tabs />
</Row>
```

image

packages/gestalt/src/Tabs.js
```diff
@@ -25,54 +52,68 @@ export default function Tabs({
...
-        styles.Tabs,
-        wrap && layout.flexWrap,
-        size === 'md' ? layout.medium : layout.large
-      )}
+    <Box
+      alignItems="center"
+      display="flex"
+      marginEnd={-1}
+      marginStart={-1}
+      minHeight={size === 'lg' ? 48 : 40}
```
christianvuerings, line 70: @jackhsu978 something odd happened here where the minHeight does change but the size of the Tab itself does not.
image

packages/gestalt/src/Tabs.js
```diff
@@ -25,54 +52,68 @@ export default function Tabs({
...
       role="tablist"
+      wrap={wrap}
+      dangerouslySetInlineStyle={{
+        __style: { userSelect: 'none' },
```
christianvuerings, line 74: @jackhsu978 although it looks the same as the original CSS, it is not. In our CSS files we autoprefix our CSS, with inline styles that's not the case.

Looking at https://caniuse.com/#feat=user-select-none we should probably add the `-ms` and `-webkit` prefix versions here as well.

image

packages/gestalt/src/Tabs.js
```diff
@@ -25,54 +52,68 @@ export default function Tabs({
...
-            aria-selected={isActive}
-            className={cs}
-            href={href}
-            {...(id ? { id } : {})}
-            key={`${i}${href}`}
-            onClick={(e: SyntheticMouseEvent<>) => handleTabClick(i, e)}
-            role="tab"
+          <Box
+            key={id || `${href}_${index}`}
+            id={id}
```
christianvuerings, line 82: @jackhsu978 what was the reason behind this change? Why did we not leave the `id` on the `anchor` tag instead?

packages/gestalt/src/Tabs.js
```diff
@@ -25,54 +52,68 @@ export default function Tabs({
...
-            <Text color={isActive ? 'white' : 'darkGray'} weight="bold">
-              {text}
-            </Text>
-          </a>
+            <Link
+              accessibilitySelected={isActive}
+              hoverStyle="none"
+              href={href}
+              onClick={({ event }) => {
+                event.preventDefault();
```
christianvuerings, line 92: @jackhsu978 could we avoid the `preventDefault` here? Ideally it's something the consumer of `<Tabs />` does.

Main use case I'm thinking of is `CMD+Click` to open the `<Tab />` in a new browser tab. If the consumer does an extra check for CMD+Click and only does `preventDefault` then, we should be good.

packages/gestalt/src/Tabs.js
```diff
@@ -25,54 +52,68 @@ export default function Tabs({
...
       })}
-    </div>
+    </Box>
   );
 }
 
 Tabs.propTypes = {
   activeTabIndex: PropTypes.number.isRequired,
+  onChange: PropTypes.func.isRequired,
+  size: PropTypes.oneOf(['md', 'lg']),
```
christianvuerings, line 110: @jackhsu978 nice spotting, thanks for adding `size` to our propTypes.

christianvuerings:
@ashleyseto could you take a look from the design side?

Main open question:
* Is the transparent pill desired? We previously made the background always white to a11y reasons? E.g. when tabs would show on a dark gray background.

packages/gestalt/src/Tabs.js
jackhsu978: you're right! thanks for catching this!

packages/gestalt/src/Tabs.js
jackhsu978: because `Link` doesn't support `id` whereas `Box` supports it. I've added the support the prop `dangerouslySetId` to `Link` component, so now the id is associated with the anchor tag.

I'm not sure if we want to add `dangerouslySetId` to `Link` component. I am thinking maybe it's safer to make it less discoverable because it seems dangerous. Any thoughts on this?

packages/gestalt/src/Tabs.js
jackhsu978: That's really good point. I changed my mind about using inline style for this. Instead of passing this as inline style to the whole Tab component wrapper, I move it to a `<Link>` CSS class that's only enabled when the `role` prop is `"tab"`. I think this is better because (1) no need to worry about vendor prefixes, (2) there's a direct logical association between `user-select: none` and the Link whose role is a tab, and (3) it's cleaner because there's less logic needed in `<Tabs>` component.

packages/gestalt/src/Tabs.js
jackhsu978: fixed. the `size` now not only controls the `minHeight`, but also the text font size. I think it makes more sense that, when the tab component is taller, the text inside of the tab component is also bigger.

packages/gestalt/src/Tabs.js
jackhsu978: @christianvuerings There are 2 problems with the current Tab:
1. when it's inside of `<Box overflow="auto">`, the blue accessibility outline gets cut off
2. the blue accessibility overlaps with the adjacent tab.

See snapshot below:

Screenshot 2020-06-29 18 11 43

My idea here is to leave enough vertical & horizontal padding such that the blue accessibility outline does not get cut off and does not overlap with the adjacent tab. See the snapshot below.
Screenshot 2020-06-29 18 09 57

The horizontal padding is set up as follows:

- tabs root element gets 0.5 boint
- each child tab element gets 0.5 boint
```
(1bt)[ TAB 1 ](1bt)[ TAB 2 ](1bt)[ TAB 3 ](1 bt)
```

Also, the `Tab` component now uses the `<Row>` component and takes the `gap` prop. If additional gap is needed, the `gap` prop would be handy.

```
(1bt)[ TAB 1 ](1bt + 2 * gap)[ TAB 2 ](1bt + 2 * gap)[ Tab 3 ](1bt)
```

What do you think?

docs/src/Tabs.doc.js
jackhsu978: yeah, good idea!

docs/src/Link.doc.js
jackhsu978: will do

jackhsu978:
> @ashleyseto could you take a look from the design side?
> 
> Main open question:
> 
> * Is the transparent pill desired? We previously made the background always white to a11y reasons? E.g. when tabs would show on a dark gray background.

Didn't know that it was as designed for a11y. I changed it back.

packages/gestalt/src/Tabs.js
jackhsu978: I changed mind. Now by default, there's no padding at all.

If the `Box` wrapping the Tab component has `overflow` prop equal to auto/scroll/hidden, the blue accessibility box shadow gets cropped, but one can fix that by adding `padding={1}` to the `Box` wrapper.

To fix blue accessibility box shadow overlapping the adjacent tab, one can fix that by adding `gap={1}` to the `Tab` component, which works similar to how `Row` works.

Given that there are fixes to both issues, I am happy with this simple solution.

docs/src/Link.doc.js
jackhsu978: updated

jackhsu978:
@christianvuerings Thank you for your feedback! I've addressed your comments. Could you take another look when you get a chance? Thanks again!

packages/gestalt/src/Link.css
```diff
@@ -16,3 +16,30 @@
...
+  position: relative;
+}
+
+.tabIsNotActive {
+  composes: whiteBg from "./Colors.css";
+}
+
+.tabIsNotActive:focus,
+.tabIsNotActive:hover {
+  background-color: #efefef;
```
jackhsu978, line 35: It was `#f1f1f1`, but it's not one of our standard colors. I'm changing it to the closest standard color.

packages/gestalt/src/Tabs.js
```diff
@@ -1,22 +1,17 @@
...
-export default function Tabs({
-  activeTabIndex,
-  onChange,
-  size = 'md',
-  tabs,
-  wrap,
-}: {|
+type Props = {|
   activeTabIndex: number,
+  gap?: 0 | 1,
```
jackhsu978, line 11: we can add more later when we discover more use cases needed.

docs/src/Tabs.doc.js
```diff
@@ -31,11 +31,18 @@ card(
...
         required: true,
         description:
           'If your app uses a tool such as react-router to navigate between pages, be sure to use onChange to navigate instead of getting a full page refresh with href',
       },
+      {
+        name: 'gap',
```
christianvuerings, line 40: @jackhsu978 just checked in with Annie Teng & @ashleyseto and both of them mentioned they would rather not have the extra gap in between.

Could you also follow up with them on the font size change?

docs/src/Tabs.doc.js
```diff
@@ -55,33 +62,53 @@ card(
...
+        <Switch
+          id="wrap"
+          onChange={() => setWrap(!wrap)}
+          switched={wrap}
+        />
+      </Row>
+      {sizes.map(([size, gap]) => (
+        <Box key={size + String(gap)}>
+          <Heading size="sm">size={size}, gap={gap}</Heading>
+          <Box maxWidth={500} overflow="auto" borderSize="sm" padding={1}>
```
christianvuerings, line 99: @jackhsu978 I do like this solution to no longer have the focus state overlap, add a padding of `1` (4px) to the container.

jackhsu978: @christianvuerings My previous iteration did prevent the focus state overlap (adding 0.5 boint padding, so that the gap between tabs is 1 boint, exactly the same as the width of the blue a11y box shadow), but I took it out because you commented that we shouldn't be adding the padding to each tab component. Do you want me to add that back? It'd be effectively adding `gap` 0.5 boint to the Row component (except that I have to do dangerouslySetInlineStyle since the `gap` prop must be an integer)

@christianvuerings @ashleyseto which one of the following 3 is preferred?
Screenshot 2020-06-30 11 13 39

jackhsu978: ok, based on Annie's feedback, we want the gap=0 variant, despite the blue accessibility outline overlapping the adjacent tab.

packages/gestalt/src/Tabs.js
jackhsu978: removed, per designer feeedback

docs/src/Link.doc.js
```diff
@@ -26,6 +26,11 @@ card(
         name: 'children',
         type: 'React.Node',
       },
+      {
+        name: 'dangerouslySetId',
```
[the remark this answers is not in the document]
jackhsu978: done

ajmoo:
@jackhsu978 There are other implications of adjusting text size with the smaller tabs. Our `button` component, for instance, has the same size text for sm, md, and lg size buttons. I like the text size change, just want to make sure we're being consistent across all components.

jackhsu978:
> @jackhsu978 There are other implications of adjusting text size with the smaller tabs. Our `button` component, for instance, has the same size text for sm, md, and lg size buttons. I like the text size change, just want to make sure we're being consistent across all components.

Got it. I changed the logic back. Thanks for the feedback.

docs/src/Tabs.doc.js
jackhsu978: Yeah, according to @ajmoo Button's size also doesn't influence the text size within. I think it's good to be consistent, so I changed the logic back.

ajmoo:
Thanks for the update, @jackhsu978 !! This looks good to me. I guess I don't quite understand why we would allow folks to wrap tabs to the next line, this isn't really something seen in product. @ashleyseto do you foresee a need for this on biz/ads?

jackhsu978:
> Thanks for the update, @jackhsu978 !! This looks good to me. I guess I don't quite understand why we would allow folks to wrap tabs to the next line, this isn't really something seen in product. @ashleyseto do you foresee a need for this on biz/ads?

@ajmoo The optional `wrap` prop was added in #361 with the following explanation:
> "The intention of this PR is to allow wrapping when there isn't enough width (too many tabs for window width). With this in mind, yeah, it'll result in tabs creating new lines as needed which is why making wrapping default might break current usages if it's not expecting extra height. I think wrapping is nicer than having things just cut off though"

I did a quick search in our codebase and found that we aren't using this optional prop anywhere.

PR #956, opened 2020-06-29
Internal: Support GitHub Codespaces

christianvuerings:
Add support for [GitHub Codespaces](https://github.com/features/codespaces/) - a new beta feature from GitHub to allow
* Editing of files in VS Code in a browser window
* Run Gestalt docs on a remote browser

## Update docs remotely
github-codespaces-gestalt

## Flow support
Screen Shot 2020-06-29 at 7 52 18 AM

## Creation log
Screen Shot 2020-06-29 at 7 46 10 AM

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,11 @@
+{
+  "css.lint.validProperties": ["composes"],
+  "editor.codeActionsOnSave": { "source.fixAll.eslint": true },
+  "editor.formatOnSave": true,
+  "eslint.packageManager": "yarn",
+  "grunt.autoDetect": "off",
+  "javascript.validate.enable": false,
+  "npm.packageManager": "yarn",
+  "prettier.packageManager": "yarn",
+  "typescript.validate.enable": false
+}
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,25 @@
+{
+  "name": "Gestalt: Node.js 12",
+  // https://github.com/microsoft/vscode-dev-containers/tree/master/containers/
+  "image": "mcr.microsoft.com/vscode/devcontainers/javascript-node:12",
+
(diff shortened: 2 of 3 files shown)


.vscode/settings.json
```diff
@@ -0,0 +1,11 @@
+{
+  "css.lint.validProperties": ["composes"],
+  "editor.codeActionsOnSave": { "source.fixAll.eslint": true },
+  "editor.formatOnSave": true,
```
chrislloyd, line 4: I wonder if we should generally be setting "formatOnSave" or whether it should be a user setting.

chrislloyd:
I had a timeout the first time opening this - though I think that might have been a Codespace issue. It definitely takes forever `yarn install`ing on every codespace creation - would it be better creating a Dockerfile that contains the install and doing it there so the layers can be cached?

chrislloyd:
I'm also getting a timeout connecting to the forwarded port. I think it might be better to run `yarn dev` in the `postCreateCommand` and move the yarn install into the container build step.

.vscode/settings.json
christianvuerings: @chrislloyd was wondering the same thing - for now I kept it the same as what we use in Pinterest's codebase

christianvuerings:
@chrislloyd interesting - what took a long time for me was the building of the image. The `yarn install` was pretty quick.
Would you be able to follow up on this PR with the docker specific changes?

For now I used the recommended way of using `yarn install`: 
image


I'm not sure about running `yarn dev` in the `postCreateCommand`. Doesn't that event only fire after you create the Codespace? From what I'm reading, it's intended for Codespaces to live for a longer period of time. Should we set `postStartCommand` to `yarn dev` instead? That way, every time the container starts, the development environment will run.

https://code.visualstudio.com/docs/remote/containers#_devcontainerjson-reference

PR #957, opened 2020-06-30
Docs: support IE11

christianvuerings:
Conditionally load a script with polyfills in IE 11 so the docs render correctly.

## Before: 'Symbol' is undefined
Screen Shot 2020-06-30 at 7 11 34 AM

## After: Documentation renders in IE11
Screen Shot 2020-06-30 at 7 11 18 AM


christianvuerings:
Confirmed the docs now work on IE11: https://deploy-preview-957--gestalt.netlify.app/
Screen Shot 2020-06-30 at 7 20 11 AM

PR #962, opened 2020-07-01
Docs: Expand Navigation subsections

mminute:
Update the docs navigation to expand the sections/subsections by default.

Potential improvements:
1. Collapse all for components
2. Make the navigation sticky so that it stays in view when scrolling through longer component docs

Screen Shot 2020-07-01 at 12 44 24 PM

- [ ] Accessibility checkup
- [x] Update documentation
- [ ] Add/update Tests

### DIFF
--- a/docs/src/components/sidebarIndex.js
+++ b/docs/src/components/sidebarIndex.js
@@ -7,6 +7,44 @@ export type sidebarIndexType = Array<{|
   subsections?: Array<{| [string]: Array<string> |}>,
 |}>;
 
+const componentSubSectionPages = {
+  displayDate: ['Avatar', 'AvatarPair', 'Badge', 'GroupAvatar', 'Table'],
+  feedBack: ['Modal', 'Pulsar', 'Spinner', 'Toast'],
+  foundations: ['Heading', 'Icon', 'Text'],
+  forms: [
+    'Button',
+    'Checkbox',
+    'DatePicker',
+    'Flyout',
+    'IconButton',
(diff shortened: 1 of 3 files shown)


docs/src/components/sidebarIndex.js
```diff
@@ -30,59 +68,37 @@ const sidebarIndex: sidebarIndexType = [
...
         subsectionName: 'Media',
-        pages: ['Image', 'Letterbox', 'Mask', 'Video'],
+        pages: componentSubSectionPages.media,
       },
       {
         subsectionName: 'Navigation',
-        pages: ['Link', 'SegmentedControl', 'Tabs'],
+        pages: componentSubSectionPages.navigation,
+      },
+      {
```
christianvuerings, line 97: @mminute I like the fact that we now see all the components by default but it's odd to see them twice. For now, I'm going to remove the `All` section - we can always add it back in later.

image

christianvuerings: This also breaks when you go `All => Avatar` & then refresh the page. Now `Data Display => Avatar` is selected.

mminute: Ah, i see.  I didn't test the refresh case.

I get what you are saying about seeing the components twice, but as someone who is familiar with gestalt I found it slower to identify the component i wanted to see in the divided list rather than an alphabetical list.  The `All` option was meant to be a bailout for those who, like me, might still want a list

PR #963, opened 2020-07-01
Styling: Prepare css for dark mode support

jennyscript:
This sets up our css generation so it's ready to support dark mode in follow up PRs
### Building two css files
Because css variables do not work on older browsers such as IE11, I'm generating a new css file with the css vars preserved and kept the old one with the existing behavior to substitute in the values during builds.
#### Why not just one file?
It's also possible to support both legacy and modern browsers in one css file by including a fallback value and the variable in each rule, but that ended up roughly doubling the css filesize.
#### Planned changes after shipping
Once this has had more testing and we consider dark mode stable, I'd like to make the "future" css the default one and rename the existing one "legacy". Anyone using gestalt who needs to support IE or doesn't need theme support can load that version of the css.
### Renamed css vars
The design team has already worked with the mobile clients to use a standard set of tokens for the colors. This updates our css variables to use the same names for consistency. We have more secondary colors on web than are currently specified, so I'll leave that update for a future PR.

Token name | Light mode value | Dark mode value
-- | -- | --
colorGray0 | #ffffff | #050505
colorGray50 | #ffffff | #272727
colorGray100 | #efefef | #494949
colorGray200 | #8e8e8e | #b8b8b8
colorGray300 | #111111 | #efefef
colorGray400 | #000000 | #ffffff
colorRed0 | #ff5247 | #e60023
colorRed100 | #e60023 | #ff5247
colorOrange0 | #ef955d | #cc4e00
colorOrange100 | #cc4e00 | #ef955d
colorGreen0 | #86da94 | #008a1c
colorGreen100 | #008a1c | #86da94
colorBlue0 | #66ccff | #0074e8
colorBlue100 | #0074e8 | #66ccff
colorPurple0 | #bf77e5 | #6c1ea6
colorPurple100 | #6c1ea6 | #bf77e5

### DIFF
--- a/docs/src/index.js
+++ b/docs/src/index.js
@@ -2,7 +2,7 @@
 import React from 'react';
 import { HashRouter, Route, Switch, Redirect } from 'react-router-dom';
 import { render } from 'react-dom';
-import 'gestalt/dist/gestalt.css';
+import 'gestalt/dist/gestalt-future.css';
 import 'gestalt-datepicker/dist/gestalt-datepicker.css';
 import App from './components/App.js';
 import CardPage from './components/CardPage.js';
--- a/packages/gestalt-core/build.js
+++ b/packages/gestalt-core/build.js
@@ -65,6 +65,7 @@ const statsPlugin = () => {
 const cssModules = (options = {}) => {
   const cssExportMap = {};
   const scopeNames = {};
+  let cssIE11 = ''; // CSS targeted toward IE11
   let css = '';
 
   const breakpoints = {
@@ -73,7 +74,31 @@ const cssModules = (options = {}) => {
     lg: '(min-width: 1312px)',
   };
 
-  const plugins = [
+  const modulesPlugin = postcssModules({
+    scopeBehavior: 'local',
+    generateScopedName: (name, filename) => {
+      const dir = relative(__dirname, filename);
+      const hash = `${dir}:${name}`;
+
+      if (!Object.prototype.hasOwnProperty.call(scopeNames, hash)) {
+        scopeNames[hash] = classnameBuilder.getMinifiedClassname(hash);
+      }
+
+      return scopeNames[hash];
+    },
+    getJSON: (filePath, exportTokens) => {
+      Object.entries(exportTokens).forEach(([className, value]) => {
+        if (value.includes('undefined')) {
+          throw new Error(
+            `${filePath} / .${className} composes from an incorrect classname`
+          );
+        }
+      });
+      cssExportMap[filePath] = exportTokens;
+    },
+  });
+
+  const pluginsIE11 = [
     postcssCssnext({
       features: {
         customMedia: {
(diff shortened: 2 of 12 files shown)


docs/src/index.js
```diff
@@ -2,7 +2,7 @@
 import React from 'react';
 import { HashRouter, Route, Switch } from 'react-router-dom';
 import { render } from 'react-dom';
-import 'gestalt/dist/gestalt.css';
+import 'gestalt/dist/gestalt-future.css';
```
jennyscript, line 5: I'd like to start dogfooding the new css in the docs. This means colors won't look right for IE11 in the docs. I *could* fix this by generating the big css file that includes both vars and fallbacks or can we say IE11 isn't supported in the docs?

packages/gestalt-core/build.js
```diff
@@ -65,6 +65,7 @@ const statsPlugin = () => {
 const cssModules = (options = {}) => {
   const cssExportMap = {};
   const scopeNames = {};
+  let cssIE11 = ''; // CSS targeted toward IE11
```
jennyscript, line 68: I thought about calling this "legacy" but I was only planning on testing in IE11 since that's the main browser we're generating this for. There could be some older Android devices we need to consider on mobile as well https://caniuse.com/#feat=css-variables

ponori:
@ajmoo Could you confirm that the colors - specifically the ones outside of the gray palette are accurate and up to date? I know we've done some additional color work recently and I just want to make sure this palette reflects the latest and greatest.

docs/src/index.js
christianvuerings: @jennyscript would there be a way to make this import conditional? We recently added support to test the docs in IE11 #957 which is useful to test IE specific docs.

An alternative is to use something like: https://www.npmjs.com/package/css-vars-ponyfill which we could also include conditionally (only for IE11)

christianvuerings:
> @ajmoo Could you confirm that the colors - specifically the ones outside of the gray palette are accurate and up to date? I know we've done some additional color work recently and I just want to make sure this palette reflects the latest and greatest.

@jennyscript could you follow up with @ajmoo & Annie Teng and verify that those are correct?

packages/gestalt/src/Colors.css
```diff
@@ -1,10 +1,13 @@
...
   /* primary colors */
-  --red: #e60023;
-  --white: #fff;
-  --lightGray: #efefef;
-  --gray: #767676;
-  --darkGray: #111;
+  --colorRed0: #ff5247;
+  --colorRed100: #e60023;
+  --colorGray0: #fff;
+  --colorGray50: #fff;
```
christianvuerings, line 6: @jennyscript this is so weird - should `--colorGray0` and `--colorGray50` have the exact same value? Can we just keep in `--colorGray0` for now?

ponori: The rationale for colorGray50 is that the dark mode needs an extra gray to address elevation in an accessible way. Happy to go through the nuances of this if you're interested.

christianvuerings: @ponori understood - it was just odd to see it here without the dark mode overrides

jennyscript: @christianvuerings Yes, it's there for the elevated value like PJ mentioned. You can see where I started adding this to Modal and Contents if that helps explain

docs/src/index.js
jennyscript: I added a polyfill for the docs (ended up using a different one for easier set up). The docs only _kind of_ work in IE11 right now, but as far as I could tell colors are showing up correctly in the docs.

packages/gestalt/src/Colors.css
ponori: @christianvuerings Yeah, totally understand. Dark mode just understandably adds some moving pieces to our equation. I wonder if there's ample rationale to slightly darken our elevated surfaces in light mode to have a consistent palette model, but I think that's beyond the scope of this effort. That said, with tokens, it should enable us to make that change more seamlessly in the future

PR #965, opened 2020-07-02
Form Actions

chrislloyd:
## TODO

- [ ] Accessibility checkup
- [ ] Update documentation
- [ ] Add/update Tests

### DIFF
--- a/packages/gestalt/src/FormActions.js
+++ b/packages/gestalt/src/FormActions.js
@@ -1,0 +1,17 @@
+// @flow strict
+import * as React from 'react';
+import Box from './Box.js';
+import Row from './Row.js';
+import Button from './Button.js';
+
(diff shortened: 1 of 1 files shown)


packages/gestalt/src/FormActions.js
```diff
@@ -0,0 +1,17 @@
...
+import * as React from 'react';
+import Box from './Box.js';
+import Row from './Row.js';
+import Button from './Button.js';
+
+export default function FormActions(): React.Node {
+  return (
+    <Box padding={2}>
+      <Row>
+        <Button text="Cancel" />
```
christianvuerings, line 11: @chrislloyd Seems like this was a test to add a layout component - we should work with design to get these added. We should also not add literal strings.

Going to close this PR for now but feel free to re-open.

PR #966, opened 2020-07-02
ZIndex: Add support for Fixed & Composite zIndexes

chrislloyd:
The `z-index` CSS property is difficult to work with as:

1. It's global. There's one index for the entire page, which goes against the composability of React's component model.
2. It's implicit. There's usually a relationship between `100` & `101` but at best that's only captured in comments and can't be verified. If `100` changes, `101` will break.

Previously, Gestalt has taken the ["War Games" approach](https://www.youtube.com/watch?v=MpmGXeAtWUw) to `z-index` - the only winning move is to not play the game. This resulted in components like `<Layer>` which help developers change the render order while still respecting the React component order. However, over time, `z-index` usage hasn't decreased and it remains one of the most commonly usages of `dangerouslySetInlineStyles` on `<Box>`. We continue to get similar kinds of incidents relating to `z-index`. Clearly this wasn't the winning move.

This PR attempts to help fix the problem by having native support for `z-index` in Gestalt, however with some principled restrictions. It attempts to solve the two issues with `z-index` (global and implicit) while also providing an escape hatch for engineers to migrate existing "dangerous" code.

It does this by providing 2 classes that "generate" a `z-index` for `<Box>` (& `<Sticky>`) - `FixedZIndex` & `CompositeZIndex`.

FixedZIndex is intended to capture existing places where we may use `z-index` today. A good example is in our `MobileModal`:

```jsx
const MODAL_Z_INDEX = 9999;
<Box dangerouslySetInlineStyle={{ __style: { zIndex: MODAL_Z_INDEX }}} />
```

This could be rewritten as:

```jsx
const MODAL_Z_INDEX = new FixedZIndex(9999);
<Box zIndex={MODAL_Z_INDEX} />
```

The goal is to capture current "fixed" indexes in a way that that's easy to search for and analyze.

The next class, `CompositeZIndex` is meant to be used in situations like in `AttributionWindowFilter.js`:

```jsx
const HEADER_ZINDEX = 600;
<Box position="absolute" dangerouslySetInlineStyle={{ __style: { zIndex: HEADER_ZINDEX + 1 } }}>
```

This could be rewritten as:

```jsx
const HEADER_ZINDEX = new FixedZIndex(600);
<Box position="absolute" zIndex={new CompositeZIndex([HEADER_ZINDEX])} />
```

While the initial pulling out of the `HEADER_ZINDEX` value was good, we have no way of knowing how that value is being used and what needs to go above it. `CompositeZIndex` helps us form a tree of `z-index` dependencies that can be analyzed with Flow.

Another common case is where the relationship is expressed in the opposite way, i.e.

```jsx
 // needs to go below whatever is at z-index 100
<Box dangerouslySetInlineStyle={{ __style: { zIndex: 100 - 1 }}} />
```

I would re-write this to make the dependency more of a purely upwards counting graph so that the artimatic here isn't complicated:

```js
const a = new FixedZIndex(99);
<Box zIndex={new CompositeZIndex([a]) />
```

A good initial outcome of this diff would be if we started transitioning existing fixed `z-index`'s to use `FixedZIndex`. This should help us explicitly capture dependencies. The next step would be to convert these `FixedZIndex`s to `CompositeZIndex`s all with one root index (`0`). Once we've got there, it should be straight forward to look at the resulting structure and see where we can carve out `<Layer>` usage.

## TODO

- [x] ~Accessibility checkup~
- [x] Update documentation
- [x] Add/update Tests
- [x] ~Pinterest Designer (for design changes)~

### DIFF
--- a/docs/src/Box.doc.js
+++ b/docs/src/Box.doc.js
@@ -204,6 +204,11 @@ card(
         defaultValue: false,
         description: `By default, flex items will all try to fit onto one line. You can change that and allow the items to wrap onto multiple lines, from top to bottom.`,
       },
+      {
+        name: 'zIndex',
+        type: 'interface { index(): number; }',
+        description: `An object representing the zIndex value of the Box.`,
+      },
     ]}
   />
 );
@@ -603,6 +608,23 @@ card(
       {props => <Box color="darkGray" width={60} height={60} {...props} />}
     </Combination>
   </Card>
+);
+
+card(
+  <Example
+    description={`
+It's possible to use box with external elements using the css \`z-index\` property by capturing those values in controlled objects. The example below shows using a \`FixedZIndex\` for a value that comes from somewhere else, and a \`CompositeZIndex\` to layer the Box on top of it.
+  `}
+    id="zindex"
+    name="ZIndex"
+    defaultCode={`
+function Example() {
+  const HEADER_ZINDEX = new FixedZIndex(100);
+  const zIndex = new CompositeZIndex([HEADER_ZINDEX]);
+  return <Box color="blue" width={60} height={60} zIndex={zIndex} />
+}
+`}
+  />
 );
 
 export default cards;
--- a/docs/src/Box.doc.js
+++ b/docs/src/Box.doc.js
@@ -206,7 +206,7 @@ card(
       },
       {
         name: 'zIndex',
-        type: 'interface { index(): number; }',
+        type: 'interface Indexable { index(): number; }',
         description: `An object representing the zIndex value of the Box.`,
       },
     ]}
--- a/docs/src/Sticky.doc.js
+++ b/docs/src/Sticky.doc.js
@@ -42,9 +42,9 @@ card(
         description: `Use numbers for pixels: top={100} and strings for percentages: top="100%"`,
       },
       {
-        name: 'dangerouslySetZIndex',
-        type: '{ __zIndex: number }',
-        defaultValue: '{ __zIndex: 1 }',
+        name: 'zIndex',
+        type: 'interface { index(): number; }',
+        description: `An object representing the zIndex value of the Sticky.`,
       },
     ]}
   />
--- a/docs/src/Sticky.doc.js
+++ b/docs/src/Sticky.doc.js
@@ -43,7 +43,7 @@ card(
       },
       {
         name: 'zIndex',
-        type: 'interface { index(): number; }',
+        type: 'interface Indexable { index(): number; }',
         description: `An object representing the zIndex value of the Sticky.`,
       },
     ]}
--- a/packages/gestalt/src/Sticky.js
+++ b/packages/gestalt/src/Sticky.js
@@ -3,6 +3,7 @@
 import * as React from 'react';
 import PropTypes from 'prop-types';
 import layout from './Layout.css';
+import { type Indexable, FixedZIndex } from './zIndex.js';
 
 type Threshold =
   | {| top: number | string |}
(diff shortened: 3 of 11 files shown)


packages/gestalt/src/index.js
```diff
@@ -59,8 +60,10 @@ export {
   Collage,
   Collection,
   Column,
+  CompositeZIndex,
```
christianvuerings, line 63: @chrislloyd could you update the docs & provide an example of how people should use `CompositeZIndex` and `FixedZIndex`?

docs/src/Sticky.doc.js
```diff
@@ -42,9 +42,9 @@ card(
         description: `Use numbers for pixels: top={100} and strings for percentages: top="100%"`,
       },
       {
-        name: 'dangerouslySetZIndex',
-        type: '{ __zIndex: number }',
-        defaultValue: '{ __zIndex: 1 }',
+        name: 'zIndex',
+        type: 'interface { index(): number; }',
```
jennyscript, line 46: Can we write this type another way like `FixedZIndex | CompositeZIndex` or `Indexable` to make it more obvious how this should be used when scanning the docs?

packages/gestalt/src/Sticky.js
```diff
@@ -20,19 +21,28 @@ type Threshold =
...
 |};
 
+const DEFAULT_ZINDEX = new FixedZIndex(1);
+
 export default function Sticky(props: Props): React.Node {
-  const { dangerouslySetZIndex = { __zIndex: 1 }, children } = props;
+  const { dangerouslySetZIndex, children } = props;
+  const zIndex =
+    props.zIndex ||
+    (dangerouslySetZIndex &&
```
jennyscript, line 35: Can we just plan to remove `dangerouslySetZIndex` totally in this release and codemod existing uses to `FixedZIndex` or is that too ambitious?

packages/gestalt/src/Sticky.js
```diff
@@ -20,19 +21,28 @@ type Threshold =
 
 type Props = {|
   children: React.Node,
+  zIndex?: Indexable,
   dangerouslySetZIndex?: {| __zIndex: number |},
   ...Threshold,
 |};
 
+const DEFAULT_ZINDEX = new FixedZIndex(1);
```
jennyscript, line 29: This will result in every sticky now having a default zIndex of 1 but before some of them may not have had a zIndex. Is this required?

packages/gestalt/src/Sticky.js
```diff
@@ -20,19 +21,28 @@ type Threshold =
...
+
 export default function Sticky(props: Props): React.Node {
-  const { dangerouslySetZIndex = { __zIndex: 1 }, children } = props;
```
chrislloyd, line 28: @jennyscript ^

packages/gestalt/src/Sticky.js
chrislloyd: I'd prefer not to couple deprecating this field and the release together. A codemod here is definitely possible but I think I'd like folks (and me) to take a moment to see if we can just fix some of the existing ones. Also - writing a codemod would require more time than I have to put into this.

docs/src/Box.doc.js
```diff
@@ -605,4 +610,21 @@ card(
   </Card>
 );
 
+card(
+  <Example
+    description={`
+It's possible to use box with external elements using the css \`z-index\` property by capturing those values in controlled objects. The example below shows using a \`FixedZIndex\` for a value that comes from somewhere else, and a \`CompositeZIndex\` to layer the Box on top of it.
```
christianvuerings, line 616: Thanks for adding the example

PR #967, opened 2020-07-06
Tabs: Add optional dot indicator

danignibus:
This PR adds the ability to render a small circular icon in a tab to indicate that the tab should be clicked per a new design we received in ads. 

notifdot1
notifdot2

## TODO

- [ ] Accessibility checkup
- [X] Update documentation
- [X] Add/update Tests
- [X] [Format PR title](https://github.com/pinterest/gestalt/#releasing): `ComponentName: Description`.

### DIFF
--- a/docs/src/Tabs.doc.js
+++ b/docs/src/Tabs.doc.js
@@ -37,9 +37,9 @@ card(
           'If your app uses a tool such as react-router to navigate between pages, be sure to use onChange to navigate instead of getting a full page refresh with href',
       },
       {
-        name: 'icon',
-        type: '"circle"',
-        description: `Icon to indicate a tab that needs attention.`,
+        name: 'indicator',
+        type: '"dot"',
+        description: `Indicate that a tab needs attention.`,
       },
       {
         name: 'size',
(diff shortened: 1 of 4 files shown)


docs/src/Tabs.doc.js
```diff
@@ -36,6 +36,11 @@ card(
         description:
           'If your app uses a tool such as react-router to navigate between pages, be sure to use onChange to navigate instead of getting a full page refresh with href',
       },
+      {
+        name: 'icon',
```
christianvuerings, line 40: @danignibus `icon` should be reserved for icons listed in https://gestalt.netlify.app/#/components/Icon

Could we rename this to `indicator: 'dot'`? /cc @ashleyseto

packages/gestalt/src/Tabs.js
```diff
@@ -11,10 +11,24 @@ type OnChangeHandler = ({|
   +activeTabIndex: number,
 |}) => void;
 
+function Circle() {
+  return (
+    <Box
+      color="red"
+      dangerouslySetInlineStyle={{ __style: { marginTop: '1px' } }}
```
christianvuerings, line 18: @danignibus could we avoid the 1px margin on the top - since it's a magic number which we want to avoid. Please check with design, if they require the `1px`, let's leave it in.

image
image

ajmoo:
This looks good to me. Sorry @christianvuerings and @danignibus I can't really speak to the `indicator: 'dot'` question above, but it sounds logical.

PR #968, opened 2020-07-06
GestaltProvider: Adding a provider for color scheme and other future context used by Gestalt

jennyscript:
This creates two new components `GestaltProvider` and `ThemeProvider` and a hook `useTheme`.

## `GestaltProvider`

This is a component that may **optionally** be set up to wrap gestalt components and provide context to them. In this PR I am just setting up support for color schemes, but we've also discussed using this provider for RTL support and to make `Link` work properly with react-router. Providers can be nested to override context set above. For now it's just a thin wrapper around `ThemeProvider`.

Open to better naming suggestions 😄 

## `ThemeProvider`

This is used internally by `GestaltProvider` to set up the theme context and generate a style tag for a custom class name. The style tag overrides the css variables we use to a new value. Options for `colorScheme` are:

- light: The standard gestalt theme
- dark: A new theme for users preferring dark mode
- userPref: Check the `prefers-color-scheme` media query to decide to use light/dark theme

Initially the color scheme will default to `light`, but once dark mode is more mature we may want to change the default to `userPref`.

## `useTheme`

This hook can be used internally to do things like change svg fills. I exported it so custom components that need to match gestalt can also use the theme.

Changing color schemes in the docs:

color-scheme

I confirmed locally that nesting `GestaltProviders` with different color schemes works correctly.

Screen Shot 2020-07-06 at 12 52 38 PM

- [x] Update documentation
- [x] Add/update Tests

### DIFF
--- a/docs/src/GestaltProvider.doc.js
+++ b/docs/src/GestaltProvider.doc.js
@@ -1,0 +1,77 @@
+// @flow strict
+import * as React from 'react';
+import Example from './components/Example.js';
+import PropTable from './components/PropTable.js';
+import PageHeader from './components/PageHeader.js';
+
+const cards = [];
+const card = c => cards.push(c);
+
+card(
+  <PageHeader
+    name="GestaltProvider"
+    description="An app may optionally have a `GestaltProvider` to set up context for components further down the tree. The first usecase is setting the color scheme, but other uses such as right to left support will be added in the future."
+  />
+);
+
+card(
+  <PropTable
+    props={[
+      {
+        name: 'colorScheme',
+        type: `"light" | "dark" | "userPref"`,
+        defaultValue: 'light',
+        description:
+          'The color scheme for components inside the provider. Specify "userPref" to use "prefers-color-scheme" media query.',
+        href: 'colorScheme',
+      },
+    ]}
+  />
+);
+
+card(
+  <Example
+    description="Specify a light or dark color scheme for components"
+    name="Color scheme"
+    id="colorScheme"
+    defaultCode={`
+function Example(props) {
+  const [scheme, setScheme] = React.useState('light')
+  const schemeOptions = [
+    {
+      value: "light",
+      label: "Light"
+    },
+    {
+      value: "dark",
+      label: "Dark"
+    },
+    {
+      value: "userPref",
+      label: "User Preference"
+    }
+  ];
+  return (
+    <GestaltProvider colorScheme={scheme}>
+      <Box color="white" padding={2}>
+        <SelectList
+          id="scheme"
+          name="scheme"
+          onChange={({ value }) => setScheme(value)}
+          options={schemeOptions}
+          placeholder="Select color scheme"
+          label="Color scheme"
+          value={scheme}
+        />
+        <Box padding={2}>
+          <Text>Some content</Text>
+        </Box>
+        <Button text="Example button" inline /> <Button color="red" text="Red Button" inline />
+      </Box>
+    </GestaltProvider>
+  );
+}`}
+  />
+);
+
+export default cards;
--- a/docs/src/GestaltProvider.doc.js
+++ b/docs/src/GestaltProvider.doc.js
@@ -19,10 +19,10 @@ card(
     props={[
       {
         name: 'colorScheme',
-        type: `"light" | "dark" | "userPreferance"`,
+        type: `"light" | "dark" | "userPreference"`,
         defaultValue: 'light',
         description:
-          'The color scheme for components inside the provider. Specify "userPreferance" to use "prefers-color-scheme" media query.',
+          'The color scheme for components inside the provider. Specify "userPreference" to use "prefers-color-scheme" media query.',
         href: 'colorScheme',
       },
     ]}
@@ -47,7 +47,7 @@ function Example(props) {
       label: "Dark"
     },
     {
-      value: "userPreferance",
+      value: "userPreference",
       label: "User Preference"
     }
   ];
--- a/docs/src/GestaltProvider.doc.js
+++ b/docs/src/GestaltProvider.doc.js
@@ -25,6 +25,12 @@ card(
           'The color scheme for components inside the provider. Specify "userPreference" to use "prefers-color-scheme" media query.',
         href: 'colorScheme',
       },
+      {
+        name: 'id',
+        type: 'string',
+        description:
+          'An optional id for your provider. If not passed in, settings will be applied as globally as possible (example: it sets color scheme variables at :root).',
+      },
     ]}
   />
 );
@@ -52,7 +58,7 @@ function Example(props) {
     }
   ];
   return (
-    <GestaltProvider colorScheme={scheme}>
+    <GestaltProvider colorScheme={scheme} id="docsExample">
       <Box color="white" padding={2}>
         <SelectList
           id="scheme"
--- a/docs/src/GestaltProvider.doc.js
+++ b/docs/src/GestaltProvider.doc.js
@@ -19,10 +19,10 @@ card(
     props={[
       {
         name: 'colorScheme',
-        type: `"light" | "dark" | "userPref"`,
+        type: `"light" | "dark" | "userPreferance"`,
         defaultValue: 'light',
         description:
-          'The color scheme for components inside the provider. Specify "userPref" to use "prefers-color-scheme" media query.',
+          'The color scheme for components inside the provider. Specify "userPreferance" to use "prefers-color-scheme" media query.',
         href: 'colorScheme',
       },
     ]}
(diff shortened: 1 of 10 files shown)


docs/src/GestaltProvider.doc.js
```diff
@@ -0,0 +1,77 @@
...
+    description="An app may optionally have a `GestaltProvider` to set up context for components further down the tree. The first usecase is setting the color scheme, but other uses such as right to left support will be added in the future."
+  />
+);
+
+card(
+  <PropTable
+    props={[
+      {
+        name: 'colorScheme',
+        type: `"light" | "dark" | "userPref"`,
```
christianvuerings, line 22: @jennyscript could we name this `userPreference`?

packages/gestalt/src/contexts/Theme.js
```diff
@@ -0,0 +1,132 @@
...
+let themeId = 0;
+const getThemeId = () => {
+  themeId += 1;
+  return themeId;
+};
+
+const getTheme = (colorScheme: ?ColorScheme) =>
+  colorScheme === 'dark' ||
+  (colorScheme === 'userPref' &&
+    window &&
```
christianvuerings, line 75: @jennyscript could we add a `typeof window` check? Otherwise this might break SSR

packages/gestalt/src/contexts/Theme.js
```diff
@@ -0,0 +1,132 @@
...
+    if (key.startsWith('color')) {
+      styles += `  --${key}: ${theme[key]};\n`;
+    }
+  });
+  return styles;
+};
+
+let themeId = 0;
+const getThemeId = () => {
+  themeId += 1;
```
christianvuerings, line 68: @jennyscript could you explain why we need a global `themId` which increments each time we call `getThemeId`? Could we think of an alternative?

packages/gestalt/src/contexts/Theme.js
```diff
@@ -0,0 +1,132 @@
...
+    ? darkModeTheme
+    : lightModeTheme;
+
+export function ThemeProvider({
+  children,
+  colorScheme,
+}: Props): React.Element<typeof ThemeContext.Provider> {
+  const [theme, setTheme] = React.useState(getTheme(colorScheme));
+  const themeIdRef = React.useRef(getThemeId());
+  const className = `gestaltTheme${themeIdRef.current}`;
```
christianvuerings, line 87: @jennyscript could we start this with an underscore to avoid potential naming conflicts.

packages/gestalt/src/contexts/Theme.js
```diff
@@ -0,0 +1,132 @@
...
+  colorGray50: '#fff',
+  colorGray100: '#efefef',
+  colorGray200: '#767676',
+  colorGray300: '#111',
+  colorGray400: '#000',
+};
+
+const darkModeTheme = {
+  name: 'darkMode',
+  colorRed0: '#e60023',
```
christianvuerings, line 42: @jennyscript would there be a way to use CSS modules in this case? And swap a classname on the depending on whether it’s light or dark theme?

ponori:
Thanks @jennyscript! 

In the example, there are some tweaks we'll need to make to how dark mode applied to the buttons, but I'm uncertain if that's within the scope of this PR. Please let me know and I can start outlining them.

christianvuerings:
@jennyscript as a follow up, would be good to update the default button (hover state)

image

ponori:
I'm signing off on the design portion and will do a thorough review of all components once there's a global toggle in the docs.

packages/gestalt/src/contexts/Theme.js
jennyscript: I added this to make sure each ThemeProvider has a unique className. This is needed so you can nest themes or use different themes in parallel on the same page. It's kind of an edge case, but seemed like a reasonable thing to support. I could also try setting it up with a timestamp or other way to get uniqueness if this approach isn't ok

packages/gestalt/src/contexts/Theme.js
jennyscript: I wanted to pass the colors down in context for example to support updating fill in some of our svgs. It seemed good to only set them in one place if they need to be in JS anyway. Is it possible to read the colors out of CSS modules to reverse where they're stored?

jennyscript:
@christianvuerings I discussed the buttons with @ponori on Slack. We're going to do an audit of all the components like `Button` to come up with a list of work needed during the Makeathon to get them working properly in dark mode. Hoping to see lots of PRs in the next few days!

packages/gestalt/src/contexts/Theme.js
chrislloyd: Perhaps this namespace could be passed  in for the moment? Thinking more about it, we might have issues with SSR not hydrating correctly + it might be difficult to apply to the `<html>` element. `<ThemeProvider namespace=":root">` and make that the default in the `GestaltProvider`.

packages/gestalt/src/contexts/Theme.jsdom.test.js
```diff
@@ -0,0 +1,159 @@
...
+        ]
+      `);
+    });
+    it('renders styling for light mode when no color scheme specified', () => {
+      const { container } = render(<ThemeProvider>Content</ThemeProvider>);
+      expect(container.querySelectorAll('style')).toMatchInlineSnapshot(`
+        NodeList [
+          <style>
+            .__gestaltTheme3 {
+          --colorRed0: #ff5247;
```
chrislloyd, line 41: Should these be `--color-red-0`? Seems like it might be more "CSS Native" (though don't feel strongly about it).

packages/gestalt/src/contexts/Theme.js
jennyscript: I updated `GroupAvatar` to use the color from context as an example of what I meant

packages/gestalt/src/contexts/Theme.js
```diff
@@ -0,0 +1,132 @@
+// @flow strict
+import * as React from 'react';
+import PropTypes from 'prop-types';
+
+export type ColorScheme = 'light' | 'dark' | 'userPreference';
+
+export const ColorSchemePropType: React$PropType$Primitive<ColorScheme> = PropTypes.oneOf(
+  ['light', 'dark', 'userPreference']
```
chrislloyd, line 8: I wonder if `userPreference` here is better expressed as the null value? If not, this needs an `isRequired`.

packages/gestalt/src/contexts/Theme.js
```diff
@@ -0,0 +1,132 @@
...
+const getTheme = (colorScheme: ?ColorScheme) =>
+  colorScheme === 'dark' ||
+  (colorScheme === 'userPreference' &&
+    typeof window !== 'undefined' &&
+    window.matchMedia &&
+    window.matchMedia('(prefers-color-scheme: dark)').matches)
+    ? darkModeTheme
+    : lightModeTheme;
+
+export function ThemeProvider({
```
chrislloyd, line 81: The naming here could be a bit more consistent with the CSS Spec: `<ColorSchemeProvider prefers="dark" />`

jennyscript: We've had enough requests for theming that I think we'll need to provide something more flexible than locking it down to only light/dark in the future. I'd prefer to keep the name more generically "theme" for now so it'll still make sense then. Also, the data in the context here is an object containing colors so it feels more "theme-y" than just being a binary color preference even though that's how the specific theme got set.

packages/gestalt/src/contexts/Theme.js
jennyscript: The `colorScheme` prop is optional. If no scheme is specified it defaults to `light` for now since dark mode looks pretty bad in some places. Once we have dark mode stable `userPreference` would be a better default

packages/gestalt/src/contexts/Theme.js
jennyscript: I like that idea. Trying it out...

packages/gestalt/src/contexts/Theme.jsdom.test.js
jennyscript: Hmm, could go either way. I like that this mostly matches the names we use in JS like `const { colorGray100 } = useTheme();` would look weird in the more css convention though they don't have to match

packages/gestalt/src/contexts/Theme.js
jennyscript: This made things much more clear, thanks. I called it `id` to be a little more standard.

packages/gestalt/src/contexts/Theme.jsdom.test.js
christianvuerings: @jennyscript please prefix all of these with `--gestalt-`. See #971 for more context

jennyscript: Updated! That did increase the filesize a bit but since it's repetitive it should be very gzip-able

PR #969, opened 2020-07-07
Video: Test video state before calling play | Default volume to Muted

simkessy:
Addressing errors as result of Chrome Browser changes in handling video: 

1. `Uncaught (in promise) DOMException: play() failed because the user didn't interact with the document first.`: https://developers.google.com/web/updates/2017/09/autoplay-policy-changes

image

Essentially: Chrome will complain if we autoplay videos with sound without any user interaction. To prevent this we should set the default volume to muted. 

2. `The play() request was interrupted by a call to pause(). https://goo.gl/LdLk22`: https://goo.gl/LdLk22

Essentially: play() is an async function (pause is not), when pause is called before play promise completes, it interrupts the promise, if play() is then called, the initial promise failed loudly ... as far as I understand it. 

Implemented solution found here: https://stackoverflow.com/questions/36803176/how-to-prevent-the-play-request-was-interrupted-by-a-call-to-pause-error

## TODO

- [x] Accessibility checkup
- [x] Update documentation
- [x] Add/update Tests


simkessy:
> Thanks for fixing these bugs! It should help cut down on our error volume.

Ha yea getting a few an hour 
image

PR #970, opened 2020-07-07
SearchField: RTL support

christianvuerings:
Add RTL (Right to Left) support to `SearchField`.

## Before
searchfield-rtl-before

## After
searchfield-rtl-after

### DIFF
--- a/packages/gestalt/src/SearchField.js
+++ b/packages/gestalt/src/SearchField.js
@@ -108,6 +108,7 @@ const SearchField = ({
             },
           }}
           left
+          right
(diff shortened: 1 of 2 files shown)


packages/gestalt/src/SearchField.js
```diff
@@ -108,6 +108,7 @@ const SearchField = ({
             },
           }}
           left
+          right
```
christianvuerings, line 111: How does this work?

https://developer.mozilla.org/en-US/docs/Web/CSS/position

image

PR #986, opened 2020-07-08
Fix Box types

chrislloyd:
We found with #966 that Box's props are effectively untyped - `any`. This lead to an incident where somebody was incorrectly setting `zIndex` on Box, but Flow didn't catch it. This fixes the Flow suppressions on Box.

## Test Plan

I know this changed worked because Flow immediately found type errors in the DatePicker due to `role` and `zIndex` being (incorrectly) marked as required props. This diff also fixes those.

### DIFF
--- a/packages/gestalt/src/Box.js
+++ b/packages/gestalt/src/Box.js
@@ -256,6 +256,7 @@ type PropType = {
   role?: string,
 
   zIndex?: Indexable,
+  ...
 };
 
 // --
--- a/packages/gestalt/src/Box.js
+++ b/packages/gestalt/src/Box.js
@@ -253,9 +253,9 @@ type PropType = {
 
   userSelect?: 'auto' | 'none',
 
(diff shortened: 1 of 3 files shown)


packages/gestalt/src/Box.js
```diff
@@ -883,8 +874,8 @@ const SizeDisplayPropType = PropTypes.oneOf([
   'inlineBlock',
 ]);
 
-// $FlowFixMe[prop-missing]
-Box.propTypes = {
+// $FlowFixMe
```
btraut, line 877: What's the reason for this `$FlowFixMe`? I'm surprised it doesn't make you label the suppression type.

chrislloyd: Flow doesn't like the `propTypes` static on the `AbstractComponent` type (which makes sense). Feel like the easiest way to address this is #987 :D

PR #987, opened 2020-07-08
Remove prop-types

chrislloyd:
We've had `prop-types` for quite a while in Gestalt. It's a useful library for doing runtime invariant checking. However, in the last few years, it's been moved out of React core and generally hasn't been maintained well by the community. At the same time, we started embracing Flow and now have extremely good coverage across both Gestalt and Pinterest's internal codebase.

This removes `prop-types` in favor of just using Flow. Flow has proven to be more expressive, catch more bugs and is actively monitored. Prop-types are now a separate source of truth that has to be maintained with little value. As I found in #986 - Box has been untyped for a while and the `prop-type` declarations provided absolutely no extra safety.

The strongest argument that I've heard for `prop-types` is that it provides quick type feedback in the Gestalt Codesandbox. I'm sympathetic to this, however I don't think that it's worth the maintenance cost.


btraut:
Probably worth updating the description on this. I love the changes, but there's historical context worth capturing here.

christianvuerings:
@chrislloyd as discussed, closing this for now until we have flowtypes for our examples. Feel free to re-open afterwards.

PR #993, opened 2020-07-08
Rename Theme -> ColorScheme

chrislloyd:
I didn't get a chance to review #968 before it was merged (my mistake!) and I had a few lingering bits of feedback. This is probably the most major - standardizing on the naming of "ColorScheme" rather than "Theme".

I wrote a larger response as to why in a  private channel but to summarize:

* "Themes" is a _very_ broad topic. By labeling this  feature as a theme, future engineers might be tempted to bake in font size changes, or button color changes (for instance). While they could be loosely similar (updating CSS variables) this Provider is extremely specific to changing the colors based on the `prefers-color-scheme` media query. If there's a common abstraction between concepts it'll be more clear when we see more examples.[1]
* Standardizes with the CSS naming convention.

This PR doesn't update the internals of ColorScheme (there are still references to Theme) - I'll update those in a followup PR. This is just a quick PR to change update the public API before we start using it in more places (and it becomes harder to change).

@jennyscript I'd really like your feedback here. I figured the quickest and easiest way to followup on the  review would be to create a quick PR. Feel free to reject if you feel strongly.

### Test Plan

CI. This does break our public API seeing as a release happened between the last PR and the first, but I don't think it's adopted anywhere yet. @christianvuerings not sure if this should be labelled as Major (because of that) or minor (because it's not really used anywhere).

[1]: Not against the idea of "theming" but would like to break down a large set of asks into particular features instead of putting them all into one mechanism.


chrislloyd:
@christianvuerings @jennyscript have you seen this kind of build failure before? I can't repro locally and the codebase doesn't contain `contexts/Theme` anywhere... Is it a GH Actions caching issue?

```

src/index.js → dist/gestalt.js, dist/gestalt.es.js...
[!] Error: Could not resolve './contexts/Theme.js' from src/Avatar.js
Error: Could not resolve './contexts/Theme.js' from src/Avatar.js
    at error (/home/runner/work/gestalt/gestalt/node_modules/rollup/dist/shared/node-entry.js:5400:30)
    at ModuleLoader.handleResolveId (/home/runner/work/gestalt/gestalt/node_modules/rollup/dist/shared/node-entry.js:12410:24)
    at ModuleLoader.<anonymous> (/home/runner/work/gestalt/gestalt/node_modules/rollup/dist/shared/node-entry.js:12298:30)
    at Generator.next (<anonymous>)
    at fulfilled (/home/runner/work/gestalt/gestalt/node_modules/rollup/dist/shared/node-entry.js:38:28)
```

https://github.com/pinterest/gestalt/pull/993/checks?check_run_id=850732542#step:7:9

jennyscript:
What about `ColorThemeProvider`? It doesn't feel right here to tie this so tightly to the css implementation because we're just setting some colors here and happen to be basing it on the media query *for now*. I'd rather just keep it `Theme` for more flexibility in the future but `ColorThemeProvider` might be a fine compromise to indicate this is only for colors but also not to tie it to the media query.

chrislloyd:
I think adding new "color themes" will have to adjust the implementation of the implementation of this component - the media query in the generated CSS will have to change because CSS doesn't support themes other than "light" and "dark". I could see two types of other color themes being added:

1. High contrast - I think this needs to be implemented in a similar way to dark mode i.e. is a feature matrix that all components have to support. It's a little under spec'd at the moment but this [recent blog post](https://hacks.mozilla.org/2020/07/adding-prefers-contrast-to-firefox/) goes into more detail. I think components will have to support both a "dark mode" and a "high contrast" dark mode. We could bundle that in with this component, but it may be easier not to and have two separate components.
2. "Change default color of active buttons" - I think this is better served with settings in the components themselves (i.e. if we need to change the Button color in forms, there should be a dedicated provider for that.

I think what makes it a bit weird for me is that I would expect a "ColorThemeProvider" to take in the actual color values and force set them. This is really clearly only targeted to the media query and also contains all the possible color values so it's not particularly generic.

```jsx
<ColorThemeProvider red100="#F00" ... />
// vs
<ColorScheme prefers="dark" />
```

jennyscript:
I think it's fine to consider this a minor release since the changed code isn't used externally yet

Issue #997, opened 2020-07-08
Mask background color needs inversion


ponori:
The mask's background is currently `rgba(0, 0, 0, .03)` - let's switch it to `rgba(255, 255, 255, .03)`

jennyscript:
Fixed by #1089

PR #1007, opened 2020-07-08
Mask: update background color

mziemer21:
## TODO

- [x] Accessibility checkup
- [x] Update documentation
- [x] Add/update Tests
- [X] Pinterest Designer (for design changes)
- [X] [Format PR title](https://github.com/pinterest/gestalt/#releasing): `ComponentName: Description`.


mziemer21:
Addresses https://github.com/pinterest/gestalt/issues/997

jennyscript:
I think in the ticket the plan was to only change the wash color for dark mode. This change affects both light and dark mode

Screen Shot 2020-07-08 at 12 16 19 PM

For the original issue though, the docs say "If you expect the masked content to be nearly white, you can apply a wash to emphasize the edge of the mask" so does it make sense to apply more whiteness to this? Seems like the intent is to add something like `darkWash` where the whiteness is applied when "masked content to be nearly black" when in dark mode and not apply the normal wash when in darkMode. @ponori could you clarify what's correct here?

ponori:
@jennyscript, Ah, I see. If the change has to be global, then this _definitely_ defeats the purpose. My thinking is that the wash could be abstracted to be light/dark mode agnostic and just be a wash that's added to apply appropriate contrast given which color scheme you're in.

jennyscript:
If it's fine for the wash to apply globally then how about adding a color variable like `--gestalt-colorWash` that would be `rgba(0, 0, 0, 0.03)` in light and `rgba(255, 255, 255, .03)` in dark? This is also used in `GroupAvatar`

mziemer21:
We also have https://github.com/pinterest/gestalt/blob/master/packages/gestalt/src/Colors.css#L233-L239
```
.lightWashBg {
  background-color: #e2e2e2;
}

.darkWashBg {
  background-color: #dadada;
}
```

jennyscript:
Looking through the codebase, both `lightWash` and `darkWash` have a decent amount of use. They're mostly being used for things like hovering. They'll also need to be flipped for dark mode to look correct. @ponori Thoughts on what these colors should be in dark mode and if we can consolidate them somehow?

christianvuerings:
Closing this out for now since the PR hasn't been updated for while. Feel free to re-open the PR once the comments are addressed.

PR #1009, opened 2020-07-08
Button: Use accessible colors for darkMode red button

### LINKED ISSUE #992: Primary button default color change
Screenshot 2020-07-08 09 28 30

This specific component is unfortunately where we need to have a dark-mode-specific treatment. The background color should be `colorRed0` (#e60023) and the text color should be `colorGray400` (#fff). These changes are needed for continuity between primary/secondary buttons as well as accessibility.

jennyscript:
Closes #992 

This checks if we're in dark mode and adjusts the colors slightly if so to keep an accessible contrast

## TODO

- [x] Accessibility checkup
- [x] Update documentation
- [x] Add/update Tests

### DIFF
--- a/packages/gestalt/src/Button.js
+++ b/packages/gestalt/src/Button.js
@@ -63,14 +63,14 @@ export default function Button(props: Props): Element<'button'> {
   const { name: colorSchemeName } = useColorScheme();
   // We need to make a few exceptions for accessibility reasons in darkMode for red buttons
   const isDarkModeRed = colorSchemeName === 'darkMode' && color === 'red';
+  const colorClass = isDarkModeRed ? 'darkModeRed' : color;
 
   const classes = classnames(styles.button, {
     [styles.sm]: size === 'sm',
     [styles.md]: size === 'md',
     [styles.lg]: size === 'lg',
     [styles.solid]: color !== 'transparent',
-    [styles[color]]: !disabled && !selected && !isDarkModeRed,
-    [styles.darkModeRed]: isDarkModeRed,
+    [styles[colorClass]]: !disabled && !selected,
     [styles.selected]: !disabled && selected,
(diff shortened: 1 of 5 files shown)


ponori:
Looks great @jennyscript! Thanks!

packages/gestalt/src/Button.js
```diff
@@ -59,13 +60,17 @@ export default function Button(props: Props): Element<'button'> {
...
+  const isDarkModeRed = colorSchemeName === 'darkMode' && color === 'red';
 
   const classes = classnames(styles.button, {
     [styles.sm]: size === 'sm',
     [styles.md]: size === 'md',
     [styles.lg]: size === 'lg',
     [styles.solid]: color !== 'transparent',
-    [styles[color]]: !disabled && !selected,
+    [styles[color]]: !disabled && !selected && !isDarkModeRed,
+    [styles.darkModeRed]: isDarkModeRed,
```
jackhsu978, line 73: can we change this to:
```
[styles[isDarkModeRed ? 'darkModeRed' : color]]: !disabled && !selected
```

packages/gestalt/src/Colors.css
```diff
@@ -40,6 +40,14 @@
   background-color: var(--gestalt-colorRed100);
 }
 
+.red0 {
```
jackhsu978, line 43: where is this used?

jennyscript: That one isn't currently used. I added it for completeness. I'll remove it since it's un-needed for now

packages/gestalt/src/Button.js
jennyscript: @jackhsu978 Yes, this wasn't quite right thanks for catching! I made a slightly different change that puts the ternary result in a variable instead to hopefully be more readable

Issue #1006, opened 2020-07-08
Video controls should not invert in dark mode


ponori:
<img width="1105" alt="Screenshot 2020-07-08 11 45 16" src="https://user-images.githubusercontent.com/50343812/86958130-89937000-c110-11ea-9fb6-cbda797f439c.png">

All colors in the video controls should remain the same in light and dark modes.

AlbertCarreras:
taking these one! Alberto Carreras

AlbertCarreras:
Addressed on https://github.com/pinterest/gestalt/pull/1010

PR #1010, opened 2020-07-08
[Video] No darkMode changes in in video controls

AlbertCarreras:
Fixes: https://github.com/pinterest/gestalt/issues/1006

Adds static 'light'-schema color via ColorSchemeProvider at the video-component level

## After demo
Kapture 2020-07-08 at 14 46 35

## TODO

- [x] Accessibility checkup
- [ ] Update documentation
- [ ] Add/update Tests

### DIFF
--- a/packages/gestalt/src/Video.js
+++ b/packages/gestalt/src/Video.js
@@ -3,6 +3,7 @@
 import * as React from 'react';
 import PropTypes from 'prop-types';
 import VideoControls from './VideoControls.js';
+import { ColorSchemeProvider } from './contexts/ColorScheme.js';
 import styles from './Video.css';
 import Box from './Box.js';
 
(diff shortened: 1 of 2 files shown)


packages/gestalt/src/Video.js
```diff
@@ -495,57 +496,59 @@ export default class Video extends React.PureComponent<Props, State> {
...
-            onPlayheadUp={this.handlePlayheadUp}
-            onPause={this.handlePause}
-            onFullscreenChange={this.toggleFullscreen}
-            onVolumeChange={this.handleVolumeChange}
-            playing={playing}
-            seek={this.seek}
-            volume={volume}
-          />
-        )}
+        <GestaltProvider id="Video" colorScheme="light">
```
AlbertCarreras, line 499: Only change is line 499

jennyscript: Instead of using `GestaltProvider` here, could you just specifically use `ColorSchemeProvider` instead? We'll eventually have other contexts set through `GestaltProvider` unrelated to colors (maybe RTL or routing?) so for internal use like this it's more clear to just update the color scheme

PR #1015, opened 2020-07-08
Docs: Reorganize sidebar

chrislloyd:
I really love the new sidebar! However, it's a bit confusing because the first thing that's presented is "Data Display" and "Avatar" which... seems like not the most important thing to introduce people to Gestalt with. This PR reorganizes the sidebar to roughly order things in the order of "frequently used".

### DIFF
--- a/docs/src/components/sidebarIndex.js
+++ b/docs/src/components/sidebarIndex.js
@@ -8,9 +8,9 @@ export type sidebarIndexType = Array<{|
 |}>;
 
 const componentSubSectionPages = {
-  displayDate: ['Avatar', 'AvatarPair', 'Badge', 'GroupAvatar', 'Table'],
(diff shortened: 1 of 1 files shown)


chrislloyd:
@ajmoo @ashleyseto hey how does this seem to ya'll? Does it affect naming anywhere else?

ajmoo:
Doesn't affect me what order you display things in! Has anyone else provided feedback that the previous layout is confusing?

chrislloyd:
@ajmoo no feedback, just figured it was weird to have "Avatar" in "Data Display" and have "Foundation" below the fold. @christianvuerings does this look good to you?

ashleyseto:
Original document had "foundations" first then everything else under it in alphabetical order. I think that makes the most sense unless we want additional categories to make the navigation deeper.

christianvuerings:
Follow up at #1051

PR #1017, opened 2020-07-09
Flyout: Fix caret appearance in dark mode

### LINKED ISSUE #983: Fix dark mode caret color on Flyout
The caret is dark black with white border instead of elevated color

jennyscript:
Closes #983 

Before
Screen Shot 2020-07-08 at 4 59 53 PM

After
Screen Shot 2020-07-08 at 4 59 27 PM

- [ ] Accessibility checkup
- [ ] Update documentation
- [ ] Add/update Tests

### DIFF
--- a/packages/gestalt/src/Contents.js
+++ b/packages/gestalt/src/Contents.js
@@ -6,6 +6,7 @@ import Caret from './Caret.js';
 import styles from './Contents.css';
 import borders from './Borders.css';
 import colors from './Colors.css';
+import { useColorScheme } from './contexts/ColorScheme.js';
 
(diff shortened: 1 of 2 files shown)


packages/gestalt/src/Contents.js
```diff
@@ -564,3 +582,14 @@ export default class Contents extends React.Component<Props, State> {
     );
   }
 }
+
+export default function Contents(props: OwnProps): React.Node {
```
chrislloyd, line 586: Instead  of having to wrap this to get the hook, could it reference the Context directly?

jennyscript: I could export a render prop context consumer and use that instead for class components. Do you think that's something we should export externally as well?

Issue #979, opened 2020-07-08
Button hover/active is incorrect in dark mode


jennyscript:
Button hover is too light in dark mode

jennyscript:
Fixed by #1089

PR #1025, opened 2020-07-09
Button: Default hover state color

simkessy:
Issue: #979 

BEFORE: 

image

image

AFTER: 

image

## TODO

- [x] Accessibility checkup
- [x] Update documentation
- [x] Add/update Tests

### DIFF
--- a/packages/gestalt/src/Button.css
+++ b/packages/gestalt/src/Button.css
@@ -1,5 +1,7 @@
 :root {
   --gestalt-colorGray200: #767676;
+  --gestalt-colorGray50: #fff;
+  --gestalt-colorGray100: #efefef;
 }
 
 .button {
--- a/packages/gestalt/src/Button.css
+++ b/packages/gestalt/src/Button.css
@@ -1,3 +1,7 @@
+:root {
+  --gestalt-colorGray200: #767676;
+}
+
 .button {
   composes: borderBox minWidth60 from "./Layout.css";
   border-radius: 24px;
@@ -56,6 +60,10 @@
 .gray:hover,
 .gray:focus {
   background-color: #e2e2e2;
+}
+
+.darkModeGray:hover {
+  background-color: var(--gestalt-colorGray50)
 }
 
 .gray:active {
--- a/packages/gestalt/src/Button.css
+++ b/packages/gestalt/src/Button.css
@@ -62,13 +62,20 @@
   background-color: #e2e2e2;
 }
 
-.darkModeGray:hover {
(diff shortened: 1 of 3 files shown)


packages/gestalt/src/Button.js
```diff
@@ -62,7 +61,8 @@ export default function Button(props: Props): Element<'button'> {
   } = props;
   const { name: colorSchemeName } = useColorScheme();
   // We need to make a few exceptions for accessibility reasons in darkMode for red buttons
-  const isDarkModeRed = colorSchemeName === 'darkMode' && color === 'red';
+  const DARK_MODE = colorSchemeName === 'darkMode';
```
jennyscript, line 64: Caps/underscore doesn't go with the naming style of the rest of the codebase. Could you name this like `isDarkMode` or `inDarkMode` maybe?

packages/gestalt/src/Button.js
```diff
@@ -124,21 +125,3 @@ export default function Button(props: Props): Element<'button'> {
   );
   /* eslint-enable react/button-has-type */
 }
-
-Button.propTypes = {
```
chrislloyd, line 128: Please don't remove these as part of this PR.

packages/gestalt/src/Button.css
```diff
@@ -62,6 +66,17 @@
   background-color: #dadada;
 }
 
+.darkModeGray:hover,
+.darkModeGray:focus{
+  background-color: var(--gestalt-colorGray50)
+
+}
+.darkModeGray:active {
+  background-color: var(--gestalt-colorGray100)
```
jennyscript, line 75: These need to be added to the file as well or they won't get built for IE

packages/gestalt/src/Button.js
```diff
@@ -76,6 +77,7 @@ export default function Button(props: Props): Element<'button'> {
     [styles.enabled]: !disabled,
     [styles.inline]: inline,
     [styles.block]: !inline,
+    [styles.darkModeGray]: isDarkMode && color === 'gray',
```
jennyscript, line 80: This needs to be part of the same logic as `colorClass` so it gets handled correctly with the disabled/selected logic. You can see the issue in the docs where in dark mode gray selected has a hover color (incorrect) but red selected does not (correct)

packages/gestalt/src/Button.css
```diff
@@ -1,3 +1,9 @@
+:root {
+  --gestalt-colorGray200: #767676;
```
jennyscript, line 2: colorGray200 isn't used in this file

PR #1035, opened 2020-07-09
Merge: Better titles and obey branch protections

chrislloyd:
I noticed that PRs (like #1032) weren't getting merged if they weren't up to date with master. This is because they have a special status check that we have set to fail if the branch isn't  up to date with master (even if the rebase is clean). Changing the "base=master" value fixes this and has Mergify rebase PRs automatically and also has it obey that branches status checks (that's why I was able to remove the explicit status-check check).

**Note:** I'll disable the requirement of PRs being up to date with master once this lands.

I also noticed that the commit description was pretty useless because people are used to updating the PR description and not have meaningful commit names because we squash them anyway. 809000b5a3f690b2f54702c65a07182cbfdddd81 is an example. I've also changed the `commit_message` var to reformat the message from the title + body of the PR which is more in  line with what  folks are used to.

## Test Plan

I've verified this config on a bunch of open PRs and it's working pretty well.


.mergify.yml
```diff
@@ -1,13 +1,12 @@
...
-      - "#status-failure=0"
+      - '-draft'
+      - '#approved-reviews-by>=1'
+      - 'base=master'
     actions:
       merge:
         strict: true
         strict_method: rebase
         method: squash
+        commit_message: title+body
```
christianvuerings, line 12: @chrislloyd do you have an example of what this looks like? What was the value before?

.mergify.yml
```diff
@@ -1,13 +1,12 @@
 pull_request_rules:
-  - name: Automatic merge
+  - name: Automatic merge to master
     conditions:
-      - "-draft"
-      - "#approved-reviews-by>=1"
-      - "approved-reviews-by=@gestalt-core"
```
christianvuerings, line 6: @chrislloyd what was the reason for removing this line? Don't we want to ensure PRs only get merged when `gestalt-core` approves?

chrislloyd: I didn't realize this before, but Mergify will respect branch protection settings that we've setup in GH: https://doc.mergify.io/actions.html#branch-protection-settings. The rules before were roughly approximating those. I could probably go further and remove the `approved-reviews-by` condition too.

.mergify.yml
chrislloyd: No example. I'm just going by the documentation: https://doc.mergify.io/actions.html#merge

.mergify.yml
chrislloyd: Actually I added these back in. Without it it'll _attempt to_ merge even if tests haven't passed which is a bit of a waste.

PR #1038, opened 2020-07-09
Docs: Added darkMode to Algolia's DocsSearch feature

AlbertCarreras:
## before
image
## after
Kapture 2020-07-14 at 12 33 07

## Test Plan

* Is it accessible? YES
* Have you involved other stakeholders (such as a Pinterest Designer)? YES: based on Datepicker darkMode adoption

### DIFF
--- a/docs/src/components/DocSearch.css
+++ b/docs/src/components/DocSearch.css
@@ -30,10 +30,12 @@
 
 .searchbox__input {
   appearance: none;
-  background: #fff;
+  background: var(--gestalt-colorGray0);
   border: 0;
   border-radius: 999px;
   box-sizing: border-box;
+  caret-color: var(--gestalt-colorGray300);
+  color: var(--gestalt-colorGray300);
   display: inline-block;
   font-family: -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto,
     Oxygen-Sans, Ubuntu, Cantarell, "Fira Sans", "Droid Sans", "Helvetica Neue",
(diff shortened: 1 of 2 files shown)


docs/src/components/DocSearch.css
```diff
@@ -1,3 +1,11 @@
+:root {
+  /* primary colors */
+  --gestalt-colorGray0: #fff;
```
christianvuerings, line 3: @AlbertCarreras please don't prefix these with gestalt - we should only do that for ones that we export in our library.

docs/src/components/DocSearch.css
```diff
@@ -1,3 +1,11 @@
+:root {
```
christianvuerings, line 1: @AlbertCarreras can we avoid `:root` and use a wrapping CSS class instead?

docs/src/components/DocSearch.css
AlbertCarreras: image

christianvuerings: @AlbertCarreras yes - we need to disable that rule in the `docs/` directory.

AlbertCarreras: @cvuerings 

The way dark mode is implemented depends on the naming of the colors variables. 

image

docs/src/components/DocSearch.css
AlbertCarreras: DONE

PR #1041, opened 2020-07-09
Internal: Add CSS variables CI checks

christianvuerings:
Improves safety of our CSS variables.

We introduced a split CSS bundle for legacy and future CSS in #963 but this resulted in potential breakages:

    await duplicateVariablesDifferentValues();
    await noVarInLegacyCSS();

### Duplicate Variables with Different Values

We already saw this issue in #971 where the font-size variables for `Text` and `Heading` got overridden resulting in heading sizes to be the same as text sizes

image

*Test*: change a color that appears multiple times in 1 location: 
```
--gestalt-colorGray100: #efefef;
```
to
```
--gestalt-colorGray100: #efefed;
```

Screen Shot 2020-07-09 at 4 19 04 PM

### CSS Variables in legacy CSS

This will break in IE11 since the compiled CSS will still contain the `var()`. We can only use `:root` for now so it does get replaced in the legacy build.

*Test*: Add this to any file CSS file
```css
.textFieldTemp {
  --gestalt-colorGrad: #111;

  color: var(--gestalt-colorGrad);
}
```

Screen Shot 2020-07-09 at 4 31 43 PM

### DIFF
--- a/scripts/cssValidate.js
+++ b/scripts/cssValidate.js
@@ -1,0 +1,65 @@
+#!/usr/bin/env node
+const fs = require('fs');
+const postcss = require('postcss');
+const globby = require('globby');
+
+const duplicateVariablesDifferentValues = async () => {
+  const files = await globby([
+    'packages/gestalt/src/**/*.css',
+    'packages/gestalt-datepicker/src/**/*.css',
+  ]);
+
+  const combined = (
+    await Promise.all(
+      files.map(async file => {
+        return await fs.promises.readFile(file, 'utf8');
+      })
+    )
+  ).join('');
+
+  const astRoot = postcss.parse(combined);
+
+  const lookup = {};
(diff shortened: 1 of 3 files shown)


scripts/cssValidate.js
```diff
@@ -0,0 +1,65 @@
...
+    )
+  ).join('');
+
+  const astRoot = postcss.parse(combined);
+
+  const lookup = {};
+  astRoot.walkDecls(/^--gestalt/, ({ prop, value }) => {
+    if (lookup[prop] && lookup[prop] !== value) {
+      throw new Error(
+        `CSS Validate error: ${prop} is defined multiple times with different values: ${lookup[prop]} & ${value}}.\nPlease make these the same`
```
jennyscript, line 26: There's an extra curly here `${value}}`

christianvuerings: @jennyscript thank you - updated

PR #1042, opened 2020-07-10
ColorScheme: Set transparent colors in dark mode

jennyscript:
Pogs using these background colors now properly inverse in dark mode. The hover colors are still not quite right in dark mode but this is an improvement

## Test Plan

Check Pogs in docs in light mode and dark mode. Make sure the "transparentDarkGray" and "white" colors reverse in dark mode

Screen Shot 2020-07-09 at 6 23 20 PM

Screen Shot 2020-07-09 at 6 23 11 PM

### DIFF
--- a/packages/gestalt/src/Colors.css
+++ b/packages/gestalt/src/Colors.css
@@ -230,15 +230,15 @@
 }
 
 .transparentDarkGray {
-  color: var(--gestalt-transparentDarkGray);
+  color: var(--gestalt-colorTransparentDarkGray);
 }
 
 .transparentDarkGrayBg {
-  background-color: var(--gestalt-transparentDarkGray);
+  background-color: var(--gestalt-colorTransparentDarkGray);
 }
 
 .transparentWhiteBg {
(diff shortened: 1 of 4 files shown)


packages/gestalt/src/Colors.css
```diff
@@ -230,15 +230,15 @@
 }
 
 .transparentDarkGray {
-  color: var(--gestalt-transparentDarkGray);
```
jackhsu978, line 233: should we clean this up?

packages/gestalt/src/Colors.css
```diff
@@ -230,15 +230,15 @@
...
+  color: var(--gestalt-colorTransparentDarkGray);
 }
 
 .transparentDarkGrayBg {
-  background-color: var(--gestalt-transparentDarkGray);
+  background-color: var(--gestalt-colorTransparentDarkGray);
 }
 
 .transparentWhiteBg {
-  background-color: var(--gestalt-transparentWhite);
```
jackhsu978, line 241: should we clean this up?

PR #1043, opened 2020-07-10
Docs: Next.js

chrislloyd:
- [x] Bold links in sidebar
- [x] Convert `_document` with extra search JS imports
- [x] Pre styling in Markdown
- [ ] `yarn start` working

PR #1044, opened 2020-07-10
Link: Add React.forwardRef

chrislloyd:
When trying to add Next.js in #1043, I've run into a few issues with the Next link and Gestalt's Link. Mostly, it expects access to the underlying ref. I think this is pretty straight-forward and a fairly safe change (seeing and not many places use Link). It also feels similar to other places where we use `forwardRef` (like `<Box/>`) in that Link is just a thin wrapper over `HTMLLinkElement`.

## Test Plan

CI

### DIFF
--- a/packages/gestalt/src/Link.js
+++ b/packages/gestalt/src/Link.js
@@ -31,22 +31,28 @@ type Props = {|
   target?: null | 'self' | 'blank',
 |};
 
-function Link({
-  accessibilitySelected,
-  children,
-  href,
-  id,
-  inline = false,
-  onBlur,
-  onClick,
-  onFocus,
-  rel = 'none',
-  role,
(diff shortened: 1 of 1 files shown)


christianvuerings:
@chrislloyd fyi adding `forwardRef` to a component is a breaking change. Let's make sure we bump the Gestalt version in Pinterest's codebase separately from any other commit when this lands.

image

chrislloyd:
Oh cool, I hadn't seen that but makes sense. Yes, lets!

christianvuerings:
Follow up at #1044 which
* Uses the same forwardRef approach as we do with other components
* Add docs
* Sets the `displayName` to `Link` (avoid having to update too many snapshots)
* Add the correct propType

PR #1045, opened 2020-07-10
AbstractEventHandler type

chrislloyd:
We've discussed this a little bit elsewhere - having to repeat the standard event format is unusual and error prone. This starts to consolidate all the event handlers to have the same enforced type. This also opens up an easy refactoring to change all event handlers to return `mixed` instead of `void` to make them place nicer with async functions.

I've updated most events across Gestalt to use it and also added a few type refinements where they were missing (filling in `SyntheticEvent<>` holes) The documentation doesn't need to change because the underlying structure is still the same.

- [x] Document `AbstractEventHandler` in a general section talking about event handlers and _why_ we have the `{event}` object instead of the normal callback format.

## Test Plan

Flow


packages/gestalt/src/AbstractEvent.js
```diff
@@ -0,0 +1,14 @@
+// @flow strict
+
+/**
+ * We use a common event payload across all components. Since it's different from standard events that people are used to from React, it's abstracted here to at least make it standard. If you have to learn something, you should only have to learn it once.
+ *
+ * Why is it non-standard? Gestalt components are controlled - meaning that the can contain state and when that changes it's reflected back to the parent. These state changes are propagated via. event handlers. Often this state doesn't represent what's in the DOM or we just want to abstract the DOM from the consumer. We felt this format struct a good balance between readability and prediability.
```
jackhsu978, line 6: "prediability" => "predictability"

packages/gestalt/src/AbstractEvent.js
```diff
@@ -0,0 +1,14 @@
...
+ */
+export type AbstractEvent<
```
jackhsu978, line 8: The `AbstractEvent` here is not an event, but a function that takes (handles) an event. I think `AbstractEventHandler` would make more sense.

packages/gestalt/src/AbstractEvent.js
```diff
@@ -0,0 +1,14 @@
...
+  T: SyntheticEvent<HTMLElement> | Event,
+  U = {||}
+> = ({|
```
jackhsu978, line 11: I know this is compatible with the existing API, but I am curious why we decided to go with the current approach.

Have we considered the following API where the first parameter is the event object and the second parameter is whatever additional data we want to return?

```
type AbstractEventHandler<T: SyntheticEvent<HTMLElement> | Event, U>
= (T, U) => void;
```

There are several benefits of this approach:

- This is more consistent with React's built-in event handler API where the first parameter is event object. 
- The second argument doesn't have to be an object
- There is no implicit requirement that the payload object cannot contain property named "event"

What do you think?

chrislloyd: I think you should either make an issue or open a PR and we can discuss further :) The intention of this PR wasn't to change the existing signature but to make it easier to change across the board in the future.

packages/gestalt/src/AbstractEvent.js
chrislloyd: I like it!

packages/gestalt/src/AbstractEvent.js
jackhsu978: sounds good. thanks!

jackhsu978:
@chrislloyd Thanks for making the changes. It looks like CI fails because [`73d4010` (#1045)](https://github.com/pinterest/gestalt/pull/1045/commits/73d4010e665768d0846cfb6018ee5cbc67e85c3d) deleted the file `AbstractEvent.js` but did not create the renamed file `AbstractEventHandler.js`. Everything else LGTM.

PR #1046, opened 2020-07-10
Video: Add captions + controls

jesselumarie:
Currently, the `captions` attribute on videos doesn't actually show captions on a video, we need to update the video text track state to show the captions.

## Test Plan
  - [x] verify that captions auto-play when provided
  - [x] verify that captions stop when toggled off
  - [x] verify captions restart when toggled on

captions

### DIFF
--- a/packages/gestalt/src/Video.js
+++ b/packages/gestalt/src/Video.js
@@ -12,6 +12,8 @@ type Source =
   | Array<{| type: 'video/m3u8' | 'video/mp4' | 'video/ogg', src: string |}>;
 
 type Props = {|
+  accessibilityHideCaptionsLabel?: string,
+  accessibilityShowCaptionsLabel?: string,
   accessibilityMaximizeLabel: string,
   accessibilityMinimizeLabel: string,
   accessibilityMuteLabel: string,
@@ -159,6 +161,8 @@ export default class Video extends React.PureComponent<Props, State> {
   player: ?HTMLDivElement;
 
   static propTypes = {
+    accessibilityHideCaptionsLabel: PropTypes.string,
+    accessibilityShowCaptionsLabel: PropTypes.string,
     accessibilityMaximizeLabel: PropTypes.string,
     accessibilityMinimizeLabel: PropTypes.string,
     accessibilityMuteLabel: PropTypes.string,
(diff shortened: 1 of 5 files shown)


packages/gestalt/src/Video.js
```diff
@@ -20,7 +20,6 @@ type Props = {|
   accessibilityUnmuteLabel: string,
   aspectRatio: number,
   captions: string,
-  children?: React.Node,
```
christianvuerings, line 23: @jesselumarie what's the reason for removing this line? Could we add it back in?

image

jesselumarie: definite rebase mistake -- good catch

packages/gestalt/src/Video.js
```diff
@@ -331,6 +337,25 @@ export default class Video extends React.PureComponent<Props, State> {
     }
   };
 
+  // Toggle captions on/off
+  toggleCaptions: () => void = () => {
+    if (this.video && this.video.textTracks && this.video.textTracks[0]) {
```
jackhsu978, line 342: nit: can we simplify the code here a little?

```js
toggleCaptions: () => {
  const [videoTrack] = this.video?.textTracks || [];
  if (videoTrack) {
    const isShowing = videoTrack.mode === 'showing';
    videoTrack.mode = isShowing ? 'disabled' : 'showing';
    this.setState({ captionsButton: isShowing ? 'disabled' : 'enabled' });
  }
};
```

packages/gestalt/src/Video.js
```diff
@@ -236,6 +238,10 @@ export default class Video extends React.PureComponent<Props, State> {
     if (playing) {
       this.play();
     }
+
+    if (this.video && this.video.textTracks && this.video.textTracks[0]) {
```
jackhsu978, line 242: nit: do we need to check the `captions` prop here?

packages/gestalt/src/Video.js
jackhsu978: lgtm

packages/gestalt/src/Video.js
```diff
@@ -499,6 +520,7 @@ export default class Video extends React.PureComponent<Props, State> {
         <ColorSchemeProvider id="Video" colorScheme="light">
           <video
             autoPlay={playing}
+            crossOrigin="anonymous"
```
jackhsu978, line 523: why is this change necessary?

packages/gestalt/src/VideoControls.js
```diff
@@ -154,6 +174,20 @@ class VideoControls extends React.Component<Props> {
             />
           </TapArea>
         </Box>
+        {captionsButton && (
+          <Box padding={2}>
+            <TapArea onTap={this.handleCaptionsChange} fullWidth={false}>
+              <Icon
+                accessibilityLabel="closed-captions"
```
christianvuerings, line 181: @jesselumarie this should be 2 props that are passed in and it should depend on whether we show / hide the captions.

Something along these lines:
```
<Icon accessibilityLabel={captionsButton === 'enabled' ? accessibilityHideCaptionsLabel : accessibilityShowCaptionsLabel} />
```

packages/gestalt/src/Video.js
jesselumarie: the track needs to match the video origin -- if it doesn't, you need to pass `anonymous`, otherwise the browser will block the captions track from being loaded.

PR #1050, opened 2020-07-14
Link: add React.forwardRef

christianvuerings:
Follow up from #1044 with the following changes:

* Uses the same forwardRef approach as we do with other components
* Add docs
* Sets the `displayName` to `Link` (avoid having to update too many snapshots)
* Add the correct propType

From @chrislloyd 
> When trying to add Next.js in #1043, I've run into a few issues with the Next link and Gestalt's Link. Mostly, it expects access to the underlying ref. I think this is pretty straight-forward and a fairly safe change (seeing and not many places use Link). It also feels similar to other places where we use `forwardRef` (like `<Box/>`) in that Link is just a thin wrapper over `HTMLLinkElement`.

## Test Plan

CI

### DIFF
--- a/packages/gestalt/src/Link.js
+++ b/packages/gestalt/src/Link.js
@@ -17,6 +17,7 @@ type TapEvent =
 type Props = {|
   accessibilitySelected?: boolean,
   children?: React.Node,
+  forwardedRef?: React.Ref<'a'>,
   hoverStyle?: 'none' | 'underline',
   href: string,
   id?: string,
@@ -34,6 +35,7 @@ type Props = {|
 function Link({
   accessibilitySelected,
   children,
+  forwardedRef,
   href,
   id,
   inline = false,
@@ -105,6 +107,7 @@ function Link({
       onTouchMove={handleTouchMove}
       onTouchCancel={handleTouchCancel}
       onTouchEnd={handleTouchEnd}
+      ref={forwardedRef}
       rel={[
         ...(target === 'blank' ? ['noopener', 'noreferrer'] : []),
         ...(rel === 'nofollow' ? ['nofollow'] : []),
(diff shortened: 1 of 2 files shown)


packages/gestalt/src/Link.js
```diff
@@ -142,6 +151,11 @@ const LinkPropTypes = {
...
 };
 
-Link.propTypes = LinkPropTypes;
+function LinkWithRef(props, ref) {
+  return <Link {...props} forwardedRef={ref} />;
+}
+LinkWithRef.displayName = 'Link';
 
-export default Link;
+export default (React.forwardRef<Props, HTMLAnchorElement>(
```
jackhsu978, line 159: The component returned from `React.forwardRef` would no longer be named `Link`. I think we might want to do something like this:
```
const LinkWithForwardRef: React$AbstractComponent<Props, HTMLAnchorElement>) = React.forwardRef<Props, HTMLAnchorElement>(LinkWithRef);
LinkWithForwardRef.displayName = 'Link';
export default LinkWithForwardRef;
```

packages/gestalt/src/Link.js
```diff
@@ -17,6 +17,7 @@ type TapEvent =
 type Props = {|
   accessibilitySelected?: boolean,
   children?: React.Node,
+  forwardedRef?: React.Ref<'a'>,
```
jackhsu978, line 20: should we do `React.Ref<HTMLAnchorElement>` for consistency? is there any difference?

christianvuerings: @jackhsu978 neither of these work:

image
image

In other places we have also used `React.Ref<'input'>` and `React.Ref<'div'>`

PR #1052, opened 2020-07-14
Pog: add optional accessibilityLabel

jackhsu978:
Add optional `accessibilityLabel` prop to `Pog` component for the following use case:

```jsx
<Link to="/following/">
  <Pog icon="people" accessibilityLabel="Following" />
</Link>
```

Screenshot 2020-07-14 15 31 53

## Test Plan

jest

### DIFF
--- a/packages/gestalt/src/Pog.js
+++ b/packages/gestalt/src/Pog.js
@@ -23,6 +23,10 @@ const SIZE_NAME_TO_ICON_SIZE_PIXEL = {
 };
 
 type Props = {|
+  // Omit accessibilityLabel if and only if an ancestor element already has the aria-label set.
+  // This is similar to having empty `alt` attributes:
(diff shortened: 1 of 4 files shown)


packages/gestalt/src/Pog.js
```diff
@@ -89,14 +94,8 @@ export default function Pog(props: Props): React.Node {
...
     <div className={classes} style={inlineStyle}>
-      {/*
-        We're explicitly setting an empty string as a label on the Icon since we
-        already have an aria-label on the button container.
-        This is similar to having empty `alt` attributes:
-        https://davidwalsh.name/accessibility-tip-empty-alt-attributes
-      */}
       <Icon
-        accessibilityLabel=""
+        accessibilityLabel={accessibilityLabel || ''}
```
christianvuerings, line 98: @jackhsu978 since we already set `accessibilityLabel = '',` on line 61, we no longer need the fallback here.

PR #1053, opened 2020-07-15
TextArea: Add forward ref (BREAKING CHANGE)

AlbertCarreras:
- Adds ref forwarding to TextArea
- Added ref example to Docs
- Added codemod helper
- Added test for ref 

[ Follows the implementation in TextField ]

### Why?
We have _0_ places in the Pinterest codebase where we set a ref on the current <TextArea />.  

However, without forwardedRefs, refs are set on the component's instance, not the actual input. So in each of those cases, people drill down to get the <textarea />.

)

### How can we find in which location we already have these legacy refs set?
Run the codemod:

```
cd ~/code/gestalt
yarn run codemod --parser=flow -t=packages/gestalt-codemods/6.2.0-7.0.0/textarea-find-deprecated-refs.js ~/code/pinboard/webapp
```

Which will output the following:

`Update legacy ref on TextArea: /home/<user>/code/pinterest/PasswordStep.js:125:12`

We supply the line number and column name so it brings you to the right location in the editor (`ctrl + P` file seach).

## Test Plan

* Is it tested? YES
~~* Is it accessible?~~
* Is it documented? YES

### DIFF
--- a/packages/gestalt/src/TextArea.js
+++ b/packages/gestalt/src/TextArea.js
@@ -12,6 +12,7 @@ import styles from './TextArea.css';
 type Props = {|
   errorMessage?: string,
   disabled?: boolean,
+  forwardedRef?: React.Ref<'textarea'>,
   hasError?: boolean,
   helperText?: string,
   id: string,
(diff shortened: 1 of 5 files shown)


packages/gestalt/src/TextArea.js
```diff
@@ -34,137 +35,118 @@ type Props = {|
...
+  return <TextArea {...props} forwardedRef={ref} />;
+}
 
-    const classes = classnames(
-      styles.textArea,
-      formElement.base,
-      disabled ? formElement.disabled : formElement.enabled,
-      hasError || errorMessage ? formElement.errored : formElement.normal
-    );
+forwardRef.displayName = 'TextArea';
```
christianvuerings, line 148: @AlbertCarreras see https://github.com/pinterest/gestalt/blob/a4c967d97d9fc93613ea4e6442c66b9c64262ca2/packages/gestalt/src/Link.js#L154-L163

we should be setting the `displayName` on the default export as well.

I recommend the following:
```
function TextAreaWithRef(props, ref) {
  return <TextArea {...props} forwardedRef={ref} />;
}
TextAreaWithRef.displayName = 'ForwardRef(TextArea)';

const TextAreaWithForwardRef: React.AbstractComponent<
  Props,
  HTMLAnchorElement
> = React.forwardRef<Props, HTMLAnchorElement>(LinkWithRef);

TextAreaWithForwardRef.displayName = 'TextArea';

export default TextAreaWithForwardRef;
```

packages/gestalt/src/TextArea.jsdom.test.js
```diff
@@ -95,6 +98,22 @@ describe('TextArea', () => {
     }
   });
 
+  it('forwards a ref to <input />', () => {
```
christianvuerings, line 101: thanks for adding this test!

packages/gestalt/src/TextArea.js
AlbertCarreras: done

PR #1054, opened 2020-07-16
Button/IconButton: compress onClick or onTouch

christianvuerings:
Add compress styles to `Button` and `IconButton`

button-iconbutton-compress

### Why not make this conditional?
To enforce consistency on how buttons and icon buttons behave. We currently have it set conditionally on `Link` and `TapArea` but want to eventually make it consistent across all components.

### Why does the PR touch so many files?
We updated the class names in `Touchable.css` so the snapshots changed.

## Test Plan

Try out the `Button` and `IconButton` out with the new compress style:
* https://deploy-preview-1054--gestalt.netlify.app/#/components/Button
* https://deploy-preview-1054--gestalt.netlify.app/#/components/IconButton

PR #1056, opened 2020-07-16
Button/IconButton: compress onClick or onTouch

christianvuerings:
Follow up from #1054 which got merged by accident because of #1055 

Add compress styles to `Button` and `IconButton`

button-iconbutton-compress

### Why not make this conditional?
To enforce consistency on how buttons and icon buttons behave. We currently have it set conditionally on `Link` and `TapArea` but want to eventually make it consistent across all components.

### Why does the PR touch so many files?
We updated the class names in `Touchable.css` so the snapshots changed.

### How do you calculate the compress style?

`(width - 4)/width`. If the button is 100px wide: (100-4)/100 = 96%

## Test Plan

Try out the `Button` and `IconButton` out with the new compress style:
* https://deploy-preview-1056--gestalt.netlify.app/#/components/Button
* https://deploy-preview-1056--gestalt.netlify.app/#/components/IconButton

### DIFF
--- a/packages/gestalt/src/useTapFeedback.js
+++ b/packages/gestalt/src/useTapFeedback.js
@@ -16,7 +16,14 @@ export const keyPressShouldTriggerTap = (
   event: SyntheticKeyboardEvent<TapTargetHTMLElement>
 ): boolean => [SPACE_CHAR_CODE, ENTER_CHAR_CODE].includes(event.charCode);
 
-export default function useTapFeedback(): {|
+export default function useTapFeedback({
+  height,
+  width,
+}: {|
+  height: ?number,
+  width: ?number,
+|}): {|
+  compressStyle: ?{| transform: string |},
   handleBlur: () => void,
   handleMouseDown: () => void,
   handleMouseUp: () => void,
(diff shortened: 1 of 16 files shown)


packages/gestalt/src/useTapFeedback.js
```diff
@@ -16,7 +16,10 @@ export const keyPressShouldTriggerTap = (
   event: SyntheticKeyboardEvent<TapTargetHTMLElement>
 ): boolean => [SPACE_CHAR_CODE, ENTER_CHAR_CODE].includes(event.charCode);
 
-export default function useTapFeedback(): {|
+export default function useTapFeedback(ref: {|
+  current: null | React$ElementRef<string>,
```
jackhsu978, line 20: instead of passing `current`, can we just pass `clientWidth`?

packages/gestalt/src/useTapFeedback.js
```diff
@@ -31,7 +34,21 @@ export default function useTapFeedback(): {|
     x: 0,
     y: 0,
   });
+
+  const [compressStyle, setCompressStyle] = React.useState(null);
+
+  React.useEffect(() => {
+    if (ref && ref.current) {
+      setCompressStyle({
+        transform: isTapping
```
jackhsu978, line 43: instead of dynamic value, can we just use "95%" instead? we used to use 95% for part of our codebase and it might be a nice simple default.

christianvuerings: @jackhsu978 the 95% doesn't work for bigger containers - they compress too much. It might work for mweb but not on desktop web.

Before
link-before-static

After
link-after-static

packages/gestalt/src/useTapFeedback.js
christianvuerings: @jackhsu978 definitely - good suggestion. Updated and passed in `width` + `height` of the components (we need to find the longest side)

jackhsu978: thanks

packages/gestalt/src/useTapFeedback.js
jackhsu978: got it. thanks!

PR #1057, opened 2020-07-16
Checkbox: Add forward ref (BREAKING CHANGE)

AlbertCarreras:
- Adds ref forwarding to Checkbox
- Added ref example to Docs
- Added test for ref 

### Why?

We have use cases in the Pinterest codebase where we set a ref on a wrapping container for Checkbox in order to anchor other components on it.
<Box><Checkbox/></Box>
 
To abstract this implementation, let's add a way for them to get it in a dryer way.

Moreover, without forwardedRefs, refs are set on the component's instance, not the actual input. So in each of those cases, people drill down to get the <input />.

)

## Test Plan

* Is it tested? YES
~~* Is it accessible?~~
* Is it documented? YES

Browser | validating the click target is still correct | validating layout looks good | it's accessible (you can hit space to toggle the checkbox) | Demo
------------ | ------------- | ------------- | ------------- | -------------
Chrome | CORRECT | CORRECT | CORRECT | Kapture 2020-07-17 at 11 22 34
Safari | CORRECT | CORRECT | FAILS unrelated to this change |Kapture 2020-07-17 at 11 25 12
IE11 | CORRECT | CORRECT | CORRECT |  Kapture 2020-07-17 at 11 21 04

### DIFF
--- a/docs/src/Checkbox.doc.js
+++ b/docs/src/Checkbox.doc.js
@@ -76,9 +76,8 @@ a checkbox and its indeterminism are independent.`,
       },
       {
         name: 'ref',
-        type: "React.Ref<'div'>",
-        description:
-          'Forward the ref to the underlying Box component wrapping the input element',
+        type: "React.Ref<'input'>",
+        description: 'Forward the ref to the underlying input element',
         href: 'refExample',
       },
       {
--- a/docs/src/Checkbox.doc.js
+++ b/docs/src/Checkbox.doc.js
@@ -75,6 +75,13 @@ a checkbox and its indeterminism are independent.`,
         type: `({ event: SyntheticInputEvent<HTMLInputElement>, checked: boolean }) => void`,
       },
       {
+        name: 'ref',
+        type: "React.Ref<'div'>",
+        description:
+          'Forward the ref to the underlying Box component wrapping the input element',
+        href: 'refExample',
+      },
+      {
         name: 'size',
         type: `"sm" | "md"`,
         defaultValue: 'md',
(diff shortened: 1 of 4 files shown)


packages/gestalt/src/Checkbox.js
```diff
@@ -100,7 +102,7 @@ export default function Checkbox({
         marginRight={-1}
       >
         <Label htmlFor={id}>
-          <Box paddingX={1} position="relative">
+          <Box ref={forwardedRef} paddingX={1} position="relative">
             <input
```
christianvuerings, line 106: @AlbertCarreras could we set the ref on the `input type="checkbox"` instead. That way people can set focus on the element if they want to.

AlbertCarreras: DONE

packages/gestalt/src/Checkbox.js
```diff
@@ -29,29 +29,34 @@ type Props = {|
...
+  } = props;
   const inputElement = React.useRef<?HTMLInputElement>(null);
   const [focused, setFocused] = React.useState(false);
   const [hovered, setHover] = React.useState(false);
 
   React.useEffect(() => {
-    if (inputElement && inputElement.current) {
+    // $FlowFixMe
+    if (forwardedRef && forwardedRef.current) {
+      // $FlowFixMe
```
AlbertCarreras, line 54: how do I fix this Flow issue?
image

packages/gestalt/src/Checkbox.js
```diff
@@ -117,7 +122,7 @@ function Checkbox({
               onFocus={() => setFocused(true)}
               onMouseEnter={handleHover}
               onMouseLeave={handleHover}
-              ref={inputElement}
+              ref={forwardedRef || inputElement}
```
christianvuerings, line 125: @AlbertCarreras I ran into this yesterday as well - the issue is that we're trying to combine `forwardRef` and an `innerRef`.

See https://github.com/pinterest/gestalt/blob/4fb07785648991a84e0f50bc5836dc3314b25514/packages/gestalt/src/TapArea.js#L65-L67 how we fixed it for `TapArea`.

packages/gestalt/src/Checkbox.js
```diff
@@ -100,7 +107,7 @@ export default function Checkbox({
         marginRight={-1}
       >
         <Label htmlFor={id}>
-          <Box paddingX={1} position="relative">
```
christianvuerings, line 103: @AlbertCarreras I do think we need `position="relative"` for browser compatibility. Could you try this change out in:
* Safari
* Chrome
* IE11

by:
* validating the click target is still correct
* validating layout looks good
* it's accessible (you can hit space to toggle the checkbox)

An alternative would be to set an explicit size on `input`

packages/gestalt/src/Checkbox.js
AlbertCarreras: DONE

packages/gestalt/src/Checkbox.js
AlbertCarreras: ALSO removing position='relative' allows Flyout to get positioned correctly. With it, 
image

docs/src/Checkbox.doc.js
```diff
@@ -74,6 +74,13 @@ a checkbox and its indeterminism are independent.`,
         name: 'onClick',
         type: `({ event: SyntheticInputEvent<HTMLInputElement>, checked: boolean }) => void`,
       },
+      {
+        name: 'ref',
+        type: "React.Ref<'div'>",
```
christianvuerings, line 79: @AlbertCarreras can we change this to `"React.Ref<'input'>"`?

AlbertCarreras: DONE

packages/gestalt/src/Checkbox.js
```diff
@@ -104,7 +113,7 @@ export default function Checkbox({
...
```
tennisonchan, line 107: This position relative is need as the input element is position absolute

PR #1060, opened 2020-07-20
CollapsibleSubsection:  Automatic Scroll to Top

### LINKED ISSUE #1059: Scroll to top CollapsibleSubsection navigation
When one navigates to a subsection in the documentation, the scroll of the page remains unchanged or does weird jumps because of page resizing. As can be seen in the image, after redirect the user ends up in random places and needs to scroll to the top to read the content. Even though this could be useful for side menu navigation, I think an automatic scroll to top after redirect would be an improvement. 

Screen Shot 2020-07-20 at 6 31 42 PM

IgorWilbert:
This PR is supposed to fix #1059.
I would like to have feedback if the scroll behavior should be updated, and if this is the most adequate way of implementing the update.

## Test Plan
Navigate to the documentation site and test if the links in the CollapsibleSubsection components of the side menu automatically scroll to top on click.

### DIFF
--- a/docs/src/components/NavLink.js
+++ b/docs/src/components/NavLink.js
@@ -26,6 +26,7 @@ const NavLink = ({ children, to, history }: Props) => {
     if (isModifiedEvent(event) || !isLeftClickEvent(event)) return;
     event.preventDefault();
     history.push(to);
+    window.scrollTo(0, 0);
   };
 
   return (


AlbertCarreras:
Thanks Igor for the suggestion! taking a look now

docs/src/components/NavLink.js
```diff
@@ -26,6 +27,7 @@ const NavLink = ({ children, to, history }: Props) => {
     if (isModifiedEvent(event) || !isLeftClickEvent(event)) return;
     event.preventDefault();
     history.push(to);
+    if (scrollToTop) window.scrollTo(0, 0);
```
AlbertCarreras, line 30: Hey Igor, I think we can just leave here, no need for the `scrollToTop` prop.

`window.scrollTo(0, 0);`

Every time we click, it scrolls to top.

IgorWilbert: @AlbertCarreras thank you for the tip and for the quick review! I updated it, please let me know if you think other changes are required.

christianvuerings:
@IgorWilbert thanks for adding in this fix!

PR #1063, opened 2020-07-20
InputButton/Button: Add forward ref (BREAKING CHANGE)

AlbertCarreras:
- Adds ref forwarding to `IconButton` & `Button`
- Added ref example to Docs
- Added test for ref 

+ Cleaned examples to use empty tags when needed.

### Why?

We have use cases in the Pinterest codebase where we set a ref on a wrapping container for `IconButton`/`Button` in order to anchor other components on it.
`<Box><IconButton/></Box>`
 
To abstract this implementation, let's add a way for them to get it in a dryer way.

Moreover, without `forwardedRef`s, refs are set on the component's instance, not the actual input. So in each of those cases, people drill down to get the innermost `<button />` element.

)
> When you start using `forwardRef` in a component library, you should treat it as a breaking change and release a new major version of your library. This is because your library likely has an observably different behavior (such as what refs get assigned to, and what types are exported), and this can break apps and other libraries that depend on the old behavior.

## Test Plan

* Is it tested? YES
~~* Is it accessible?~~
* Is it documented? YES

### DIFF
--- a/packages/gestalt/src/Button.js
+++ b/packages/gestalt/src/Button.js
@@ -1,6 +1,6 @@
 // @flow strict
 
-import React, { useRef, type Element } from 'react';
+import * as React from 'react';
 import classnames from 'classnames';
 import PropTypes from 'prop-types';
 import Box from './Box.js';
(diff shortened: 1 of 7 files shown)


packages/gestalt/src/Button.js
```diff
@@ -1,6 +1,6 @@
 // @flow strict
 
-import React, { useRef, type Element } from 'react';
+import * as React from 'react';
```
christianvuerings, line 3: @AlbertCarreras curious, what was the reason for this change?

AlbertCarreras: I was trying to stick to one pattern in particular. TBDiscussed

PR #1066, opened 2020-07-21
Docs: fix broken CodeSandbox links

jackhsu978:
This PR fixes CodeSandbox links in docs, which do not work since #1058

before
Screenshot 2020-07-21 13 37 06

after
Screenshot 2020-07-21 13 38 05

## Test Plan

manually tested on localhost

### DIFF
--- a/docs/src/components/Example.js
+++ b/docs/src/components/Example.js
@@ -32,20 +32,20 @@ const exportDefaultMaybe = ({ code }) =>
 
 export default Demo;`;
 
+const dedupeArray = <T>(arr: Array<T>): Array<T> => [...new Set(arr)];
+
 const handleCodeSandbox = async ({ code, title }) => {
   const gestaltComponents = Object.keys(gestalt);
   const additionalGestaltComponents = ['DatePicker'];
 
-  const usedComponents = [
-    ...new Set(
-      code.match(/<((\w+))/g).map(component => component.replace('<', ''))
+  const usedComponents = dedupeArray([
+    ...(code.match(/<((\w+))/g) || []).map(component =>
+      component.replace('<', '')
     ),
-    ...new Set(
-      code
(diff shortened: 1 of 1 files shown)


docs/src/components/Example.js
```diff
@@ -32,20 +32,20 @@ const exportDefaultMaybe = ({ code }) =>
...
 
+const dedupeArray = <T>(arr: Array<T>): Array<T> => [...new Set(arr)];
+
 const handleCodeSandbox = async ({ code, title }) => {
   const gestaltComponents = Object.keys(gestalt);
   const additionalGestaltComponents = ['DatePicker'];
 
-  const usedComponents = [
-    ...new Set(
-      code.match(/<((\w+))/g).map(component => component.replace('<', ''))
```
jackhsu978, line 41: `match()` call can return `null`

docs/src/components/Example.js
```diff
@@ -32,20 +32,20 @@ const exportDefaultMaybe = ({ code }) =>
...
-    ),
-    ...new Set(
-      code
-        .match(/(new FixedZIndex)|(new CompositeZIndex)/g)
```
jackhsu978, line 45: `match()` call can return `null`

docs/src/components/Example.js
```diff
@@ -32,20 +32,20 @@ const exportDefaultMaybe = ({ code }) =>
 
 export default Demo;`;
 
+const dedupeArray = <T>(arr: Array<T>): Array<T> => [...new Set(arr)];
+
 const handleCodeSandbox = async ({ code, title }) => {
   const gestaltComponents = Object.keys(gestalt);
   const additionalGestaltComponents = ['DatePicker'];
 
-  const usedComponents = [
```
jackhsu978, line 39: we are concatenating two deduped arrays. I think it makes more sense to dedupe _after_ concatenating two arrays.

PR #1067, opened 2020-07-22
Typeahead: Refactor to show all options when there's a defaultValue, renamed defaultItem + removed searchField prop, fix forwardedRef + test coverage, add example to Docs, 

simkessy:
- Changes one default value behavior:
  - When the Typeahead has a default value, the initial click into the component shows only the filtered items based on the initial value
  - The change with make it so when the Typeahead has an item, this item is not used to filter the initial set of results
- Renamed the prop `defaultItem` to just `value` and `data` to `options`
- Removed `searchField` prop
- Added  Example for ref prop in Docs. 
- Added test to cover forwardRef (ref prop) functionality. 
- Refactored internal code for cleanup and small fixes.

## Test Plan

- Original functionality: https://gestalt.netlify.app/#/components/Typeahead 
    - In the second example when you click in the Typeahead, only `Label-4` is displayed 
- In the update, all labels should be displayed when Typeahead is clicked on despite `item` being set
    - https://deploy-preview-1067--gestalt.netlify.app/#/components/Typeahead

### DIFF
--- a/docs/src/Typeahead.doc.js
+++ b/docs/src/Typeahead.doc.js
@@ -26,7 +26,7 @@ card(
         href: 'basicExample',
       },
       {
-        name: 'defaultItem',
+        name: 'item',
         type: '{ label: string, value: string }',
         description: 'The default item set in the Typeahead',
         required: false,
@@ -192,7 +192,7 @@ function Example(props) {
         id="Typeahead-example-defaultItem"
         noResultText="No Results"
         data={options}
-        defaultItem={defaultOption}
+        item={defaultOption}
         placeholder="Select a Label"
         onChange={handleOnChange}
         value={item}
--- a/docs/src/Typeahead.doc.js
+++ b/docs/src/Typeahead.doc.js
@@ -75,7 +75,7 @@ card(
       },
       {
         name: 'onSelect',
-        type: '({ event: SyntheticInputEvent<>, item: Object }) => void',
+        type: '({ event: SyntheticInputEvent<>, value: string }) => void',
         required: false,
         description: 'Callback when you select an item ',
         href: 'basicExample',
@@ -94,7 +94,7 @@ card(
       },
       {
         name: 'ref',
-        type: "React.Ref<'div'>",
+        type: "React.Ref<'input'>",
         description:
           'Forward the ref to the underlying component container element',
         href: 'refExample',
@@ -188,6 +188,38 @@ function Example(props) {
         onSelect={handleSelect}
       />
     </>
+  );
+}`}
+  />
+);
+
+card(
+  <Example
+    id="defaultItemExample2"
+    name="Ref Example"
+    defaultCode={`
+function TypeaheadExample() {
+  const ref = React.useRef();
+  const [option, setOption] = React.useState();
+  return (
+    <Row gap={2}>
+      <Typeahead
+        label="Select  your favorite shape"
+        id="favorite-shape"
+        noResultText="No Results"
+        options={[{label:'square', value:'square'}, {label:'circle', value:'circle'}]}
+        onSelect={p => ref.current.focus()}
+        placeholder="Select a shape"
+      />
+       <Typeahead
+        label="Select  your favorite color"
+        id="favorite-color"
+        noResultText="No Results"
+        options={[{label:'red', value:'red'}, {label:'blue', value:'blue'}]}
+        placeholder="Select a color"
+        ref={ref}
+     />
+    </Row>
   );
 }`}
   />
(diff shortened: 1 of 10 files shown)


ashleyseto:
Approved from design side

docs/src/Typeahead.doc.js
```diff
@@ -26,7 +26,7 @@ card(
         href: 'basicExample',
       },
       {
-        name: 'defaultItem',
```
christianvuerings, line 29: @simkessy this would be a semver major change. If we do want to change the name, could you add a codemod so we (and others) can easily upgrade gestalt in a repo?

A good example is https://github.com/pinterest/gestalt/blob/master/packages/gestalt-codemods/1.8.0-1.9.0/heading-size-replace.js

packages/gestalt/src/Typeahead.js
```diff
@@ -1,3 +1,5 @@
+/* eslint-disable eslint-comments/disable-enable-pair */
```
christianvuerings, line 1: @simkessy could we remove these 2 eslint disable comments?

If it's still required, please use `eslint-disable-next-line` or `eslint-disable-line` instead to make it specific to a line.

AlbertCarreras:
@simkessy Hey Kessy, if you're going to change the name of the prop -codemod included- I suggest considering `value`. We should keep consistency across all components and most components use that naming for that specific funcitonality. The ones you mentioned, plus Searchfield, DatePicker, etc. 

For our Gestalt users, having a consistent API is better so that they don't have to remember the specific naming of props for different components. `value` prop behavior should be predictable.
 
I think it's a good change, but I think we should go with `value` rather than `item`.

AlbertCarreras:
K:
_Hey I was thinking value as well but my concern was that we were returning an object
So we'd end up with something like value.value or value.label_

A:
What about `options` + `value`
`SelectList` uses `options` instead of `data`
Could you make `item` a string instead?

I’m checking `SelectList`;  `Typeahead`  is a searchable `SelectList`  so they should have more or less the same API. You can probably check the implementation there to use `value` (`defaultItem`) as a key to match it  in `options` (`data`)

AlbertCarreras:
I think the default option is still the only one showing.

AlbertCarreras:
nit change

could you keep `Typeahead` prefix on both or none of them? Just to stay consistent

`
import TypeaheadInputField from './TypeaheadInputField.js';
import Option from './TypeaheadOption.js';
`

AlbertCarreras:
Inside `handleKeyNavigation` image

AlbertCarreras:
These lines. , you could just out a 0.
image

AlbertCarreras:
Flow cleanup here:
image

AlbertCarreras:
You can simplify handlers:

image
image
image

AlbertCarreras:
Can you give a use case for searchField = 'value'?  In my brain, this seems counterintuitive*, but I guess this prop is here for some reason, I would add an example of a use case with searchField = 'value'  Otherwise, I would remove it. *Because you are only going to search base on the displayed label, not the hidden value that the user doesn’t see. 

image

AlbertCarreras:
findDefaultOption

remove the eslint disable there, just change the name of the argument to `defaultOption`

image

AlbertCarreras:
`= id`  is dead code
image

AlbertCarreras:
We ccan simplify this

1.  image
2. 
image

AlbertCarreras:
I think the ref from the API is targeting the outer Box, try the solutions we've implemented in TapArea, Checkbox, RadioButton...
image

AlbertCarreras:
nit refactor 
image

AlbertCarreras:
image
 

add forwardRef

image


check Checkbox, etc

AlbertCarreras:
on TypeaheadInputField

no need for if statement

image

AlbertCarreras:
image

AlbertCarreras:
nit 
image

AlbertCarreras:
could you use `IconButton` here? play with size + padding to find the right ratio between icon size and total size.

image

AlbertCarreras:
image

AlbertCarreras:
no need for `direction="row"`, it's the default value
image

AlbertCarreras:
darkMode
image

AlbertCarreras:
nit edge bug. https://www.loom.com/share/6fff552b3a90478eb68ee692c79bbec7

AlbertCarreras:
New changes LGTM!

packages/gestalt-codemods/11.0.0-12.0.0/typeahead-update-props.js
```diff
@@ -0,0 +1,65 @@
+/*
+ * Log an error when a `ref` is specified on `<TextArea />`
```
christianvuerings, line 2: @simkessy could we update the description of this codemod?

PR #952, opened 2020-06-26
Docs: Refactored Gestalt Docs Sidebar Menu

AlbertCarreras:
- Refactored sidebar menu implementing:

1. grouping by component type
2. menu sections (such as `Getting Started` and `Components`)
3. sidebar menu responsiveness

## TODO

- [-] Accessibility checkup
- [x] Update documentation
- [-] Add/update Tests
- [-] Pinterest Designer (for design changes)
- [x] [Format PR title](https://github.com/pinterest/gestalt/#releasing): `ComponentName: Description`.

PR #1072, opened 2020-07-23
Docs: remove sticky header

christianvuerings:
The `<Sticky />` header causes issues in our docs. It was introduced as part of #952 /cc @AlbertCarreras 

## Before
Screen Shot 2020-07-23 at 2 22 06 PM

## After
Screen Shot 2020-07-23 at 2 21 26 PM

### DIFF
--- a/docs/src/components/Header.js
+++ b/docs/src/components/Header.js
@@ -7,7 +7,6 @@ import {
   IconButton,
   Tooltip,
   Link as GestaltLink,
-  Sticky,
 } from 'gestalt';
 import DocSearch from './DocSearch.js';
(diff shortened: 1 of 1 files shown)


AlbertCarreras:
I don't see the issue. What's the issue?  The example looks like `Layout`
image

AlbertCarreras:
What do you think about an intermediate solution?  https://github.com/pinterest/gestalt/pull/1074
I personally find pretty helpful the sticky header.

christianvuerings:
@AlbertCarreras it's a bit magic but I'm ok with it for now. Going to close this PR

PR #1073, opened 2020-07-23
useFocusVisible: Only show focus ring on keyboard input

christianvuerings:
Adds `useFocusVisible` a React hook we can use in all of our components to only show the blue outline (focus ring) when using the keyboard.

In this first PR we only apply it to `<Button />`

focus-outline-keyboard-only

### Click on a `<Button />` and don't see a focus ring:
useFocusVisible-mouse

### Use the `TAB` key to get to a button and do see the focus ring:
useFocusVisible-keyboard

### DIFF
--- a/packages/gestalt/src/Button.css
+++ b/packages/gestalt/src/Button.css
@@ -5,11 +5,9 @@
 
 .button {
   composes: borderBox minWidth60 from "./Layout.css";
-  border-radius: 24px;
-}
-
-.solid {
   composes: noBorder from "./Borders.css";
+  border-radius: 24px;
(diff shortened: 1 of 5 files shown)


AlbertCarreras:
@christianvuerings Interesting. is this a standard?

christianvuerings:
> @christianvuerings Interesting. is this a standard?

Yes, this is https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible - however, since browser support is so low, we have to implement this in JavaScript right now.

https://caniuse.com/#search=focus%20visible

PR #1074, opened 2020-07-24
Docs: Fix Header to remove sticky header on reduced window size

AlbertCarreras:
The `<Sticky />` header causes issues in our docs. It was introduced as part of #952 /cc @AlbertCarreras 

However,  it is convenient to have a sticky header on Gestalt Docs due to the amount of scrolling that is required when reading the Docs. 

A proposed solution is to keep the sticky header on full windows and remove it when windows downsize to a size that doesn't allow the correct display of Modals.

## Before
Screen Shot 2020-07-23 at 2 22 06 PM

## After
Kapture 2020-07-23 at 18 19 17

### DIFF
--- a/docs/src/components/Header.js
+++ b/docs/src/components/Header.js
@@ -18,7 +18,7 @@ type Props = {|
   onChangeColorScheme: () => void,
 |};
 
-export default function Header({ colorScheme, onChangeColorScheme }: Props) {
+function Header({ colorScheme, onChangeColorScheme }: Props) {
   const [isRTL, setIsRTL] = React.useState(false);
   const { isSidebarOpen, setIsSidebarOpen } = useSidebarContext();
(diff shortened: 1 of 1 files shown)


docs/src/components/Header.js
```diff
@@ -35,105 +35,134 @@ export default function Header({ colorScheme, onChangeColorScheme }: Props) {
...
       </Box>
+    </Box>
+  );
+}
+
+export default function StickyHeader({
+  colorScheme,
+  onChangeColorScheme,
+}: Props) {
+  const isReducedHeight = () => window.innerHeight < 709;
```
christianvuerings, line 143: @AlbertCarreras 709 is definitely a magic number, I'm ok with this approach for now but it will break for other use cases.

PR #1076, opened 2020-07-24
Callout: Add component

norrise120:
## Test Plan
Wrote tests for component, matching pattern of testing for other components. 
Updated docs and verify that docs looked as expected.

Screen Shot 2020-07-24 at 10 10 24 AM

### DIFF
--- a/docs/src/Callout.doc.js
+++ b/docs/src/Callout.doc.js
@@ -1,0 +1,100 @@
+// @flow strict
+import * as React from 'react';
+import PropTable from './components/PropTable.js';
+import Example from './components/Example.js';
+import PageHeader from './components/PageHeader.js';
+
+const cards = [];
+const card = c => cards.push(c);
+
+card(
+  <PageHeader
+    name="Callout"
+  />
+);
+
+card(
+  <PropTable
+    props={[
+      {
+        name: 'description',
+        type: 'string',
+        required: true,
+      },
+      {
+        name: 'iconAccessibilityLabel',
+        type: 'string',
+        required: true,
+      },
+      {
+        name: 'onDismiss',
+        type: '() => void',
+        required: true,
+      },
+      {
+        name: 'primaryLink',
+        type: '{| href: string, label: string |}',
+      },
+      {
+        name: 'secondaryLink',
+        type: '{| href: string, label: string |}',
+      },
+      {
+        name: 'style',
+        type: `"error" | "info" | "warning"`,
+        required: true,
+      },
+      {
+        name: 'title',
+        type: 'string',
+      },
+    ]}
+  />
+);
+
+card(
+  <Example
+    name="Info Example"
+    defaultCode={`
+<Callout
+    style="info" 
+    iconAcessibilityLabel=""
+    title="The thing you were doing is now done"
+    description="Get a badge, show up in more experiences and more."
+    primaryLink={{href: "https://pinterest.com", label:"Get started"}} 
+    secondaryLink={{href: "https://pinterest.com", label:"Learn more"}} 
+  />
+`}
+  />
+);
+
+card(
+  <Example
+    name="Warning Example"
+    defaultCode={`
+<Callout
+  style="warning" 
+  iconAcessibilityLabel=""
+  onDismiss={()=>{}}
+  description="This feature will be removed in two weeks."
+  primaryLink={{href: "https://pinterest.com", label:"Learn more"}}
+/>
+  `}
+  />
+);
+
+card(
+  <Example
+    name="Error Example"
+    defaultCode={`
+<Callout
+  style="error" 
+  iconAcessibilityLabel=""
+  onDismiss={()=>{}}
+  description="Oops"
+/>
+  `}
+  />
+);
+
+export default cards;
--- a/docs/src/Callout.doc.js
+++ b/docs/src/Callout.doc.js
@@ -18,18 +18,12 @@ card(
         required: true,
       },
       {
-        name: 'dismissIconAccessibilityLabel',
-        type: 'string',
-        required: true,
+        name: 'dismissButton',
+        type: '{| accessibilityLabel: string, onDismiss: () => void, |}',
       },
       {
         name: 'iconAccessibilityLabel',
         type: 'string',
-        required: true,
-      },
-      {
-        name: 'onDismiss',
-        type: '() => void',
         required: true,
       },
       {
@@ -61,12 +55,15 @@ card(
     defaultCode={`
 <Callout
   style="info" 
-  dismissIconAcessibilityLabel="Dismiss banner"
   iconAcessibilityLabel="Info icon"
   title="The thing you were doing is now done"
   description="Get a badge, show up in more experiences and more."
   primaryLink={{href: "https://pinterest.com", label:"Get started"}} 
-  secondaryLink={{href: "https://pinterest.com", label:"Learn more"}} 
+  secondaryLink={{href: "https://pinterest.com", label:"Learn more"}}
+  dismissButton={{
+    accessibilityLabel: 'Dismiss banner',
+    onDismiss: ()=>{},
+  }}
   />
 `}
   />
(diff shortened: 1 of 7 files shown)


docs/src/components/sidebarIndex.js
```diff
@@ -2,14 +2,14 @@
 
 export type sidebarIndexType = Array<{|
   sectionPathname: string,
-  displayName: string,
-  pages?: Array<string>,
-  subsections?: Array<{| [string]: Array<string> |}>,
+    displayName: string,
+      pages ?: Array < string >,
+      subsections ?: Array < {| [string]: Array<string> |}>,
```
norrise120, line 7: My editor kept doing this when I saved. I'll update when I address feedback and make sure to remove these changes

packages/gestalt/src/Callout.js
```diff
@@ -0,0 +1,111 @@
...
+        href: string,
+          label: string,
+  |},
+onDismiss: () => void,
+  secondaryLink ?: {|
+    href: string,
+      label: string,
+  |},
+style: 'error' | 'info' | 'warning',
+  title ?: string,
```
norrise120, line 24: My editor did this whenever I saved. I'll update to make sure it's formatted correctly when I address feedback

docs/src/Callout.doc.js
```diff
@@ -0,0 +1,100 @@
...
+    ]}
+  />
+);
+
+card(
+  <Example
+    name="Info Example"
+    defaultCode={`
+<Callout
+    style="info"
```
christianvuerings, line 60: @norrise120 could you enforce 2 spaces indentation? (just like you did in the example below)

docs/src/Callout.doc.js
```diff
@@ -0,0 +1,100 @@
...
+  />
+);
+
+card(
+  <Example
+    name="Info Example"
+    defaultCode={`
+<Callout
+    style="info" 
+    iconAcessibilityLabel=""
```
christianvuerings, line 61: @norrise120 let's fill this out, use something like `Dismiss banner`

docs/src/Callout.doc.js
```diff
@@ -0,0 +1,100 @@
...
+        required: true,
+      },
+      {
+        name: 'onDismiss',
+        type: '() => void',
+        required: true,
+      },
+      {
+        name: 'primaryLink',
+        type: '{| href: string, label: string |}',
```
christianvuerings, line 36: @norrise120 could we also add an `onClick` handler? That way we can integrate these links with react-router in the Pinterest code base.

Try to use the same type for `onClick` as we use on `<Link />`:

```
  onClick?: AbstractEventHandler<
```

packages/gestalt/src/Callout.js
```diff
@@ -0,0 +1,111 @@
...
+|};
+
+const CALLOUT_TYPE_ATTRIBUTES = {
+  info: {
+    icon: 'check-circle',
+    color: 'blue',
+    backgroundColor: '#EBF4FE',
```
christianvuerings, line 31: @ashleyseto @anniemochi could we spec out what these colors should be in dark mode?

packages/gestalt/src/Callout.test.js
```diff
@@ -0,0 +1,84 @@
...
+    const tree = create(
+      <Callout
+        description="Insert a clever info callout message here"
+        iconAccessibilityLabel="info icon"
+        onDismiss={() => { }}
+        style="info"
+        title="A Title"
+      />
+    ).toJSON();
+    expect(tree).toMatchSnapshot();
```
christianvuerings, line 53: @norrise120 thanks for adding these tests!

packages/gestalt/src/Callout.test.js
```diff
@@ -0,0 +1,84 @@
...
+        onDismiss={() => { }}
+        primaryLink={{ href: "pinterest.com", label: "Visit Pinterest" }}
+        secondaryLink={{ href: "pinterest.com/help", label: "Learn more" }}
+        style="info"
+        title="A Title"
+      />
+    ).toJSON();
+    expect(tree).toMatchSnapshot();
+  });
+});
```
christianvuerings, line 84: @norrise120 could we also add a `.jsdom` test which tests the dismiss functionality?

ashleyseto:
Hey. Thanks for taking this on! A few things that need adjustment based on spec:

- The header (bold text) should be heading small
- Icon size should be 32px
- There is some padding around the icon button to close. Currently it is 16px in spec but we might need to adjust slightly on how it looks in code.
- 8px of space should be between the buttons for the example
- I'd like one of the examples to be without an x to close. We have the info, error, warning and the minimal option without an x.

packages/gestalt/src/Callout.js
ashleyseto: At the moment, there is not a dark mode spec for this as it is a non-standard color so I would leave it identical in dark mode to ship.

ashleyseto:
Tagging in @ajmoo who did the original design spec for a final pass. 

One other adjustment, the text block is 24px from the icon and the full width of the text block is max 600px.

ajmoo:
aside from what @ashleyseto called out this looks good to me. "oops" is my favorite.

packages/gestalt/src/Callout.js
christianvuerings: This is what currently happens in dark mode:
image

So we do need to figure out a way forward. Right now the text colors and everything else automatically updates. The easiest way forward would be to have dark mode specific colors for these non-standard colors.

ashleyseto: Looks like more than just color isn't right in that above screenshot. My immediate recommendation since dark mode was not in scope when this was designed would be that it would look identical in dark and light mode much like primary buttons are the same in both.

ashleyseto:
Screen Shot 2020-07-27 at 8 14 22 AM

Hey a few adjustments.

- The 24px was meant to be in addition to the 600px specified for the text box width. Sorry that wasn't clear. The total space for text and padding of that box then would be 648px
- The buttons are still meant to be aligned to the right of the callout as the component grows in size.

ashleyseto:
Screen Shot 2020-07-27 at 9 17 55 AM

Chatted with Taylor and it seems we need to add a different look to for the gestalt-sm breakpoint for m-web. The buttons need to shift under the textbox and the icon would align to the top. Added to the Figma as well.

ashleyseto:
@taylorannsimpson for visibility

norrise120:
Updated darkmode so that text is readable. I left the button untouched, but I can update to make it match the lightmode version exactly if preferred. 
Screen Shot 2020-07-27 at 11 07 49 PM

> Screen Shot 2020-07-27 at 9 17 55 AM
> 
> Chatted with Taylor and it seems we need to add a different look to for the gestalt-sm breakpoint for m-web. The buttons need to shift under the textbox and the icon would align to the top. Added to the Figma as well.

If ok with everyone, I would prefer to handle the mweb updates in a different PR later this week/early next week.

christianvuerings:
> Updated darkmode so that text is readable. I left the button untouched, but I can update to make it match the lightmode version exactly if preferred.
> Screen Shot 2020-07-27 at 11 07 49 PM
> 
That's totally fine

packages/gestalt/src/Callout.js
```diff
@@ -0,0 +1,159 @@
...
+              {title}
+            </Heading>
+          </Box>
+        )}
+        <Text color={isDarkMode ? 'white' : 'darkGray'}>{description}</Text>
+      </Box>
+      {secondaryLink && (
+        <Box paddingX={1} marginStart="auto">
+          <Box padding={4} flex="none">
+            <Link href={secondaryLink.href}>
```
christianvuerings, line 107: @norrise120 could we add the `onClick` we passed through? `onClick={secondaryLink.onClick}`

ashleyseto:
One other minor change. The padding on the right where the buttons are should be 48px.

Current
Screen Shot 2020-07-28 at 11 08 00 AM

Desired
Screen Shot 2020-07-28 at 11 07 50 AM

We'll handle the responsive part of the component in a separate PR then. Thanks :)

AlbertCarreras:
Notes:

**API**
- is iconAccessibilityLabel needed? I've seen other components not adding accessibility labels to internal icons.
- should primaryLink focus the box instead of the link? It's a confusing linkButton.
- primaryLink/secondaryLin, should they allow target?

**Docs**
- We need a brief description

PR #1077, opened 2020-07-26
Video: Fix failing CORS access and preventing playback 

simkessy:
Video's where sources where from crossorigins failed to load. 

Additional information how cross-origin behaves:

image

https://stackoverflow.com/questions/41069330/with-script-crossorigin-anonymous-why-is-a-script-blocked-by-cors-policy

### DIFF
--- a/packages/gestalt/src/Video.js
+++ b/packages/gestalt/src/Video.js
@@ -525,7 +525,6 @@ export default class Video extends React.PureComponent<Props, State> {
         <ColorSchemeProvider id="Video" colorScheme="light">
           <video
             autoPlay={playing}
-            crossOrigin="anonymous"
             loop={loop}
             muted={volume === 0}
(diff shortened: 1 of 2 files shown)


christianvuerings:
This bug was introduced in https://github.com/pinterest/gestalt/pull/1046#discussion_r454048460

@jesselumarie Heads up - this will break the previous subtitles fix you made at #1046 but using `anonymous` broke playback for videos.

Moving forward, we should host subtitles on the same domain as the video OR work with traffic to proxy them through the same domain.

chrislloyd:
@christianvuerings is there an integration test or similar that we could add to prevent this regression happening again?

christianvuerings:
@chrislloyd yes, definitely. One thing which would also be useful is to detect dead links in our docs.

PR #1083, opened 2020-07-28
Icon: Add drag-drop icon

ladyd252:
Adding a new icon as required for Board Tools feature.

## Test Plan

Checked documentation on localhost and confirmed new icon appears in the Icon section.

Screen Shot 2020-07-28 at 11 50 25 AM


packages/gestalt/src/icons/drag-drop.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="24" height="24" xmlns="http://www.w3.org/2000/svg">
+<path d="M0 10.3636V21.2727C0 22.7793 1.22073 24 2.72727 24H13.6364C15.1429 24 16.3636 22.7793 16.3636 21.2727V16.3636H21.2727C22.7793 16.3636 24 15.1429 24 13.6364V2.72727C24 1.22073 22.7793 0 21.2727 0H10.3636C8.85709 0 7.63636 1.22073 7.63636 2.72727V7.63636H2.72727C1.22073 7.63636 0 8.85709 0 10.3636ZM2.72727 10.3636H7.63636V13.6364C7.63636 15.1429 8.85709 16.3636 10.3636 16.3636H13.6364V21.2727H2.72727V10.3636Z" fill="#767676"/>
```
AlbertCarreras, line 2: hey. Dana, can we remove `fill="#767676"`. The fill doesn't matter that much since we only use the path in from the SVG.

See 
https://github.com/pinterest/gestalt/blob/a6f7152ed194c68d93d4e8ef626755517d132865/packages/gestalt/rollup.config.js#L36

and https://github.com/pinterest/gestalt/pull/916/commits/048209a40f021486b3ec20ac1413307529bd39b6

AlbertCarreras:
left comment but LGTM

packages/gestalt/src/icons/drag-drop.svg
ladyd252: Updated to remove fill!

christianvuerings: @AlbertCarreras This is actually incorrect - we do want the `fill` (also see our other icons) since it should still show up similar to our other icons when we look at the individual SVGs.

I updated the PR + optimized the SVG

PR #1084, opened 2020-07-28
Layer: Fix flow type return on render

jonluca:


chrislloyd:
@jonluca can you change the return of this so it's either `?Portal`? Boolean here seems like more of a bug than a feature.

jonluca:
> @jonluca can you change the return of this so it's either `?Portal`? Boolean here (because of the `&&` short circuit) seems like more of a bug than a feature.

Agreed. I'll have it return `React.Portal | null`

PR #1085, opened 2020-07-29
Flowtest: Add flow specific test files

christianvuerings:
Instead of add flow tests to `.test.js` files and having to add a useless `expect` statement, let's create flow specific test files

## Test Plan

* Flow passed on CI

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -88,6 +88,12 @@
       }
     },
     {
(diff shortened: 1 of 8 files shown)


christianvuerings:
@bishwei would you be able to take a look? Made the change to `flowtest.js` files after your feedback.

Issue #1086, opened 2020-07-29
Box: marginEnd/marginStart do not support responsive  sizes sm, md, lg


AlbertCarreras:
![image](https://user-images.githubusercontent.com/10593890/88840284-14f0a600-d191-11ea-9214-4c27515d280a.png)
![image](https://user-images.githubusercontent.com/10593890/88840293-191cc380-d191-11ea-84ab-8a7512ee3eff.png)

![image](https://user-images.githubusercontent.com/10593890/88840367-2fc31a80-d191-11ea-9cff-fc454ab9e58b.png)

![image](https://user-images.githubusercontent.com/10593890/88840307-1f12a480-d191-11ea-8d4c-2ec51b945080.png)

PR #1087, opened 2020-07-29
Box: added support to responsive marginEnd marginStart

AlbertCarreras:
Fixes https://github.com/pinterest/gestalt/issues/1086

Added support to responsive `marginEnd` `marginStart` for sm/md/lg window size

### Before

image

### After 
Kapture 2020-07-29 at 13 14 29
Kapture 2020-07-29 at 13 27 49


chrislloyd:
Do we already ship the CSS helpers for these properties?

PR #1088, opened 2020-07-30
Callout: Added responsiveness design

AlbertCarreras:
### Figma Design
image

### Before

image

### After

image

### DIFF
--- a/docs/src/Callout.doc.js
+++ b/docs/src/Callout.doc.js
@@ -94,7 +94,7 @@ card(
 <Callout
   type="error"
   iconAccessibilityLabel="Error icon"
-  description="Oops"
+  description="This action can't be undone."
 />
   `}
   />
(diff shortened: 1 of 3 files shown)


ashleyseto:
Assuming this is just for the responsiveness. Looking good! Only one small thing is on the smaller size we dropped the internal padding from 32px to 24px.

AlbertCarreras:
@ashleyseto I checked the margins. Let me know if I'm missing something.
Kapture 2020-07-31 at 11 24 23

christianvuerings:
@AlbertCarreras lgtm - not specific to your diff but one thing I really wish we had is something like https://github.com/marcj/css-element-queries . Instead of relying on the browser window sizes it relies on the element size

ashleyseto:
I saw this was merged but the design hadn't been updated. I'm still seeing the content at 32px

Screen Shot 2020-07-31 at 4 38 37 PM

PR #1089, opened 2020-07-30
[Dark Mode] Add hover/active color variables

jennyscript:
This sets up css variables for most of the remaining colors so they can be overridden in dark mode and accessed in apps using gestalt to implement custom components with the same palette. I've added hover and active versions of the existing colors that are used on `Button` and `Pog`. These are now also available in context through `useColorScheme`. I used the existing values for light mode.

Because I didn't have dark mode color equivalents I made these up. **Design review** for these colors would be great! These just need to be "good enough" for now until we have more official colors. This will allow us to start adding dark mode support internally in the main repo before exposing this to end users.

The only component that still has hardcoded hex colors in the css file is `Pulsar`.

## Test Plan

I've added the rest of the colors to `Button` docs to test these new colors out.


packages/gestalt/src/FormElement.css
```diff
@@ -4,7 +4,7 @@
 
   /* hover colors */
   --gestalt-colorGray150Hovered: #d0d0d0;
-  --gestalt-colorRed100Hovered: #d80021;
+  --gestalt-colorRed100Hovered: #ad081b;
```
jennyscript, line 7: This one was already in here with the new name. I went with the same red for hover both here and the button for consistency but could break it out again if this won't work.

packages/gestalt/src/contexts/ColorScheme.js
```diff
@@ -35,34 +45,54 @@ const lightModeTheme = {
...
   colorGray150: '#585858',
   colorGray200: '#ababab',
+  colorGray200Active: '#9b9b9b',
+  colorGray200Hovered: '#919191',
   colorGray300: '#efefef',
   colorGray400: '#fff',
   colorTransparentDarkGray: 'rgba(255, 255, 255, 0.8)',
   colorTransparentGray60: 'rgba(250, 250, 250, 0.5)',
-  colorTransparentGray100: 'rgba(0, 0, 0, 0.5)',
+  colorTransparentGray100: 'rgba(250, 250, 250, 0.6)',
```
jennyscript, line 92: I noticed this color looked bad during testing and I think it's not in the official palette anyway

christianvuerings:
Arg - looks like mergify merged without checking for the release type. @chrislloyd any idea why this might be going on?
image

ashleyseto:
@ajmoo Said he could help out here.

Looks like we need a darkmode focus color as well.

chrislloyd:
Sorry about that folks! Fixed above. Next PR merged should create a release with this commit in it.

ajmoo:
I've got some notes so far…

- Secondary (lightGray) `Buttons` should be #404040, hover state #535353 and active state #666666
- `IconButton` should be #535353 hover state and #666666 active state
- Noticing a bug with `button`: after clicking, the hover state color doesn't clear.
- I don't love the blue/red+dark text pairing on , but don't know how to resolve this one right now. cc @anniemochi 
- The white shadows on `DatePicker` and elsewhere look off to me. @anniemochi is it okay to keep shadows black in dark mode?
- Should the accessibility focus indicator have transparency? Looks wrong against a dark background. How does #9CC2FB look? @anniemochi 
- hover state for `DatePicker` should be #535353 to align with button hover state
- Are you addressing `SelectList`, `TextField`, `TextArea`, and `SearchField` in this PR?
- `SegmentedControl` doesn't have a hover state. @anniemochi is that ok? More noticeable in dark mode.
- `Tab` hover state should be #535353

ajmoo:
Oh yeah, and our Primary red `button`…
#CF001F hover
#B8001C active

anniemochi:
Responding to the notes @ajmoo tagged me in…

1. Could we keep white text on a blue or red background?

anniemochi:
Responding to the notes @ajmoo tagged me in…

- Could we keep white text on a blue or red background?
- Yep, use black for the `DatePicker` shadow
- Focus indicator is looking good
- I think the `SegmentedControl` should have a hover state, but that should can addressed separately from this PR (cc @ashleyseto)

PR #1091, opened 2020-07-30
Box: Added responsiveness to sm/md/lg 0px margins

AlbertCarreras:
## Why supporting 0 boint values for responsive margins

image
- `marginStart`, `marginEnd`, `marginBottom`, `marginTop`, `marginRight`, `marginLeft` accept values in the range -12 < 0 < 12 boints.
- `0` boint values were skipped and NOT mapped to CSS variables.

- However,  responsive margins `[sm/md/lg]margin[Start, End, Bottom/Top/Right/Left]` are used to overwrite the default margin. By ignoring `0` boint values, there was no solution to remove margins upon resizing.

During instance, the following example would work

```
margin={12}
mdMargin={1}
```
but the following example would NOT work

```
margin={12}
mdMargin={0}
```

### Changes

- Added mapped classnames for responsive margins `[sm/md/lg]margin[Start, End, Bottom/Top/Right/Left]` (**including** defaults `margin[Start/End/Bottom/Top/Right/Left]`) 
- Fixed & Updated script to autogenerate `packages/gestalt/src/boxWhitespace.css`
- Updates snapshots for components internally using 0-boint values on margins.
image
image

### Before
Kapture 2020-07-30 at 0 43 30

### After
Kapture 2020-07-30 at 0 42 06


packages/gestalt/src/Box.js
```diff
@@ -279,7 +280,9 @@ const transformNumberOrPassthrough = (selector: string) => (
   m: Margin
 ): Style => {
   if (typeof m === 'number') {
-    return bind(rangeWithoutZero(selector), whitespace)(m);
+    return /^sd|md|lg/.test(selector)
```
christianvuerings, line 283: @AlbertCarreras does this need to be `sm` instead of `sd`? Would be great to add a few tests.

Also, do we need the regex check? Can we just do `return bind(rangeWithZero(selector), whitespace)(m);` which would also allow `margin={0}` to be supported?

AlbertCarreras: Following our previous conversation, passing all instances to **rangeWithZero** to support default margins with 0 point values.

packages/gestalt/src/__snapshots__/Tabs.test.js.snap
```diff
@@ -2,7 +2,7 @@
 
 exports[`<Tabs /> matches snapshot with default props 1`] = `
 <div
-  className="box"
+  className="box marginEnd0 marginStart0"
```
christianvuerings, line 5: TIL looks like this is happening because we use `<Row />` in `<Tabs />`. (no changes required)

AlbertCarreras: Yeah, left comment in changes.

PR #1096, opened 2020-07-30
Flowtests: Add more tests to components + fix GestaltProvider flowtype

christianvuerings:
Add more flow tests to our Gestalt components.

We did find an issue with `GestaltProvider` where the flowtype was incorrect.

### DIFF
--- a/packages/gestalt/src/Card.flowtest.js
+++ b/packages/gestalt/src/Card.flowtest.js
@@ -4,5 +4,5 @@ import Card from './Card.js';
 
 const Valid = <Card />;
 
-// $FlowExpectedError[prop-missing] Cannot create `Card` element because property `nonexisting` is missing in  props.
+// $FlowExpectedError[prop-missing]
 const NonExistingProp = <Card nonexisting={33} />;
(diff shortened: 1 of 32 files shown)


packages/gestalt/src/GestaltProvider.js
```diff
@@ -10,7 +10,7 @@ import {
 type Props = {|
   children: React.Node,
   colorScheme?: ColorScheme,
-  id: ?string,
+  id?: string,
```
christianvuerings, line 13: @jennyscript fyi

Discovered this issue when trying to type `<GestaltProvider />` without an id.

packages/gestalt/src/Card.flowtest.js
```diff
@@ -4,5 +4,5 @@ import Card from './Card.js';
 
 const Valid = <Card />;
 
-// $FlowExpectedError[prop-missing] Cannot create `Card` element because property `nonexisting` is missing in  props.
+// $FlowExpectedError[prop-missing]
```
jennyscript, line 7: Just wondering why remove the additional info from the comments here?

christianvuerings: @jennyscript will be too cumbersome to maintain - I looked at flow-typed where they also moved away from adding these extra comments.

PR #1098, opened 2020-07-30
Button/Checkbox/IconButton/Link/RadioButton/Switch/TapArea: useFocusVisible

christianvuerings:
Follow up from #1073 where we only show the blue outline (focus ring) when using the keyboard.

Other changes:
* Convert Switch to React hooks (in order to use `useFocusVisible`)
* TapArea: no longer prevent bubbling up /cc @jackhsu978 in #906 

## Test Plan

For each of these components:

* Button
* Checkbox
* IconButton
* Link
* RadioButton
* Switch
* TapArea

Verify the following:

* Click on the component => You should see no focus style
* Tab to set the focus on the component => You should see the focus style

### DIFF
--- a/packages/gestalt/src/TapArea.js
+++ b/packages/gestalt/src/TapArea.js
@@ -9,6 +9,8 @@ import getRoundingClassName, {
   type Rounding,
 } from './getRoundingClassName.js';
 import { type AbstractEventHandler } from './AbstractEventHandler.js';
+import focusStyles from './Focus.css';
+import useFocusVisible from './useFocusVisible.js';
 
 type TapStyle = 'none' | 'compress';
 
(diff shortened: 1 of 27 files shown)


packages/gestalt/src/TapArea.js
```diff
@@ -116,7 +121,6 @@ function TapArea({
         if (!disabled && onFocus) {
           onFocus({ event });
         }
-        event.stopPropagation();
```
christianvuerings, line 119: @jackhsu978 could you verify that this is ok to remove? You added it back in #906 for the following reason:

> ### Why stops onFocus event from bubbling up?
> See diff D304752
>> "This makes it so that with the STL ... on a pin is focused the whole pin is not focused."

We do need the focus event to bubble up in this case in order for `useFocusVisible` to work properly. Now that we only show focus styles on keyboard navigation, I hope this is ok.

christianvuerings:
@ashleyseto @anniemochi Could you both take a look and verify the focus styles work as expected?

anniemochi:
Kept getting myself stuck in the code snippets 😅, but the indicators look good! Only change I would consider is making 8px the minimum corner radius, in other words, applying an 8px corder radius to the focus indicator around `link`

packages/gestalt/src/TapArea.js
jackhsu978: I think the approach sounds good. The side effect here is not too implicit.

christianvuerings:
> Kept getting myself stuck in the code snippets 😅, but the indicators look good! Only change I would consider is making 8px the minimum corner radius, in other words, applying an 8px corder radius to the focus indicator around `link`

Thank you for going through these @anniemochi - will follow up on the border radius change in a follow up commit. It is a bit controversial since it might be break if the link content is square (e.g. an image)

PR #1100, opened 2020-07-31
useReducedMotion: Accessibility hook to support prefers-reduced-motion

christianvuerings:
Adds `useReducedMotion` an accessibility hook to support prefers-reduced-motion

useReducedMotion

## Test Plan

Enable / disable the reduce motion settings on your OS and ensure the animation stops / starts.


docs/src/components/sidebarIndex.js
```diff
@@ -31,6 +31,11 @@ const sidebarIndex: sidebarIndexType = [
     sectionPathname: 'setup',
     pages: ['GestaltProvider'],
   },
+  {
+    sectionName: 'Accessibility',
```
chrislloyd, line 35: Should this be "Hooks" or similar? We don't have too many a11y components.

christianvuerings: @chrislloyd Just wait for it, this is only the beginning... Also plan on exporting `useFocusVisible` and adding `VisuallyHidden` + `useId` soon

AlbertCarreras:
@cvuerings Box does not support animation, so this hook is to be used with div tags or dangerouslyInlineStyle?

christianvuerings:
> @cvuerings Box does not support animation, so this hook is to be used with div tags or dangerouslyInlineStyle?

It is to be used within Gestalt (if we add any animation) + in other repos where animation is used. That might be through div tags with `style`, `dangerouslyInlineStyle` on `Box`, using `className` or any other library they use.

packages/gestalt/src/useReducedMotion.js
```diff
@@ -0,0 +1,27 @@
...
+      setMatch(mediaQuery.matches);
+    };
+    handleChange();
+    mediaQuery.addEventListener('change', handleChange);
+    return () => {
+      mediaQuery.removeEventListener('change', handleChange);
+    };
+  }, [supportsMatchMedia]);
+  return matches;
+}
```
jennyscript, line 27: Feels like we should have a test for this though I guess it'd just be testing the browser API so maybe not

docs/src/useReducedMotion.doc.js
```diff
@@ -0,0 +1,62 @@
...
+
+const cards = [];
+const card = c => cards.push(c);
+
+card(
+  <PageHeader
+    name="useReducedMotion"
+    description={`
+      \`useReducedMotion\` uses the <a href="https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion" target="_blank">prefers-reduced-motion</a> CSS media feature to detect
+      if the user has requested that the system minimize the amount of non-essential motion.
```
jennyscript, line 14: Can we provide more guidance here on when devs should use this? Not everyone will click the link for a full explanation

christianvuerings: @jennyscript added - mind taking another look?

packages/gestalt/src/useReducedMotion.js
christianvuerings: @jennyscript good callout - added tests.

@jennyscript @AlbertCarreras fyi this PR now also ads `@testing-library/react-hooks` to start testing hooks we export.

PR #1103, opened 2020-07-31
Internal: Refactored imports from namespacing * to explicit Default Exports and Named Values Importing

AlbertCarreras:
### Why are we moving from `import * as React` to `import React, { ... }`

- Consistency across Gestalt codebase.
- Explicit Default Imports and Named Values Importing is consistent with the importing format used in all other cases in our Pinboard/webapp codebase. 
- Explicit Default Imports and Named Values helps with readability; all used React features are listed on top. For instance, we only have to check the first line of code to know if a component is using `React.forwardRef` instead of scrolling all the way to the bottom of the file.


christianvuerings:
@AlbertCarreras thank you for doing this, some more resources on this topic:
https://github.com/airbnb/javascript#modules--no-wildcard

PR #1107, opened 2020-08-03
Button/IconButton/CheckBox/RadioButton: refactored forwardRef

AlbertCarreras:
- Refactored `forwardRef` implementation to prevent rendering 2 components and fix `displayName` in:

1. Button
2. IconButton
3. CheckBox
4. RadioButton

- Fix Doc examples for  CheckBox, RadioButton

### DIFF
--- a/docs/src/Checkbox.doc.js
+++ b/docs/src/Checkbox.doc.js
@@ -191,7 +191,7 @@ function CheckboxExample() {
 
 card(
   <Example
-    id="ref"
+    id="refExample"
     name="Example: ref"
     description={`The innermost \`Checkbox\` element can be accessed via \`ref\``}
     defaultCode={`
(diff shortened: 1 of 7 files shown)


packages/gestalt/src/Button.js
```diff
@@ -176,21 +178,16 @@ function Button(props: Props): Element<'button'> {
     </button>
   );
   /* eslint-enable react/button-has-type */
-}
+});
 
-Button.propTypes = {
+// $FlowFixMe Flow(InferError)
```
christianvuerings, line 183: @AlbertCarreras wondering if we can figure out a fix for this. Also can we use specific comments? `// $FlowFixMe [issue]`

docs/src/Checkbox.doc.js
```diff
@@ -193,42 +193,98 @@ card(
   <Example
     id="refExample"
```
christianvuerings, line 194: @AlbertCarreras fyi changed this back to `refExample` - we use the ID to link to specific sections of the page
image

christianvuerings:
@mergifyio refresh

PR #1109, opened 2020-08-03
VSCode: add recommended extensions

christianvuerings:
In #1108 @ashleyseto ran into an issue where VSCode was incorrectly formatting code (not using prettier). The root cause is that the prettier extension is not installed.

image

After this lands, users will see the following upon restarting VSCode:
image


christianvuerings:
@Mergifyio refresh

PR #1112, opened 2020-08-04
Switch: Implemented forwardRef (BREAKING CHANGE)

AlbertCarreras:
- Adds ref forwarding to Switch
- Added ref example to Docs
- Added test for ref 

- Added Label prop to Switch. The Docs example was demoing how to add a Label for accessibility. Implemented Label prop following CheckBox/RadioButton pattern.

### Why?

Without forwardedRefs, refs are set on the component's instance, not the actual input. So in each of those cases, people drill down to get the input.

)

### DIFF
--- a/docs/src/Switch.doc.js
+++ b/docs/src/Switch.doc.js
@@ -35,6 +35,10 @@ card(
         href: 'basicExample',
       },
       {
+        name: 'label',
+        type: 'string',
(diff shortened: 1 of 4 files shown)


christianvuerings:
Closing this out for now since we don't have a use for it yet in the pinterest codebase.

PR #1118, opened 2020-08-04
Focus Styles: override CSS outline if a global one is specified

christianvuerings:
## Before
button-focus-outline-before

## After
button-focus-outline-after

## Test Plan

1) Add the following global CSS to the docs CSS (e.g. `DocSearch.css`)
```
:focus {
  outline: 1px auto #d1d1d1;
}
```
2) Ensure you don't see a double / light gray focus style


AlbertCarreras:
reproduced and tested. it solved it following testing plan.

PR #1122, opened 2020-08-05
DatePicker: Fix proptype bug

AlbertCarreras:

### DIFF
--- a/packages/gestalt-datepicker/src/DatePicker.js
+++ b/packages/gestalt-datepicker/src/DatePicker.js
@@ -212,7 +212,7 @@ DatePickerWithForwardRef.propTypes = {
   nextRef: PropTypes.oneOfType([
     PropTypes.func,
     PropTypes.shape({
-      current: PropTypes.instanceOf(Node),
+      current: PropTypes.node,
     }),
   ]),
   onChange: PropTypes.func.isRequired,


packages/gestalt-datepicker/src/DatePicker.js
```diff
@@ -212,7 +212,7 @@ DatePickerWithForwardRef.propTypes = {
   nextRef: PropTypes.oneOfType([
     PropTypes.func,
     PropTypes.shape({
-      current: PropTypes.instanceOf(Node),
+      current: PropTypes.node,
```
jackhsu978, line 215: The reference cannot be any React node, but has to be an HTMLDivElement, right? Should this be `PropTypes.elementType`?

christianvuerings: Going to r+ to unblock the Gestalt release (once we SSR our docs we will be removing propTypes anyhow)

AlbertCarreras: Hey @jackhsu978, we can probably be more specific with `PropTypes.elementType`, `PropTypes.node` is more general but did the job of fixing the bug. https://reactjs.org/docs/typechecking-with-proptypes.html
image

PR #1124, opened 2020-08-05
Docs: single subdirectory URLs

christianvuerings:
Internal request by couple of developers.

Before
`https://gestalt.netlify.app/layout/Box`
After
`https://gestalt.netlify.app/Box`

Pros
* Allows developers to quickly go to a component without knowing which subsection it is in.

Cons
* Means that every section in itself needs to be unique. E.g. we can't do `layout/Box` and `forms/Box`.


christianvuerings:
@mergifyio refresh

PR #1128, opened 2020-08-05
Docs: Navigation- Add an option to list components alphabetically

mminute:
Add a select list to the documentation navigation bar to allow users to choose whether the Gestalt components are organized by category or alphabetically.

## Test Plan
Go to the Gestalt documentation.  In the side nav bar observe that there is a select list with `Categorized` selected.
The components listed below the select list should be organized by category (ex. Foundation: [Heading, Icon, Text])
Select `Alphabetical` in the select list.
The components should update to be alphabetical.

GestaltAlpha

### DIFF
--- a/docs/src/components/Navigation.js
+++ b/docs/src/components/Navigation.js
@@ -1,27 +1,95 @@
 // @flow strict
-import React from 'react';
-import { Box } from 'gestalt';
+import React, { useState } from 'react';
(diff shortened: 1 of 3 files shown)


AlbertCarreras:
I love this. I'd suggest removing  `Component Organization` as we categorize based on functionality (not everything is a component, we have hooks, providers, etc) 

Alpabetically, typo.

PR #1133, opened 2020-08-07
WIP Row/Stack: Re-implement `gap` using CSS

dangerismycat:
The current implementation of `gap` is very limiting, in that children cannot declare any flex properties (e.g. `flex="grow"`) that are respected by the parent. This PR re-implements `gap` using the CSS child combinator, thereby adding the padding styles to the existing child styles rather than inserting an additional div.

This also required extracting Box's transforms for use in Row and Stack. Given the exports, Flow had a lot to say about this, which I updated as best I could but still had to add a handful of suppressions.


dangerismycat:
something went crazy with git, closing this PR and opening a new one from a fresh branch

PR #1135, opened 2020-08-07
Datepicker: standardized date format & placeholder date format across…

AlbertCarreras:
… locale

- `date format` is used to format the displayed date and as default placeholder following [datepicker best practices]( https://uxplanet.org/date-picker-design-best-practices-41bd522f10a5).
- Locale format provided by `date-fns` was inconsistent across locales. 
- We added a small library to provide a consistent date format across locales.

### Before
image
image

### After
image
image

### DIFF
--- a/packages/gestalt-datepicker/src/DatePicker.js
+++ b/packages/gestalt-datepicker/src/DatePicker.js
@@ -14,7 +14,7 @@ import { Icon, Box, Label, Text } from 'gestalt';
 import PropTypes from 'prop-types';
 import DatePickerTextField from './DatePickerTextField.js';
 import styles from './DatePicker.css';
-import format from './format.js';
+import dateFormat from './dateFormat.js';
 import { LocaleDataPropTypes, type LocaleData } from './LocaleDataTypes.js';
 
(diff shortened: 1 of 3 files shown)


packages/gestalt-datepicker/src/format.js
```diff
@@ -0,0 +1,36 @@
+// @flow strict-local
```
christianvuerings, line 1: @AlbertCarreras could you change this to `// @flow strict`?

packages/gestalt-datepicker/src/DatePicker.js
```diff
@@ -14,6 +14,7 @@ import { Icon, Box, Label, Text } from 'gestalt';
 import PropTypes from 'prop-types';
 import DatePickerTextField from './DatePickerTextField.js';
 import styles from './DatePicker.css';
+import format from './format.js';
```
christianvuerings, line 17: @AlbertCarreras could we rename this to `dateFormat`?

AlbertCarreras: DONE

PR #1136, opened 2020-08-07
ZIndexClasses: note on docs re extracting index values from classes

AlbertCarreras:

### DIFF
--- a/docs/src/ZIndexClasses.doc.js
+++ b/docs/src/ZIndexClasses.doc.js
@@ -1,6 +1,6 @@
 // @flow strict
 import React from 'react';
-import { Box, Heading, Stack, Text } from 'gestalt';
+import { Box, Heading, Icon, Row, Stack, Text } from 'gestalt';
 import Example from './components/Example.js';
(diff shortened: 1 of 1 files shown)


docs/src/ZIndexClasses.doc.js
```diff
@@ -108,4 +108,31 @@ function ZIndexBoxExample() {
...
 );
 
+card(
+  <Stack gap={2}>
+    <Heading size="md">Note</Heading>
+    <Text>
+      FixedZIndex and CompositeZIndex work with Box and Sticky components. To
+      stay consistent across your codebase using zIndex classes, you can extract
+      zIndex values from both zIndex classes in cases where the zIndex receptor
+      does not accept zIndex classes.
```
jennyscript, line 118: I'd call out here that this is an escape hatch that should only be used in cases like needing to work with a 3rd party library. The more "gestalt-y" way to handle this is wrap whatever needs a zIndex in a Box

AlbertCarreras: image

PR #1137, opened 2020-08-07
Sticky: add height prop

jackhsu978:
`<Sticky>` container always have the same height has the height of its content. This adds the `height` prop which allows the container to have a different height than its content.

Screen Shot 2020-08-07 at 2 12 17 PM

### DIFF
--- a/packages/gestalt/src/Sticky.js
+++ b/packages/gestalt/src/Sticky.js
@@ -20,16 +20,17 @@ type Threshold =
     |};
 
 type Props = {|
-  children: Node,
-  zIndex?: Indexable,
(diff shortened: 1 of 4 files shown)


packages/gestalt/src/Sticky.js
```diff
@@ -20,16 +20,17 @@ type Threshold =
     |};
 
 type Props = {|
+  ...Threshold,
   children: Node,
-  zIndex?: Indexable,
   dangerouslySetZIndex?: {| __zIndex: number |},
-  ...Threshold,
+  height?: number,
```
christianvuerings, line 26: @jackhsu978 should we also make this a number or string? Like we do on Box?

image

PR #1138, opened 2020-08-10
[Pog]: Add red to prop type

simkessy:
The color red was missing in the prop types causing an erorr:

image


AlbertCarreras:
Thanks!

PR #1139, opened 2020-08-10
Avatar: Remove person icon fallback variant

nachofassini:
We decided to move forward with the `first initial` version of the Avatar when the user has no profile pic

- Removes person icon fallback variant
- Removes `__dangerouslyUseDefaultIcon` prop

Before we had the possibility to choose between person icon or first initial variants for default avatar 
Before

We decided to deprecate the person icon variant. So the first initial is the one to show when the user has no profile pic
After

Tested locally

### DIFF
--- a/packages/gestalt/src/Avatar.js
+++ b/packages/gestalt/src/Avatar.js
@@ -5,7 +5,6 @@ import Box from './Box.js';
 import Icon from './Icon.js';
 import Image from './Image.js';
 import Mask from './Mask.js';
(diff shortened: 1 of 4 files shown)


AlbertCarreras:
Hey Ignacio, could we put here a before and after?? Thank you!

AlbertCarreras:
Also, didn't find any usage of `__dangerouslyUseDefaultIcon`
image

AlbertCarreras:
LGTM Thanks!

PR #1140, opened 2020-08-11
SelectList: Add disabled to options prop

norrise120:
This is an update adding in the ability to disable options within SelectList as supported in the native select https://developer.mozilla.org/en-US/docs/Web/HTML/Element/option

## Test Plan
Tested locally to verify that it works as expected and added two tests.

Screen Shot 2020-08-10 at 5 04 22 PM

### DIFF
--- a/packages/gestalt/src/SelectList.js
+++ b/packages/gestalt/src/SelectList.js
@@ -27,6 +27,7 @@ type Props = {|
   options: Array<{
     label: string,
     value: string,
+    disabled?: boolean,
   }>,
   placeholder?: string,
   size?: 'md' | 'lg',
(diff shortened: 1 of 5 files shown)


packages/gestalt/src/SelectList.js
```diff
@@ -162,7 +164,11 @@ export default class SelectList extends Component<Props, State> {
               </option>
             )}
             {options.map(option => (
-              <option key={option.value} value={option.value}>
+              <option
+                key={option.value}
+                value={option.value}
+                disabled={option.disabled || false}
```
AlbertCarreras, line 170: do we need `|| false`, the default is false already? Is Flow complaining without it?

norrise120: Looks like it wasn't needed! Updated accordingly--thanks!

AlbertCarreras: Thanks

AlbertCarreras:
LGTM thanks!

PR #1142, opened 2020-08-12
IconButton/Button: replaced disabled with aria-disabled for accessibility

AlbertCarreras:
We all know the disabled attribute. In most cases it’s used on a button. You could have a form with a button at the bottom. This button should be disabled as long as all forms haven’t been filled out. It will look like this:

`<button type="submit" disabled>Submit</button>
`
The disabled attribute often appears a little bit dimmed or transparent. Of course, you can style this as well. Either way, sighted users will get some sort of visual feedback that he/she is not able to click on this button. A click on this button will do no good. Actually it does nothing at all. It’s inactive.

And what about visually impaired users? They don’t get any feedback at all. By using the tabulator key and navigating within a form the screen readers will ignore the existence of any disabled button.

Now close your eyes and imagine the screen reader reads out all the inputs of a form. You know you're within a form. And normally a form gets submitted by hitting a button. But our form depends on the fact that every input has to be filled out - or something else has to happen before it can be submitted. Thus we disable the button until all the conditions have been fulfilled. All sighted persons will get it. The button is semi-transparent. But when a screen reader doesn't ever reach the button, then he will never indicate that there's a button.

In this case, we can use an `aria-disabled="true"`  instead of the good old disabled.

https://a11y-101.com/development/aria-disabled

### DIFF
--- a/packages/gestalt/src/Button.js
+++ b/packages/gestalt/src/Button.js
@@ -145,8 +145,8 @@ const ButtonWithForwardRef: React$AbstractComponent<
       aria-expanded={accessibilityExpanded}
       aria-haspopup={accessibilityHaspopup}
       aria-label={accessibilityLabel}
+      aria-disabled={disabled}
       className={classes}
-      disabled={disabled}
       name={name}
       onBlur={handleBlur}
       onClick={event => onClick && onClick({ event })}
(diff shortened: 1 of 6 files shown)


AlbertCarreras:
https://stackoverflow.com/questions/38059140/when-to-use-the-disabled-attribute-vs-the-aria-disabled-attribute-for-html-eleme/38059790

PR #1150, opened 2020-08-17
Video: add crossOrigin option

jesselumarie:
  - Adding an option to choose the `crossOrigin` mode for videos so that users can use captions from a different domain if needed
  - Additional Context: I broke Gestalt video for [some folks](https://github.com/pinterest/gestalt/pull/1046#discussion_r454048460), and @simkessy [fixed them](https://github.com/pinterest/gestalt/pull/1077)
* What kind of feedback do you want?
  - Make sure changes seem reasonable

## Test Plan
  - [x] add captions and verify that they still work with `crossOrigin='anonymous'` passed in
  - [x] Make sure documentation Video still works locally

### DIFF
--- a/docs/src/Video.doc.js
+++ b/docs/src/Video.doc.js
@@ -89,6 +89,12 @@ card(
           Instead it serves to add overlays on top of the html video element, while still being under the video controls.`,
       },
       {
+        name: 'crossOrigin',
+        type: 'string',
+        description:
+          "Designate CORS behavior for the video element, one of 'use-credentials' or 'anonymous'. Note: when not passed in, CORS checks are disabled.",
+      },
+      {
         name: 'controls',
         type: 'boolean',
         description: 'Show the video player controls',
--- a/docs/src/Video.doc.js
+++ b/docs/src/Video.doc.js
@@ -92,7 +92,7 @@ card(
         name: 'crossOrigin',
         type: "'use-credentials' | 'anonymous'",
         description:
-          "Designate CORS behavior for the video element. When not passed in, CORS checks are disabled.",
+          'Designate CORS behavior for the video element. When not passed in, CORS checks are disabled.',
       },
       {
         name: 'controls',
--- a/docs/src/Video.doc.js
+++ b/docs/src/Video.doc.js
@@ -90,9 +90,9 @@ card(
       },
       {
         name: 'crossOrigin',
-        type: 'string',
+        type: "'use-credentials' | 'anonymous'",
         description:
-          "Designate CORS behavior for the video element, one of 'use-credentials' or 'anonymous'. Note: when not passed in, CORS checks are disabled.",
+          "Designate CORS behavior for the video element. When not passed in, CORS checks are disabled.",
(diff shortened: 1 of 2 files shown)


docs/src/Video.doc.js
```diff
@@ -88,6 +88,12 @@ card(
         description: `This \`children\` prop is not same as children inside the native html \`video\` element.
           Instead it serves to add overlays on top of the html video element, while still being under the video controls.`,
       },
+      {
+        name: 'crossOrigin',
+        type: 'string',
+        description:
+          "Designate CORS behavior for the video element, one of 'use-credentials' or 'anonymous'. Note: when not passed in, CORS checks are disabled.",
+      },
```
AlbertCarreras, line 96: type: 'use-credentials' | 'anonymous',

please make it explicit to the 2 strigs.

packages/gestalt/src/Video.js
```diff
@@ -172,6 +172,7 @@ export default class Video extends PureComponent<Props, State> {
     aspectRatio: PropTypes.number.isRequired,
     captions: PropTypes.string.isRequired,
     children: PropTypes.node,
+    crossOrigin: PropTypes.string,
```
AlbertCarreras, line 176: PropTypes.oneOf(['use-credentials' | 'anonymous'']),

packages/gestalt/src/Video.js
```diff
@@ -203,11 +204,13 @@ export default class Video extends PureComponent<Props, State> {
   };
 
   static defaultProps: {|
+    crossOrigin?: 'use-credentials' | 'anonymous',
     playbackRate: number,
     playing: boolean,
     preload: 'auto' | 'metadata' | 'none',
     volume: number,
   |} = {
+    crossOrigin: undefined,
```
AlbertCarreras, line 214: if it's undefined, we might not need a defaultProp.

packages/gestalt/src/Video.js
```diff
@@ -503,6 +506,7 @@ export default class Video extends PureComponent<Props, State> {
       aspectRatio,
       captions,
       children,
+      crossOrigin,
```
AlbertCarreras, line 510: without default props or given value, this is already `undefined`.

docs/src/Video.doc.js
AlbertCarreras: description: "Designate CORS behavior for the video element. When not passed in, CORS checks are disabled."

packages/gestalt/src/Video.js
AlbertCarreras: double checking this one...

packages/gestalt/src/Video.js
```diff
@@ -205,13 +205,11 @@ export default class Video extends PureComponent<Props, State> {
   };
 
   static defaultProps: {|
-    crossOrigin?: 'use-credentials' | 'anonymous',
```
AlbertCarreras: defaultProps as undefined might not be needed

PR #1154, opened 2020-08-20
Docs: add sidebar alphabetization toggle

christianvuerings:
Based on #1128 from @mminute with the following changes:

* Move toggle to the header (where have have all the other toggles)
* Save sort preference in `localStorage` (so it stays consistent between refreshes)
* Don't exclude `Getting Started` (more scalable but does mean we don't only sort components / hooks)

## Test Plan

Test on netlify:
* Toggle the sort preference in the header
* Refresh the site + see that sort preference stays the same

### DIFF
--- a/docs/src/components/App.js
+++ b/docs/src/components/App.js
@@ -10,6 +10,8 @@ type Props = {|
   children?: React.Node,
 |};
 
+const localStorageOrganizedByKey = 'gestalt-sitebar-organized-by';
+
 export default function App(props: Props) {
   const { children } = props;
   const [isSidebarOpen, setIsSidebarOpen] = useState(false);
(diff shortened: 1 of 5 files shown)


christianvuerings:
@mminute could you take a look and see if this also meets your use case?

docs/src/components/App.js
```diff
@@ -10,18 +10,41 @@ type Props = {|
   children?: React.Node,
 |};
 
+const localStorageOrganizedByKey = 'gestalt-sitebar-organized-by';
```
AlbertCarreras, line 13: nit: is this a typo?  `gestalt-**sidebar**-organized-by`

docs/src/components/App.js
```diff
@@ -10,18 +10,41 @@ type Props = {|
...
+
 export default function App(props: Props) {
   const { children } = props;
   const [isSidebarOpen, setIsSidebarOpen] = useState(false);
   const [colorScheme, setColorScheme] = useState('light');
 
   useTracking('UA-12967896-44');
 
+  const [sidebarOrganisedBy, setSitebarOrganizedBy] = useState<
```
AlbertCarreras, line 22: typo? set**Sidebar**OrganizedBy

docs/src/components/Header.js
```diff
@@ -100,6 +105,34 @@ function Header({ colorScheme, onChangeColorScheme }: Props) {
...
+            text={`Sidebar: ${
+              sidebarOrganisedBy === 'categorized'
+                ? 'Alphabetical'
+                : 'Categorize'
+            }`}
+          >
+            <Box display="flex" alignItems="center">
+              <IconButton
+                size="md"
+                accessibilityLabel="Toggle side bar categorization"
```
AlbertCarreras, line 119: accessibilityLabel=Toggle **sidebar** categorization

AlbertCarreras: LGTM otherwise

docs/src/components/App.js
christianvuerings: @AlbertCarreras thanks for spotting - updated all of these

Issue #617, opened 2020-01-06
Adding error message handling to SearchField


themindfuldev:
We are using `<SearchField>` as a form field, to select dynamically populated values, retrieved from a search term.

That's a required field for us, but since SearchField doesn't render error messages, we rebuilt a similar-looking error message on our parent component. 

However, this causes 2 problems:

1. On the rendered UI, we can't really add `aria-invalid` to the `<input>` field (since that's controlled by SearchField ), neither `aria-describedby` which should point to the error message.
1. Additional maintenance required on our error message look and feel, should gestalt's error message change.

Hence, I'd like to propose making SearchField behave like TextField for error state/error message, by:

- Adding 2 new optional props `hasError` and `errorMessage`. 
- Consuming them to populate `aria-invalid`, `aria-describedby` and including `<FormErrorMessage>`.
 
I'd be happy to work on it once I get an agreement from your side.

By doing this, we will be able to start verifying error states using `aria-invalid` on unit tests using RTL.

christianvuerings:
@themindfuldev Feel free to go ahead an open a PR for this, we should use a similar approach as to TextField and TextArea

themindfuldev:
Thank you so much @szatakia and @CalvinChhour !

PR #1155, opened 2020-08-20
SearchField: Add errorMessage

szatakia:
To add error message to SearchField Component

* What is the context surrounding this PR? https://github.com/pinterest/gestalt/issues/617
* Have you [formatted the PR title](https://github.com/pinterest/gestalt/#releasing)? Yes

image

## Test Plan

How can reviewers verify this is good to merge?

* Is it tested? Yes
* Is it accessible? Yes
* Is it documented? Yes
* Have you involved other stakeholders (such as a Pinterest Designer)? @themindfuldev

### DIFF
--- a/docs/src/SearchField.doc.js
+++ b/docs/src/SearchField.doc.js
@@ -68,6 +68,15 @@ card(
         name: 'value',
         type: 'string',
       },
+      {
+        name: 'errorMessage',
+        type: 'string',
+      },
+      {
+        name: 'hasError',
+        type: 'boolean',
+        defaultValue: false,
+      },
     ]}
   />
 );
--- a/docs/src/SearchField.doc.js
+++ b/docs/src/SearchField.doc.js
@@ -72,11 +72,6 @@ card(
         name: 'errorMessage',
         type: 'string',
       },
-      {
-        name: 'hasError',
-        type: 'boolean',
-        defaultValue: false,
-      },
     ]}
   />
 );
--- a/packages/gestalt/src/SearchField.js
+++ b/packages/gestalt/src/SearchField.js
@@ -5,8 +5,10 @@ import PropTypes from 'prop-types';
 import classnames from 'classnames';
 import layout from './Layout.css';
 import styles from './SearchField.css';
+import formElement from './FormElement.css';
 import Box from './Box.js';
 import Icon from './Icon.js';
+import FormErrorMessage from './FormErrorMessage.js';
 import { type AbstractEventHandler } from './AbstractEventHandler.js';
 
 type Props = {|
@@ -25,6 +27,8 @@ type Props = {|
   placeholder?: string,
   size?: 'md' | 'lg',
   value?: string,
+  errorMessage?: string,
+  hasError?: boolean,
 |};
 
 const SearchFieldWithForwardRef: React$AbstractComponent<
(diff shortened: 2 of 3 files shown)


CalvinChhour:
This is @szatakia's task for the Web Mentorship program, we chatted with @themindfuldev in a meeting about this.
https://github.com/pinterest/gestalt/issues/617

packages/gestalt/src/SearchField.js
```diff
@@ -81,82 +87,92 @@ const SearchFieldWithForwardRef: React$AbstractComponent<
...
-        value={value}
-      />
-      {hasValue && (
-        <button
-          className={styles.clear}
-          onClick={handleClear}
-          tabIndex={-1}
-          type="button"
-        >
+    <span>
```
christianvuerings, line 105: Instead of wrapping this in a `<span />` - could we see it if works with a React Fragment?

szatakia: Changed this to React Fragment

szatakia:
@christianvuerings Can you review this again? I addressed your comment

docs/src/SearchField.doc.js
```diff
@@ -68,6 +68,15 @@ card(
         name: 'value',
         type: 'string',
       },
+      {
+        name: 'errorMessage',
```
christianvuerings, line 72: @szatakia could you explain (perhaps share a screenshot) why we need both `hasError` and `errorMessage`? It's confusing to have both, e.g. what should happen when `hasError` is `false` and `errorMessage` is set to a string?

For now I would suggest to only keep `errorMessage` and remove `hasError`:
* It makes it clear to the user why we show an error.
* Consistent with other text inputs: `TextField` / `TextArea`

themindfuldev: Hi @christianvuerings I originally suggestedincluding `hasError` for parity with `TextField` since it's here: https://github.com/pinterest/gestalt/blob/master/packages/gestalt/src/TextField.js#L21

However, I noticed it does not appear on Gestalt Docs so I wonder if we really need them all. Do you have any thoughts on that?

To me it seems it's a way to highlight the field with error border when there is no message to be displayed? Not really sure

christianvuerings: @themindfuldev you are right - it's in the code but not documented. Let's remove `hasError` from `SearchField` for now.

I looked at the stats for the Pinterest codebase and we only have 6 instances of `hasError` compared to 154 instances of `errorMessage` on `TextField`.

image

szatakia: I did that just to make it consistent with error-handling capabilities of [TextField](https://github.com/pinterest/gestalt/blob/954da26894743e2d6abf5be85125c34b1743728a/packages/gestalt/src/TextField.js#L21). But it makes sense. I will simplify this and will create a PR for TextField too.

christianvuerings: @szatakia thank you! For `TextField`, we should first update the 6 places in our Pinterest codebase where we still use `hasError` and replace those with `errorMessage`.

szatakia: Removed `hasError` from SearchField

christianvuerings:
@szatakia looks great - thank you for making the updates. /cc @themindfuldev

PR #1158, opened 2020-08-21
Image: convert to function component

christianvuerings:
Convert `Image` to a function component. Main reason is to get rid of the duplicated flow types in `defaultProps`. This recently caused an issue with `SelectList` in #1151 

## Test Plan

Image renders correctly

### DIFF
--- a/packages/gestalt/src/Image.js
+++ b/packages/gestalt/src/Image.js
@@ -1,15 +1,16 @@
 // @flow strict
-import React, { PureComponent, type Node } from 'react';
+import React, { type Node, useEffect, useCallback } from 'react';
 import PropTypes from 'prop-types';
 import Box from './Box.js';
 import styles from './Image.css';
+import usePrevious from './usePrevious.js';
 
 const shouldScaleImage = fit => fit === 'cover' || fit === 'contain';
 
 type Props = {|
   alt: string,
   children?: Node,
-  color: string,
+  color?: string,
   fit?: 'contain' | 'cover' | 'none',
   importance?: 'high' | 'low' | 'auto',
   loading?: 'eager' | 'lazy' | 'auto',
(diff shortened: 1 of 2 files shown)


packages/gestalt/src/Image.js
```diff
@@ -22,126 +23,107 @@ type Props = {|
...
-    importance?: 'high' | 'low' | 'auto',
-    loading?: 'eager' | 'lazy' | 'auto',
-  |} = {
-    color: 'transparent',
-    fit: 'none',
-    importance: 'auto',
-    loading: 'auto',
-  };
-
-  componentDidMount() {
```
AlbertCarreras, line 54: is this one inside 'useEffect'? `useEffect` has a different conditinal `shouldScaleImage(this.props.fit))`

packages/gestalt/src/Image.js
```diff
@@ -22,126 +23,107 @@ type Props = {|
...
-      image.onerror = this.handleError;
-      image.src = this.props.src;
+      image.onload = handleLoad;
+      image.onerror = handleError;
+      image.src = src;
     }
-  }
+  }, [handleError, handleLoad, src]);
+
+  const previousSrc = usePrevious(src);
```
AlbertCarreras, line 62: useRef used to store info and replace componentDidUpdate

packages/gestalt/src/Image.js
christianvuerings: @AlbertCarreras line 70 combines the previous `componentDidMount` and `componentDidUpdate`.

packages/gestalt/src/Image.js
christianvuerings: That's correct - as explained in the React docs: https://reactjs.org/docs/hooks-faq.html#how-to-get-the-previous-props-or-state

christianvuerings:
Closing for now since this PR hasn't been touched in a while

PR #1159, opened 2020-08-21
WIP: visual diff testing

mziemer21:


christianvuerings:
@mziemer21 thanks for putting this proof of concept together! Instead of using jest + puppeteer, we should use Cypress + a visual diff plug-in. That way we stick to 1 integration test framework and it'll be easier to maintain moving forward.

https://docs.cypress.io/guides/tooling/visual-testing.html
https://docs.cypress.io/plugins/#visual-testing

An example of a provider we could use is Applitools who does have an open source license: 
https://applitools.com/contact/sales/open-source/

PR #1162, opened 2020-08-24
Docs: update margin/padding for markdown elements

christianvuerings:
After #1145 the docs for each component look off (indented description):

image

/cc @AlbertCarreras 

### After
image

### DIFF
--- a/docs/src/components/Markdown.css
+++ b/docs/src/components/Markdown.css
@@ -1,20 +1,15 @@
-.Markdown p, pre, li {
-  margin-left: 30px;
-}
-
-.Markdown p, pre {
-  margin-bottom: 20px;
-  margin-top: 20px;
-}
-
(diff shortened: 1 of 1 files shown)


docs/src/components/Markdown.css
```diff
@@ -1,22 +1,17 @@
...
-  margin-right: 30px;
-}
-
 .Markdown pre {
   background-color: rgb(236, 236, 236);
   overflow: auto;
   padding: 5px;
 }
 
+.Markdown ul + pre {
```
christianvuerings, line 7: @AlbertCarreras fyi this uses the `Adjacent sibling combinator` https://developer.mozilla.org/en-US/docs/Web/CSS/Adjacent_sibling_combinator

So we only indent the code (`pre`) when the previous child is a `ul`.

### Regular `pre`
image


### `pre` after a `ul`
image

PR #1167, opened 2020-08-25
SelectList: convert to function component

christianvuerings:
Convert `SelectList` to a function component. Main reason is to get rid of the duplicated flow types in `defaultProps`. This recently caused an issue with `SelectList` in #1151 

## Test Plan
* SelectList tests pass
* SelectList works as expected in the docs

### DIFF
--- a/packages/gestalt/src/SelectList.js
+++ b/packages/gestalt/src/SelectList.js
@@ -1,5 +1,5 @@
 // @flow strict
-import React, { Component, type Node } from 'react';
+import React, { useState, type Node } from 'react';
 import PropTypes from 'prop-types';
 import classnames from 'classnames';
 import Box from './Box.js';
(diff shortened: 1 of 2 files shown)


packages/gestalt/src/SelectList.js
```diff
@@ -36,147 +36,128 @@ type Props = {|
...
-          >
-            <Icon
-              icon="arrow-down"
-              size={12}
-              color={disabled ? 'gray' : 'darkGray'}
-              accessibilityLabel=""
-            />
-          </Box>
-          <select
-            aria-describedby={errorMessage && focused ? `${id}-error` : null}
```
christianvuerings, line 148: Previously `focused` was never set to `true`, which was a bug. So we added `onFocus`

PR #1168, opened 2020-08-26
Docs: Add in page content navigation

christianvuerings:
Add in page content navigation to the Gestalt Docs.

gestalt-in-page-navigation

## Notes

* Updated multiple pages like `Table` / `Pulsar` / `Icon` to ensure the headings are unique.
* Converted the `.md` file imports since it became to hacky to also add in content nav. It also allows us to keep our docs more consistent moving forward /cc @AlbertCarreras 

### Scrolling

1. Go to any doc page
2. Make sure your screen is wide enough to see the content nav
3. Scroll down
4. Make sure the correct item highlights in the content nav

### Clicking a content nav item

1. Go to any doc page
2. Make sure your screen is wide enough to see the content nav
3. Click on a content nav item
4. Make sure it scrolls to the correct item on the page


AlbertCarreras:
Can we enable scrolling on the right sidebar?

christianvuerings:
> Can we enable scrolling in the right sidebar?

@AlbertCarreras Could you explain (perhaps share a screenshot / gif) why that would be necessary? Is it mostly for screens with a smaller height?

ashleyseto:
> > Can we enable scrolling in the right sidebar?
> 
> @AlbertCarreras Could you explain (perhaps share a screenshot / gif) why that would be necessary? Is it mostly for screens with a smaller height?

This is not a normal side navigation panel. It should not have independent scrolling from the page. If you content list is so long that you can't see the entire nav on a normal computer, then something is wrong with the page.

ashleyseto:
I like that we got the select state in. A couple notes:

- Can we align the nav to the top so it lines up with the start of the content?
Screen Shot 2020-08-27 at 8 43 10 AM

- The select color should be pine and not green. Green is not accessible on white.
- I noticed a gray box appearing on the anchor link headers when we jump. I don't think this is necessary.

Other than that looks great!

christianvuerings:
@ashleyseto Thank you - addressed the comments:

> I like that we got the select state in. A couple notes:
> 
> * Can we align the nav to the top so it lines up with the start of the content?

Updated - do know that it might be a few pixels off because of the way headings are rendered on web (due to line-height) image

> * The select color should be pine and not green. Green is not accessible on white.

Updated - thanks for spotting that!
image

> * I noticed a gray box appearing on the anchor link headers when we jump. I don't think this is necessary.

Removed the gray box appearing

AlbertCarreras:
@christianvuerings @ashleyseto  In Box, if the height is not enough or for very long lists, it's hard to get to the last items. 
Kapture 2020-08-27 at 10 27 59

ashleyseto:
> @christianvuerings @ashleyseto In Box, if the height is not enough or for very long lists, it's hard to get to the last items.
> Kapture 2020-08-27 at 10 27 59

I think box is a great example of where writing should be able to help. There are a lot of IA issues in the Gestalt docs and I think box is one of the good examples of where improvements are really needed.

ashleyseto:
> > @christianvuerings @ashleyseto In Box, if the height is not enough or for very long lists, it's hard to get to the last items.
> > Kapture 2020-08-27 at 10 27 59
> 
> I think box is a great example of where writing should be able to help. There are a lot of IA issues in the Gestalt docs and I think box is one of the good examples of where improvements are really needed.

We could reduce the text size to the smallest size for a V1 launch if it is a huge concern but box lists also the same nave item "Combinations" 3 times.

PR #1169, opened 2020-08-27
Button: v.2. Added support for href and segmented Flow type

AlbertCarreras:
Changes documented here: [TDD](https://paper.dropbox.com/doc/Gestalt-Button-V.2--A6cM36Ezpyk_HZcDKcVTs74HAg-gNyH1dOoWYwvPfqm3Kgme)

- Added support to href to Button.
- Segmented Flow for each Button type to make Flow Type stricter in Button.
- Upgraded Buttons Docs.
- Added test.
- Remove `name` from `Button`

### DIFF
--- a/docs/src/Button.doc.js
+++ b/docs/src/Button.doc.js
@@ -1,6 +1,6 @@
 // @flow strict
 import React, { type Node } from 'react';
-import { Button, Stack, Text } from 'gestalt';
+import { Button } from 'gestalt';
 import PropTable from './components/PropTable.js';
 import Example from './components/Example.js';
 import Combination from './components/Combination.js';
@@ -8,16 +8,6 @@ import PageHeader from './components/PageHeader.js';
 
 const cards: Array<Node> = [];
 const card = c => cards.push(c);
-
-const buildDescription = (lines: Array<string>): Node => (
-  <Stack gap={1}>
-    {lines.map((line, idx) => (
-      <Text key={idx} color="gray">
-        {line}
-      </Text>
-    ))}
-  </Stack>
-);
 
 card(
   <PageHeader
@@ -34,53 +24,52 @@ card(
         type: 'string',
         required: false,
         defaultValue: null,
-        description: buildDescription([
+        description: [
           'Supply a short, descriptive label for screen-readers to replace button texts that do not provide sufficient context about the button component behavior. Texts like `Click Here,` `Follow,` or `Read More` can be confusing when a screen reader reads them out of context. In those cases, we must pass an alternative text to replace the button text.',
           'Use IconButton if you only need a button with an icon and no text.',
-          'Optional with button-type and submit-type buttons.',
-          'Accessibility: It populates aria-label.  Screen readers read the `accessibilityLabel` prop, if present, instead of `text`.',
-        ]),
-        href: 'accessibilityI',
+          'Accessibility: It populates aria-label. Screen readers read the `accessibilityLabel` prop, if present, instead of `text`.',
+        ],
+        href: 'accessibilityLabel',
       },
       {
         name: 'accessibilityControls',
         type: 'string',
         required: false,
         defaultValue: null,
-        description: buildDescription([
+        description: [
           'Specify the `id` of an associated element (or elements) whose contents or visibility are controlled by a button component so that screen reader users can identify the relationship between elements.',
-          'Optional with button-type buttons.',
+          'Optional with button-role + button-type buttons.',
           'Accessibility: It populates aria-controls.',
-        ]),
-        href: 'accessibilityII',
+        ],
+        href: 'accessibility',
       },
       {
         name: 'accessibilityExpanded',
         type: 'boolean',
         required: false,
         defaultValue: null,
-        description: buildDescription([
+        description: [
           'Indicate that a button component hides or exposes collapsible components and expose whether they are currently expanded or collapsed.',
-          'Optional with button-type buttons.',
+          'Optional with button-role + button-type buttons.',
           'Accessibility: It populates aria-expanded.',
-        ]),
-        href: 'accessibilityII',
+        ],
+        href: 'accessibility',
       },
       {
         name: 'accessibilityHaspopup',
         type: 'boolean',
         required: false,
         defaultValue: null,
-        description: buildDescription([
+        description: [
           'Indicate that a button component controls the appearance of menus and expose whether they are currently opened or closed.',
-          'Optional with button-type buttons.',
+          'Optional with button-role + button-type buttons.',
           'Accessibility: It populates aria-haspopup.',
-        ]),
-        href: 'accessibilityII',
+        ],
+        href: 'accessibility',
       },
       {
         name: 'color',
-        type: `gray' | 'red' | 'blue' | 'transparent' | 'white'`,
+        type: `'gray' | 'red' | 'blue' | 'transparent' | 'white'`,
         required: false,
         defaultValue: 'gray',
         description: 'Primary colors to apply to the button background.',
@@ -100,11 +89,11 @@ card(
         type: '$Keys<typeof icons>',
         required: false,
         defaultValue: false,
-        description: buildDescription([
+        description: [
           'Add a Gestalt icon to be displayed after the button text. Sometimes an icon can help clarify the usage of a button. Menus are a common use case.',
           'Accessibility: Icons on buttons are not accessible for screen readers.',
           'Use IconButton if you only need buttons with icons and no text.',
-        ]),
+        ],
         href: 'iconEnd',
       },
       {
@@ -121,10 +110,10 @@ card(
         type: '({ event: SyntheticMouseEvent<> }) => void',
         required: false,
         defaultValue: null,
-        description: buildDescription([
+        description: [
           'Callback fired when a button component is clicked (pressed and released) with a mouse or keyboard.',
-          'Required with button-type buttons.',
-        ]),
+          'Required with button-role + button-type buttons.',
+        ],
         href: 'selected',
       },
       {
@@ -141,10 +130,10 @@ card(
         type: 'string',
         required: true,
         defaultValue: null,
-        description: buildDescription([
-          'Text to render inside the button to convey the function and purpose of the button. The button text has a fix size.',
+        description: [
+          'Text to render inside the button to convey the function and purpose of the button. The button text has a fixed size.',
           'Accessibility: Screen readers read the `accessibilityLabel` prop, if present, instead of the `text` prop. See `accessibilityLabel` for more info.',
-        ]),
+        ],
         href: 'basic-button',
       },
       {
@@ -152,25 +141,23 @@ card(
         type: `'white' | 'darkGray' | 'blue' | 'red'`,
         required: false,
         defaultValue: null,
-        description: buildDescription([
+        description: [
           'Primary colors to apply to the text inside the Button.',
           'Text colors are automatically paired with the defined button `color`.',
           '`textColor` should only be used when using a white background color; it may become necessary to manually override the text color to match that of the parent`s background. Do not override the text color for any button besides those with white background as it breaks design standards.',
-        ]),
+        ],
         href: 'textColor',
       },
       {
         name: 'type',
-        type: `'submit' | 'button' | 'link'`,
+        type: `'submit' | 'button'`,
         required: false,
         defaultValue: 'button',
-        description: buildDescription([
-          'Select a button variant:',
-          '-`button`: Use to render simple push buttons with no default behavior and control custom functionality inside the `onClick` callback. The button is rendered as an button or  `<button> tag`.',
-          '-`submit`: Use to submit forms.  The button is inside/associated with a form. The button is rendered as an button or `<button> tag`.',
-          '-`link`: Use for buttons to act like links. The button is rendered as an anchor or `<a> tag`.',
-          'Required with submit-type and link-type buttons.',
-        ]),
+        description: [
+          'Select a type of button-role button:',
+          '-`button`: Use to render simple push buttons with no default behavior and control custom functionality inside the `onClick` callback.',
+          '-`submit`: Use to submit forms. The button is inside/associated with a form.',
+        ],
         href: 'type',
       },
       {
@@ -178,21 +165,21 @@ card(
         type: 'boolean',
         required: false,
         defaultValue: false,
-        description: buildDescription([
+        description: [
           'Control the "selected" state of a button component to toggle binary states.',
-          'Optional with button-type buttons.',
-        ]),
-        href: '',
+          'Optional with button-role + button-type buttons.',
+        ],
+        href: 'selected',
       },
       {
         name: 'href',
         type: 'string',
         required: false,
         defaultValue: null,
-        description: buildDescription([
-          'Specify a redirect url.',
-          'Required with link-type buttons.',
-        ]),
+        description: [
+          'Specify a link URL.',
+          'Required with link-role buttons.',
+        ],
         href: 'type',
       },
       {
@@ -205,11 +192,24 @@ card(
         href: 'ref',
       },
       {
+        name: 'role',
+        type: `'button' | 'link'`,
+        required: false,
+        defaultValue: 'button',
+        description: [
+          `Select a button variant:`,
+          `- 'button': Use to render 'submit' or 'button'-type buttons. The button is rendered as a '<button>'.`,
+          `- 'link': Use for buttons to act like links. The button is rendered as an '<a>'.`,
+          `Required with link-role buttons.`,
+        ],
+        href: 'type',
+      },
+      {
         name: 'rel',
         type: `'none' | 'nofollow'`,
         required: false,
         defaultValue: 'none',
-        description: 'Optional with link-type buttons.',
+        description: 'Optional with link-role buttons.',
         href: 'type',
       },
       {
@@ -217,13 +217,13 @@ card(
         type: `null | 'self' | 'blank'`,
         required: false,
         defaultValue: 'null',
-        description: buildDescription([
+        description: [
           'Define the frame or window to open the anchor defined on `href`:',
           '- `null` opens the anchor in the same window.',
           '- `blank` opens the anchor in a new window.',
           '- `self` opens an anchor in the same frame.',
-          'Optional with link-type buttons.',
-        ]),
+          'Optional with link-role buttons.',
+        ],
         href: 'type',
       },
     ]}
@@ -240,19 +240,19 @@ card(
 
 card(
   <Example
-    name="Types"
+    name="Types & Roles"
     id="type"
     defaultCode={`
 <Row gap={1}>
-    <Tooltip text="button-type button">
-      <Button onClick={() => {}} text="Clear search history" inline />
-    </Tooltip>
-    <Tooltip text="submit-type button">
-      <Button type="submit" name='satisfaction-questionaire' text="Submit your response" inline />
-    </Tooltip>
-    <Tooltip text="link-type button">
-      <Button type="link" target='blank' href='https://www.pinterest.com' text="Visit pinterest.com" inline />
-    </Tooltip>
+  <Tooltip text="default button: button-role + button-type button">
+    <Button onClick={() => {}} text="Clear search history" inline />
+  </Tooltip>
+  <Tooltip text="form button variant: button-role + submit-type button">
+    <Button type="submit" name='satisfaction-questionaire' text="Submit your response" inline />
+  </Tooltip>
+  <Tooltip text="link button variant: link-role button + button-type">
+    <Button role="link" target='blank' href='https://www.pinterest.com' text="Visit pinterest.com" inline />
+  </Tooltip>
 </Row>
 `}
   />
@@ -291,88 +291,8 @@ card(
     name="Icons"
     id="iconEnd"
     defaultCode={`
-function MenuButtonExample() {
-  const [open, setOpen] = React.useState(false);
-  const anchorRef = React.useRef();
-
-  return (
-    <>
-      <Box display="inlineBlock" ref={anchorRef}>
-        <Button
-          accessibilityExpanded={!!open}
-          accessibilityHaspopup
-          iconEnd={open ? "arrow-down" : "arrow-forward"}
-          inline
-          onClick={() => setOpen(!open)}
-          text="Menu"
-        />
-      </Box>
-
-      {open && (
-        <Layer>
-          <Flyout
-            anchor={anchorRef.current}
-            idealDirection="down"
-            onDismiss={() => setOpen(false)}
-            positionRelativeToAnchor={false}
-            size="md"
-          >
-            <Box direction="column" display="flex" padding={2}>
-              <Box padding={2}>
-                <Text weight="bold">
-                  Option 1
-                </Text>
-              </Box>
-              <Box padding={2}>
-                <Text weight="bold">
-                  Option 2
-                </Text>
-              </Box>
-            </Box>
-          </Flyout>
-        </Layer>
-      )}
-    </>
-  );
-}
+    <Button iconEnd='download' text="Download CVS file" inline />
 `}
-  />
-);
-
-card(
-  <Example
-    id="ref"
-    name="ref"
-    defaultCode={`
-function ButtonFlyoutExample() {
-  const [open, setOpen] = React.useState(false);
-  const [checked, setChecked] = React.useState(false);
-
-  const anchorRef = React.useRef(null);
-
-  return (
-    <>
-      <Button
-        inline
-        onClick={() => setOpen(true)}
-        ref={anchorRef}
-        text="Show Flyout"
-      />
-      {open && (
-        <Flyout
-          anchor={anchorRef.current}
-          idealDirection="right"
-          onDismiss={() => setOpen(false)}
-          shouldFocus={false}
-        >
-          <Box padding={3}>
-            <Text weight="bold">This is a Button with an anchor ref to a Flyout component</Text>
-          </Box>
-        </Flyout>
-      )}
-    </>
-  );
-}`}
   />
 );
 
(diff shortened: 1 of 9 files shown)


docs/src/Button.doc.js
```diff
@@ -127,7 +127,7 @@ card(
   <Example
     name="Example"
     defaultCode={`
-<Button text="Medium sized button" inline />
+<Button href="http://www.pinterest.com" text="Medium sized button" target='blank' type="link" inline />
```
christianvuerings, line 130: @AlbertCarreras let's not make this the default example and add a new one instead.

packages/gestalt/src/Button.js
```diff
@@ -28,54 +29,65 @@ const DEFAULT_TEXT_COLORS = {
...
+  name?: string,
+  type: 'submit',
+|};
+
+type LinkButtonType = {|
+  ...BaseButton,
+  href: string,
+  rel?: 'none' | 'nofollow',
+  target?: null | 'self' | 'blank',
+  type: 'link',
```
christianvuerings, line 68: @AlbertCarreras just so I understand. Does this mean developers would have to add `type="link"` each time they add the `href` tag?

packages/gestalt/src/Button.js
```diff
@@ -143,6 +155,85 @@ const ButtonWithForwardRef: React$AbstractComponent<
...
+    const { href, rel, target } = props;
+
+    return (
+      <Link
+        hoverStyle="none"
+        href={href}
+        ref={innerRef}
+        rel={rel}
+        target={target}
+        inline
```
christianvuerings, line 168: @AlbertCarreras curious to know what happens for non-inline buttons. Should this be conditional based on the passed in props?

packages/gestalt/src/Button.js
```diff
@@ -143,6 +155,85 @@ const ButtonWithForwardRef: React$AbstractComponent<
...
+        rounding="pill"
+      >
+        <div
+          style={{ alignItems: 'center', display: 'flex' }}
```
christianvuerings, line 172: @AlbertCarreras let's add it to a CSS file instead of using inline styles. (perf)

packages/gestalt/src/Button.js
```diff
@@ -143,6 +155,85 @@ const ButtonWithForwardRef: React$AbstractComponent<
...
+        onMouseUp={handleMouseUp}
+        onTouchCancel={handleTouchCancel}
+        onTouchEnd={handleTouchEnd}
+        onTouchMove={handleTouchMove}
+        onTouchStart={handleTouchStart}
+        ref={innerRef}
+        type={type}
+        {...(compressStyle ? { style: compressStyle } : {})}
+      >
+        {iconEnd ? iconEndComponent(iconEnd) : buttonText}
```
christianvuerings, line 224: @AlbertCarreras does this mean we can not have `iconEnd` on `Link`? I think that's a pretty big use case (e.g. external links that look like a button)

packages/gestalt/src/Button.js
```diff
@@ -28,54 +29,65 @@ const DEFAULT_TEXT_COLORS = {
...
     disabled = false,
-    iconEnd,
     inline = false,
-    name,
     onClick,
     selected = false,
     size = 'md',
     text,
     textColor: textColorProp,
-    type = 'button',
```
christianvuerings, line 77: @AlbertCarreras could we leave `type = 'button'` as the default? Otherwise the default becomes "submit" and this PR becomes a breaking change which will most likely cause bugs in Pinterest's codebase.

image

See https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#attr-type

docs/src/Button.doc.js
AlbertCarreras: Oh I know, I did this because it's how I'm testing while developing

packages/gestalt/src/Button.js
AlbertCarreras: It means that. We could just use href to define the type we return but I think for communication clarity it's easier to explain that we have 3 types each of them requiring specific props. `submit` requires name, `link` requires href, etc. 

My thoughts, we could use just href though but href excludes type submit/button then and it has to be described rather than being explicit with type link.

packages/gestalt/src/Button.js
AlbertCarreras: WIP. The finishings were not done.

packages/gestalt/src/Button.js
AlbertCarreras: Ok. Makes sense. It was easy for me to see if I was correcting it.

packages/gestalt/src/Button.js
AlbertCarreras: Agree. While coding it got in the mindset that icons were not needed for links. Again, WIP, finishings were not done.

packages/gestalt/src/Button.js
AlbertCarreras: I probably missed this. The intention was to leave the button as default. The draft was more of the first exploratory implementation. But thanks for catching that.

packages/gestalt/src/Button.js
AlbertCarreras: DONE

packages/gestalt/src/Button.js
AlbertCarreras: DONE

packages/gestalt/src/Button.js
AlbertCarreras: DONE

packages/gestalt/src/Button.js
AlbertCarreras: See line 231
image

AlbertCarreras: DONE

docs/src/Button.doc.js
```diff
@@ -9,281 +9,287 @@ import PageHeader from './components/PageHeader.js';
 const cards: Array<Node> = [];
 const card = c => cards.push(c);
 
+const buildDescription = (lines: Array<string>): Node => (
```
christianvuerings, line 12: @AlbertCarreras Instead of adding this to `Button.doc.js`, could we move it to `PropTable`? Basically if `description` is an array of strings, execute this logic.

docs/src/Button.doc.js
```diff
@@ -9,281 +9,287 @@ import PageHeader from './components/PageHeader.js';
...
+      {
+        name: 'accessibilityLabel',
+        type: 'string',
+        required: false,
+        defaultValue: null,
+        description: buildDescription([
+          'Supply a short, descriptive label for screen-readers to replace button texts that do not provide sufficient context about the button component behavior. Texts like `Click Here,` `Follow,` or `Read More` can be confusing when a screen reader reads them out of context. In those cases, we must pass an alternative text to replace the button text.',
+          'Use IconButton if you only need a button with an icon and no text.',
+          'Optional with button-type and submit-type buttons.',
+          'Accessibility: It populates aria-label.  Screen readers read the `accessibilityLabel` prop, if present, instead of `text`.',
```
christianvuerings, line 41: @AlbertCarreras could you replace the double spaces? `.  ` => `. `

docs/src/Button.doc.js
```diff
@@ -9,281 +9,287 @@ import PageHeader from './components/PageHeader.js';
...
+        ]),
+        href: 'accessibilityI',
```
christianvuerings, line 43: @AlbertCarreras instead, can we be more descriptive and link to `accessibilityLabel` instead?

Same for `accessibilityII`, which could be something like `accessibility` as the ID and `Accessibility: controls & expanded` as the title.

docs/src/Button.doc.js
```diff
@@ -9,281 +9,287 @@ import PageHeader from './components/PageHeader.js';
...
+          'Indicate that a button component controls the appearance of menus and expose whether they are currently opened or closed.',
+          'Optional with button-type buttons.',
+          'Accessibility: It populates aria-haspopup.',
+        ]),
+        href: 'accessibilityII',
       },
       {
         name: 'color',
-        type: `"blue" | "gray" | "red" | "transparent" | "white"`,
+        type: `gray' | 'red' | 'blue' | 'transparent' | 'white'`,
```
christianvuerings, line 83: @AlbertCarreras I think there is a `'` missing before `gray`

docs/src/Button.doc.js
```diff
@@ -9,281 +9,287 @@ import PageHeader from './components/PageHeader.js';
...
+          'Add a Gestalt icon to be displayed after the button text. Sometimes an icon can help clarify the usage of a button. Menus are a common use case.',
+          'Accessibility: Icons on buttons are not accessible for screen readers.',
+          'Use IconButton if you only need buttons with icons and no text.',
+        ]),
         href: 'iconEnd',
       },
       {
         name: 'inline',
         type: 'boolean',
+        required: false,
```
christianvuerings, line 113: @AlbertCarreras curious why we added `required: false`, is that not the default?

docs/src/Button.doc.js
```diff
@@ -9,281 +9,287 @@ import PageHeader from './components/PageHeader.js';
...
+          'Display a button in different sizes. Size only changes top & bottom padding modifying the component height: sm (32px), md (40px), lg (48px).',
+        href: 'height',
       },
       {
         name: 'text',
         type: 'string',
         required: true,
+        defaultValue: null,
+        description: buildDescription([
+          'Text to render inside the button to convey the function and purpose of the button. The button text has a fix size.',
```
christianvuerings, line 145: `fix size` => `fixed size`

docs/src/Button.doc.js
```diff
@@ -9,281 +9,287 @@ import PageHeader from './components/PageHeader.js';
...
+        defaultValue: null,
+        description: buildDescription([
+          'Primary colors to apply to the text inside the Button.',
+          'Text colors are automatically paired with the defined button `color`.',
+          '`textColor` should only be used when using a white background color; it may become necessary to manually override the text color to match that of the parent`s background. Do not override the text color for any button besides those with white background as it breaks design standards.',
+        ]),
         href: 'textColor',
       },
       {
         name: 'type',
```
christianvuerings, line 163: @AlbertCarreras thought a bit more about it over the weekend and I don't think we should switch on `type`. Mostly because we'll want a consistent API with `TapArea`.

What do you think about keeping `type` as is, with `button | submit` and introducing `role` with `button | link`?

```jsx
<TapArea role="link" href="https://pinterest.com" />
<Button role="link" href="https://pinterest.com" />
```

docs/src/Button.doc.js
```diff
@@ -9,281 +9,287 @@ import PageHeader from './components/PageHeader.js';
...
-        type: `"submit" | "button"`,
+        type: `'submit' | 'button' | 'link'`,
+        required: false,
         defaultValue: 'button',
-        href: 'types',
+        description: buildDescription([
+          'Select a button variant:',
+          '-`button`: Use to render simple push buttons with no default behavior and control custom functionality inside the `onClick` callback. The button is rendered as an button or  `<button> tag`.',
```
christianvuerings, line 169: Replace `an` with `a` and remove double spaces after `or`

```
The button is rendered as an button or  `<button> tag`.
```

```
The button is rendered as a button or `<button> tag`.
```

docs/src/Button.doc.js
```diff
@@ -9,281 +9,287 @@ import PageHeader from './components/PageHeader.js';
...
-        type: "React.Ref<'button'>",
-        description: 'Forward the ref to the underlying button element',
-        href: 'refExample',
+        required: false,
+        defaultValue: null,
+        description: buildDescription([
+          'Callback fired when a button component is clicked (pressed and released) with a mouse or keyboard.',
+          'Required with button-type buttons.',
+        ]),
+        href: 'selected',
```
christianvuerings, line 128: @AlbertCarreras should this `href` move to the `selected` prop?

docs/src/Button.doc.js
```diff
@@ -9,281 +9,287 @@ import PageHeader from './components/PageHeader.js';
...
+        ]),
+        href: '',
+      },
+      {
+        name: 'href',
+        type: 'string',
+        required: false,
+        defaultValue: null,
+        description: buildDescription([
+          'Specify a redirect url.',
```
christianvuerings, line 193: @AlbertCarreras not really a redirect, let's rephrase this as `Specify a link URL.`

docs/src/Button.doc.js
```diff
@@ -9,281 +9,287 @@ import PageHeader from './components/PageHeader.js';
...
-    <Box display="flex" direction="row" column={6} paddingX={2}>
-      <Button color="transparent" text="Later" />
-    </Box>
-    <Box column={6} paddingX={2}>
-      <Button color="white" text="Learn more" />
-    </Box>
-  </Box>
-</Box>
+<Row gap={1}>
+    <Tooltip text="button-type button">
```
christianvuerings, line 247: @AlbertCarreras could we fix the indentation to be 2 spaces?

packages/gestalt/src/Button.flowtest.js
```diff
@@ -4,8 +4,10 @@ import Button from './Button.js';
 
 const Valid = <Button text="Next" />;
 
-// $FlowExpectedError[prop-missing]
```
christianvuerings, line 7: @AlbertCarreras could we add a couple more tests of buttons that are correct and incorrect to test the flow types?

packages/gestalt/src/Button.js
```diff
@@ -28,54 +29,67 @@ const DEFAULT_TEXT_COLORS = {
...
   textColor?: 'white' | 'darkGray' | 'blue' | 'red',
-  type?: 'submit' | 'button',
 |};
 
+type ButtonType = {|
+  ...BaseButton,
+  accessibilityControls?: string,
+  accessibilityExpanded?: boolean,
+  accessibilityHaspopup?: boolean,
+  accessibilityLabel?: string,
```
christianvuerings, line 54: @AlbertCarreras `accessibilityLabel?: string` should be moved to `BaseButton`. It's totally fine to set a label on a `Link` but not required.

We can also add it to `Link` if that prop isn't on that component yet.

packages/gestalt/src/Button.js
```diff
@@ -142,7 +157,89 @@ const ButtonWithForwardRef: React$AbstractComponent<
...
+
   /* eslint-disable react/button-has-type */
+  if (props.type === 'link') {
+    const { href, rel, target } = props;
+
+    return (
+      <Link
+        hoverStyle="none"
+        href={href}
+        onClick={({ event }) => {
```
christianvuerings, line 190: @AlbertCarreras could we extract this `onClick` + the one below into 1 `handleClick`?

```jsx
function handleClick(event) {
  if (onClick) {
    onClick({event})
  }
}
///
<Link onClick={handleClick} />
```

packages/gestalt/src/Button.js
```diff
@@ -142,7 +157,89 @@ const ButtonWithForwardRef: React$AbstractComponent<
...
+          if (onClick) {
+            onClick({ event });
+          }
+        }}
+        ref={innerRef}
+        rel={rel}
+        target={target}
+        inline={inline}
+        rounding="pill"
+        tapStyle="compress"
```
christianvuerings, line 200: @AlbertCarreras doesn't this mean we have 2 compress styles?
1. Within `classes` on the wrapping `div`
2. Within `Link`

If so, we probably want to keep the one on `Link` and make the one on `classes` conditional.

```
[touchableStyles.tapCompress]: props.type !== 'link' && !disabled && isTapping
```

packages/gestalt/src/Button.js
```diff
@@ -142,7 +157,89 @@ const ButtonWithForwardRef: React$AbstractComponent<
...
+            icon={iconEnd}
+            size={SIZE_NAME_TO_PIXEL[size]}
+          />
+        </Box>
+      </Box>
+    );
+  };
+
   /* eslint-disable react/button-has-type */
+  if (props.type === 'link') {
```
christianvuerings, line 183: @AlbertCarreras instead of `props.type`, can we compare with `type` and destructure above?

packages/gestalt/src/__snapshots__/Toast.test.js.snap
```diff
@@ -106,7 +106,7 @@ exports[`<Toast /> Text + Image + Button 1`] = `
           onTouchEnd={[Function]}
           onTouchMove={[Function]}
           onTouchStart={[Function]}
-          type="button"
+          type="Button"
```
christianvuerings, line 109: @AlbertCarreras this is a bug - it should be `button` lowercase instead.

christianvuerings:
@AlbertCarreras seeing a double scrollbar on the proptable now. Could you investigate?

FYI we do need the `overflow` auto so the table becomes scrollable with a horizontal scrollbar.

image

docs/src/Button.doc.js
```diff
@@ -295,8 +301,7 @@ function MenuButtonExample() {
         <Button
           accessibilityExpanded={!!open}
           accessibilityHaspopup
-          color="white"
-          iconEnd="arrow-down"
+          iconEnd={open ? "arrow-down" : "arrow-forward"}
```
christianvuerings, line 304: @AlbertCarreras could we revert the dropdown icon change? Instead, we should toggle the `selected` state when the dropdown is open (please verify with design)

docs/src/Button.doc.js
```diff
@@ -428,42 +449,58 @@ function A11yExDisclosure() {
...
-function ButtonFlyoutExample() {
-  const [open, setOpen] = React.useState(false);
-  const [checked, setChecked] = React.useState(false);
+function AccessibilityExample() {
+  const [isFollowing, setIsFollowing] = React.useState(false);
 
-  const anchorRef = React.useRef(null);
+  return (
+    <Button
+      accessibilityLabel={isFollowing ? "Following Alberto on Pinterest" : "Follow Alberto on Pinterestlow"}
```
christianvuerings, line 460: @AlbertCarreras could we make this
```jsx
accessibilityLabel={isFollowing ? "Unfollow Alberto on Pinterest" : "Follow Alberto on Pinterest"}
text={isFollowing ? "Unfollow" : "Follow"}
```

docs/src/Button.doc.js
```diff
@@ -428,42 +449,58 @@ function A11yExDisclosure() {
...
       <Button
+        accessibilityControls={COLLAPSIBLE_MENU}
+        accessibilityExpanded={isOpen}
+        accessibilityHasPopup="true"
         inline
-        color="red"
-        onClick={() => setOpen(true)}
-        ref={anchorRef}
-        text="Submit"
+        iconEnd={isOpen ? "arrow-down" : "arrow-forward"}
```
christianvuerings, line 487: @AlbertCarreras same here - let's check with design whether we actually want to toggle the icon

docs/src/Button.doc.js
AlbertCarreras: That makes more sense. Thanks

docs/src/Button.doc.js
AlbertCarreras: You think... no... there isnt!

docs/src/Button.doc.js
AlbertCarreras: DONE

docs/src/Button.doc.js
```diff
@@ -12,217 +12,390 @@ const card = c => cards.push(c);
...
     defaultCode={`
-<Button text="Medium sized button" inline />
+<Row gap={1}>
+  <Tooltip text="default button: button-role + button-type button">
+    <Button onClick={() => {}} text="Clear search history" inline />
+  </Tooltip>
+  <Tooltip text="form button variant: button-role + submit-type button">
+    <Button type="submit" name='satisfaction-questionaire' text="Submit your response" inline />
+  </Tooltip>
+  <Tooltip text="link button variant: link-role button + button-type">
```
christianvuerings, line 253: Could we simplify the tooltip text? Just show `Link button` / `Submit button` / `Default button` in the tooltip. Also don't think `link-role button + button-type` is what we want

docs/src/Button.doc.js
```diff
@@ -12,217 +12,390 @@ const card = c => cards.push(c);
...
         name: 'size',
-        type: `"sm" | "md" | "lg"`,
+        type: `'sm' | 'md' | 'lg'`,
+        required: false,
         defaultValue: 'md',
-        description: 'sm: 32px, md: 40px, lg: 48px',
-        href: 'combinations',
+        description:
+          'Display a button in different sizes. Size only changes top & bottom padding modifying the component height: sm (32px), md (40px), lg (48px).',
+        href: 'height',
```
christianvuerings, line 126: @AlbertCarreras could we rename `height` to `sizes`. It not only updates the top/bottom padding but also left/right.

packages/gestalt/src/Button.js
```diff
@@ -122,9 +139,11 @@ const ButtonWithForwardRef: React$AbstractComponent<
       [styles.selected]: !disabled && selected,
       [styles.disabled]: disabled,
       [styles.enabled]: !disabled,
```
christianvuerings, line 141: @AlbertCarreras looks like a disabled `<Button href="" disabled />` does not work as expected:

* The button (link) is still clickable
* It still opens the link

image

packages/gestalt/src/Button.jsdom.test.js
```diff
@@ -13,10 +13,21 @@ describe('Button', () => {
     expect(mockOnClick).toHaveBeenCalled();
   });
 
-  it('forwards a ref to the innermost button element', () => {
-    const ref = React.createRef();
+  it('forwards a ref to the innermost button element with type button', () => {
```
christianvuerings, line 16: Could we add a test for a disabled Button with `href`?

docs/src/Button.doc.js
AlbertCarreras: DONE.

docs/src/Button.doc.js
AlbertCarreras: DONE

docs/src/Button.doc.js
AlbertCarreras: DONE

packages/gestalt/src/Button.js
AlbertCarreras: can't, it triggers Flow errors

docs/src/Button.doc.js
AlbertCarreras: DONE

docs/src/Button.doc.js
AlbertCarreras: DONE Yeah, didn't notice lg also changes left/right padding

packages/gestalt/src/Button.js
AlbertCarreras: I added _disable prop to Link so I can do `href={ _disable ? undefined : href}`

We can use this approach for id, role and, accessibilitySelected so we know they are private props and therefore not documented

packages/gestalt/src/Button.jsdom.test.js
AlbertCarreras: DONE

packages/gestalt/src/Button.js
AlbertCarreras: DONE

AlbertCarreras:
> @AlbertCarreras seeing a double scrollbar on the proptable now. Could you investigate?
> 
> FYI we do need the `overflow` auto so the table becomes scrollable with a horizontal scrollbar.
> 
> image
> 
> Looks like it's caused by `buildDescription`

DONE

docs/src/components/PropTable.js
```diff
@@ -98,7 +98,10 @@ export default function PropTable({
 
   return (
     <Card id="Props" name="Props" showHeading={showHeading}>
-      <Box overflow="auto">
+      <Box
+        overflow="auto"
+        dangerouslySetInlineStyle={{ __style: { overflowY: 'hidden' } }}
```
christianvuerings, line 103: @AlbertCarreras 🤭  - not for this diff but I've been thinking we should add `overflow="autoX"` and `overflow="autoY"` to `Box` so we don't need to do this hack.

packages/gestalt/src/Button.js
```diff
@@ -142,7 +161,108 @@ const ButtonWithForwardRef: React$AbstractComponent<
...
+  }
+
   /* eslint-disable react/button-has-type */
+  if (props.role === 'link') {
+    const { href, rel, target } = props;
+
+    return (
+      <Link
+        accessibilityLabel={accessibilityLabel}
+        _disabled={disabled}
```
christianvuerings, line 215: @AlbertCarreras Let's not make this a pattern and add `_` props for "private" props. People can still use them.

Instead, let's add `disabled` to `Link`

docs/src/Button.doc.js
```diff
@@ -12,217 +12,410 @@ const card = c => cards.push(c);
...
       },
       {
         name: 'inline',
         type: 'boolean',
+        required: false,
         defaultValue: false,
-        href: 'widths',
-      },
-      {
-        name: 'name',
```
christianvuerings, line 84: @AlbertCarreras let's make it explicit that we removed `name` from `Button` by adding a codemod which removes this prop. Even though we might not use it in the Pinterest codebase, we should still be due diligent.

christianvuerings:
@AlbertCarreras feel free to self merge after:
* Renaming `_disabled` on `Link` to `disabled`
* Adding a codemod to remove `name` from all `<Button />`s

docs/src/Button.doc.js
AlbertCarreras: @cvuerings, I didn't remove it. Just undocumented.

AlbertCarreras: We still have a few instances on Pinboard that we could rename to `id`

packages/gestalt/src/Button.js
AlbertCarreras: DONE

PR #1170, opened 2020-08-27
Docs: Add 'edit this page' link to every doc page

christianvuerings:
image

Makes it easier for people to make updates to a specific page. They only have to click the link on that page + make the updates on GitHub.

## Test Plan

1. Go to `/Box` in the Gestalt Docs: https://deploy-preview-1170--gestalt.netlify.app/Box
2. Scroll down to the bottom of the page
3. Click on `Edit this page`
4. It opens https://github.com/pinterest/gestalt/blob/master/docs/src/Box.doc.js

### DIFF
--- a/docs/src/components/CardPage.js
+++ b/docs/src/components/CardPage.js
@@ -1,48 +1,61 @@
 // @flow strict
 import React from 'react';
-import { Box, FixedZIndex, Sticky } from 'gestalt';
+import { Box, FixedZIndex, Link, Sticky, Row, Text } from 'gestalt';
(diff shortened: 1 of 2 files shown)


ashleyseto:
Since we don't tend to recommend icons for these scenarios in product, I think we can remove the icon. Do we want to be more specific such as "Edit page on GitHub"?

christianvuerings:
@ashleyseto updated - thank you!

Mostly added the icon to differentiate between the actual content of the page + the edit link

PR #1175, opened 2020-08-28
Modal: Add forwardRef

nachofassini:
The purpose of this PR is to add React `forwardRef` to `Modal` in order to be able to reference the Modal container from parent components

## Test Plan

Tested locally
There are no UI changes
Renders as expected

Screen Shot 2020-08-28 at 3 40 04 PM

### DIFF
--- a/docs/src/Modal.doc.js
+++ b/docs/src/Modal.doc.js
@@ -54,6 +54,12 @@ card(
         href: 'sizesExample',
       },
       {
+        name: 'ref',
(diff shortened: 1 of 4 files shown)


AlbertCarreras:
Hey Nacho, Could you document ref in the docs, add an example in the Docs, and a test? Follow the pattern from all the components with forwardRef. Let me know if  you have questions.

AlbertCarreras:
LGTM

PR #1178, opened 2020-08-31
Docs: fix Divider overlap on bottom of the page

christianvuerings:
### Before
See divider overlap on `SegmentedControl`

Screen Shot 2020-08-31 at 10 02 43 AM

### After
No divider overlap on `SegmentedControl`

Screen Shot 2020-08-31 at 10 02 55 AM

### DIFF
--- a/docs/src/components/App.js
+++ b/docs/src/components/App.js
@@ -65,10 +65,6 @@ export default function App(props: Props): Node {
               <Box padding={4} mdPadding={6} lgPadding={8} width="100%">
                 {children}
               </Box>
-            </Box>
-          </Box>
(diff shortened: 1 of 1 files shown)


docs/src/components/App.js
```diff
@@ -65,10 +65,6 @@ export default function App(props: Props): Node {
               <Box padding={4} mdPadding={6} lgPadding={8} width="100%">
                 {children}
               </Box>
-            </Box>
-          </Box>
-          {document.location.href.includes('netlify') ? (
```
christianvuerings, line 70: Lesson learned - our local docs should looks as similar as possible as the ones hosted on netlify. So we discover these issues early on.

PR #1180, opened 2020-09-01
Link: added accessibilityLabel / Docs: implemented support to Arrays in description 

AlbertCarreras:

### DIFF
--- a/docs/src/Link.doc.js
+++ b/docs/src/Link.doc.js
@@ -54,6 +54,17 @@ card(
         type: '() => void',
       },
       {
+        name: 'accessibilityLabel',
+        type: 'string',
+        required: false,
+        defaultValue: null,
+        description: [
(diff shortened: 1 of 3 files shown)


docs/src/Link.doc.js
```diff
@@ -126,8 +137,9 @@ card(
 
 card(
   <Example
+    id="accessibility"
     description={`
-    When providing the content for the link, avoid phrases like "click here" or "go to".
+    When providing the content for the link, avoid phrases like "click here" or "go to" or use accessibilityLabel.
```
christianvuerings, line 142: @AlbertCarreras can we remove the `or use accessibilityLabel`? We do want them to use `accessibilityLabel` when it's required

AlbertCarreras: DONE

PR #1181, opened 2020-09-01
Modal: Prevent showing zoom out icon if closeOnOutsideClick is false

nachofassini:
This fix the problem of showing the zoom out icon even when `closeOnOutsideClick` is set to `false`.
That behavior might confuse user thinking that if they click outside the modal this will be closed.

## Test Plan

Tested locally

Screen Shot 2020-09-01 at 3 11 42 PM
Screen Shot 2020-09-01 at 3 10 58 PM

### DIFF
--- a/docs/src/Modal.doc.js
+++ b/docs/src/Modal.doc.js
@@ -35,6 +35,7 @@ card(
         type: 'boolean',
         description: 'Close the modal when you click outside of it',
         defaultValue: true,
+        href: 'closeOnOutsideClickExample',
       },
       {
         name: 'footer',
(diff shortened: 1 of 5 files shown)


docs/src/Modal.doc.js
```diff
@@ -177,6 +181,49 @@ function Example(props) {
   />
 );
 
+card(
```
christianvuerings, line 184: @nachofassini fyi moved this example down - shouldn't be the first one that people see.

packages/gestalt/src/Modal.css
```diff
@@ -1,12 +1,15 @@
 .backdrop {
   composes: absolute left0 top0 right0 bottom0 from "./Layout.css";
-  composes: zoomOut from "./Cursor.css";
   background: rgba(0, 0, 0, 0.4);
   height: 100%;
   opacity: 0.9;
   overflow-y: scroll;
 }
 
+.zoomout {
```
christianvuerings, line 9: fyi renamed to `zoomOut`

docs/src/Modal.doc.js
```diff
@@ -177,6 +181,49 @@ function Example(props) {
...
+    description={`
+      Sometimes we need the user to complete some required action at a Modal.
+      We can ensure that by preventing the user closing the modal.
+      For that you can set: \`closeOnOutsideClick\` to \`false\`.
+    `}
+    defaultCode={`
+function Example(props) {
+  const [showModal, setShowModal] = React.useState(false);
+  return (
+    <>
```
christianvuerings, line 197: fyi no longer wrapped this in a `Box` with negative margin left. We only did that in the other examples because there were multiple buttons.

nachofassini:
Thanks @AlbertCarreras and @christianvuerings for those improvements!!

PR #1182, opened 2020-09-02
VideoControls: convert to function component

christianvuerings:
Convert `VideoControls` to a function component:

* Makes the code more readable
* Better flow type support

## Test Plan

1. Unit tests pass
2. Go to `/Video#videoUpdatesExample` and ensure the video controls still work

### DIFF
--- a/packages/gestalt/src/VideoControls.js
+++ b/packages/gestalt/src/VideoControls.js
@@ -1,5 +1,5 @@
 // @flow strict
-import React, { Component, type Node } from 'react';
+import React, { type Node } from 'react';
 import PropTypes from 'prop-types';
 import Box from './Box.js';
 import Icon from './Icon.js';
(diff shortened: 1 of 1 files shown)


packages/gestalt/src/VideoControls.js
```diff
@@ -87,12 +86,11 @@ class VideoControls extends Component<Props> {
       | SyntheticMouseEvent<HTMLDivElement>
       | SyntheticKeyboardEvent<HTMLDivElement>,
   |}) => {
-    const { onFullscreenChange } = this.props;
     event.stopPropagation();
     onFullscreenChange();
   };
 
-  handlePlayingChange: ({|
+  const handlePlayingChange: ({|
```
AlbertCarreras, line 93: Could we add

`type eventType =  SyntheticMouseEvent<HTMLDivElement> | SyntheticKeyboardEvent<HTMLDivElement`

and reuse it?

AlbertCarreras:
@christianvuerings 

Should we try to standardize all `onChange`-type  types to `AbstractEventHandler<SyntheticMouseEvent<HTMLButtonElement>>`

christianvuerings:
@AlbertCarreras agreed that those would be good changes but out of scope for this PR.

PR #1185, opened 2020-09-02
[WIP] IconButton: Added support for href and segmented Flow type

AlbertCarreras:
- Added support to href to IconButton.
- Segmented Flow for each IconButton type to make Flow Type stricter in IconButton.
- Upgraded IconButton Docs to add new props. >> There will be follow up
- Added test.

### DIFF
--- a/docs/src/IconButton.doc.js
+++ b/docs/src/IconButton.doc.js
@@ -79,6 +79,17 @@ card(
         Icon.`,
       },
       {
+        name: 'href',
+        type: 'string',
+        required: false,
+        defaultValue: null,
+        description: [
+          'Specify a link URL.',
+          'Required with link-role buttons.',
+        ],
+        href: 'roles',
+      },
+      {
         name: 'onClick',
         type: '({ event: SyntheticMouseEvent<> }) => void',
       },
@@ -89,10 +100,31 @@ card(
         href: 'paddingCombinations',
       },
       {
+        name: 'rel',
+        type: `'none' | 'nofollow'`,
+        required: false,
+        defaultValue: 'none',
+        description: 'Optional with link-role buttons.',
+        href: 'roles',
+      },
+      {
         name: 'ref',
-        type: "React.Ref<'button'>",
+        type: `React.Ref<'button'> | React.Ref<'a'>`,
         description: 'Forward the ref to the underlying button element',
         href: 'refExample',
+      },
+      {
+        name: 'role',
+        type: `'button' | 'link'`,
+        required: false,
+        defaultValue: 'button',
+        description: [
+          `Select a button variant:`,
+          `- 'button': Use to render 'submit' or 'button'-type buttons. The button is rendered as a '<button>'.`,
+          `- 'link': Use for buttons to act like links. The button is rendered as an '<a>'.`,
+          `Required with link-role buttons.`,
+        ],
+        href: 'roles',
       },
       {
         name: 'selected',
@@ -106,6 +138,20 @@ card(
         description: `xs: 24px, sm: 32px, md: 40px, lg: 48px, xl: 56px`,
         defaultValue: 'md',
         href: 'sizeCombinations',
+      },
+      {
+        name: 'target',
+        type: `null | 'self' | 'blank'`,
+        required: false,
+        defaultValue: 'null',
+        description: [
+          'Define the frame or window to open the anchor defined on `href`:',
+          '- `null` opens the anchor in the same window.',
+          '- `blank` opens the anchor in a new window.',
+          '- `self` opens an anchor in the same frame.',
+          'Optional with link-role buttons.',
+        ],
+        href: 'roles',
       },
     ]}
   />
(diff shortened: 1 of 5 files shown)


docs/src/IconButton.doc.js
```diff
@@ -126,6 +172,43 @@ card(
...
+    id="roles"
+    defaultCode={`
+function Example() {
+  const [disabled, setDisabled] = React.useState(false);
+
+  return (
+    <Stack gap={3}>
+      <Row gap={1}>
+        <Tooltip text="Default IconButton">
+          <IconButton accessibilityLabel='Default IconButton' icon='share' onClick={() => {}} disabled={disabled} />
```
christianvuerings, line 187: @AlbertCarreras could we change the example code to use double quotes? `<IconButton accessibilityLabel="Default IconButton"``

packages/gestalt/src/IconButton.js
```diff
@@ -87,9 +107,75 @@ const IconButtonWithForwardRef: React$AbstractComponent<
...
+  };
+
+  if (props.role === 'link') {
+    const { href, rel, target } = props;
+
+    return (
+      <Link
+        accessibilityLabel={accessibilityLabel}
+        disabled={disabled}
+        inline
```
christianvuerings, line 148: @AlbertCarreras could you double check whether this is what we want? Currently on the `IconButton role="button"` we set it to `display="block"`

image

But we set it to `inline` for `role="link"`

image

Example:
```
function Example() {

  return (
    <>
      <IconButton accessibilityLabel='Default IconButton' icon='share' onClick={() => {}} />
      <IconButton accessibilityLabel='Default IconButton' icon='share' onClick={() => {}} />
      <IconButton accessibilityLabel='Link IconButton' icon='link' role="link" target="blank" href="https://www.pinterest.com" />
      <IconButton accessibilityLabel='Link IconButton' icon='link' role="link" target="blank" href="https://www.pinterest.com" />
    </>
  );
}
```

Results in this:
image

packages/gestalt/src/IconButton.js
```diff
@@ -87,9 +107,75 @@ const IconButtonWithForwardRef: React$AbstractComponent<
...
+        href={href}
+        onClick={handleLinkClick}
+        ref={innerRef}
+        rel={rel}
+        rounding="circle"
+        tapStyle={disabled ? undefined : 'compress'}
+        target={target}
+      >
+        <div
```
christianvuerings, line 157: @AlbertCarreras setting the classes on an inner div breaks the focus styles, e.g. the icon background does not become darker when focussed.

For `role="button"`
image


For `role="link"`
image

We have the same issue with `<Button />`, so we might have to move away from using `<Link />` and use a regular `<a />` instead. And I would recommend we focus on getting `<Button />` right first.

packages/gestalt/src/IconButton.js
```diff
@@ -87,9 +107,75 @@ const IconButtonWithForwardRef: React$AbstractComponent<
...
+          onMouseEnter={() => setHovered(true)}
+          onMouseLeave={() => {
+            setActive(false);
+            setHovered(false);
+          }}
+          className={classes}
```
christianvuerings, line 163: Why did we not add `onMouseUp` and `onMouseDown`? It might be good to extra all of these functions (since they are now re-used`.

`onMouseEnter={handleMouseEnter}`

packages/gestalt/src/IconButton.js
```diff
@@ -87,9 +107,75 @@ const IconButtonWithForwardRef: React$AbstractComponent<
...
+    if (onClick) {
+      onClick({ event });
+    }
+  }
+
+  function handleLinkClick({ event }) {
+    handleClick(event);
+  }
+
+  const renderPogComponent = (selected?: boolean): Node => {
```
christianvuerings, line 124: @AlbertCarreras this looks like a React component but it's not. Instead can we make it a real component?

`const PogWithSelected = ({props}): Node =>`

`<PogWithSelected selected="..." />`

packages/gestalt/src/IconButton.jsdom.test.js
```diff
@@ -1,13 +1,58 @@
...
+    );
+    getByRole('button').click();
+    expect(mockOnClick).toHaveBeenCalled();
+  });
+
+  it('renders a button and forwards a ref to the innermost <button> element', () => {
+    const ref = createRef();
     render(<IconButton disabled accessibilityLabel="test" ref={ref} />);
     expect(ref.current instanceof HTMLButtonElement).toEqual(true);
+    // $FlowIgnore[prop-missing]
```
christianvuerings, line 20: @AlbertCarreras could we remove these flow ignores? For every one we add, we need to think about ways to avoid it. Looks like we do the same for `<Button / >`, so we should fix those as well:

```js
    expect(ref.current instanceof HTMLAnchorElement).toEqual(true);
    if (ref.current instanceof HTMLAnchorElement) {
      expect(ref.current?.href).toEqual('http://www.pinterest.com/');
    }
```

packages/gestalt/src/Button.js
```diff
@@ -248,7 +247,7 @@ const ButtonWithForwardRef: React$AbstractComponent<
         onTouchStart={handleTouchStart}
         ref={innerRef}
         style={compressStyle}
-        type={type}
+        type={type} // eslint-disable-line react/button-has-type
```
christianvuerings, line 250: @AlbertCarreras let's remove this `eslint-disable-line` completely and explicitly pass in the type:

```
if (props.type === 'submit') {
  return <button type="submit" />
}
```

AlbertCarreras: DONE in Button fix.

docs/src/IconButton.doc.js
AlbertCarreras: DONE

packages/gestalt/src/IconButton.js
AlbertCarreras: DONE

packages/gestalt/src/IconButton.js
AlbertCarreras: DONE

packages/gestalt/src/IconButton.js
AlbertCarreras: DONE

packages/gestalt/src/IconButton.js
AlbertCarreras: DONE

packages/gestalt/src/IconButton.jsdom.test.js
AlbertCarreras: DONE

PR #1186, opened 2020-09-03
ButtonGroup: Add component

christianvuerings:
# ButtonGroup

A simple, straightforward component to group a series of buttons.

Screen Shot 2020-09-03 at 7 52 05 AM

## Notes
* Didn't use `<Stack />` or `<Row />` since we need a combination of both. There should be horizontal & vertical (when it wraps) spacing (8px).

## Test Plan

- [x] Unit tests pass
- [ ] Approved by design /cc @ashleyseto

### DIFF
--- a/packages/gestalt/src/ButtonGroup.js
+++ b/packages/gestalt/src/ButtonGroup.js
@@ -1,0 +1,41 @@
+// @flow strict
+
+import React, { Children, type Node } from 'react';
+import PropTypes from 'prop-types';
+import Box from './Box.js';
+
+function ButtonGroup({ children }: {| children?: Node |}): Node {
+  const count = Children.count(children);
+
+  if (count === 0) {
+    return null;
+  }
+  if (count === 1) {
+    return <>{children}</>;
+  }
+
+  return (
+    <Box
+      marginStart={-1}
+      marginRight={-1}
+      marginTop={-1}
+      marginBottom={-1}
+      display="flex"
+      wrap
+    >
+      {Children.map(children, child =>
+        child !== null && child !== undefined ? (
+          <Box paddingX={1} paddingY={1}>
+            {child}
+          </Box>
+        ) : null
+      )}
+    </Box>
+  );
+}
+
+ButtonGroup.propTypes = {
+  children: PropTypes.node,
+};
+
+export default ButtonGroup;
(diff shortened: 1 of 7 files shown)


ashleyseto:
Yay button group! I know this is layout technically but I'm wondering if we should put it still under forms so people find it. When someone is looking for buttons, will they know we have button group in another section?

ashleyseto:
Adding also the same comment from forms group to the fact that we won't want the padding all the way around as it will cause alignment issues on the page.

christianvuerings:
> Yay button group! I know this is layout technically but I'm wondering if we should put it still under forms so people find it. When someone is looking for buttons, will they know we have button group in another section?

@ashleyseto Agreed - I was doubting about this too. Moved it to `Forms` instead

image

> Adding also the same comment from forms group to the fact that we won't want the padding all the way around as it will cause alignment issues on the page.

@ashleyseto in this case there was no padding all the way around. Only between the buttons (whether they are wrapped or not). It might be confusing that this is the case in the docs, since each docs example does have a 16px padding (not part of the component).

The `Wrap` example might make this more clear. Here we show a 1px border around the container which has a fixed width of `150px` (to showcase what wrapping looks like)

image

ashleyseto (approved):

packages/gestalt/src/ButtonGroup.js
```diff
@@ -0,0 +1,41 @@
...
+  return (
+    <Box
+      marginStart={-1}
+      marginRight={-1}
+      marginTop={-1}
+      marginBottom={-1}
+      display="flex"
+      wrap
+    >
+      {Children.map(children, child =>
```
AlbertCarreras, line 26: Should we check the children are Buttons? 
image

christianvuerings: @AlbertCarreras when we tried that in the past, it always ended up being to restrictive (e.g checking with flow / doing a typeof):
* When people want to experiment with a new button in the Pinterest codebase.
* When someone wants to wrap the button with a flyout: `<Flyout><Button / ></Flyout`
* When you wrap it in a conditional, depending on an experiment `{inExperiment ? <Button /> : null}`

AlbertCarreras: Makes sense.

AlbertCarreras (approved):

PR #1187, opened 2020-09-03
Layouts: Add form layout example

christianvuerings:
# Layouts

* Add a "Layouts" page
* Add a `Form Layout` example which is responsive, works with RTL and dark mode

layouts-form-layout

## Test Plan

- [ ] Test on netlify as soon as the preview goes out
- [ ] Approved by design /cc @ashleyseto @anniemochi

### DIFF
--- a/docs/src/Layouts.doc.js
+++ b/docs/src/Layouts.doc.js
@@ -1,0 +1,122 @@
+// @flow strict
+import React, { type Node } from 'react';
+import Example from './components/Example.js';
+import PageHeader from './components/PageHeader.js';
+
+const cards: Array<Node> = [];
+const card = c => cards.push(c);
+
+card(
+  <PageHeader
+    name="Layouts"
+    description="A list of easy-to-copy layouts which have been battle tested."
+  />
+);
+
+card(
+  <Example
+    description="Responsive &amp; RTL friendly form layout."
+    name="Form Layout"
+    defaultCode={`
+<Box
+  display="flex"
+  marginLeft={-3}
+  marginRight={-3}
+  marginBottom={-3}
+  marginTop={-3}
+  wrap
+  width="100%"
+  padding={3}
+  direction="column"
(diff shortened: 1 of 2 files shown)


ashleyseto:
I think it looks correct. I'm not sure about the implementation though. I would expect it to be 24px bottom rather than around. It is adding space to the left and the right that might not be wanted when trying to line up content inside containers like the Module that should already have padding defined.

christianvuerings:
> I think it looks correct. I'm not sure about the implementation though. I would expect it to be 24px bottom rather than around. It is adding space to the left and the right that might not be wanted when trying to line up content inside containers like the Module that should already have padding defined.

@ashleyseto agreed - I removed the padding around so it will work within other containers which already have padding set. As to the bottom (and left/right) - we do use symmetrical padding in Gestalt since:
* It automatically looks good when the elements wrap one below the other
* Works with RTL
* There is no need to know whether the next row will be rendered which helps with composability of each row. (We see this use case quite often on the Pinterest settings page)

ashleyseto (approved):

AlbertCarreras:
@christianvuerings https://www.loom.com/share/2a0615667601470b8cc39bb314e9b00a Not an expert on accessibility at all. Should the screen reader read the form title?.

christianvuerings:
> @christianvuerings https://www.loom.com/share/2a0615667601470b8cc39bb314e9b00a Not an expert on accessibility at all. Should the screen reader read the form title?.

@AlbertCarreras Yes, that is correct and it does read it when you instruct the screenreader to do so.

See https://webaim.org/articles/voiceover/#reading for how to navigate with voiceover. When running through, I usually use
`VO` + `→`: `Read next item`
When you simply tab through, it will only focus on the elements that can have focus (e.g. TextField / Button)

screenreader-form-layouts

AlbertCarreras (approved):

PR #1189, opened 2020-09-04
Layer: convert to function component

christianvuerings:
Convert `<Layer />` to a function component. Makes the code more readable + consistent with other components in our codebase.

## Test Plan

1. Jest tests pass for `<Layer />`
2. Go to `/Layer` in the docs & verify the layer shows + hides correctly

image

### DIFF
--- a/packages/gestalt/src/Layer.js
+++ b/packages/gestalt/src/Layer.js
@@ -1,54 +1,36 @@
 // @flow strict
-import { Component, type Portal, type Node } from 'react';
+import { useRef, useState, type Portal, type Node, useEffect } from 'react';
 // flowlint-next-line untyped-import:off
 import { createPortal } from 'react-dom';
 
-type Props = {|
+export default function Layer({
+  children,
+}: {|
   children: Node,
-|};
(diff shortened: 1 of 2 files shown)


packages/gestalt/src/Layer.js
```diff
@@ -1,54 +1,36 @@
...
-  el: HTMLDivElement;
-
-  constructor(props: Props) {
-    super(props);
     if (typeof document !== 'undefined' && document.createElement) {
-      this.el = document.createElement('div');
-    } else {
-      // eslint-disable-next-line no-console
-      console.warn(
-        'Using Layer without document present. Children will not be rendered.'
```
jennyscript, line 28: The idea for this was to provide a warning if you tried the use `Layer` during SSR since that wouldn't work. Do you think that isn't useful?

christianvuerings: @jennyscript this is the only place where we do this and I think it would be better for it not to become a pattern. Ideally we would solve this through linting instead of adding `console.warn` / `console.log`

PR #1188, opened 2020-09-03
Button: deprecated textColor prop values

AlbertCarreras:
- Deprecated `textColor` prop color values used for `white` & `transparent` background buttons. White/Transparent backgrounds default to `darkGray`.
- New color value `color = 'transparentWhiteText'` for transparent background buttons with white text 
image

## Before
image
image

## After
image
image

# Codemod

## From
```
       <Box>
      <Button text="Medium-sized button" inline textColor="red"/>
      <Button text="Medium-sized button" inline color="transparent"/>
      <Button text="Medium-sized button" inline textColor="white" color="transparent" />
      <Button text="Medium-sized button" inline textColor="white" color="red"/>
      <Renamed text="Medium-sized button" inline textColor="red"/>
    </Box>
```
## To
```
    <Box>
      <Button text="Medium-sized button" inline />
      <Button text="Medium-sized button" inline color="transparent"/>
      <Button text="Medium-sized button" inline color="transparentWhiteText" />
      <Button text="Medium-sized button" inline color="red" />
      <Renamed text="Medium-sized button" inline />
    </Box>
```

Codemod helper, run
yarn codemod --parser=flow -t=packages/gestalt-codemods/12.6.0-12.7.0/button-remove-textColor-prop.js relative/path/to/your/code

PR #1190, opened 2020-09-04
Button: link-role button logic refactor

AlbertCarreras:
This PR:

- creates an `InternalLink` component, a `Link` variant to use in `Button` (next in `IconButton`, `TapArea`). This solution fixes: https://github.com/pinterest/gestalt/pull/1185#discussion_r483676872

- refactores `iconEnd` function into `IconEnd` component

- other minor refactors


packages/gestalt/src/Button.js
```diff
@@ -247,10 +246,19 @@ const ButtonWithForwardRef: React$AbstractComponent<
         onTouchMove={handleTouchMove}
         onTouchStart={handleTouchStart}
         ref={innerRef}
-        style={compressStyle}
-        type={type}
+        style={compressStyle || undefined}
```
AlbertCarreras, line 249: https://github.com/pinterest/gestalt/pull/1188/files#r483664035

packages/gestalt/src/WrapperLink.js
```diff
@@ -0,0 +1,163 @@
+// @flow strict
```
christianvuerings, line 1: @AlbertCarreras could we rename this to `InternalLink`

AlbertCarreras: DONE

christianvuerings (approved):

PR #1191, opened 2020-09-04
Internal: make incremental builds 4x faster

christianvuerings:
Make incremental builds (when developing locally) 4x faster than before.

### Before: 13 seconds
before-13s

### After: 3 seconds
after-3s

## Notes

We had to disable rollup watch mode in #877 because it kept appending to the output CSS file instead of updating what is required which caused for an inconsistent dev experience and having to restart a lot.

We fixed this issue by using a cache and setting the id to the CSS filename.

## Downsides

Since gestalt-datepicker relies on gestalt, we will momentarily see errors on every incremental change. Those do get resolved though, as soon as `gestalt.js` is build.

## Test Plan

* Run `yarn start` locally
* Make a change
* Changes should take about 3 seconds (compared to 13+ seconds before)

### DIFF
--- a/packages/gestalt-core/build.js
+++ b/packages/gestalt-core/build.js
@@ -1,15 +1,12 @@
 // eslint-disable-next-line flowtype/require-valid-file-annotation
-import babel from 'rollup-plugin-babel';
+import babel from '@rollup/plugin-babel';
 import cssnano from 'cssnano';
-import filesize from 'rollup-plugin-filesize';
-import gzip from 'gzip-size';
-import json from 'rollup-plugin-json';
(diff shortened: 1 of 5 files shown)


packages/gestalt-core/build.js
```diff
@@ -38,35 +35,11 @@ const svgPath = () => ({
   },
 });
 
-const statsPlugin = () => {
```
christianvuerings, line 41: Removed all the stats plugins since they accounted for 1-2 seconds for every incremental change

jennyscript (approved):

PR #1193, opened 2020-09-04
Row, Stack: Re-implement for better `gap`

dangerismycat:
The current implementation of `gap` is very limiting, in that children cannot declare any flex properties (e.g. `flex="grow"`) that are respected by the parent. This PR re-implements `gap` using the CSS child combinator, thereby adding the padding styles to the existing child styles rather than inserting an additional div.

This also required extracting Box's transforms for use in Row and Stack. Given the now-exported functions, Flow had a lot to say about this, which I updated as best I could but still had to add a handful of suppressions.

(Apologies for the size of this PR, but Box is a pretty verbose component)


packages/gestalt/src/Box.js
```diff
@@ -23,30 +23,56 @@ import React, {
...
-  range,
-  toggle,
-  mapping,
-  rangeWithZero,
-  rangeWithoutZero,
-} from './transforms.js';
-import { getRoundingStyle } from './getRoundingClassName.js';
+import { buildStyles } from './boxTransforms.js';
 import { type Indexable } from './zIndex.js';
+import {
```
dangerismycat, line 28: moved all types to boxTypes for easier use in Row/Stack

packages/gestalt/src/Box.js
```diff
@@ -273,480 +194,29 @@ type PropType = {
 
 /*
 
-# Transformers
```
dangerismycat, line 276: moved to boxTransforms for use in Row/Stack

packages/gestalt/src/Box.js
```diff
@@ -273,480 +194,29 @@ type PropType = {
...
-  },
-};
-
-/*
-
 # The Component
 
 */
 
-const contains = (key, arr) => arr.indexOf(key) >= 0;
```
dangerismycat, line 692: moved to /utils

packages/gestalt/src/Box.js
```diff
@@ -273,480 +194,29 @@ type PropType = {
...
-const BoxWithForwardRef: AbstractComponent<
-  PropType,
+const BoxWithForwardRef: AbstractComponent<Props, HTMLDivElement> = forwardRef<
+  Props,
   HTMLDivElement
-> = forwardRef<PropType, HTMLDivElement>(function Box(
-  props,
-  ref
-): Element<'div'> {
-  // Flow can't reason about the constant nature of Object.keys so we can't use
```
dangerismycat, line 718: also moved to boxTransforms

packages/gestalt/src/Row.js
```diff
@@ -39,37 +42,23 @@ type Props = {|
...
-  justifyContent = 'start',
-  width,
-  ...rest
-}: Props): Node {
+// Defaults for align-items and justify-content set in base styles
+export default function Row({ children, gap = 0, ...rest }: Props): Node {
+  const baseClasses = classnames(
+    styles.row,
+    styles[`horizontalGap${gap}`],
+    styles[`horizontalGap${gap} > *`]
```
dangerismycat, line 50: using the child combinator to prevent needing to add another div. this allows children to set styles like `flex="grow"`

packages/gestalt/src/boxTypes.js
```diff
@@ -0,0 +1,253 @@
+// @flow strict
+import PropTypes from 'prop-types';
```
dangerismycat, line 2: moved all Box Flow types here, and most proptypes as well. a central location should make future updates easier

packages/gestalt/src/stackDisallowedProps.js
```diff
@@ -0,0 +1,80 @@
+// @flow strict
+
+const disallowedProps = [
+  // Default
+  'onClick',
+  'className',
+  'style',
+  // Extra Box props
+  'dangerouslySetInlineStyle',
```
dangerismycat, line 9: this…isn't great. but the alternative is switching to an allowlist, which would involve auditing every non-standard prop currently passed to Box. though verbose, this seems like the better solution

packages/gestalt/src/utils/omit.js
```diff
@@ -0,0 +1,16 @@
+// @flow strict
+
+const contains = (key, arr) => arr.indexOf(key) >= 0;
+// $FlowExpectedError[unclear-type]
+const omit = (keys: Array<string>, obj: Object): Object =>
```
dangerismycat, line 5: we can't know anything about the object used here, so I don't see a way around this suppression

packages/gestalt/src/boxTransforms.js
```diff
@@ -0,0 +1,648 @@
...
+
+# Transformers
+
+This is where the meat and the bones of Box's transforms are. You can read more about the DSL in `./transforms.js`, but basically they are a small declarative way of specifying how a property (i.e. `marginTop={4}`) gets turned into a CSS class (`marginTop4`).
+
+There's a little preamble here, but it culminates in a big object mapping the actual property names to the transformer values.
+
+*/
+
+const display: Functor<Display> = mapping({
```
dangerismycat, line 62: this…is unfortunate. Flow couldn't figure out the types for almost any of these when defined in the object literal, so I had to define them here to type the function outputs. why Flow was losing the types from the function definitions, I don't know.

packages/gestalt/src/boxTransforms.js
```diff
@@ -0,0 +1,648 @@
...
+const columnRange = range('Col');
+
+/*
+
+It's preferable to put new properties into that object directly just so it's easier to read.
+
+*/
+
+export const propToFn = {
+  xs: (value: ?ResponsiveProps): Style => {
```
dangerismycat, line 439: contrary to the other fields, Flow gave me a lot of grief when I tried to abstract these to a `buildResponsiveStyles` helper, so they remain inline

packages/gestalt/src/boxTransforms.js
```diff
@@ -0,0 +1,648 @@
...
+
+This is where it all comes together. This function takes the base styles for the component,
+the component's props, and any disallowed props. It outputs the passthrough props (after
+removing disallowed and given props), as well as an object with the combined classNames
+and styles from the given props.
+
+*/
+
+// $FlowExpectedError[unclear-type]
+export function buildStyles<T: Object>(
```
dangerismycat, line 603: we know that the generic type will be an object, but we don't know anything else about it

packages/gestalt/src/boxTransforms.js
```diff
@@ -0,0 +1,648 @@
...
+  // to it (which is typically small) instead of the total number of possible
+  // properties (~30 or so). While it may ~feel~ like Box is inefficient, its
+  // biggest performance impact is on startup time because there's so much code
+  // here.
+
+  // eslint-disable-next-line no-restricted-syntax
+  for (const prop in props) {
+    if (
+      Object.prototype.hasOwnProperty.call(propToFn, prop) &&
+      !omitProps.includes(prop)
```
dangerismycat, line 635: this is new, ensuring that disallowed props aren't just not passed through, but also aren't applied (this is what keeps the disallowed Box props from working)

packages/gestalt/src/Box.js
```diff
@@ -761,180 +231,17 @@ And we're done here :)
...
-  }),
-  md: PropTypes.exact({
-    display: PropTypes.oneOfType([PropTypes.bool, SizeDisplayPropType]),
-    column: PropTypes.number,
-  }),
-  lg: PropTypes.exact({
-    display: PropTypes.oneOfType([PropTypes.bool, SizeDisplayPropType]),
-    column: PropTypes.number,
-  }),
+  xs: ResponsivePropsPropType,
```
AlbertCarreras, line 241: this was removed! something is incorrect here.

AlbertCarreras: make sure we are not reintroducing this ones.

packages/gestalt/src/stackDisallowedProps.js
```diff
@@ -0,0 +1,80 @@
...
+  'xs',
```
AlbertCarreras, line 10: 'xs', 'sm', 'md', 'lg' were removed from `Box`

christianvuerings:
@dangerismycat thanks for putting this together.

Main feedback is that Stack / Row now will start to leak abstractions since we use the CSS immediate child selector. For instance in this case, how much padding is there on the inner boxes?

```jsx
<Stack gap={1}>
  <Box paddingY={3}>Test 1</Box>
  <Box paddingY={3}>Text 2</Box>
</Stack>
```

packages/gestalt/src/Stack.js
```diff
@@ -39,37 +42,23 @@ type Props = {|
...
-  justifyContent = 'center',
-  width,
-  ...rest
-}: Props): Node {
+// Defaults for align-items and justify-content set in base styles
+export default function Stack({ children, gap = 0, ...rest }: Props): Node {
+  const baseClasses = classnames(
+    styles.stack,
+    styles[`verticalGap${gap}`],
+    styles[`verticalGap${gap} > *`]
```
christianvuerings, line 50: @dangerismycat do we need this here? I think we should be able to remove it since `verticalGap > *` is not a valid classname.

packages/gestalt/src/__snapshots__/Row.test.js.snap
```diff
@@ -2,172 +2,44 @@
...
 exports[`Row aligns items 1`] = `
 <div
-  className="box marginEnd0 marginStart0"
-  style={
-    Object {
-      "height": undefined,
-      "width": undefined,
-    }
-  }
+  className="itemsEnd row horizontalGap0 horizontalGap0 > *"
```
christianvuerings, line 5: @dangerismycat same for horizontalGap, I think we can remove the `horizontalGap0 > *`

packages/gestalt/src/Stack.js
dangerismycat: agreed, removed

packages/gestalt/src/__snapshots__/Row.test.js.snap
dangerismycat: agreed, removed

packages/gestalt/src/Row.js
```diff
@@ -39,37 +42,24 @@ type Props = {|
...
-  width,
-  ...rest
-}: Props): Node {
+// Defaults for align-items and justify-content set in base styles
+export default function Row({ children, gap = 0, ...rest }: Props): Node {
+  const hasGap = gap > 0;
+
+  const baseClasses = classnames(styles.row, {
+    [`horizontalGap${gap}`]: hasGap,
+    [`horizontalGap${gap} > *`]: hasGap,
```
christianvuerings, line 51: @dangerismycat can we remove this line? Otherwise we see the following:

```
className="row horizontalGap4 horizontalGap4 > *"
```

So `horizontalGap4 > *` should get removed

packages/gestalt/src/Stack.js
```diff
@@ -39,37 +42,24 @@ type Props = {|
...
-  width,
-  ...rest
-}: Props): Node {
+// Defaults for align-items and justify-content set in base styles
+export default function Stack({ children, gap = 0, ...rest }: Props): Node {
+  const hasGap = gap > 0;
+
+  const baseClasses = classnames(styles.stack, {
+    [`verticalGap${gap}`]: hasGap,
+    [`verticalGap${gap} > *`]: hasGap,
```
christianvuerings, line 51: @dangerismycat same here, can we remove this line?

christianvuerings (approved):

PR #1198, opened 2020-09-08
Internal: fix 'undefined' CSS in future CSS

christianvuerings:
Bug caused by #1191 - the legacy CSS was correct but the future CSS was not.

We were seeing the following CSS in the future CSS output:
```
undefined.Qoc{padding-bottom:8px}
```

### Before

image

### After

image

## Test Plan

https://deploy-preview-1198--gestalt.netlify.app/TextField
* Gestalt form components have 8px padding below the label.

### DIFF
--- a/packages/gestalt-core/build.js
+++ b/packages/gestalt-core/build.js
@@ -125,7 +125,7 @@ const cssModules = (options = {}) => {
       });
       const process = postcssParser.process(code, opts).then(result => {
         // Set CSS for specific file
-        cssCache[id] += result.css;
+        cssCache[id] = result.css;
       });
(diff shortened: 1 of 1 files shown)


packages/gestalt-core/build.js
```diff
@@ -125,7 +125,7 @@ const cssModules = (options = {}) => {
       });
       const process = postcssParser.process(code, opts).then(result => {
         // Set CSS for specific file
-        cssCache[id] += result.css;
+        cssCache[id] = result.css;
```
christianvuerings, line 128: Compare with line 116 where we do this correctly:

```
cssIE11Cache[id] = result.css;
```

PR #1201, opened 2020-09-09
Internal: BREAKING only build css file with css vars

jennyscript:
After experimenting in the main repo with the gestalt-future.css file that has css vars in it, we've determined this is safe to ship. I'm marking this as a breaking change because any app using gestalt that needs to support older browsers such as IE11 will need to use a polyfill. I've already added this one to our app's build for legacy browsers: https://github.com/nuxodin/ie11CustomProperties

Possible improvement: We still have `:root` declarations spread across a few css files. This is nice from a code organization perspective (border width vars defined in Borders.css, color vars defined in Colors.css for example) but makes less sense in the output css file. Perhaps we could combine the declarations into one Variables.css file? Some vars like the ones specific to the datepicker could be scoped to a specific class rather than globally as well. This PR is safe to ship, so I could make these changes in a follow up.

I also noticed markdown syntax highlighting wasn't working when I added documentation for the polyfill. This fixes that too.

## Test Plan
The docs are using the new css file. I've confirmed the docs look as expected and I can see the css vars in the built css file. Switching to dark mode works correctly indicating the css vars are being overwritten by the color scheme provider properly.

### DIFF
--- a/docs/src/Faq.doc.js
+++ b/docs/src/Faq.doc.js
@@ -115,16 +115,17 @@ import { Button, Text } from 'gestalt';
         set the polyfill up in the docs which should go in the &lt;head /&gt; of
         your html.
         <Markdown
-          text="
+          text={`
 ~~~js
 // Load polyfills for IE 11
-if (/MSIE \d|Trident.*rv:/.test(navigator.userAgent)) {
-document.write(
-'<script src=https://polyfill.io/v3/polyfill.min.js?features=default><\/script>' +
-'<script src=https://cdn.jsdelivr.net/gh/nuxodin/ie11CustomProperties@4.1.0/ie11CustomProperties.min.js><\/script>'
-);
+if (/MSIE \\d|Trident.*rv:/.test(navigator.userAgent)) {
+  document.write(
+    '<script src="https://polyfill.io/v3/polyfill.min.js?features=default"></script>' +
+    '<script src="https://cdn.jsdelivr.net/gh/nuxodin/ie11CustomProperties@4.1.0/ie11CustomProperties.min.js"></script>'
+  );
 }
-~~~"
+~~~
+          `}
         />
       </Text>
     </Stack>
--- a/docs/src/Faq.doc.js
+++ b/docs/src/Faq.doc.js
@@ -102,8 +102,28 @@ padding 0 .. 12
         Add the following to your import declarations:
         <Markdown
           text="
-~~~bash
+~~~js
 import { Button, Text } from 'gestalt';
+~~~"
+        />
+      </Text>
+
+      <Heading size="sm">What&apos;s required to support IE11?</Heading>
+      <Text>
+        Gestalt supports IE11 currently, but you will need to use a polyfill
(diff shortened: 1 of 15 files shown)


christianvuerings (commented):
@jennyscript thanks for putting this together! A couple of comments

1. We'll need to update `./scripts/cssValidate.js` and remove `noVarInLegacyCSS`, otherwise the build will fail
2. Could you add a section in the FAQ on how to make Gestalt work for IE11? E.g. adding the following to the `<head />`:

```
<script type="text/javascript">
      // Load polyfills for IE 11
      if (/MSIE \d|Trident.*rv:/.test(navigator.userAgent))
        document.write(
          '<script src="https://polyfill.io/v3/polyfill.min.js?features=default"><\/script>' +
            '<script src="https://cdn.jsdelivr.net/gh/nuxodin/ie11CustomProperties@4.1.0/ie11CustomProperties.min.js"><\/script>'
        );
    </script>
```

3. Totally agree that it would be nice if we can now extract these variables in a separate file, especially since they are otherwise repeated in the output. We could do that in a follow up PR

docs/src/components/Markdown.js
```diff
@@ -55,6 +51,13 @@ export default function Markdown({ text }: Props): Node {
 
   const html = marked(stripIndent(text), {
     renderer,
+    highlight(code, language) {
+      const validLanguage = highlightjs.getLanguage(language)
```
christianvuerings, line 55: @jennyscript thanks for fixing this!

christianvuerings (approved):

docs/src/Faq.doc.js
```diff
@@ -102,8 +102,28 @@ padding 0 .. 12
...
+        <Markdown
+          text="
+~~~js
+// Load polyfills for IE 11
+if (/MSIE \d|Trident.*rv:/.test(navigator.userAgent)) {
+document.write(
+'<script src=https://polyfill.io/v3/polyfill.min.js?features=default><\/script>' +
+'<script src=https://cdn.jsdelivr.net/gh/nuxodin/ie11CustomProperties@4.1.0/ie11CustomProperties.min.js><\/script>'
+);
+}
```
jennyscript, line 126: I tried a few things in here to get code indentation working, but any time I tried it broke line breaks. Any idea what I need to change here to get indentation showing?

christianvuerings:
@jennyscript lgtm - we do have a merge conflict now in `docs/src/index.js` - once that is resolved this should be good to merge

docs/src/Faq.doc.js
christianvuerings: @jennyscript Where we defined the `<Markdown />` component, we should use template literals. So

```
<Markdown text={`
```

instead of
```
<Markdown text="
```

christianvuerings:
@jennyscript I fixed the merge conflict + line breaks.

docs/src/Faq.doc.js
jennyscript: Thanks for fixing this!

PR #1204, opened 2020-09-09
Internal: update most dev dependencies

christianvuerings:
* Update most dev dependencies to their latest version
* Move dev dependencies up from `gestalt` + `gestalt-datepicker` packages

Ignored dev dependencies:
* Flow: to maintain compatibility with the pinterest codebase
* Prettier: will be done in a separate commit since their will be major styling changes

## Test Plan

- CI tests pass
- Netlify site shows up correctly


packages/gestalt-datepicker/package.json
```diff
@@ -27,20 +27,6 @@
     "prepublish": "rollup -c rollup.config.js",
     "watch": "rollup -c rollup.config.js --watch"
   },
-  "devDependencies": {
```
christianvuerings, line 30: Moved these up - otherwise we would have to update them in 2 locations each time + we might forget 1 location.

PR #1207, opened 2020-09-09
Callout: implement Link Button on primary/secondary links

AlbertCarreras:
- Replaced links with link-role Buttons
- Implemented accessibility labels for primary/secondary links
- Updated Docs API

## Before
Kapture 2020-09-09 at 18 32 16

## After
Kapture 2020-09-09 at 18 34 54


christianvuerings (approved):

christianvuerings:
@AlbertCarreras could you take a look at the failing tests? Looks like we need to update the snapshots

PR #1209, opened 2020-09-10
Table.Header: fix stacking context on sticky header

christianvuerings:
### Before
table-header-sticky-before

### After
table-header-sticky-after

## Test Plan

1. Go to `/Table#stickyHeader` in the gestalt docs
2. Scroll down the table
3. Ensure the header is always on top

### DIFF
--- a/packages/gestalt/src/Table.css
+++ b/packages/gestalt/src/Table.css
@@ -19,6 +19,7 @@
   background-color: var(--g-colorGray0);
   position: sticky;
   top: 0;
+  z-index: 1;
 }
 
 .thead tr:last-child th {
--- a/docs/src/Table.doc.js
+++ b/docs/src/Table.doc.js
@@ -196,9 +196,12 @@ card(
     id="stickyHeader"
     name="Example: Sticky header"
     defaultCode={`
-<Table maxHeight={110}>
+<Table maxHeight={200}>
   <Table.Header sticky>
     <Table.Row>
+      <Table.HeaderCell>
+        <Text weight="bold">Image</Text>
+      </Table.HeaderCell>
       <Table.HeaderCell>
         <Text weight="bold">Name</Text>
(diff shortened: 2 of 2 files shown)


packages/gestalt/src/Table.css
```diff
@@ -19,6 +19,7 @@
   background-color: var(--g-colorGray0);
   position: sticky;
   top: 0;
+  z-index: 1;
```
chrislloyd, line 22: Not 100% sure about this, but should it be using the z-index helpers so we can track this better?

AlbertCarreras: could we do something like

`  return props.sticky ? <Sticky zIndex={new FixedZIndex(1)} > <thead className={cs}>{props.children}</thead> </Sticky > : <thead className={cs}>{props.children}</thead>;
`

christianvuerings: @chrislloyd that's the approach I went with before, to add a `zIndex` prop to `<Table.Header />`. But the zIndex needs to be set on the `th`, not the `thead`. The `th` is also not aware of the `thead` so that's why I ended up with this initial fix (set the zIndex to 1).

See https://css-tricks.com/position-sticky-and-table-headers/
> You can’t position: sticky; a <thead>. Nor a <tr>. But you can sticky a <th>

@AlbertCarreras no, sadly that's not correct html. the `thead` should be an immediate child of the `table`. If we wrap it in `<Sticky />`, that's no longer the case.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/thead

image

chrislloyd: Ah great to see you already tried this. I was just suggesting using it internally might be better than using the CSS var directly but I don't have a strong opinion :)

AlbertCarreras (approved):

PR #1211, opened 2020-09-14
TapArea: reimplement tapStyle

AlbertCarreras:
## Test Plan
- TapArea only compresses only with tapStyle='compress'
- Button compress behavior is not affected.


packages/gestalt/src/InternalLink.js
```diff
@@ -81,6 +82,7 @@ const InternalLinkWithForwardRef: AbstractComponent<
     rel,
     rounding,
     size,
+    tapStyle = 'compress',
```
AlbertCarreras, line 85: default for Button/IconButton

packages/gestalt/src/TapArea.js
```diff
@@ -89,6 +90,7 @@ const TapAreaWithForwardRef: React$AbstractComponent<
     onMouseLeave,
     onTap,
     rounding = 0,
+    tapStyle = 'none',
```
AlbertCarreras, line 93: default is none for TapArea

docs/src/TapArea.doc.js
```diff
@@ -220,26 +220,37 @@ card(
...
       },
+      {
+        name: 'tapStyle',
+        type: `none | 'compress'`,
+        required: false,
+        defaultValue: 'none',
+        description: [
+          'Set a compressing behavior when the TapArea is clicked / touched.',
+          `- 'none' does not compress TapArea.`,
+          `- 'compress' scales down TapArea by 1%.`,
```
christianvuerings, line 238: @AlbertCarreras I would remove the `by 1%`

We actually compress it this way: https://github.com/pinterest/gestalt/blob/4f520db60d005670c8db7b6bee1500fa2e1a0419/packages/gestalt/src/useTapFeedback.js#L48

The reason for doing that is so we do not shrink large objects too much

AlbertCarreras: DONE

docs/src/TapArea.doc.js
```diff
@@ -220,26 +220,37 @@ card(
...
+        name: 'tapStyle',
+        type: `none | 'compress'`,
+        required: false,
+        defaultValue: 'none',
+        description: [
+          'Set a compressing behavior when the TapArea is clicked / touched.',
+          `- 'none' does not compress TapArea.`,
+          `- 'compress' scales down TapArea.`,
+        ],
+        href: 'tapStyle',
```
christianvuerings, line 240: @AlbertCarreras this link doesn't seem to work - should it be `roles` instead?

AlbertCarreras: DONE

christianvuerings (approved):

docs/src/TapArea.doc.js
AlbertCarreras: I was going to create it's own example, but add it there at the end.

PR #1213, opened 2020-09-15
Revert "Row, Stack: Re-implement for better `gap` (#1193)"

christianvuerings:
This reverts commit 72078fffc1748134c50e96d3e5c65312213cc3ad.

We no longer saw a gap between the items when using `Row`.

image

Probably due to
image


dangerismycat (approved):
whoops 😓

PR #1215, opened 2020-09-16
Internal: Add Cypress integration tests

christianvuerings:
First step of adding Cypress integration tests for Gestalt.

Next steps:
* Use https://github.com/cypress-io/netlify-plugin-cypress to run these on every netlify build
* Add https://www.npmjs.com/package/cypress-axe to enable accessibility integration tests 

## Test Plan

* Cypress runs + passes after the netlify build


christianvuerings:
Follow up at #1220 to implement these integration tests with GitHub actions instead.

PR #1216, opened 2020-09-16
Row, Stack: Update `children` type

dangerismycat:
Given how Row/Stack implement `gap` on their children, each child needs to be a React Element — styles won't be applied if given an unwrapped string or other non-Element Nodes.

This PR updates the `children` type for both components, and fixes one non-standard usage in the docs (no visual change that I can detect in the [preview](https://deploy-preview-1216--gestalt.netlify.app/Installation)). The Flow suppressions aren't ideal, but we really don't care what kind of Element we're getting, just that each child is an Element.

### DIFF
--- a/packages/gestalt/src/Row.js
+++ b/packages/gestalt/src/Row.js
@@ -28,8 +28,7 @@ type Props = {|
   alignContent?: AlignContent,
   alignItems?: AlignItems,
   alignSelf?: AlignSelf,
-  // $FlowExpectedError[unclear-type]
-  children?: ChildrenArray<Element<any>>,
+  children?: ChildrenArray<?Element<*>>,
   flex?: Flex,
   gap?: Padding,
   height?: Dimension,
--- a/packages/gestalt/src/Row.js
+++ b/packages/gestalt/src/Row.js
@@ -1,5 +1,5 @@
 // @flow strict
-import React, { type Node } from 'react';
+import React, { type ChildrenArray, type Element, type Node } from 'react';
 import PropTypes from 'prop-types';
 import classnames from 'classnames';
 import disallowedProps from './stackDisallowedProps.js';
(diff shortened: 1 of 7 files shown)


packages/gestalt/src/Row.js
```diff
@@ -28,7 +28,8 @@ type Props = {|
   alignContent?: AlignContent,
   alignItems?: AlignItems,
   alignSelf?: AlignSelf,
-  children?: Node,
+  // $FlowExpectedError[unclear-type]
+  children?: ChildrenArray<Element<any>>,
```
christianvuerings, line 32: @dangerismycat could we change these to `children?: ChildrenArray<Element<*>>`? That way we might be able to get rid of the `$FlowExpectedError`

christianvuerings (commented):
@dangerismycat thanks for the quick turnaround!


Could you also add a `.flowtest.js` for both `<Stack />` and `<Row />`?

`Stack.flowtest.js` example:
```jsx
// @flow strict
import React from 'react';
import Stack from './Stack.js';

const ValidSingle = (
  <Stack gap={1}>
    <div />
  </Stack>
);

const ValidMultiple = (
  <Stack gap={1}>
    <div />
    <div />
    <div />
    <div />
  </Stack>
);

const test = true;
const Testing = <Stack>{test ? <div /> : null}</Stack>;

// $FlowExpectedError[incompatible-type]
const MissingProp = <Stack gap={1}>test</Stack>;
```

packages/gestalt/src/Row.js
christianvuerings: @dangerismycat actually, should this be `ChildrenArray<?Element<*>>` (note the `?`) so we also support this case:
```
ChildrenArray<?Element<*>>
```
What do you think?

dangerismycat: works for me!

christianvuerings (approved):

PR #1224, opened 2020-09-17
Sticky: remove deprecated dangerouslySetZIndex

christianvuerings:
`dangerouslySetZIndex` has been removed from the Pinterest codebase.

## Changes

* Remove `dangerouslySetZIndex` from `<Sticky />`
* Add a codemod to find legacy usages when other people upgrade
* Add flow test for `<Sticky />`

## Test Plan

* `dangerouslySetZIndex` is no longer available on Sticky

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -4,6 +4,7 @@
 [ignore]
 <PROJECT_ROOT>/packages/gestalt-codemods/.*/__testfixtures__
 <PROJECT_ROOT>/cypress
+<PROJECT_ROOT>/node_modules/cypress/
 
 [options]
 esproposal.optional_chaining=enable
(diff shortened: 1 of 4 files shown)


.flowconfig
```diff
@@ -4,6 +4,7 @@
 [ignore]
 <PROJECT_ROOT>/packages/gestalt-codemods/.*/__testfixtures__
 <PROJECT_ROOT>/cypress
+<PROJECT_ROOT>/node_modules/cypress/
```
christianvuerings, line 7: Added this since Cypress does not expose flow types but the VSCode extension keeps crashing on it.

AlbertCarreras (approved):

PR #1227, opened 2020-09-22
Video: add `objectFill`  as a prop

valbooth28:
## Background
When Story Pins started using the Gestalt `Video` component for MP4s (experiment gated), employees noticed that the "poster", or placeholder while the video loads, was being displayed distorted. 

### Gif reproduction:
distortion

### Screenshot for emphasis.
Screen Shot 2020-09-15 at 4 47 54 PM copy

## Changes
Setting the style as 'object-fill: contain` seems to fix the issue, so this PR adds the specific `object-fill` part of the style as an optional prop.
Screen Shot 2020-09-16 at 6 18 20 PM

## Test Plan
Internal e2e testing confirms this will be a fix, and you can locally observe the style is passed correctly:
Screen Shot 2020-09-21 at 5 22 14 PM

### DIFF
--- a/packages/gestalt/src/Video.js
+++ b/packages/gestalt/src/Video.js
@@ -11,6 +11,8 @@ type Source =
   | string
   | Array<{| type: 'video/m3u8' | 'video/mp4' | 'video/ogg', src: string |}>;
 
+type ObjectFit = 'fill' | 'contain' | 'cover' | 'none' | 'scale-down';
+
 type Props = {|
   accessibilityHideCaptionsLabel?: string,
   accessibilityShowCaptionsLabel?: string,
@@ -26,6 +28,7 @@ type Props = {|
   children?: Node,
   controls?: boolean,
   loop?: boolean,
+  objectFit?: ObjectFit,
   onDurationChange?: ({|
     event: SyntheticEvent<HTMLVideoElement>,
     duration: number,
@@ -515,6 +518,7 @@ export default class Video extends PureComponent<Props, State> {
       children,
       crossOrigin,
       loop,
+      objectFit,
       playing,
       playsInline,
       poster,
@@ -523,9 +527,7 @@ export default class Video extends PureComponent<Props, State> {
       volume,
     } = this.props;
     const { currentTime, duration, fullscreen, captionsButton } = this.state;
-
     const paddingBottom = (fullscreen && '0') || `${(1 / aspectRatio) * 100}%`;
-
     return (
       <div
         ref={this.setPlayerRef}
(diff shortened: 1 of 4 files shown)


packages/gestalt/src/Video.js
```diff
@@ -11,6 +11,8 @@ type Source =
   | string
   | Array<{| type: 'video/m3u8' | 'video/mp4' | 'video/ogg', src: string |}>;
 
+type ObjectFit = 'fill' | 'contain' | 'cover' | 'none' | 'scale-down';
```
valbooth28, line 14: MDN for reference: https://developer.mozilla.org/en-US/docs/Web/CSS/object-fit

packages/gestalt/src/Video.js
```diff
@@ -549,6 +551,8 @@ export default class Video extends PureComponent<Props, State> {
             onSeeked={this.handleSeek}
             onTimeUpdate={this.handleTimeUpdate}
             onProgress={this.handleProgress}
+            // $FlowIssue facebook/flow#8448
```
valbooth28, line 554: facebook/flow#8448

christianvuerings: @valbooth28 thanks for the PR! Made a few updates:
* Remove the $FlowIssue - see https://github.com/facebook/flow/issues/8186#issuecomment-555747680
* Updated the docs to include the `objectFit` prop
* Added tests to ensure objectFit gets rendered

Also know that with every flow issue, we require the exact exception, so instead of something like `// $FlowFixMe`, we should write `// $FlowFixMe[prop-missing]`. See https://github.com/gajus/eslint-plugin-flowtype/issues/453 for the request to make this an ESLint rule

packages/gestalt/src/Video.js
```diff
@@ -549,7 +552,10 @@ export default class Video extends PureComponent<Props, State> {
             onSeeked={this.handleSeek}
             onTimeUpdate={this.handleTimeUpdate}
             onProgress={this.handleProgress}
-            {...(crossOrigin ? { crossOrigin } : null)}
+            {...(objectFit ? { style: { 'object-fit': objectFit } } : null)}
+            {...((crossOrigin ? { crossOrigin } : { ...null }): {|
```
christianvuerings, line 556: @AlbertCarreras / @jennyscript fyi - a pattern we might see more often when we spread on a component. It's pretty verbose but better than adding a `$FlowIssue` / `$FlowFixMe`: https://github.com/facebook/flow/issues/8186#issuecomment-555747680

The `{ ...null }` seems odd but it's what we have to do to make it work with exact types

christianvuerings (approved):

PR #1228, opened 2020-09-22
Internal: fix `flowtype/require-exact-type` errors

jackhsu978:
We added the linter `flowtype/require-exact-type` in #946, but we disable the linter in several places. 

For the same reasons we added the `flowtype/require-exact-type` linter, disabling these errors is actually very bad because the downstream projects do not really respect the exact-by-default logic and would still treat `{ }` as inexact. We should avoid using `{ }` to flow running in downstream projects correctly know whether object type is exact and inexact without any ambiguity. 

This PR fixes all linter errors. Note that in some cases we actually meant to use inexact types. I have made sure that they are correctly typed as inexact.

## Test Plan

Flow

### DIFF
--- a/packages/gestalt-datepicker/src/LocaleDataTypes.js
+++ b/packages/gestalt-datepicker/src/LocaleDataTypes.js
@@ -4,39 +4,37 @@ import PropTypes from 'prop-types';
 // flowlint unclear-type:off
 
 // LocaleData type from https://github.com/date-fns/date-fns/blob/81ab18785146405ca2ae28710cdfbb13a294ec50/src/locale/af/index.js.flow
-/* eslint-disable flowtype/require-exact-type */
-export type LocaleData = {
+export type LocaleData = {|
   code?: string,
   formatDistance?: (...args: Array<any>) => any,
 
   formatRelative?: (...args: Array<any>) => any,
-  localize?: {
+  localize?: {|
     ordinalNumber: (...args: Array<any>) => any,
     era: (...args: Array<any>) => any,
     quarter: (...args: Array<any>) => any,
     month: (...args: Array<any>) => any,
     day: (...args: Array<any>) => any,
     dayPeriod: (...args: Array<any>) => any,
-  },
-  formatLong?: {
+  |},
+  formatLong?: {|
     date: (...args: Array<any>) => any,
     time: (...args: Array<any>) => any,
     dateTime: (...args: Array<any>) => any,
-  },
-  match?: {
+  |},
+  match?: {|
     ordinalNumber: (...args: Array<string>) => any,
     era: (...args: Array<any>) => any,
     quarter: (...args: Array<any>) => any,
     month: (...args: Array<any>) => any,
     day: (...args: Array<any>) => any,
     dayPeriod: (...args: Array<any>) => any,
-  },
-  options?: {
+  |},
+  options?: {|
     weekStartsOn?: 0 | 1 | 2 | 3 | 4 | 5 | 6,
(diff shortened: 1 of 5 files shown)


packages/gestalt/src/Masonry.js
```diff
@@ -84,10 +84,11 @@ const VIRTUAL_BUFFER_FACTOR = 0.7;
 
 const layoutNumberToCssDimension = n => (n !== Infinity ? n : undefined);
 
-// eslint-disable-next-line flowtype/require-exact-type
-export default class Masonry<T: {}> extends ReactComponent<Props<T>, State<T>> {
-  // eslint-disable-next-line flowtype/require-exact-type
-  static createMeasurementStore<T1: {}, T2>(): MeasurementStore<T1, T2> {
+export default class Masonry<T: { ... }> extends ReactComponent<
```
jackhsu978, line 87: This is incorrectly annotated as exact (the default). Changing it to inexact.

packages/gestalt/src/MeasurementStore.js
```diff
@@ -1,8 +1,7 @@
 // @flow strict
 import type { Cache } from './Cache.js';
 
-// eslint-disable-next-line flowtype/require-exact-type
-export default class MeasurementStore<T: {} | $ReadOnlyArray<mixed>, V>
+export default class MeasurementStore<T: { ... } | $ReadOnlyArray<mixed>, V>
```
jackhsu978, line 4: This is incorrectly annotated as exact (the default). Changing it to inexact.

packages/gestalt/src/SelectList.js
```diff
@@ -12,12 +12,11 @@ import Icon from './Icon.js';
 import styles from './SelectList.css';
 import { type AbstractEventHandler } from './AbstractEventHandler.js';
 
-// eslint-disable-next-line flowtype/require-exact-type
-type Options = Array<{
+type Options = Array<{|
```
jackhsu978, line 15: This is exact, but we should explicitly declare it as exact so that the downstream projects with [`exact-by-default`](https://flow.org/en/docs/config/options/#toc-exact-by-default-boolean) and [`ambiguous-object-type`](https://flow.org/en/docs/linting/rule-reference/#toc-ambiguous-object-type) both enabled will not throw errors.

packages/gestalt-datepicker/src/LocaleDataTypes.js
```diff
@@ -4,39 +4,37 @@ import PropTypes from 'prop-types';
 // flowlint unclear-type:off
 
 // LocaleData type from https://github.com/date-fns/date-fns/blob/81ab18785146405ca2ae28710cdfbb13a294ec50/src/locale/af/index.js.flow
-/* eslint-disable flowtype/require-exact-type */
```
jackhsu978, line 7: Declare them as exact types explicitly

packages/gestalt/src/Masonry.js
```diff
@@ -56,12 +55,10 @@ type Props<T> = {|
...
     | ((
-        // eslint-disable-next-line flowtype/require-exact-type
-        ?{
+        ?{|
           from: number,
-        }
-        // eslint-disable-next-line flowtype/require-exact-type
-      ) => void | boolean | {}),
+        |}
+      ) => void | boolean | { ... }),
```
jackhsu978, line 61: changing from exact to inexact since consumers may be returning an object. although this should have meant `{||}` because of the `exact_by_default` option, flow running in downstream projects may not correctly respect that, so it's safer to change to make this inexact.

in the future, we could just change the return type to `void` because the return value isn't used at all.

AlbertCarreras (approved):

PR #1236, opened 2020-09-29
Sheet: Control animation behavior with AnimationContext

themindfuldev:
This PR extracts the animation handling logic from `<Sheet>` into `<AnimationController>` while providing a number of new options:
- A clear API to separate callbacks for `onDismissStart` (when a dismiss animation begins) and `onDismissEnd` (when a dismiss animation is over and only then unmount/cleanup will take place)
- A Context Provider `AnimationController` and a custom hook `useAnimation` to make it easier to control this animation behavior on internal components.
- 4 distinct animation states: in -> postIn -> out -> postOut
- A plug-and-play configuration where all the behavior works fine when an `<AnimationController>` is not provided
- Respecting `useReducedMotion` for a11y

## Test Plan
- Tested on `<Sheet>` docs
- Tested with OS reduced motion activated


packages/gestalt/src/AnimationController.js
```diff
@@ -0,0 +1,88 @@
...
+    [setAnimationState, shouldAnimate]
+  );
+
+  useEffect(() => {
+    if (animationState === 'post-out') {
+      onDismissEnd();
+    }
+  }, [animationState, onDismissEnd]);
+
+  return animationState !== 'post-out' ? (
```
dangerismycat, line 60: For readability, consider phrasing this as an early return:
```
if (animationState === 'post-out') {
  return null;
}

return ( ...
```

packages/gestalt/src/AnimationController.js
```diff
@@ -0,0 +1,88 @@
...
+    }
+  }, [animationState, setAnimationState]);
+  return {
+    animationState,
+    onAnimationEnd: animationState ? onAnimationEnd : null,
+  };
+}
+
+export default {
+  AnimationController,
```
dangerismycat, line 86: Should the controller itself be the default return?

dangerismycat (approved):
This is great!

packages/gestalt/src/AnimationController.js
themindfuldev: It makes sense, I agree :)

christianvuerings:
Checked in with @themindfuldev offline - we will be importing the animation controller inside of `Sheet`

docs/src/AnimationController.doc.js
```diff
@@ -0,0 +1,171 @@
...
+const cards: Array<Node> = [];
+const card = c => cards.push(c);
+
+card(
+  <PageHeader
+    name="AnimationController"
+    description={`An \`<AnimationController>\` is a wrapper to control animation of its wrapped components. 
+    It works with a \`Context.Provider\` which holds the state for \`animationState\`. 
+    Finally, it provides the custom hook \`useAnimation\` which provides an object \`{ animationState, onAnimationEnd }\` where:
+    - \`animationState\`: the current animation state. Possible values are: \`null\`, \`"in"\`, \`"post-in"\`, \`"out"\`, \`"post-out"\`.
```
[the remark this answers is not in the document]
themindfuldev: Changed :)

themindfuldev:
Updated docs: https://deploy-preview-1236--gestalt.netlify.app/Sheet

docs/src/Sheet.doc.js
```diff
@@ -307,27 +377,27 @@ function SheetRefExample() {
...
               accessibilitySheetLabel="Focused sheet"
-              onDismiss={() => { setShowSheet(!showSheet) }}
+              onDismiss={() => setShouldShow(false)}
               ref={sheetRef}
               size="md"
             >
               <Box color="white" minHeight={400} padding={8}>
                 <Box marginBottom={4}>
-                  <Heading size="md">Focused content</Heading>
+                  <Heading size="md">Focused content</Heading>
```
christianvuerings, line 394: @themindfuldev looks like it's adding a bunch of extra whitespace after certain lines, can we remove that?
image

christianvuerings (approved):

docs/src/Sheet.doc.js
themindfuldev: Sure - removing as part of https://github.com/pinterest/gestalt/pull/1242

PR #1238, opened 2020-09-29
ActivationCard: Add component

norrise120:
Example of Activation Cards
Screen Shot 2020-09-29 at 11 25 34 AM

In Darkmode
Screen Shot 2020-09-29 at 11 26 07 AM

## Test Plan
Wrote tests for component, matching pattern of testing for other components.
Updated docs and verify that docs looked as expected.


packages/gestalt/src/ActivationCard.js
```diff
@@ -0,0 +1,147 @@
...
+  status: 'notStarted' | 'pending' | 'needsAttention' | 'complete',
+  statusMessage: string,
+  title?: string,
+|};
+
+const STATUS_ICONS = {
+  notStarted: undefined,
+  pending: { symbol: 'clock', color: 'gray' },
+  needsAttention: { symbol: 'workflow-status-problem', color: 'red' },
+  complete: { symbol: 'check-circle', color: 'green' },
```
norrise120, line 43: Would love some feedback/recommendation on how to handle the "complete" status. The design feels significantly different from the other statuses to me. I wonder if it would be better to have a separate return statement or if I should use a lot of have a lot of `isComplete ?` ternaries through current return statement.

Screen Shot 2020-09-29 at 11 38 27 AM

packages/gestalt/src/ActivationCard.css
```diff
@@ -0,0 +1,11 @@
+html[dir="rtl"] .rtlPos {
```
norrise120, line 1: Here (and throughout this PR) I'm following the pattern set by Callout

packages/gestalt/src/ActivationCard.js
AlbertCarreras: what about rendering something like:
` {isCompleted  ? <Uncompleted href={href} etc. > : <Completed >}`

docs/src/ActivationCard.doc.js
```diff
@@ -0,0 +1,123 @@
...
+
+card(
+  <PageHeader
+    name="ActivationCard"
+    description="Activation cards are used in groups to communicate a user’s stage in a series of steps toward an overall action."
+  />
+);
+
+card(
+  <Example
```
AlbertCarreras, line 18: <Row gap={2} alignSelf='stretch'>
  <ActivationCard
    status="pending"
    statusMessage="Pending"
    title="Claim your website"
    message="We will notify you via email as soon as your site has been successfully claimed."
    linkData={{href: "https://pinterest.com", label:"Claim your website now"}}
  />
  <ActivationCard
    status="needsAttention"
    statusMessage="Needs attention"
    title="Tag is unhealthy"
    message="Oops! Your tag must be healthy to continue."
    linkData={{accessibilityLabel:"Learn more about how tag health", href: "https://pinterest.com", label:"Learn more"}}
    dismissButton={{
      accessibilityLabel: 'Dismiss card',
      onDismiss: ()=>{},
    }}
  />
</Row>

AlbertCarreras: Also, there's an issue with the dismissButton 
image

AlbertCarreras: Check implementation in callout, not happening there 
image

AlbertCarreras:
There's also an issue with responsiveness 
image

docs/src/ActivationCard.doc.js
```diff
@@ -0,0 +1,123 @@
...
+    name="Example"
+    defaultCode={`
+<Box display="flex" justifyContent="between">
```
christianvuerings, line 21: @norrise120 could we remove the `justifyContent="between"` and use symmetrical padding instead?
```
<Box
  display="flex"
  marginStart={-1}
  marginEnd={-1}
>
  <Box paddingX={1}>
    <ActivationCard>
  </Box>
  <Box paddingX={1}>
    <ActivationCard>
  </Box>
</Box>
```

packages/gestalt/src/ActivationCard.js
```diff
@@ -0,0 +1,147 @@
...
+  status,
+  statusMessage,
+  title,
+}: Props): Node {
+  const isStarted = status !== 'notStarted';
+  const icon = STATUS_ICONS[status];
+
+  return (
+    <Box flex="grow" borderSize="sm" rounding={4} padding={6} maxWidth={400}>
+      <Box display="flex" alignContent="center">
```
christianvuerings, line 87: @norrise120 Don't we want

image

instead of

image

If so, we should set a flex grow on a `<Box>` which wraps the title

docs/src/ActivationCard.doc.js
```diff
@@ -0,0 +1,123 @@
...
+  <ActivationCard
+    status="pending"
+    statusMessage="Pending"
+    title="Claiming your website"
+    message="We will notify you via email as soon as your site has been successfully claimed"
+    linkData={{href: "https://pinterest.com", label:"Learn more"}}
```
christianvuerings, line 27: Can we call this `link` instead of `linkData`?

packages/gestalt/src/ActivationCard.js
```diff
@@ -0,0 +1,147 @@
...
+        </Box>
+      )}
+      {message && (
+        <Text color="gray" size="md">
+          {message}
+        </Text>
+      )}
+      {linkData && <ActivationCardLink data={linkData} />}
+      {dismissButton && (
+        <div className={classnames(styles.rtlPos)}>
```
christianvuerings, line 113: @norrise120 since we use position `absolute` here, we need a wrapper which is set to `position: relative`.

Before:
image

After:
image

docs/src/ActivationCard.doc.js
christianvuerings: @norrise120 Figured it out, we need a `100%` height on the `<ActivationCard />` to make this work.

Feel free to check out the changes here: https://github.com/christianvuerings/gestalt/commit/76685a9b0851f512ba500b3ddce3424d9278f2aa

image

/cc @AlbertCarreras

norrise120: Updated! I also added `width="100%"` to each of the boxes holding the activation cards so that they are the same size (otherwise they look a little wonky, especially if one has a lot less content). Let me know though if there is a better way to achieve the same effect.

Before
Screen Shot 2020-10-05 at 1 10 03 PM

After
Screen Shot 2020-10-05 at 1 11 46 PM

taylorannsimpson:
**Design nits**
- the small icons should be 24px instead of 16px
- for the complete card, the icon should float to the vertical center of the card, not the top
- - the icon should have 16px margin right
- the container has an outline but I thought we were using a drop shadow? If Gestalt design decided on outline you can ignore this comment

norrise120:
Made requested design updates!

> * the container has an outline but I thought we were using a drop shadow? If Gestalt design decided on outline you can ignore this comment

My understanding is that drop shadow isn't currently supported by gestalt and we were just going to do a simple border until it was added

docs/src/ActivationCard.doc.js
```diff
@@ -0,0 +1,162 @@
...
+        ],
+        href: '',
+      },
+      {
+        name: 'title',
+        type: 'string',
+        required: true,
+        defaultValue: null,
+        description: [
+          'Heading to render inside the activation card above the message to convey the activation card topic to the user. The message text has a fixed size.',
```
christianvuerings, line 154: @norrise120 Can we remove `The message text has a fixed size` (looks like that should be for the `message` prop)

docs/src/ActivationCard.doc.js
```diff
@@ -0,0 +1,162 @@
...
+  `}
+  />
+);
+
+card(
+  <Example
+    name="Needs attention and Complete Cards"
+    defaultCode={`
+<Box display="flex" marginStart={-1} marginEnd={-1}>
+  <Box paddingX={1} width="100%">
```
christianvuerings, line 58: @norrise120 could we remove the `width="100%"`? Would be good to keep this consistent across examples.

christianvuerings (approved):

PR #1240, opened 2020-09-30
ExpandableModule component

galaxy0101:
## Test Plan
unit test
manual on https://deploy-preview-1240--gestalt.netlify.app


docs/src/ExpandableModule.doc.js
```diff
@@ -0,0 +1,108 @@
...
+    props={[
+      {
+        name: 'items',
+        type:
+          'Array<{| title: string, icon: ?string, summary: ?Array<string>, hasError: ?boolean, iconAccessibilityLabel: ?string, children: ?React.Node |}>',
+        required: true,
+      },
+    ]}
+  />
+);
```
AlbertCarreras, line 32: Hey Grace, could you add all the remaining props here?

AlbertCarreras:
Hey Grace, some feedback here https://www.loom.com/share/6d309d63f36949f6b6e2f0a5f0365713

packages/gestalt/src/ExpandableModule.css
```diff
@@ -0,0 +1,5 @@
+.expandablemodule {
```
simkessy, line 1: `.expandableModule {}`

packages/gestalt/src/ExpandableModule.js
```diff
@@ -0,0 +1,45 @@
...
+import ExpandableModuleBase from './ExpandableModuleBase.js';
+import styles from './ExpandableModule.css';
+
+type Props = {|
+  items: Array<{|
+    title: string,
+    icon?: string,
+    iconAccessibilityLabel?: string,
+    summary?: Array<string>,
+    type?: string,
```
simkessy, line 14: type should have options? Im not sure what the different types are:

`type: "error" | "warning" | "info"`  maybe?

packages/gestalt/src/ExpandableModule.js
```diff
@@ -0,0 +1,45 @@
...
+import Divider from './Divider.js';
+import ExpandableModuleBase from './ExpandableModuleBase.js';
+import styles from './ExpandableModule.css';
+
+type Props = {|
+  items: Array<{|
+    title: string,
+    icon?: string,
+    iconAccessibilityLabel?: string,
+    summary?: Array<string>,
```
simkessy, line 13: Does summary need to be an array of strings? Can summary be a node?

simkessy:
@ashleyseto should there be a hover state around the entire menu row. There's no indication on what you're hovering on.

image

simkessy:
Should this be called Accordion? 
https://getbootstrap.com/docs/4.0/components/collapse/

packages/gestalt/src/ExpandableModuleBase.js
```diff
@@ -0,0 +1,108 @@
...
+
+const EXPANDABLE_TYPE_ATTRIBUTES = {
+  info: {
+    color: 'darkGray',
+  },
+  error: {
+    icon: 'workflow-status-problem',
+    color: 'red',
+    accessibilityLabel: 'Error icon',
+  },
```
simkessy, line 28: @ashleyseto  is there a warning state?

simkessy:
The alignment is a little off 
image

packages/gestalt/src/ExpandableModuleBase.js
```diff
@@ -0,0 +1,108 @@
...
+  icon?: string,
+  iconAccessibilityLabel?: string,
+  summary?: Array<string>,
+  isCollapsed: boolean,
+  onModuleClicked?: boolean => void,
+  type?: string,
+  children?: Node,
+|};
+
+const EXPANDABLE_TYPE_ATTRIBUTES = {
```
simkessy, line 20: nit: Can rename to something shorter like "EXPANDABLE_TYPES"

docs/src/ExpandableModule.doc.js
```diff
@@ -0,0 +1,108 @@
...
+
+card(
+  <PropTable
+    Component={ExpandableModule}
+    props={[
+      {
+        name: 'items',
+        type:
+          'Array<{| title: string, icon: ?string, summary: ?Array<string>, type: ?string, iconAccessibilityLabel: ?string, children: ?React.Node |}>',
```
christianvuerings, line 27: Instead, could we do something like `<ExpandableModule.Item title="" icon="" />`? It would make the component more composable.

docs/src/ExpandableModule.doc.js
```diff
@@ -0,0 +1,108 @@
...
+        required: true,
+      },
+    ]}
+  />
+);
+
+card(
+  <Example
+    name="Example 1"
```
christianvuerings, line 36: Let's call this one `Example` - the most basic example.

docs/src/ExpandableModule.doc.js
```diff
@@ -0,0 +1,108 @@
...
+    </Box>
+  );
+}
+`}
+  />
+);
+
+card(
+  <Example
+    name="Example 2"
```
christianvuerings, line 57: We can call this `Multiple items`

docs/src/ExpandableModule.doc.js
```diff
@@ -0,0 +1,108 @@
...
+    </Box>
+  );
+}
+`}
+  />
+);
+
+card(
+  <Example
+    name="Example 3"
```
christianvuerings, line 88: Let's call this `Error State`

packages/gestalt/src/ExpandableModule.css
```diff
@@ -0,0 +1,5 @@
...
+  background: rgba(0, 0, 0, 0);
```
christianvuerings, line 2: @galaxy0101 Could we remove this line? It basically adds a transparent background

packages/gestalt/src/ExpandableModule.css
```diff
@@ -0,0 +1,5 @@
...
+  box-shadow: 0 0 8px rgba(0, 0, 0, 0.1);
```
christianvuerings, line 3: @galaxy0101 could we extract this into `./Borders.css`? And call it `boxShadow`? We only want 1 box shadow in Gestalt.

packages/gestalt/src/ExpandableModule.js
christianvuerings: Agreed with @simkessy - ideally we would be more strict

packages/gestalt/src/ExpandableModuleBase.js
```diff
@@ -0,0 +1,108 @@
...
+  info: {
+    color: 'darkGray',
+  },
+  error: {
+    icon: 'workflow-status-problem',
+    color: 'red',
+    accessibilityLabel: 'Error icon',
```
christianvuerings, line 27: @galaxy0101 these should be passed in as props for i18n

packages/gestalt/src/ExpandableModuleBase.js
```diff
@@ -0,0 +1,108 @@
...
+    if (onModuleClicked) {
+      onModuleClicked(!isCollapsed);
+    }
+  };
+  return (
+    <>
+      <TapArea onTap={handleModuleClick}>
+        <Box display="flex">
+          <Box display="flex" flex="grow" marginEnd={6}>
+            <Box column={isCollapsed ? 6 : 12} display="flex">
```
christianvuerings, line 51: Would be better if we can use symmetrical padding:

```
<Box display="flex" marginStart={-2} marginRight={-2}>
  <Box paddingX={2}></Box>
  <Box paddingX={2}></Box>
  <Box paddingX={2}></Box>
</Box>
```

packages/gestalt/src/ExpandableModuleBase.js
```diff
@@ -0,0 +1,108 @@
...
+import React, { type Node } from 'react';
+import Box from './Box.js';
+import IconButton from './IconButton.js';
+import Icon from './Icon.js';
+import TapArea from './TapArea.js';
+import Text from './Text.js';
+
+type Props = {|
+  title: string,
+  icon?: string,
```
christianvuerings, line 11: @galaxy0101 let's make sure we can only use Gestalt icons (see `Icon.js`: `icon?: $Keys<typeof icons>`)

packages/gestalt/src/ExpandableModuleBase.js
```diff
@@ -0,0 +1,108 @@
...
+  iconAccessibilityLabel?: string,
+  summary?: Array<string>,
+  isCollapsed: boolean,
+  onModuleClicked?: boolean => void,
```
christianvuerings, line 15: @galaxy0101 can we make this required instead of optional?

packages/gestalt/src/ExpandableModuleBase.js
```diff
@@ -0,0 +1,108 @@
...
+      onModuleClicked(!isCollapsed);
+    }
+  };
+  return (
+    <>
+      <TapArea onTap={handleModuleClick}>
+        <Box display="flex">
+          <Box display="flex" flex="grow" marginEnd={6}>
+            <Box column={isCollapsed ? 6 : 12} display="flex">
+              {EXPANDABLE_TYPE_ATTRIBUTES[type].icon && (
```
christianvuerings, line 52: @galaxy0101 it looks like the `EXPANDABLE_TYPE_ATTRIBUTES[type].icon` is competing with `icon` below. Seriously wondering if we need to make the `type` part of this component

packages/gestalt/src/ExpandableModuleBase.js
```diff
@@ -0,0 +1,108 @@
...
+                    </Text>
+                  </Box>
+                ))}
+              </Box>
+            )}
+          </Box>
+          <Box>
+            {children && (
+              <IconButton
+                accessibilityLabel="arrow icon to expand or collapse the module"
```
christianvuerings, line 95: @galaxy0101 these should be passed in for a11y and ideally change based on whether we want to show or hide the items.

packages/gestalt/src/ExpandableModuleBase.jsdom.test.js
```diff
@@ -0,0 +1,74 @@
...
+
+    const expandButton = screen.getByRole('button', {
+      name: /arrow icon to expand or collapse the module/i,
+    });
+    fireEvent.click(expandButton);
+    expect(props.onModuleClicked).toHaveBeenCalledWith(false);
+  });
+
+  it('should render icon correctly', () => {
+    props = {
```
christianvuerings, line 29: @galaxy0101 instead of mutating props, can we make it more explicit?

```
const propsWithIcon = {
   ...baseProps,
   icon: 'lock',
   iconAccessibilityLabel: 'test label'
}
```

packages/gestalt/src/ExpandableModuleBase.js
```diff
@@ -0,0 +1,108 @@
...
+  children,
+}: Props): Node {
+  const handleModuleClick = () => {
+    if (onModuleClicked) {
+      onModuleClicked(!isCollapsed);
+    }
+  };
+  return (
+    <>
+      <TapArea onTap={handleModuleClick}>
```
christianvuerings, line 48: @galaxy0101 didn't we say it would only open when you click on the arrow down/up?

packages/gestalt/src/ExpandableModuleBase.js
galaxy0101: 1) we will need an expandable card without summary in ads manager. 
2) we will need an expandable card without icon or iconAccessibilityLabel too
3) onModuleClicked can be made required, qq: why do we want to make onModuleClicked required?

packages/gestalt/src/ExpandableModuleBase.js
galaxy0101: Yes. added arrowIconAccessibilityLabel to differentiate from titleIconAccessibilityLabel

packages/gestalt/src/ExpandableModuleBase.jsdom.test.js
galaxy0101: fixed

packages/gestalt/src/ExpandableModuleBase.js
galaxy0101: we said to expand it when click on the entire bar, not the entire block, particularly not the entire children area in TDD.

packages/gestalt/src/ExpandableModule.css
galaxy0101: Fixed.

packages/gestalt/src/ExpandableModule.js
galaxy0101: Updated.
But we don't have a design for Warning state, so I put
type: "error" | "info" for now

packages/gestalt/src/ExpandableModule.js
galaxy0101: Based on the discussion in TDD meeting, we decided to use array of strings instead of Node. We may revisit it on next meeting

packages/gestalt/src/ExpandableModuleBase.js
galaxy0101: we don't have Warning state in the original design. We will revisit in the next meeting.

packages/gestalt/src/ExpandableModuleBase.js
galaxy0101: I think EXPANDABLE_TYPE_ATTRIBUTES is more accurate.

packages/gestalt/src/ExpandableModule.css
galaxy0101: Removed

packages/gestalt/src/ExpandableModule.js
```diff
@@ -0,0 +1,61 @@
...
+import styles from './ExpandableModule.css';
+
+type Props = {|
+  items: Array<{|
+    title: string,
+    icon?: string,
+    arrowIconAccessibilityLabel: string,
+    titleIconAccessibilityLabel?: string,
+    summary?: Array<string>,
+    type: 'error' | 'info',
```
christianvuerings, line 15: @galaxy0101: `type?: 'error' | 'info'`

packages/gestalt/src/ExpandableModuleBase.js
```diff
@@ -0,0 +1,112 @@
...
+
+export default function ExpandableModuleBase({
+  title,
+  icon,
+  arrowIconAccessibilityLabel,
+  titleIconAccessibilityLabel,
+  summary,
+  isCollapsed,
+  onModuleClicked,
+  type,
```
christianvuerings, line 40: @galaxy0101 `type = 'info'`

packages/gestalt/src/ExpandableModuleBase.js
```diff
@@ -0,0 +1,112 @@
...
+                truncate
+                color={EXPANDABLE_TYPE_ATTRIBUTES[type].color}
+              >
+                {title}
+              </Text>
+            </Box>
+            {summary && isCollapsed && (
+              <Box column={6}>
+                {summary.map((item, i) => (
+                  <Box key={i} marginBottom={1}>
```
christianvuerings, line 87: @galaxy0101 Let's see if we can use `<Stack />` or don't add the marginBottom of `{1}` to the last item in summary.

packages/gestalt/src/ExpandableModuleBase.js
```diff
@@ -0,0 +1,112 @@
...
+                      {item}
+                    </Text>
+                  </Box>
+                ))}
+              </Box>
+            )}
+          </Box>
+          <Box>
+            {children && (
+              <IconButton
```
christianvuerings, line 98: @galaxy0101 could we change this with a regular `Icon`

`<Box padding={2}><Icon>`

packages/gestalt/src/ExpandableModuleBase.js
christianvuerings: @galaxy0101 https://react-accessible-accordion.springload.co.nz/ - let's make sure we add the correct `aria-` roles (expanded)

packages/gestalt/src/ExpandableModuleBase.js
```diff
@@ -0,0 +1,112 @@
...
+  arrowIconAccessibilityLabel,
+  titleIconAccessibilityLabel,
+  summary,
+  isCollapsed,
+  onModuleClicked,
+  type,
+  children,
+}: Props): Node {
+  const handleModuleClick = () => {
+    if (onModuleClicked) {
```
christianvuerings, line 44: @galaxy0101 we should be able to remove this check.

packages/gestalt/src/ExpandableModuleBase.js
themindfuldev: @galaxy0101 changing this to an Icon will also fix the following keyboard accessibility concern, where each element (title + icon) is getting focused twice for the exact same action:

Screen Shot 2020-10-05 at 6 21 06 PM

packages/gestalt/src/ExpandableModuleBase.js
```diff
@@ -0,0 +1,112 @@
...
+      onModuleClicked(!isCollapsed);
+    }
+  };
+  return (
+    <>
+      <TapArea onTap={handleModuleClick}>
```
themindfuldev, line 50: Please use `accessibilityExpanded` on `<TapArea>` to conform with the WAI-ARIA Design Pattern of "Accordion" from https://www.w3.org/TR/wai-aria-practices-1.1/#accordion

themindfuldev: More info on https://gestalt.netlify.app/TapArea#accessibility

Something else that can be considered is to add `accessibilityControls` as it's also part of WAI-ARIA Design Pattern of "Accordion" but that also requires giving an `id` to the content area and that may overcomplicate things here, so I'd leave that just for thought.

packages/gestalt/src/ExpandableModuleBase.js
```diff
@@ -0,0 +1,112 @@
...
+                    color={EXPANDABLE_TYPE_ATTRIBUTES[type].color}
+                  />
+                </Box>
+              )}
+              {icon && (
+                <Box paddingX={2}>
+                  <Icon
+                    icon="lock"
+                    accessibilityLabel={
+                      titleIconAccessibilityLabel || 'Title icon'
```
themindfuldev, line 70: I wonder if passing an english fallback sentence can be dangerous because it doesn't support i18n use cases on Pinterest code base. I would strongly advise on not providing a fallback string here to really force users to pass a localized string.

themindfuldev (changes requested):

packages/gestalt/src/ExpandableModuleBase.js
galaxy0101: Yes. We will use Icon instead.

packages/gestalt/src/ExpandableModuleBase.js
galaxy0101: Yes. Fixed

packages/gestalt/src/ExpandableModuleBase.js
galaxy0101: <Stack gap={}
could remove the marginBottom of the last item

packages/gestalt/src/ExpandableModuleBase.js
galaxy0101: 1) Our expandable looks like the one in
Collapsing the last expanded item
If you set allowZeroExpanded to true then a solitary expanded item may be collapsed again.
2) will add accessibility_ExpandableModule_spec.js

AlbertCarreras: re request on clarification `on let's make sure we add the correct aria- roles (expanded) `https://www.loom.com/share/c98f07484ce84eb0830b3cb40fcf33d9

christianvuerings:
Follow up is at #1253

PR #1242, opened 2020-10-03
Sheet: Add subHeading prop [Depends on #1236]

themindfuldev:
Depends on https://github.com/pinterest/gestalt/pull/1236

This PR adds a new prop "subHeading" to support a use case where certain designs ask for a sub-heading docked under the heading.

This "subHeading" prop shall take a Node or a Render prop to behave similar to "footer" and "children" so it can contain exitpoints which may trigger "out" animations.

## Test Plan
Tested on <Sheet> docs
Tested with OS reduced motion activated

### DIFF
--- a/packages/gestalt/src/Sheet.js
+++ b/packages/gestalt/src/Sheet.js
@@ -44,24 +44,39 @@ import StopScrollBehavior from './behaviors/StopScrollBehavior.js';
 import sheetStyles from './Sheet.css';
 import TrapFocusBehavior from './behaviors/TrapFocusBehavior.js';
 
-type SheetProps = {|
+type SheetMainProps = {|
   accessibilityDismissButtonLabel: string,
   accessibilitySheetLabel: string,
   children: Node,
   closeOnOutsideClick?: boolean,
   footer?: Node,
-  heading?: string,
   onDismiss: () => void,
   size?: 'sm' | 'md' | 'lg',
 |};
 
+type SheetProps = {|
+  ...SheetMainProps,
+  heading?: string,
+  subHeading?: Node,
+|};
+
 type NodeOrRenderProp = Node | (({| onDismissStart: () => void |}) => Node);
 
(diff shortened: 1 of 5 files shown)


packages/gestalt/src/Sheet.js
```diff
@@ -19,27 +43,52 @@ import Row from './Row.js';
...
   closeOnOutsideClick?: boolean,
   footer?: Node,
-  heading?: string,
   onDismiss: () => void,
   size?: 'sm' | 'md' | 'lg',
 |};
 
+type SheetProps = {|
+  ...SheetMainProps,
+  heading?: string,
```
dangerismycat, line 59: Should `heading` be required? Would we ever want to display a Sheet without a heading?

packages/gestalt/src/Sheet.js
```diff
@@ -64,53 +113,60 @@ const DismissButton = ({
...
+): Node {
   const {
     accessibilityDismissButtonLabel,
     accessibilitySheetLabel,
     children,
     closeOnOutsideClick = true,
-    onDismiss,
     footer,
-    heading,
+    heading = undefined,
```
dangerismycat, line 135: Default values are only triggered when the field is undefined, so these default values aren't doing anything.

dangerismycat (approved):

packages/gestalt/src/Sheet.js
themindfuldev: Yes, great catch! I only added them to keep flow happy, from the way I was using Union Types with Flow - but it seems it's no longer necessary, will remove it :)

packages/gestalt/src/Sheet.js
themindfuldev: The initial figma design that I received has a case without a heading, that's why I did it like that. Will share you the file on Slack.

christianvuerings (approved):

PR #1244, opened 2020-10-05
Button/TapArea/IconButton: add tabIndex prop

AlbertCarreras:
For accessibility reasons, sometimes we have to remove components from the sequential keyboard navigation.

We use implement `tabIndex` https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex in `Button`, `TapArea`, and `IconButton`.

Added tests and updated documentation for each component to cover all tabIndex combinations.

## Test Plan
- Added test coverage to check for tabIndex in the innermost component
- Tested in Docs https://www.loom.com/share/91f4c25c87bb4be69579084ff5199446

### DIFF
--- a/docs/src/Button.doc.js
+++ b/docs/src/Button.doc.js
@@ -31,18 +31,40 @@ card(
     defaultCode={`
 function Example() {
   const [disabled, setDisabled] = React.useState(false);
+  const [tabIndex, setTabIndex] = React.useState(false);
 
   return (
     <Stack gap={3}>
       <Row gap={1}>
         <Tooltip text="Default button">
-          <Button onClick={() => {}} text="Clear search history" inline disabled={disabled} />
+          <Button
+            onClick={() => {}}
+            text="Clear search history"
+            inline
+            disabled={disabled}
+            tabIndex={tabIndex ? -1 : 0}
+          />
         </Tooltip>
         <Tooltip text="Submit button">
-          <Button type="submit" name="satisfaction-questionaire" text="Submit your response" inline disabled={disabled} />
+          <Button
+            type="submit"
+            name="satisfaction-questionaire"
+            text="Submit your response"
+            inline
+            disabled={disabled}
+            tabIndex={tabIndex ? -1 : 0}
+          />
         </Tooltip>
         <Tooltip text="Link button">
-          <Button role="link" target="blank" href="https://www.pinterest.com" text="Visit pinterest.com" inline disabled={disabled} />
+          <Button
+            role="link"
+            target="blank"
+            href="https://www.pinterest.com"
+            text="Visit pinterest.com"
+            inline
+            disabled={disabled}
+            tabIndex={tabIndex ? -1 : 0}
+          />
         </Tooltip>
       </Row>
       <Row gap={1}>
@@ -54,6 +76,18 @@ function Example() {
         <Box paddingX={2} flex="grow">
           <Label htmlFor="disable-buttons">
             <Text>Disable buttons</Text>
+          </Label>
+        </Box>
+      </Row>
+      <Row gap={1}>
+        <Switch
+          onChange={() => setTabIndex(!tabIndex)}
+          id="unreachable-buttons"
+          switched={tabIndex}
+        />
+        <Box paddingX={2} flex="grow">
+          <Label htmlFor="unreachable-buttons">
+            <Text>Remove from keyboard navigation with tabIndex</Text>
           </Label>
         </Box>
       </Row>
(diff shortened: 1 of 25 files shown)


packages/gestalt/src/InternalLink.js
```diff
@@ -224,10 +226,10 @@ const InternalLinkWithForwardRef: AbstractComponent<
         ...(target === 'blank' ? ['noopener', 'noreferrer'] : []),
         ...(rel === 'nofollow' ? ['nofollow'] : []),
       ].join(' ')}
+      tabIndex={removeFromSequentialKeyboardNavigation ? '-1' : '0'}
       {...(tapStyle === 'compress' && compressStyle && !disabled
         ? { style: compressStyle }
         : {})}
-      tabIndex={isTapArea && !disabled ? '0' : null}
```
AlbertCarreras, line 230: the default is always 0.

docs/src/Button.doc.js
```diff
@@ -218,6 +218,37 @@ function ButtonFlyoutExample() {
   />
 );
 
+card(
+  <Example
+    name="Accessibility: sequential keyboard navigation"
+    id="removeFromSequentialKeyboardNavigation"
```
christianvuerings, line 224: @AlbertCarreras Instead of adding a boolean prop, can we pass in `tabIndex` and set it to an enum of `0` or `-1`? Main reasons:
* It allows us to add e.g. `1` (even though it is an anti-pattern) if we do end up needing it
* Avoids boolean props: https://spicefactory.co/blog/2019/03/26/how-to-avoid-the-boolean-trap-when-designing-react-components/

AlbertCarreras: DONE

packages/gestalt/src/Button.js
```diff
@@ -214,6 +216,7 @@ const ButtonWithForwardRef: React$AbstractComponent<
         onClick={handleLinkClick}
         ref={innerRef}
         rel={rel}
+        tabIndex={tabIndex}
```
christianvuerings, line 219: @AlbertCarreras we should also add `tabIndex` to the `<button />` element below

packages/gestalt/src/IconButton.js
```diff
@@ -184,6 +186,7 @@ const IconButtonWithForwardRef: React$AbstractComponent<
         onMouseLeave={handleOnMouseLeave}
         ref={innerRef}
         rel={rel}
+        tabIndex={tabIndex}
```
christianvuerings, line 189: @AlbertCarreras we should also add `tabIndex` to the `<button />` element below

packages/gestalt/src/TapArea.js
```diff
@@ -189,6 +191,7 @@ const TapAreaWithForwardRef: React$AbstractComponent<
         onMouseLeave={handleLinkOnMouseLeave}
         ref={innerRef}
         rel={rel}
+        tabIndex={tabIndex}
```
christianvuerings, line 194: @AlbertCarreras we should also add `tabIndex` to the `<div />` element below.

If the TapArea is disabled, it should always be `null`
If the TapArea is not disabled, it should default to `0` (TAB-able) but if the `tabIndex` prop set to `-1`, it should be `-1`

packages/gestalt/src/InternalLink.js
christianvuerings: @AlbertCarreras I think this change is wrong, when a TapArea is disabled, the `tabIndex` should be `null` or `-1`

```
<TapArea
        role="link"
        target="blank"
        href="https://www.pinterest.com"
        disabled
      >
```

image

AlbertCarreras: MMM I read the in absence of a tabIndex, the default is 0. I'll double check 

[If not specified, it takes the default value 0.](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex)

christianvuerings:
@AlbertCarreras looks like snapshot tests need to get updated

packages/gestalt/src/InternalLink.js
christianvuerings: @AlbertCarreras that is true for elements which are already focussable (e.g. `<a />`, `<button />`) but not for others (e.g. `<div />`)

packages/gestalt/src/Button.js
AlbertCarreras: DONE

packages/gestalt/src/IconButton.js
AlbertCarreras: done

packages/gestalt/src/TapArea.js
AlbertCarreras: done

packages/gestalt/src/__snapshots__/TapArea.test.js.snap
```diff
@@ -163,7 +163,7 @@ exports[`TapArea supports press style 1`] = `
   onTouchMove={[Function]}
   onTouchStart={[Function]}
   role="button"
-  tabIndex="0"
+  tabIndex={0}
```
christianvuerings, line 166: @AlbertCarreras heads up that the `tabIndex` change will result in a lot of snapshot updates in the Pinterest codebase

christianvuerings (approved):

PR #1252, opened 2020-10-07
Table: Add ExpandableRow

faisalgedi:
The purpose of this PR is to create the functionality of an Expandable Row for the Gestalt Table along with a hover state. The row expands and collapses through the click of an arrow icon button.

Test Plan:
Using the docs for manual testing and unit tests.


christianvuerings:
@faisalgedi Looks like
* The button size to expand/collapse a row is too big
* We should have consistent padding around each expandable row (24px => 6 boints)

image
image

packages/gestalt/src/TableRowExpandable.js
```diff
@@ -0,0 +1,51 @@
...
+            icon={expanded ? 'arrow-up' : 'arrow-down'}
+            iconColor="darkGray"
+            onClick={() => setExpanded(!expanded)}
+          />
+        </td>
+        {props.children}
+      </tr>
+      {expanded && (
+        <tr>
+          <td colSpan={Object.keys(props.children).length}>
```
christianvuerings, line 44: @faisalgedi Are we sure this logic is correct? In some of the examples, the colspan should be 1 higher

image

packages/gestalt/src/Table.css
```diff
@@ -29,3 +29,11 @@
 .tbody tr:not(:last-child) td {
   border-bottom: 1px solid var(--g-colorGray150);
 }
+
+.hoverShadeGray {
+  background-color: var(--g-colorGray150);
```
christianvuerings, line 34: @faisalgedi this should be `#efefef` or `--g-colorGray100` instead of `--g-colorGray150`

packages/gestalt/src/TableRowExpandable.js
```diff
@@ -0,0 +1,51 @@
...
+  const cs =
+    hovered && hoverStyle === 'gray'
+      ? cx(styles.hoverShadeGray)
+      : cx(styles.hoverNone);
+
+  return (
+    <>
+      <tr
+        className={cs}
+        onMouseEnter={() => handleHover(true)}
```
christianvuerings, line 29: @faisalgedi we can handle this in CSS instead of having to do a mouseEnter / mouseLeave

```
.hoverShadeGray:hover {
  background-color: var(--g-colorGray150);
}
```

and no need to have `.hoverNone`, we can just leave it as an empty string:

```
const classes = hoverStyle === 'gray'
      ? cx(styles.hoverShadeGray)
      : '';
```

packages/gestalt/src/TableRowExpandable.js
```diff
@@ -0,0 +1,51 @@
...
+        onMouseLeave={() => handleHover(false)}
+      >
+        <td>
+          <IconButton
+            accessibilityLabel=""
```
christianvuerings, line 34: @faisalgedi we should pass these props on the root of the `Table.ExpandableRow` component:

```
accessibilityExpandLabel="Click to expand the row"
accessibilityCollapseLabel="Click to collapse the row"
```

both props should be required.

docs/src/Table.doc.js
```diff
@@ -288,6 +288,181 @@ card(
...
+        <Text weight="bold">House</Text>
+      </Table.HeaderCell>
+      <Table.HeaderCell>
+        <Text weight="bold">Birthday</Text>
+      </Table.HeaderCell>
+    </Table.Row>
+  </Table.Header>
+  <Table.Body>
+    <Table.RowExpandable
+    expandedContents=<Box maxWidth={236} padding={2} column={12}>
```
christianvuerings, line 341: @faisalgedi could we also use the curly bracket here?

```
expandedContents={<
```
instead of
```
expandedContents=<
```

Also recommend copy/pasting the example in a separate file + running prettier on it

docs/src/Table.doc.js
```diff
@@ -288,6 +288,181 @@ card(
...
+card(
+  <Example
+    name="Example: Table Row Expanable"
+    defaultCode={`
+<Table>
+  <Table.Header>
+    <Table.Row>
+    <Table.HeaderCell>
+  <Box display="visuallyHidden">
+    <Text weight="bold">Open/Close row</Text>
```
christianvuerings, line 325: @faisalgedi thanks for adding this in

christianvuerings:
@faisalgedi overall looks good, a few more comments:
* Could we add a flowtest
* Please make sure the tests on CI pass, right now they are failing

packages/gestalt/src/TableRowExpandable.js
christianvuerings: `accessibilityLabel={expanded ? accessibilityCollapseLabel : accessibilityExpandLabel}`

packages/gestalt/src/TableRowExpandable.js
```diff
@@ -0,0 +1,52 @@
...
+
+  return (
+    <>
+      <tr className={cs}>
+        <td>
+          <IconButton
+            accessibilityLabel={
+              expanded ? accessibilityCollapseLabel : accessibilityExpandLabel
+            }
+            icon={expanded ? 'arrow-up' : 'arrow-down'}
```
christianvuerings, line 35: @faisalgedi we also need `accessibilityExpanded` and `accessibilityControls` on this `IconButton`.

```
accessibilityExpanded={expanded}
accessibilityControls={id}
```

The id should then be specified on the expanded `tr`:

```
      {expanded && (
        <tr id={id}>
          <td colSpan={React.Children.count(children) + 1}>
            <Box padding={6}>{expandedContents}</Box>
          </td>
        </tr>
      )}
```

So we should add an `id` and make it required for every `Table.ExpandableRow`

christianvuerings (approved):

PR #1256, opened 2020-10-08
Checkbox/RadioButton: Add options to provide subtext and/or image for label

ayeshakmaz:
Add the ability to both Checkbox and RadioButton to provide `subtext` and/or `image`

Screen Shot 2020-10-08 at 3 52 23 PM
Screen Shot 2020-10-08 at 3 52 34 PM
Screen Shot 2020-10-08 at 3 52 41 PM

## Test Plan

- Added snapshot tests for new functionality
- Tested with a screenreader and keyboard
- Documented new options

### DIFF
--- a/docs/src/Checkbox.doc.js
+++ b/docs/src/Checkbox.doc.js
@@ -162,26 +162,26 @@ function CheckboxExample() {
 
   return (
     <Box margin={-2}>
-      <Box padding={2}>
+      <Box padding={1}>
         <Checkbox
           checked={checkedEn}
           id="coral"
           label="Coral"
           subtext="Botanical art in coral and green"
-          image={<Box height={100} width={80} rounding={3}><Image alt="" src="https://i.ibb.co/7bQQYkX/stock2.jpg" fit="container"/></Box>}
+          image={<Box height={100} width={80} paddingX={1}><Image alt="" src="https://i.ibb.co/7bQQYkX/stock2.jpg" fit="container" naturalWidth={1} naturalHeight={1}/></Box>}
           name="favorite art"
           onChange={({ checked }) => {
             setCheckedEn(checked);
           }}
         />
       </Box>
-      <Box padding={2}>
+      <Box padding={1}>
         <Checkbox
           checked={checkedSp}
           id="blue"
           label="Blue"
           subtext="Typography in blue"
-          image={<Box height={100} width={80} rounding={3}><Image alt="" src="https://i.ibb.co/jVR29XV/stock5.jpg" fit="container"/></Box>}
+          image={<Box height={100} width={80} paddingX={1}><Image alt="" src="https://i.ibb.co/jVR29XV/stock5.jpg" fit="container" naturalWidth={1} naturalHeight={1}/></Box>}
           name="favorite art"
           onChange={({ checked }) => {
             setCheckedSp(checked);
--- a/docs/src/Checkbox.doc.js
+++ b/docs/src/Checkbox.doc.js
@@ -95,6 +95,108 @@ function CheckboxExample() {
 
 card(
   <Example
+    id="subtext"
+    description="Here is an example of a group of checkboxes with additional subtext applied."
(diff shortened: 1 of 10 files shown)


christianvuerings:
@ayeshakmaz @ashleyseto should be add 4px spacing between the `subtext` and `label`? Right now it looks very tight.

image

docs/src/Checkbox.doc.js
```diff
@@ -93,6 +93,108 @@ function CheckboxExample() {
...
+
+  return (
+    <Box margin={-2}>
+      <Box padding={1}>
+        <Checkbox
+          checked={checkedEn}
+          id="coral"
+          label="Coral"
+          subtext="Botanical art in coral and green"
+          image={<Box height={100} width={80} paddingX={1}><Image alt="" src="https://i.ibb.co/7bQQYkX/stock2.jpg" fit="container" naturalWidth={1} naturalHeight={1}/></Box>}
```
christianvuerings, line 171: @ayeshakmaz instead of setting the `paddingX` on the inner `Box`, is that something we should wrap around `image`?

```
{image && <Box paddingX={1}>{image}</Box>}
```

Related question: should there be a `maxWidth` on the image? @ashleyseto

docs/src/Checkbox.doc.js
```diff
@@ -93,6 +93,108 @@ function CheckboxExample() {
...
+    name="Example: With Images"
+    defaultCode={`
+function CheckboxExample() {
+    const [checkedEn, setCheckedEn] = React.useState(false);
+    const [checkedSp, setCheckedSp] = React.useState(false);
+    const [checkedCh, setCheckedCh] = React.useState(false);
+
+  return (
+    <Box margin={-2}>
+      <Box padding={1}>
```
christianvuerings, line 165: @ayeshakmaz could we make sure the negative margin corresponds with the `padding` value?

```
<Box margin={-2}>
  <Box padding={2}>
```

docs/src/Checkbox.doc.js
ayeshakmaz: There's currently no size limit spec'd on the image size (according to the [figma](https://www.figma.com/file/EWNPd8z7knrIIrQrd8bAr6/Radio-%26-Checkbox-improvements?node-id=0%3A1)), so I figure if users need to use a Box anyway, then the padding can be up to the user when they set that width and height (though I do recognize that that can produce some inconsistencies) We can at least add guidance in the docs around what padding might be most optimal?

ayeshakmaz:
@christianvuerings Updated the spacing for now, let me know what you think @ashleyseto 

Screen Shot 2020-10-09 at 3 52 09 PM


I'll wait and see what Ashley says about the image maxSize as well before encapsulating the image in a Box

ashleyseto:
> @christianvuerings Updated the spacing for now, let me know what you think @ashleyseto
> 
> Screen Shot 2020-10-09 at 3 52 09 PM
> 
> I'll wait and see what Ashley says about the image maxSize as well before encapsulating the image in a Box

4px spacing definitely looks better. There is some built in spacing at times in Figma which makes it difficult to do the mocks pixel to code perfect.

As far as default size, it is difficult to know exactly what all the sizes are that we'd need. We could release this with a default width option and a custom option. I'd say a good default width would be potentially 40px based on the example I've seen. Happy to also talk out some options too before implementation.

ashleyseto (approved):

docs/src/RadioButton.doc.js
```diff
@@ -50,32 +50,148 @@ function RadioButtonExample() {
...
+  );
+}
+`}
+  />
+);
+
+card(
+  <Example
+    description="
+    Here is an example of a group of radio buttons with subtext.
```
christianvuerings, line 91: NIT: No need to say `Here is an example of a ...` - instead we could just name it `A group of radio buttons with subtext`

christianvuerings (approved):

PR #1263, opened 2020-10-16
Table: Add onExpand function prop to RowExpandable

faisalgedi:
## Summary
The purpose of this PR is to add the onExpand function prop to the Table.RowExpandable component. 

An example use case for this prop would be if a user wanted to request an API call **only** when the row is expanded. 

## Test Plan
I used the docs and created more tests to make sure the function prop works as needed.

### DIFF
--- a/docs/src/Table.doc.js
+++ b/docs/src/Table.doc.js
@@ -335,7 +335,8 @@ card(
       },
       {
         name: 'onExpand',
-        type: `({ event: SyntheticMouseEvent<HTMLButtonElement> | SyntheticKeyboardEvent<HTMLButtonElement>, expanded: boolean }) => void`,
+        type: `({ event: SyntheticMouseEvent<HTMLButtonElement> | SyntheticKeyboardEvent<HTMLButtonElement> | SyntheticMouseEvent<HTMLAnchorElement>
+          | SyntheticKeyboardEvent<HTMLAnchorElement> expanded: boolean }) => void`,
         description: [
           'Callback fired when the expand arrow-down button component is clicked with a mouse or keyboard.',
         ],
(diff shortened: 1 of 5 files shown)


docs/src/Table.doc.js
```diff
@@ -341,193 +348,211 @@ card(
...
-              <Link href="https://pinterest.com">
-                <Box paddingX={3} paddingY={2}>
-                  Luna's Info
-                </Box>
-              </Link>
-            </Text>
-          </Card>
-        </Box>
+    function RowExpandableExample() {
+      const [iconShown, setIconShown] = React.useState(false);
```
christianvuerings, line 352: @faisalgedi Instead of showing an icon, could we make it more explicit and toggle some text? E.g.

`<Text>Row expanded</Text>` 

image

christianvuerings:
@faisalgedi looks like we have a flow failure on the build, I would try the following:
```
    | SyntheticKeyboardEvent<HTMLButtonElement>
```

image

docs/src/Table.doc.js
```diff
@@ -333,6 +333,14 @@ card(
         type: 'string',
         required: true,
       },
+      {
+        name: 'onExpand',
+        type: `({ event: SyntheticMouseEvent<HTMLButtonElement> | SyntheticKeyboardEvent<HTMLButtonElement> | SyntheticMouseEvent<HTMLAnchorElement>
+          | SyntheticKeyboardEvent<HTMLAnchorElement> expanded: boolean }) => void`,
+        description: [
+          'Callback fired when the expand arrow-down button component is clicked with a mouse or keyboard.',
```
christianvuerings, line 341: @faisalgedi could we update this to

> 'Callback fired when the expand button component is clicked.'

* We also do a callback with the `arrow-up` button
* We can't technically 'click' with a keyboard.

christianvuerings (approved):

PR #1264, opened 2020-10-16
Docs: New What's New landing page

ayeshakmaz:
Create a new "What's New" landing page to showcase updates to Gestalt
- Also updates link styling for markdown sections

This is V1, in V2 I plan to add ability to link to codemods in github using a label in the PR

### DIFF
--- a/docs/src/What's New.doc.js
+++ b/docs/src/What's New.doc.js
@@ -1,0 +1,74 @@
+// @flow strict
+import React, { type Node, useEffect } from 'react';
+import { Box, Text, Link, Stack, Row } from 'gestalt';
+import Markdown from './components/Markdown.js';
+import PageHeader from './components/PageHeader.js';
+
+const cards: Array<Node> = [];
+const card = c => cards.push(c);
+
+function Changelog() {
+  const [changelogData, setChangelogData] = React.useState('Loading...');
+
+  const getChangelog = () => {
+    fetch('https://cdn.jsdelivr.net/gh/pinterest/gestalt@master/CHANGELOG.md')
+      .then(response => {
+        if (!response.ok) {
+          return '# There was error loading the changelog, please try again later.';
+        }
+        return response.text();
+      })
+      .then(data => {
+        setChangelogData(data);
+      });
+  };
+
+  useEffect(() => {
+    getChangelog();
+  });
+
+  return (
+    <Box>
+      <PageHeader name="What's New 🎉" showSourceLink={false} />
+      <Stack gap={2}>
+        <Row gap={2}>
+          <Link inline target="blank" href="https://npmjs.org/package/gestalt">
+            <img
+              src="https://img.shields.io/npm/v/gestalt.svg?label=gestalt"
+              alt="Gestalt NPM package version badge"
+            />
+          </Link>
+
+          <Link
+            inline
+            target="blank"
+            href="https://npmjs.org/package/gestalt-datepicker"
+          >
+            <img
+              src="https://img.shields.io/npm/v/gestalt-datepicker.svg?label=gestalt-datepicker"
+              alt="Gestalt DatePicker NPM package version badge"
+            />
+          </Link>
+        </Row>
+        <Text>
+          Gestalt is a set of React UI components that enforces Pinterest’s
+          design language. We use it to streamline communication between
+          designers and developers by enforcing a bunch of fundamental UI
+          components. This common set of components helps raise the bar for UX
+          &amp; accessibility across Pinterest.
+        </Text>
+        <Text>
+          Find information below on all new and updated components by version
+          number, as well as available codemods to help upgrade between
+          versions.
+        </Text>
+      </Stack>
+
+      <Markdown text={changelogData} type="changelog" />
+    </Box>
+  );
+}
+
+card(<Changelog />);
+
+export default cards;
--- a/docs/src/What's New.doc.js
+++ b/docs/src/What's New.doc.js
@@ -8,7 +8,9 @@ const cards: Array<Node> = [];
 const card = c => cards.push(c);
 
 function Changelog() {
-  const [changelogData, setChangelogData] = React.useState('Loading...');
+  const [changelogData, setChangelogData] = React.useState(
+    'Loading changelog from GitHub...'
+  );
 
   useEffect(() => {
     const fetchChangelog = async () => {
@@ -18,7 +20,7 @@ function Changelog() {
       setChangelogData(
         result.ok
           ? await result.text()
-          : '# There was error loading the changelog, please try again later.'
+          : 'There was error loading the changelog, please try again later.'
       );
     };
 
--- a/docs/src/What's New.doc.js
+++ b/docs/src/What's New.doc.js
@@ -10,22 +10,20 @@ const card = c => cards.push(c);
 function Changelog() {
   const [changelogData, setChangelogData] = React.useState('Loading...');
 
-  const getChangelog = () => {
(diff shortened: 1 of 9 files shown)


christianvuerings:
@ayeshakmaz looks like this test is now failing (expected since we updated the first page we show):
image

christianvuerings:
Could we remove the npm badges + Gestalt description from the installation page (since we now have it on the what's new page)?

docs/src/What's New.doc.js
```diff
@@ -0,0 +1,74 @@
...
+const card = c => cards.push(c);
+
+function Changelog() {
+  const [changelogData, setChangelogData] = React.useState('Loading...');
+
+  const getChangelog = () => {
+    fetch('https://cdn.jsdelivr.net/gh/pinterest/gestalt@master/CHANGELOG.md')
+      .then(response => {
+        if (!response.ok) {
+          return '# There was error loading the changelog, please try again later.';
```
christianvuerings, line 17: @ayeshakmaz looks like this does not work as expected, it keep showing 'Loading...' for me.

You can try it out by:
1) Loading the `What's new page`
2) Going to a different page
3) Disconnect from the internet
4) Go back to the `What's new page`
image

docs/src/What's New.doc.js
```diff
@@ -0,0 +1,74 @@
...
+        }
+        return response.text();
+      })
+      .then(data => {
+        setChangelogData(data);
+      });
+  };
+
+  useEffect(() => {
+    getChangelog();
```
christianvuerings, line 27: @ayeshakmaz could we use async / await and inline the function since we don't use it anywhere else?

```
useEffect(() => {
    const fetchChangelog = async () => {
      const result = await fetch('https://cdn.jsdelivr.net/gh/pinterest/gestalt@master/CHANGELOG.md');
      setChangelogData(response.ok ? await response.text() : '# There was error loading the changelog, please try again later.');
    };
 
    fetchChangelog();
}, []);
```

docs/src/components/Markdown.js
```diff
@@ -28,27 +29,65 @@ const stripIndent = (str: string): string => {
...
+const formatCmpName = listitem => {
+  const cmpNamesToIgnore = ['doc', 'docs', 'internal', 'codemod'];
+  const nameAndUpdate = listitem.split(':', 2);
+  const cmpName = nameAndUpdate[0];
+  if (
+    !cmpNamesToIgnore.includes(cmpName.toLowerCase()) &&
+    nameAndUpdate.length > 1
+  ) {
+    const cmpURL = `https://gestalt.netlify.app/${cmpName}`;
+    const formattedListItem = `<a href="${cmpURL}">${cmpName}:</a>${nameAndUpdate[1]}`;
```
christianvuerings, line 57: @ayeshakmaz could we keep the `:` out of the `<a />` so that's not clickable?

image

docs/src/components/Markdown.js
```diff
@@ -28,27 +29,65 @@ const stripIndent = (str: string): string => {
...
+      pullRequestNumRegex,
+      pullRequestAnchor
+    );
+    return listItemWithLink;
+  }
+  return listitem;
+};
+
+const formatCmpName = listitem => {
+  const cmpNamesToIgnore = ['doc', 'docs', 'internal', 'codemod'];
```
christianvuerings, line 49: @ayeshakmaz this doesn't always seem to work, especially when there is more than 1 component.

`ActivationCard/Callout` links to https://gestalt.netlify.app/ActivationCard/Callout - which is not a valid page.

Might be worthwhile to dynamically pull in all the Component Pages + replace those

image

docs/src/components/Markdown.js
```diff
@@ -28,27 +29,65 @@ const stripIndent = (str: string): string => {
...
+              </a>`
             : ''
         }
-        ${input}
       </h${level}>`;
   };
 
+  if (type === 'changelog') {
+    renderer.listitem = listitem => {
+      let formattedListItem = formatCmpName(listitem);
```
christianvuerings, line 85: @ayeshakmaz instead of using a `let`, could we make this immutable with `const` or combine the function calls?

```
renderer.listitem = listitem => makePullRequestLink(formatCmpName(listitem));
```

docs/src/components/Markdown.js
```diff
@@ -28,27 +29,65 @@ const stripIndent = (str: string): string => {
...
+    const listItemWithLink = listitem.replace(
+      pullRequestNumRegex,
+      pullRequestAnchor
+    );
+    return listItemWithLink;
+  }
+  return listitem;
+};
+
+const formatCmpName = listitem => {
```
christianvuerings, line 48: @ayeshakmaz let's try to be explicit and name this `formatComponentName`

docs/src/components/Markdown.css
```diff
@@ -18,8 +18,35 @@ html[dir="rtl"] .Markdown ul + pre {
...
+  align-items: baseline;
+  display: flex;
+  scroll-margin-top: 75px;
+}
+
+.Markdown h3 {
+  color: var(--g-colorGray200);
+}
+
+.anchor {
```
christianvuerings, line 40: @ayeshakmaz could we also scope these to `.Markdown`?

`.Markdown .anchor`
and
`.Markdown .headerLink`

docs/src/What's New.doc.js
ayeshakmaz: OO yes, sure thing

docs/src/components/Markdown.js
ayeshakmaz: Yeah this one is funky - it still does work, in the sense that it will just bring you to the first cmp's page (ActivationCard in that example ^) but with a wonky URL. I can make an update to look for slashes, and link each cmp individually, what do you think?

docs/src/components/Markdown.css
ayeshakmaz: Sure thing, any reason we don't use Sass?

docs/src/What's New.doc.js
ayeshakmaz: Ahh this is because the default value for the changelog data is `Loading...` 

If you disconnect from the internet, it should serve from cache, but if I'm mistaken that fetch wouldn't happen at all in offline mode right?

christianvuerings: Afaik the fetch should happen but might fail - one thing we can try is to wrap it in a `try/catch`

docs/src/components/Markdown.css
christianvuerings: @ayeshakmaz we haven't need it so far. Ideally these would be Gestalt components instead of custom HTML tags.

christianvuerings:
@ayeshakmaz thanks for all the updates!

Few more comments:

## Could we ensure there is no extra space, so no extra `-` at the end of these links?

`/What's%20New#14-11-2-oct-7-2020`
instead of
`/What's%20New#14-11-2-oct-7-2020-`

## Reverts currently break the linking:

image

The first link goes to https://gestalt.netlify.app/Revert%20%22Button which does not exist. I am wondering if we can add an integration test to check for broken links.

docs/src/components/Markdown.js
```diff
@@ -6,8 +6,11 @@ import highlightjs from 'highlight.js';
 import './Markdown.css';
 import 'highlight.js/styles/a11y-light.css';
 
+const sidebarIndex = require('./sidebarIndex.js');
```
christianvuerings, line 9: @ayeshakmaz nit - can we use a regular `import` in this case? In `scripts/` we do have to use `require` since those don't go through webpack

ayeshakmaz: oh whoops sure thing

christianvuerings (approved):

PR #1265, opened 2020-10-19
Layer: fix issue with Flyout when zIndex gets set

christianvuerings:
When we set a zIndex on a `<Layer />` which wraps a `<Flyout />`, the flyout currently disappears.
This happens because of the `position: absolute` which is implicitly set when we set `zIndex` on `<Layer />`

## Test Plan

For one of the examples in https://gestalt.netlify.com/Flyout

1. Set `zIndex` prop on the wrapping `<Layer />`
```
<Layer zIndex={new FixedZIndex(2)}>
  <Flyout
```
2. Ensure the `Flyout` shows on the screen

### Before: Flyout no longer shows shows
Screen Shot 2020-10-19 at 1 48 59 PM

### After: Flyout shows
Screen Shot 2020-10-19 at 1 48 53 PM


dangerismycat:
I was having similar issues with Layer and Sheet, where the contents of the Layer (the Sheet) would disappear when the user interacted with the Sheet, causing a re-render. I solved that issue by memoizing the zIndex:
```
  const memoizedZIndex = useMemo(() => new CompositeZIndex([headerZIndex]), [headerZIndex]);
```
In this case, headerZIndex is a FixedZIndex defined outside the component with an imported const:
```
const headerZIndex = new FixedZIndex(HEADER_ZINDEX);
```

AlbertCarreras (approved):

PR #1268, opened 2020-10-22
Docs: Update accessibility tests for all docs pages

ayeshakmaz:
## Summary

Remove .skip from most Docs pages

Typeahead and Tabs are still skipped because the underlying component has accessibility issues. I will address those in a separate ticket/PR

- Also added ability to specify aria-labelledby for RadioButtons and Checkboxes, which makes our docs more accessible, but will likely be useful in the future as well for the components themselves

## Test Plan

Ran all Cypress a11y tests for docs


cypress/integration/accessibility_Table_spec.js
```diff
@@ -4,14 +4,17 @@ describe('Table Accessibility check', () => {
...
     cy.configureAxe({
       rules: [
         {
           id: 'color-contrast',
           enabled: false,
         },
+        {
+          id: 'scrollable-region-focusable',
+          enabled: false,
+        },
```
ayeshakmaz, line 17: Added this to avoid confusion for the sticky header example

docs/src/Pulsar.doc.js
```diff
@@ -78,9 +78,7 @@ class FlyoutExample extends React.Component {
             top: 10,
             left: 10,
             pointerEvents: "none", }}>
-            <TapArea onTap={({ event }) => this.handleClick(event)} rounding="circle" fullWidth={false}>
               <Pulsar paused={this.state.open} />
-            </TapArea>
```
ayeshakmaz, line 83: @christianvuerings I'm unclear about the purpose of this TapArea here - happy to put it back if it's needed

cypress/integration/accessibility_Table_spec.js
christianvuerings: @ayeshakmaz could you add a comment explaining why we disable it?

docs/src/Pulsar.doc.js
christianvuerings: @ayeshakmaz no need for it, this should be handled by the `<Button />`

docs/src/Checkbox.doc.js
```diff
@@ -327,16 +327,28 @@ card(
     disabled={[false, true]}
     indeterminate={[false, true]}
     size={['sm', 'md']}
+    hasCheckerboard={false}
   >
     {(props, i) => (
-      <Checkbox id={`example-${i}`} onChange={() => {}} {...props} />
+      <Checkbox
+        id={`example-${i}`}
+        accessibilityLabelledBy={`combinations-label-${i}`}
```
christianvuerings, line 335: @ayeshakmaz we've been trying to avoid using `accessibilityLabelledBy` and use `id` + `<Label htmlFor />` instead. 

Looks like we only use `accessibilityLabelledBy` in the `<Combination />`.

docs/src/components/Combination.js
```diff
@@ -117,7 +117,7 @@ export default function Combination({
             alignItems="center"
           >
             {showValues && (
-              <Box marginBottom={2}>
+              <Box marginBottom={2} id={`${id || ''}-label-${i}`}>
```
christianvuerings, line 120: @ayeshakmaz this is a bit of magic - using the `-label-`, would there be a way to avoid that? E.g. not using `accessibilityLabelledBy`

packages/gestalt/src/TypeaheadInputField.js
```diff
@@ -170,7 +170,9 @@ const TypeaheadInputFieldWithForwardRef: React$AbstractComponent<
             right
           >
             <Icon
-              accessibilityLabel=""
+              accessibilityLabel={
+                !hasValue ? 'Show options' : 'Clear selection'
```
christianvuerings, line 174: @ayeshakmaz we can not have any English strings in any Gestalt component since they would not be internationalized.

Instead we add extra accessibility labels, e.g. `accessibilityModalLabel`

ayeshakmaz: Removed this for now, and will address in[ PDS-2337](https://jira.pinadmin.com/browse/PDS-2337)

docs/src/components/Combination.js
ayeshakmaz: Switched this up to use a new prop on Combinations `useLabelForValues`, let me know what you think

docs/src/components/Combination.js
```diff
@@ -15,6 +15,7 @@ type Props = {
   showHeading?: boolean,
   showValues?: boolean,
   stacked?: boolean,
+  useLabelForValues?: boolean,
```
christianvuerings, line 18: @ayeshakmaz instead of making this a boolean prop, can we pass in a `labelPrefix`? Using a boolean results in only being able to use 1 `Combination` with `useLabelForValues` on a page. If we pass something like `radiobutton-size-example` to `labelPrefix`, we can have multiple `Combination` elements with `useLabelForValues` on a page.

ayeshakmaz: ahhhh yes, great call - I was also thinking of using the ID, but this will give us the most flexibility, I'll make that update!

PR #1270, opened 2020-10-26
Internal: improve package.json with repository / keywords + exports information

christianvuerings:
https://www.skypack.dev/view/gestalt surfaced a couple of fields we should add to our `package.json` file:

* `repository`: Make it easy to see where the code lives when you've install the package
* `keywords`: Improves discoverability of the the package
* `exports`: Explicitly states which files are meant to be imported:

Before
```
import "../node_modules/gestalt/dist/gestalt.css";
```

After
```
import "gestalt/css";
```

## Test Plan

CI Passes

### DIFF
--- a/packages/gestalt/package.json
+++ b/packages/gestalt/package.json
@@ -8,6 +8,14 @@
   "jsnext:main": "dist/gestalt.es.js",
   "module": "dist/gestalt.es.js",
   "style": "dist/gestalt.css",
+  "exports": {
+    ".": {
+      "import": "./dist/gestalt.es.js"
+    },
+    "./css": {
+      "import": "./dist/gestalt.css"
+    }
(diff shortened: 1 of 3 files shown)


packages/gestalt/package.json
```diff
@@ -31,5 +39,15 @@
     "last 2 versions",
     "not IE < 11",
     "not <1%"
-  ]
+  ],
+  "keywords": [
+    "react",
+    "react component",
+    "pinterest",
+    "ui library"
```
ayeshakmaz, line 47: Should we add "design system" here?

christianvuerings: Definitely - added in https://github.com/pinterest/gestalt/pull/1271

PR #1271, opened 2020-10-26
Internal: Add 'design system' to package.json keywords

christianvuerings:
Follow up from #1270

https://github.com/pinterest/gestalt/pull/1270#discussion_r512159520


ayeshakmaz (approved):
👍

PR #1272, opened 2020-10-27
Docs: persist dark mode and RTL settings across refreshes

christianvuerings:
## Changes

* Persist dark mode and RTL settings across refreshes (using local storage)
* Introduces `useLocalStorage` to make it easier to work with local storage in our docs. Forked from https://usehooks.com/useLocalStorage/

## Test Plan

1. Go to `/`
2. Switch to dark mode
3. Reload

Expected result: dark mode is still enabled.


ayeshakmaz (approved):
Looks and works great 👍

PR #1273, opened 2020-10-27
Tests: Fix flaky test on master

christianvuerings:
Issue was that in the sidebar we wrap a `TextField` in a `<div />` which we don't do on the `What's New` Page

image

## Test Plan

* CI tests pass


ayeshakmaz:
This will also be fixed by #1268

christianvuerings:
> This will also be fixed by #1268

@ayeshakmaz I did see this test fail on #1268 - it's currently blocking that PR getting merged

image

PR #1274, opened 2020-10-29
Callout: V2 Redesign for responsiveness

AlbertCarreras:
Callout: V2 Redesign for responsiveness

Before 
Kapture 2020-10-29 at 16 24 10

After
Kapture 2020-10-29 at 16 22 15

https://www.loom.com/share/696bcb05f2e7495089a2c3c2907ff0cb

### DIFF
--- a/docs/src/Callout.doc.js
+++ b/docs/src/Callout.doc.js
@@ -21,15 +21,15 @@ card(
 <Callout
   type="info"
   iconAccessibilityLabel="Info icon"
-  title="Your business account was successfully created!"
-  message="Get a badge, show up in more shopping experiences and more. Apply to the Verified Merchant Program—it’s free!"
+  title="Your business account was created!"
+  message="Apply to the Verified Merchant Program!"
   primaryLink={{href: "https://pinterest.com", label:"Get started"}}
   secondaryLink={{href: "https://pinterest.com", label:"Learn more"}}
   dismissButton={{
     accessibilityLabel: 'Dismiss banner',
     onDismiss: ()=>{},
   }}
-  />
+/>
 `}
   />
 );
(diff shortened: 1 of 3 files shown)


docs/src/Callout.doc.js
```diff
@@ -18,18 +18,20 @@ card(
...
-  title="Your business account was successfully created!"
-  message="Get a badge, show up in more shopping experiences and more. Apply to the Verified Merchant Program—it’s free!"
-  primaryLink={{href: "https://pinterest.com", label:"Get started"}}
-  secondaryLink={{href: "https://pinterest.com", label:"Learn more"}}
-  dismissButton={{
-    accessibilityLabel: 'Dismiss banner',
-    onDismiss: ()=>{},
-  }}
-  />
+  <Box margin={1}>
```
christianvuerings, line 21: @AlbertCarreras could we avoid the wrapping `<Box />`? Ideally this would be easiest copy/paste example and we don't want a 4px margin around each Callout each time

ashleyseto (approved):

docs/src/Callout.doc.js
christianvuerings: Should be good to merge once this comment is addressed

AlbertCarreras: Thanks for catching this. I left this dead code that I used for testing 2 Callouts for the demo videos

christianvuerings:
@AlbertCarreras When we wrap, the buttons butt up against the text. /cc @ashleyseto 

`Learn More` and `Program`: 
image

You can fix this by using vertical symmetrical padding on the text and on the buttons: https://gestalt.netlify.app/Layouts#Form-Layout

Going to r+ since design approved but would be good to follow up and fix this

christianvuerings (approved):

PR #1275, opened 2020-10-30
Docs: Update GitHub source link for component pages

ayeshakmaz:
Update "Source" link to "View in GitHub" with icon

Screen Shot 2020-10-30 at 4 00 46 PM

## Test Plan

Cypress a11y test pass


docs/src/components/PageHeader.js
```diff
@@ -43,9 +43,29 @@ export default function ComponentHeader({
...
-            <Link href={githubUrl(fileName || name)} inline>
-              Source
+          <Text color="gray" align="center">
+            <Link
+              href={githubUrl(fileName || name)}
+              target="blank"
+              accessibilityLabel={`View ${name} in GitHub`}
+            >
+              <Box display="flex" alignItems="center">
+                <svg
```
christianvuerings, line 53: @ayeshakmaz can we use a Gestalt `Icon` for this? See `dangerouslySetSvgPath`

christianvuerings:
@ashleyseto Could you take a look at this from design?

AlbertCarreras:
The clickable area is pretty large 
image

Could we add `inline` so it does not get width 100%?
image

AlbertCarreras (approved):

ashleyseto:
> @ashleyseto Could you take a look at this from design?

I think we're making changes to the docs in the wrong order. It should be proposed idea, design then implementation. Long term, I wouldn't put this under the title. I'd probably be aligning it top right or somewhere else on the page because it breaks up the name of the component and the definition. Short term, I'd remove the icon.

christianvuerings:
Closing this for now - let's address when we re-design the docs

ayeshakmaz:
@ashleyseto @christianvuerings I'd look at this as more of a prototyping/experiment opportunity - given some of the feedback we got, I think it'd be great to make a small incremental change here, rather than wait for the redesign, and at least improve the language here. What do ya'll think about switching to "View in GitHub" for now? (without the icon)

PR #1283, opened 2020-11-05
Upsell: Add component

norrise120:
This PR introduces the new Upsell component: 
Examples: 
Screen Shot 2020-11-05 at 12 17 41 PM
Screen Shot 2020-11-05 at 12 17 33 PM
Screen Shot 2020-11-05 at 12 16 25 PM

This is adapted from the Callout component, with minor adjustments.


packages/gestalt/src/Upsell.js
```diff
@@ -0,0 +1,164 @@
...
+              </Box>
+            )}
+            <Text>{message}</Text>
+          </Box>
+        </Box>
+        <Box mdDisplay="flex" marginStart="auto" mdPaddingY={3}>
+          {secondaryLink && (
+            <UpsellLink type="secondary" data={secondaryLink} />
+          )}
+          {primaryLink && (
```
norrise120, line 116: One thing to note is that in the design spec for Upsell (and Callout) there is 48px of margin on the right side of this button. This is _not_ consistent with the current implementation of Callout where there is only 32px on margin on the right side. I opted to be consistent with the current implementation of Callout, but let me know if I should make this 48px, make both 48px or leave it as is.

docs/src/Upsell.doc.js
```diff
@@ -0,0 +1,139 @@
+// @flow strict
+import React, { type Node } from 'react';
+import PropTable from './components/PropTable.js';
+import Example from './components/Example.js';
+import PageHeader from './components/PageHeader.js';
+
+const cards: Array<Node> = [];
+const card = (c) => cards.push(c);
+
+card(<PageHeader name="Upsell" />);
```
norrise120, line 10: Component description coming as soon as I hear back from design!

packages/gestalt/src/Upsell.js
```diff
@@ -0,0 +1,164 @@
...
+  return (
+    <Box
+      display="flex"
+      alignItems="center"
+      justifyContent="center"
+      paddingX={1}
+      marginTop={type === 'primary' && stacked ? 2 : undefined}
+      mdMarginTop="auto"
+      mdMarginBottom="auto"
+      rounding={type === 'primary' ? 'pill' : undefined}
```
AlbertCarreras, line 57: we don't need this line. It's probably dead code from before having LinkButtons

norrise120: Removed! And also removed from `Callout.js`

packages/gestalt/src/Upsell.js
AlbertCarreras: Correct. Both Upsell and Callout have a 48px padding on the right while a 32px padding on the left. 
I don't know if it's the finest solution but this fixes it/
```
        <Box mdDisplay="flex" marginStart="auto" mdMarginEnd={4}>
          {secondaryLink && (
            <CalloutLink type="secondary" data={secondaryLink} />
          )}
          {primaryLink && (
            <CalloutLink
              stacked={!!secondaryLink}
              type="primary"
              data={primaryLink}
            />
          )}

```

packages/gestalt/src/Upsell.js
AlbertCarreras: Thank you!

AlbertCarreras:
This is a note for the Upsell and Callout. As a followup to this, (and before extra work on Upsell) we could create an InternalBanner component that would share most of the internal logic. It would have a variant prop defining `callout` or `upsell` and it would preestablish some of the design differences and flowtype segmentation. Otherwise, we have 2 sibling components and much higher maintenance.

AlbertCarreras:
image

Just copy paste `cypress/integration/accessibility_Callout_spec.js` and change the names to Upsell

packages/gestalt/src/Upsell.js
norrise120: Updated here and Callout! Thank you!!

Screen Shot 2020-11-05 at 3 13 37 PM
Screen Shot 2020-11-05 at 3 13 25 PM

AlbertCarreras (approved):

packages/gestalt/src/Upsell.js
```diff
@@ -99,7 +99,12 @@ export default function Upsell({
           mdMarginBottom={primaryLink || secondaryLink ? 0 : undefined}
           mdPaddingY={3}
         >
-          <Box marginBottom="auto" marginTop="auto" maxWidth={648} paddingX={6}>
+          <Box
+            marginBottom="auto"
+            marginTop="auto"
+            maxWidth={648}
+            marginEnd={6}
```
AlbertCarreras, line 106: For reference, 6 boints  because that was what we had on Callout, with a paddingX=6 around the message
image

AlbertCarreras (approved):

ashleyseto (approved):

PR #1284, opened 2020-11-06
Button: Add iconStart and iconColor property 

simkessy:
We have requirement for a button with icons at the start

image

## Test Plan
It is tested, as accessible as iconEnd and documented 

image


simkessy:
Rejected by designers

PR #1288, opened 2020-11-10
Flex: Replace Row/Stack with Flex

dangerismycat:
This PR replaces the existing Row and Stack components with a new component, Flex. Flex also has a sub-component, FlexItem, accessed as `<Flex.Item>`. This solves our current problems with Row/Stack:

- By unifying to a single component with a `direction` prop, we avoid repeating code and make for easier imports for users.
- If the `gap` prop is set on Flex, children will be conditionally wrapped with FlexItem. `gap` padding is set on that FlexItem using a specific class, not the dangerous `*` universal child selector. This prevents collisions with child padding.
- For each child, if the child is already a FlexItem, it will not be wrapped. This allows using `<Flex.Item>` directly, where the user can set `alignSelf` or `flex` props. `padding` is _not_ a valid prop on this component.

# TODO
- [x] Stress test to make sure nothing breaks and we can achieve desired layouts
- [x] Add codemod to convert current Stack / Row to `Flex`
- [x] Update docs with plenty of examples, including using FlexItem
- [x] Update tests


packages/gestalt/src/boxTypes.js
```diff
@@ -62,6 +62,32 @@ export type Display =
...
+  | 4
+  | 5
+  | 6
+  | 7
+  | 8
+  | 9
+  | 10
+  | 11
+  | 12
+  | 13
```
christianvuerings, line 79: @dangerismycat do we need to go all the way to `24` here? All of these will require extra generated CSS.

Looks like previously we only went to `12`

image

packages/gestalt/src/Flex.js
```diff
@@ -0,0 +1,123 @@
...
+  gap = 0,
+  justifyContent = 'center',
+  ...rest
+}: Props): Node {
+  const children = gap
+    ? Children.map(childrenProp, (child, index) => {
+        if (child === null || child === undefined) {
+          return null;
+        }
+        return wrapWithComponent({
```
christianvuerings, line 79: @dangerismycat would there be a way to make this an actual React component?

packages/gestalt/src/FlexItem.js
```diff
@@ -0,0 +1,35 @@
...
+import styles from './Flex.css';
+import {
+  AlignSelfPropType,
+  FlexPropType,
+  type AlignSelf,
+  type Flex,
+} from './boxTypes.js';
+
+export type Props = {|
+  alignSelf?: AlignSelf,
```
christianvuerings, line 14: Seems like a good start, I do imagine this list growing over time.

docs/src/Flex.doc.js
```diff
@@ -1,6 +1,6 @@
 // @flow strict
```
christianvuerings, line 1: @dangerismycat could we add more info on how to use `Flex.Item`?

packages/gestalt/src/boxTypes.js
dangerismycat: The current implementation applies the gap value to each side, so 1 boint actually means 8px between elements, with a max of 96px between elements. However, this wasn't very flexible, since it was impossible to set a gap of odd numbers of actual boints.

The new implementation splits the gap value between each adjoining pair of elements, allowing for a single boint gap. However, this means the largest gap is only 48px, which is pretty small. I bumped this up to 24 to allow for a larger total gap, similar to what we can currently achieve. (This is also necessary to be a non-breaking change, though I don't know if we have any `gap={12}` usages.)

Happy to change it back if we want to be more restrictive instead, but I fear that will lead to lower adoption.

packages/gestalt/src/Flex.js
dangerismycat: Sure. Does that gain us anything? The avalanche of Flow errors I had to deal with while adapting this code means we're definitely not losing type safety. 😅 It accepts a generic type that it seems to be able to infer even without setting explicitly.

packages/gestalt/src/FlexItem.js
dangerismycat: The list of props on FlexItem? What else do you think would be needed?

docs/src/Flex.doc.js
dangerismycat: Definitely! I haven't touched the docs or tests aside from fixing them to not be broken — was planning on updating once we've nailed down the implementation. I'll add that to the TODOs above.

packages/gestalt/src/boxTypes.js
christianvuerings: @dangerismycat Thanks for explaining, we should also make the divide by 2 change in the codemod. I know in certain places we will pass `gap` dynamically but then we can throw an error to manually check.

Since we only have one place in the pinterest codebase which uses a legacy `gap={12}`, we should stick to only allowing up to 12 (since that's what we allow on other components as well). I know that limits the usage but we can always add it in if there is a big demand.

packages/gestalt/src/Flex.js
christianvuerings: @dangerismycat mostly making it easier to read through and see that it returns a React Element. This is a NIT, not a must have (since we can always change it later)

packages/gestalt/src/FlexItem.js
christianvuerings: @dangerismycat things like flex basis, margins (auto): https://css-tricks.com/the-peculiar-magic-of-flexbox-and-auto-margins/ . For now, let's leave it as is, this is a good start.

dangerismycat: Sounds good. I'm not familiar with flex-basis, but it seems useful in some situations. We should add the ability to set auto margins soon, but I think that API will need some more thought, since I don't think we want to open up the full `margin` attribute.

packages/gestalt/src/Flex.js
dangerismycat: Cool cool, we can refactor later since it's internal.

packages/gestalt/src/boxTypes.js
dangerismycat: I discussed with @ashleyseto and we agree, so I'll change this back to 12. We can up it to 24 in the future if needed.

packages/gestalt-codemods/x.x.x/flex-update-from-row.js
```diff
@@ -0,0 +1,112 @@
...
+
+    if (!touchableSpecifier) {
+      return;
+    }
+
+    touchableLocalIdentifierName =
+      touchableSpecifier && touchableSpecifier.local.name;
+
+    const newSpecifiers = [
+      // Strip out Touchable import
```
christianvuerings, line 33: @dangerismycat Could we update these comments?

packages/gestalt-codemods/x.x.x/flex-update-from-row.js
```diff
@@ -0,0 +1,112 @@
...
+      if (node.openingElement.name.name !== touchableLocalIdentifierName) {
+        return;
+      }
+
+      // Double 'gap'
+      node.openingElement.attributes = node.openingElement.attributes.map(
+        (attr) => {
+          const attribute = attr;
+          if (attribute.name && attribute.name.name === 'gap') {
+            const doubledVal = attribute.value.expression.value * 2;
```
christianvuerings, line 71: @dangerismycat What if the value of `gap` is not a number?

```jsx
const gap = 2;
//
<Row gap={gap} />
```

In those cases, we usually throw an exception to modify it manually: https://git.io/Jklij

christianvuerings:
@dangerismycat could we add an a11y test for `Flex`? Seeing this in the CI output:

```
The following doc page does not have an accessibility integration test: Flex
```

image

packages/gestalt/src/Flex.js
```diff
@@ -0,0 +1,123 @@
...
+  height: DimensionPropType,
+  justifyContent: JustifyContentPropType,
+  maxHeight: DimensionPropType,
+  maxWidth: DimensionPropType,
+  minHeight: DimensionPropType,
+  minWidth: DimensionPropType,
+  overflow: OverflowPropType,
+  width: DimensionPropType,
+  wrap: PropTypes.bool,
+};
```
ayeshakmaz, line 123: Could we give Flex and FlexItem display names? Could be helpful when debugging

packages/gestalt-codemods/x.x.x/flex-update-from-stack.js
```diff
@@ -0,0 +1,119 @@
+/**
+ * Converts
+ *  <Stack gap={x} />
+ * to
+ *  <Flex alignItems="start" justifyContent="center" gap={2x} />
```
ayeshakmaz, line 5: Should this also add in `direction="column"` to maintain the Stack?

dangerismycat: 🤦 Indeed it should 😅

christianvuerings (approved):

packages/gestalt/src/utils/wrapWithComponent.js
```diff
@@ -0,0 +1,68 @@
...
+
+  return Components.some(
+    // $FlowFixMe[incompatible-call]
+    (AComponent) => typeof type !== 'string' && isComponent(AComponent, type)
+  );
+}
+
+// Conditionally wraps `element` in `Component` if it is not already an instance of
+// `Component`. If `props` is passed, those will be added as props on the
+// wrapped component. If `element` is nullish, the component is returned unwrapped.
```
AlbertCarreras, line 49: Is this correct? `If `element` is nullish, the component is returned unwrapped.`, it returns null if so

packages/gestalt/src/Flex.js
dangerismycat: Oh, sorry, I missed this earlier. It shouldn't matter; displayName is really only useful when returning an anonymous component, or one that's wrapped in something like forwardRef. DevTools should pick up the names of these components from the function declaration just fine. (The same is true for function expressions.)

packages/gestalt/src/utils/wrapWithComponent.js
dangerismycat: Nope, it's not. I'll put together a lil PR to update. Good catch!

PR #1289, opened 2020-11-10
Modal: add padding to container and margin top to children container

diegofontenelle:
# Modal spacing update

- Add 16px padding to container (x axis)
- Add 24px padding to Header container (y axis)
- Add 16px margin top to children container 
- Add 16px padding to footer container (y axis)

Proposed:
image

Implemented:
image
image

On smaller windows: 
image

Sheet: 
image

## Test Plan

Manual run

### DIFF
--- a/packages/gestalt/src/Modal.js
+++ b/packages/gestalt/src/Modal.js
@@ -40,7 +40,7 @@ function Header({ heading }: {| heading: string | Node |}) {
   }
 
   return (
-    <Heading size="md" accessibilityLevel={1}>
+    <Heading size="sm" accessibilityLevel={1}>
       {heading}
     </Heading>
   );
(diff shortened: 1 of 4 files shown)


christianvuerings:
@diegofontenelle thanks for the PR! @anniemochi could you verify from the design side?

Couple of notes:
* We should also update `Sheet` which uses similar padding
* The current change breaks scrolling (happens on windows with smaller heights). So the padding still needs to be on the outward component, it can just be less than before.
image

diegofontenelle:
> @diegofontenelle thanks for the PR! @anniemochi could you verify from the design side?
> 
> Couple of notes:
> 
> * We should also update `Sheet` which uses similar padding
> * The current change breaks scrolling (happens on windows with smaller heights). So the padding still needs to be on the outward component, it can just be less than before.
>   image

Hi Chris! Thanks for reviewing, I have updated the PR and added new images showing how it looks right now. Let me know if this is good or you meant something else

anniemochi (approved):

christianvuerings:
Discussed with @anniemochi and @diegofontenelle: We will communicate this change internally first before merging

diegofontenelle:
Making small changes to the padding of the children container on both components after reviewing the designs with @anniemochi

diegofontenelle:
After further investigating the modal spacing adjustments we were proposing in relation to modal/sheet examples from the Monetization org, the Gestalt team has determined that we should not make those spacing changes after all (because all of the Monetization modal/sheet titles and content are left-aligned, changing the header/footer padding creates awkward alignment issues which would take a lot more effort to address).

PR #1291, opened 2020-11-18
Toast: Update documentation around Text property

ayeshakmaz:
Updated the docs to hopefully avoid confusion when users specify a text color within the Toast component's Text prop

## Test Plan

Manually tested

### DIFF
--- a/docs/src/Toast.doc.js
+++ b/docs/src/Toast.doc.js
@@ -56,7 +56,48 @@ card(
 card(
   <Example
     id="textOnlyExample"
-    name="Example: Text only"
+    name="Example: Simple Text"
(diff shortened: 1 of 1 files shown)


docs/src/Toast.doc.js
```diff
@@ -46,7 +46,7 @@ card(
         name: 'text',
         type: 'string | Array<string>',
```
AlbertCarreras, line 47: This is incorrect, the type is   `text: string | Element<*>,`

ayeshakmaz: I was wondering about that - will update : )

AlbertCarreras (approved):

PR #1293, opened 2020-11-30
Box: Allow zero padding overrides

christianvuerings:
We currently do not allow for `padding` to be overwritten when hitting a breakpoint (we do for `margin`). This is an issue for responsive designs, where we do want that flexibility.

## Test Plan

1. Enter the following code in the playground
```jsx
<Box
  padding={2}
  mdPadding={0}
  color="green"
>
  <Box width={40} height={40} color="red" />
</Box>
```

2. Expected: on small screens we show a padding, on bigger screens we remove the padding

### Before
padding-zero-before

### After
padding-zero-after


packages/gestalt/src/boxWhitespace.css
```diff
@@ -41,6 +41,16 @@ html[dir="rtl"] .marginEnd0 {
...
+.paddingY0 {
+  padding-bottom: calc(var(--g-boint) * 0);
+  padding-top: calc(var(--g-boint) * 0);
+}
+
+.paddingX0 {
+  padding-left: calc(var(--g-boint) * 0);
+  padding-right: calc(var(--g-boint) * 0);
+}
+
```
ayeshakmaz, line 53: Mentioned this in Slack too, but I think we should at least add a description to the prop table about this ability (for both padding and margin), and even better would be to add an example to the documentation about responsive padding/margin

christianvuerings: @ayeshakmaz updated - thank you. Also added an example

docs/src/Box.doc.js
```diff
@@ -10,40 +10,6 @@ import Card from './components/Card.js';
 const cards: Array<Node> = [];
 const card = (c) => cards.push(c);
 
-const marginProps = [
```
christianvuerings, line 13: Inline these since they are only used once

docs/src/components/PropTable.js
```diff
@@ -171,10 +171,11 @@ export default function PropTable({
                     },
                     i
                   ) => {
+                    const hasSecondRow = description || responsive;
```
christianvuerings, line 174: Changes here fix an issue with the PropTable.

### Before

Screen Shot 2020-11-30 at 4 00 42 PM


### After
Screen Shot 2020-11-30 at 4 00 56 PM

christianvuerings:
@ayeshakmaz updated to `propNameHasSecondRow`

PR #1294, opened 2020-12-01
[PROTOTYPE] Internal: Pinned Components feature in Docs

AlbertCarreras:
New Docs feature that enables users to pin up to 3 sections on top of the navigation sidebar for quick access to the most visited ones.

[This feature is in response to some comments from this year Gestalt survey results]

CA4A5E05-F4EB-407D-A0B3-5EBD6FE079B7

B297F68B-2D68-4A1F-B2EC-3837BDF38D51

Also, 
- Centralized Nav Sidebar context
- Fixed isSidebarOpen logic
- Reduced Nav Sidebar height under md breakpoint

## Before
image

image

## After
image

image


docs/src/components/NavLink.js
```diff
@@ -19,10 +19,10 @@ const NavLink = ({ children, to, history }: Props) => {
   const href = history.createHref({
     pathname: to,
   });
-  const { isSidebarOpen, setIsSidebarOpen } = useSidebarContext();
+  const { isSidebarOpen, setIsSidebarOpen } = useNavigationSidebarContext();
 
   const handleClick = ({ event }) => {
-    setIsSidebarOpen(!isSidebarOpen);
+    setIsSidebarOpen(isSidebarOpen && !isSidebarOpen);
```
christianvuerings, line 25: @AlbertCarreras Should this just be `false`?

docs/src/components/PageHeader.js
```diff
@@ -31,17 +41,59 @@ export default function ComponentHeader({
   fileName,
   showSourceLink = true,
 }: Props): Node {
+  const { pinnedSection, setPinnedSection } = useNavigationSidebarContext();
+
+  function handlePinnedComponents() {
+    const arrayPinned = JSON.parse(pinnedSection);
+    if (!arrayPinned.includes(fileName || name) && arrayPinned.length < 3) {
+      return setPinnedSection(
```
christianvuerings, line 49: @AlbertCarreras one thing to think about is what happens if a component name changes? E.g. when we changed `GestaltProvider` to `Provider`

docs/src/components/PageHeader.js
```diff
@@ -31,17 +41,59 @@ export default function ComponentHeader({
...
+        JSON.stringify([...arrayPinned, fileName || name])
+      );
+    }
+    return setPinnedSection(
+      JSON.stringify(arrayPinned.filter((item) => item !== (fileName || name)))
+    );
+  }
+
+  const pinnedSectionArr = JSON.parse(pinnedSection);
+  const pinnedSectionMax = pinnedSectionArr.length < 3;
```
christianvuerings, line 59: @AlbertCarreras might be good to have the max number of pinned items in a const on the top of the file.

christianvuerings:
Really like this functionality! Might be good to also get a thumbs up from design before shipping. @ashleyseto @anniemochi

docs/src/components/NavLink.js
AlbertCarreras: @christianvuerings  This was a fix! We only want to close it if it was opened with the menu button. Otherwise every time you click an option it changes. Sometimes I would find the sidebar open when reducing the screen, and it was very annoying. You want to click the hen by button first, then make it disappear after selecting an option

docs/src/components/PageHeader.js
AlbertCarreras: Oh yeah that makes size

AlbertCarreras: Sense

docs/src/components/PageHeader.js
AlbertCarreras: Mmmm good point

AlbertCarreras: We should be able to reset in those cases

docs/src/components/NavLink.js
AlbertCarreras: set to 'false'

docs/src/components/PageHeader.js
```diff
@@ -31,17 +41,59 @@ export default function ComponentHeader({
...
+              }
+            >
+              <IconButton
+                accessibilityLabel="Pin this section in navigation sidebar"
+                icon="angled-pin"
+                iconColor="gray"
+                onClick={() => handlePinnedComponents()}
+                padding={1}
+                size="xs"
+                selected={pinnedSectionArr.includes(fileName || name)}
```
christianvuerings, line 88: @AlbertCarreras could we change this to either `fileName` or `name`? So we can

PR #1295, opened 2020-12-01
Internal: Accessibility improvements: Enable color contrast checks, better log output and upgrade axe-core

christianvuerings:
## Changes

* Upgrade to the latest versions of `cypress`, `cypress-axe` and `axe-core`.
* Enable color contrast checks on most components
* Improve logging when an exception occurs (easier to debug)

## Notes

* We no longer have to use the forked `@christianv/cypress-axe` version of the library.
* Cypress-axe no longer includes `axe-core. It's listed as a peer dependency so we should install it ourselves.
* Actual a11y fixes are out of scope for this PR. This PR is all about avoiding regressions moving forward.

## Test Plan

* CI a11y tests pass


cypress/integration/accessibility_Video_spec.js
```diff
@@ -8,7 +8,7 @@ describe('Video Accessibility check', () => {
     cy.configureAxe({
       rules: [
         {
-          id: 'color-contrast',
+          id: 'aria-progressbar-name',
```
christianvuerings, line 11: New check by axe-core. Not fixing it as part of this diff but let's make sure we don't introduce this issue in other components.

AlbertCarreras (approved):

PR #1297, opened 2020-12-02
Internal: Docs improvements; Contexts, header options, + minor fixes

AlbertCarreras:
- Centralized context in new AppContext
- Fixed isSidebarOpen logic
- Reduced Nav Sidebar height under md breakpoint
- Added header options under md breakpoint inside navigation sidebar + scroll to top with sidebar
- Renamed  zIndex Classes section

BEFORE/AFTER EXPLAINED: https://www.loom.com/share/b63efcb322c4404e86ffb93e4e17945e

## Before

image

## After

image


docs/src/components/appContext.js
```diff
@@ -0,0 +1,87 @@
...
+const localStorageColorSchemeKey = 'gestalt-color-scheme';
+const localStorageTextDirectionKey = 'gestalt-text-direction';
+
+type SidebarOrganisedBy = 'categorized' | 'alphabetical';
+type PropTableVariant = 'collapsed' | 'expanded';
+type ColorScheme = 'light' | 'dark' | 'userPreference';
+type DirectionScheme = 'ltr' | 'rtl';
+
+export type AppContextType = {|
+  isSidebarOpen: boolean,
```
christianvuerings, line 17: @AlbertCarreras This means we now have 1 big context object which contains all context options instead of more dedicated context.

The pros right now is that we have it all in 1 location but the downside is that this will most likely grow into an even bigger object over time, a bit like legacyContent in pinterest's codebase.

docs/src/components/Navigation.js
```diff
@@ -39,7 +39,13 @@ export default function Navigation(): Node {
   return (
     <Box role="navigation">
       {isSidebarOpen && (
-        <Box display="block" mdDisplay="none" padding={4}>
+        <Box
+          height={350}
```
christianvuerings, line 43: @AlbertCarreras thanks - definitely looks way better on smaller screens now

christianvuerings (approved):

PR #1298, opened 2020-12-02
Upsell: Add image prop

norrise120:
This PR adds an optional image prop that can be used to add in an image or icon to the left side of the banner.

Icon and Image examples 
Screen Shot 2020-12-03 at 3 32 36 PM
Screen Shot 2020-12-03 at 3 32 55 PM

On smaller screens
Screen Shot 2020-12-03 at 3 33 25 PM
Screen Shot 2020-12-03 at 3 33 32 PM

## Test Plan
Added additional tests and manually verified that components respond to different screen sizes as expected.

### DIFF
--- a/docs/src/Upsell.doc.js
+++ b/docs/src/Upsell.doc.js
@@ -133,7 +133,7 @@ card(
     defaultCode={`
 <Upsell
   title="Give $30, get $60 in ads credit"
-  message="When your friend spends their first $30 on ads, you’ll earn $60 of ads credit, and they’ll get $30 of ads credit, too"
+  message="Earn $60 of ads credit, and give $30 of ads credit to a friend"
   primaryLink={{href: "https://pinterest.com", label:"Send invite"}}
   dismissButton={{
     accessibilityLabel: 'Dismiss banner',
@@ -153,7 +153,7 @@ card(
     defaultCode={`
 <Upsell
   title="Stay healthy and safe"
-  message="Please practice social distancing, and check out our resources for adapting to these times."
+  message="Check out our resources for adapting to these times."
   primaryLink={{href: "https://pinterest.com", label:"Visit"}}
   dismissButton={{
     accessibilityLabel: 'Dismiss banner',
@@ -162,7 +162,7 @@ card(
   imageData={{
       component: 
         <Image
-          alt="Please practice social distancing, and check out our resources for adapting to these times."
+          alt="Check out our resources for adapting to these times."
           color="rgb(231, 186, 176)"
           naturalHeight={751}
           naturalWidth={564}
--- a/docs/src/Upsell.doc.js
+++ b/docs/src/Upsell.doc.js
@@ -40,14 +40,14 @@ card(
         href: '',
       },
       {
-        name: 'image',
-        type: 'React.Node',
+        name: 'imageData',
+        type: '{| component: typeof Image | typeof Icon, width?: number, |}',
         required: false,
         defaultValue: null,
         description: [
-          'React node to render on left side of banner. Max width is 128 px.',
+          'Either <Image /> or <Icon /> to render on left side of banner. Width is not used with Icon. Image width defaults to 128 px. Max width of image is 128 px.',
         ],
-        href: '',
+        href: 'Image',
       },
       {
         name: 'primaryLink',
@@ -138,7 +138,9 @@ card(
     accessibilityLabel: 'Dismiss banner',
     onDismiss: ()=>{},
   }}
-  image={<Icon icon="pinterest" accessibilityLabel="Pin" color="darkGray" size={32}/>}
+  imageData={{
+    component: <Icon icon="pinterest" accessibilityLabel="Pin" color="darkGray" size={32}/>
+  }}
 />
 `}
   />
@@ -156,15 +158,17 @@ card(
     accessibilityLabel: 'Dismiss banner',
     onDismiss: ()=>{},
   }}
-  image={
-      <Image
-        alt="Please practice social distancing, and check out our resources for adapting to these times."
-        color="rgb(231, 186, 176)"
-        naturalHeight={751}
-        naturalWidth={564}
-        src="https://i.ibb.co/7bQQYkX/stock2.jpg"
-      />
-    }
+  imageData={{
+      component: 
+        <Image
+          alt="Please practice social distancing, and check out our resources for adapting to these times."
+          color="rgb(231, 186, 176)"
+          naturalHeight={751}
+          naturalWidth={564}
+          src="https://i.ibb.co/7bQQYkX/stock2.jpg"
+        />,
+      width: 128,
+    }}
 />
 `}
   />
--- a/docs/src/Upsell.doc.js
+++ b/docs/src/Upsell.doc.js
@@ -41,7 +41,8 @@ card(
       },
       {
         name: 'imageData',
-        type: '{| component: typeof Image | typeof Icon, width?: number, |}',
+        type:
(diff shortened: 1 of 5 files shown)


docs/src/Upsell.doc.js
```diff
@@ -39,6 +39,16 @@ card(
         ],
         href: '',
       },
+      {
+        name: 'image',
```
norrise120, line 43: Not convinved this is the best prop name, as it's used for both images and icons. Open to suggestions here!

norrise120 (commented):
I'm not quite sure how to fix the layout here for the small display (having the image move to be above the title instead of to the left) and would love some guidance there!

Current implementation:
Screen Shot 2020-12-01 at 10 36 13 PM

Desired:
Screen Shot 2020-12-01 at 10 37 36 PM

docs/src/Upsell.doc.js
```diff
@@ -39,6 +39,16 @@ card(
...
+        type: 'React.Node',
+        required: false,
+        defaultValue: null,
+        description: [
+          'React node to render on left side of banner. Max width is 128 px.',
+        ],
+        href: '',
```
christianvuerings, line 50: @norrise120 we probably do want to link to the image example

packages/gestalt/src/Upsell.js
```diff
@@ -99,11 +101,23 @@ export default function Upsell({
           mdMarginBottom={primaryLink || secondaryLink ? 0 : undefined}
           mdPaddingY={3}
         >
+          {image && (
+            <Box
+              marginBottom={0}
+              marginTop={0}
+              mdMarginBottom="auto"
```
christianvuerings, line 108: Nice CSS trick!

Might be good to also think about what we should do on smaller screens
image

docs/src/Upsell.doc.js
```diff
@@ -116,4 +126,53 @@ card(
...
+<Upsell
+  title="Stay healthy and safe"
+  message="Please practice social distancing, and check out our resources for adapting to these times."
+  primaryLink={{href: "https://pinterest.com", label:"Visit"}}
+  dismissButton={{
+    accessibilityLabel: 'Dismiss banner',
+    onDismiss: ()=>{},
+  }}
+  image={
+    <Box
```
christianvuerings, line 160: @norrise120 1 way to avoid this wrapping `Box` is to use a column based approach (please confirm whether this works for design):

```jsx
<Box display="flex" direction="row">
  {image && (
    <Box maxWidth={128} column={3}>
      {image}
    </Box>
  )}
  <Box column={image ? 9 : 12}>Content</Box>
</Box>
```

docs/src/Upsell.doc.js
AlbertCarreras: Hey Erica, played around with the code. These are some thoughts and comments. https://www.loom.com/share/ce017b46c28a45c6984a563d38788d9c

AlbertCarreras:
Awesome, almost finished! 

I was wondering if we want to allow a [<Mask/>](https://gestalt.netlify.app/Mark)! Maybe we can add an extra optional key in the `imageData`, mask: { rounding: "circle" | 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8, wash: boolean}

packages/gestalt/src/Upsell.js
```diff
@@ -70,50 +77,89 @@ const UpsellLink = ({
 
 export default function Upsell({
   dismissButton,
+  imageData,
   message,
   primaryLink,
   secondaryLink,
   title,
 }: Props): Node {
+  const isImage = imageData && imageData.component.type === Image;
```
AlbertCarreras, line 86: im getting an error here.

can we do 
`const isImage = imageData?.component instanceof Image;`

packages/gestalt/src/Upsell.js
```diff
@@ -70,50 +77,89 @@ const UpsellLink = ({
...
-            marginEnd={6}
-          >
-            {title && (
-              <Box marginBottom={2}>
-                <Heading size="sm">{title}</Heading>
+          {imageData && (
+            <Box
+              marginBottom={4}
+              smMarginBottom={0}
+              width={isImage ? Math.min(imageData.width, 128) : undefined}
```
AlbertCarreras, line 120: can we do 

`width={isImage ? Math.min(imageData.width || 128, 128) : undefined}`

`Math.min` doesnt take `undefined`

packages/gestalt/src/Upsell.js
```diff
@@ -149,6 +194,10 @@ Upsell.propTypes = {
     accessibilityLabel: PropTypes.string.isRequired,
     onDismiss: PropTypes.func.isRequired,
   }),
+  imageData: PropTypes.exact({
```
AlbertCarreras, line 197:
``` $FlowFixMe[signature-verification-failure] flow 0.135.0 upgrade
  dismissButton: PropTypes.shape({
    accessibilityLabel: PropTypes.string.isRequired,
    onDismiss: PropTypes.func.isRequired,
  }),
  imageData: PropTypes.shape({
    component: PropTypes.node.isRequired,
    width: PropTypes.number,
  }),

```

packages/gestalt/src/Upsell.js
```diff
@@ -27,6 +29,10 @@ type Props = {|
     accessibilityLabel: string,
     onDismiss: () => void,
   |},
+  imageData?: {|
+    component: typeof Image | typeof Icon,
```
AlbertCarreras, line 33: can we do here:

`import React, { type Element, type Node } from 'react';
`

`    component: Element<typeof Image | typeof Icon>,
`

AlbertCarreras:
Can you update the in flow test?

```
// @flow strict
import React from 'react';
import Upsell from './Upsell.js';
import Icon from './Icon.js';
import Box from './Box.js';

const Valid = (
  <Upsell
    message="Upsell message"
    imageData={{ component: <Icon accessibilityLabel="test" /> }}
  />
);

// $FlowExpectedError[prop-missing]
const NonExistingProp = <Upsell nonexisting={33} />;

// $FlowExpectedError[prop-missing]
const MissingProp = <Upsell />;

const InvalidImage = (
  // $FlowExpectedError[incompatible-type]
  <Upsell message="Upsell message" imageData={{ component: <Box /> }} />
);

```

packages/gestalt/src/Upsell.js
```diff
@@ -70,50 +77,89 @@ const UpsellLink = ({
...
-          <Box
-            marginBottom="auto"
-            marginTop="auto"
-            maxWidth={648}
-            marginEnd={6}
-          >
-            {title && (
-              <Box marginBottom={2}>
-                <Heading size="sm">{title}</Heading>
+          {imageData && (
```
AlbertCarreras, line 116: `imageData?.component` here?

ashleyseto:
Hey looks good! I think functionally the only adjustment we wanted to make was the button order on the mobile size. We wanted to put the primary on top and the secondary on the bottom. This was something we're looking to update in the callout as well after some feedback.

Screen Shot 2020-12-04 at 3 07 50 PM

---

Not on the functionality side, I took a stab at rewriting a couple examples until we widen our docs so designers don't get confused by the break point. If you could update the text so it doesn't wrap for now, that'd be helpful for them.

Screen Shot 2020-12-04 at 3 05 26 PM

Screen Shot 2020-12-04 at 3 07 23 PM

---

Are we handling the form inputs in a separate PR?

norrise120:
> Are we handling the form inputs in a separate PR?

Yes! Plan is to get this one merged and then I'll follow up with two PRs. The first will be making updates to Callout so that it is consistent with the changes here (button responsiveness and shifting icon to be above content when on small display). The second display will handle adding forms to Upsell

ashleyseto (approved):

AlbertCarreras:
LGTM!

AlbertCarreras (approved):

PR #1300, opened 2020-12-02
Docs: Add search autodiscovery (OpenSearch)

christianvuerings:
Adds OpenSearch support to the Gestalt docs so users don't have to manually set up their browsers. Thanks to @ayeshakmaz for pointing us in this direction

https://developer.mozilla.org/en-US/docs/Web/OpenSearch
https://www.chromium.org/tab-to-search

## Test Plan

1) Go to https://gestalt.netlify.app/ once
2) Type `Gest` in the URL bar
3) You should see an autocomplete for Gestalt

image

4) Hit the `TAB` key
5) Type `Box`
6) It takes you to the `Box` docs page

Note: we'll only be able to test this once `opensearch.xml` is on the prod docs

### DIFF
--- a/docs/public/index.html
+++ b/docs/public/index.html
@@ -9,6 +9,12 @@
     <meta name="theme-color" content="#000000" />
     <link rel="manifest" href="%PUBLIC_URL%/manifest.json" />
     <link rel="shortcut icon" href="%PUBLIC_URL%/pinterest_favicon.png" />
+    <link
+      rel="search"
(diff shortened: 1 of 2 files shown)


AlbertCarreras (approved):

ayeshakmaz:
🎉 Super cool!

ayeshakmaz (approved):
🎉 Super cool!

PR #1301, opened 2020-12-02
Tag component

ohariz:
## Screenshots

Screen Shot 2020-12-03 at 2 56 38 PM

### States
Screen Shot 2020-12-03 at 2 56 54 PM

## Docs
https://paper.dropbox.com/doc/Tag-Component--BAdeKYZILgfmiSC_5cuWjOl0Ag-0lQCinJD5Ok1kXpkJqlXT
https://paper.dropbox.com/doc/Gestalt-Tag--BAbKmQvyCBritpD1GMhuN4pfAg-uceyUciJMgYCjtMRZkmwr

## Figma:
https://www.figma.com/file/Dp6tZGQigleqTA8LGCtcdj/Tags

## Test Plan

Tests and docs included


ashleyseto:
Looking good so far!

I think just some small pixel changes and the disabled state being adjusted.

- The error icon should be 12px in size.
- This is very nit picky but I think we're going to have to do some tweaking of the "New" text. It is sitting a bit low. Can we bring it up 1px so it is more visually centering within the tag?

ohariz:
> The error icon should be 12px in size.
👍 
> Can we bring it up 1px so it is more visually centering within the tag?
The line-height is 17px, so the centering is rounding this down. @christianvuerings, how do we feel about adding margins that aren't multiples of 4 to deal with this sort of thing?

packages/gestalt/src/Tag.js
```diff
@@ -0,0 +1,117 @@
...
+import Icon from './Icon.js';
+import useFocusVisible from './useFocusVisible.js';
+import focusStyles from './Focus.css';
+import touchableStyles from './Touchable.css';
+import styles from './Tag.css';
+
+type Props = {|
+  text: string,
+  errorMessage?: string,
+  ...
```
AlbertCarreras, line 17: nice here. i didn't know this trick to keep it dryier.

AlbertCarreras:
rtl breaks corners on the left

image

packages/gestalt/src/Tag.js
```diff
@@ -0,0 +1,117 @@
...
+  const {
+    disabled = false,
+    errorMessage,
+    onRemove = null,
+    removeIconAccessibilityLabel = '',
+    text,
+  } = props;
+
+  const bgColor = errorMessage ? 'red' : 'lightGray';
+  let fgColor = 'darkGray';
```
AlbertCarreras, line 44: what does fg stand for?

AlbertCarreras: fore·ground

packages/gestalt/src/Tag.js
```diff
@@ -0,0 +1,117 @@
...
+              color={fgColor}
+              icon="workflow-status-problem"
+              size={12}
+            />
+          </Box>
+        )}
+        <Box
+          marginStart={2}
+          marginEnd={disabled ? 2 : 1}
+          dangerouslySetInlineStyle={{ __style: { marginBottom: 1 } }}
```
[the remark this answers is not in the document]
AlbertCarreras: @ohariz

I think we can replace this `dangerouslySetInlineStyle` and use `<Text inline>` instead. It solves the issue with the margin-bottom.

https://www.loom.com/share/13ad5b2b9cb547e2b24116541b969a74

AlbertCarreras:
We havent' implemented max witdh 
image 

now:
image

AlbertCarreras:
Hey, found a cleaner way of fixing the margin-bottom pixel: https://www.loom.com/share/13ad5b2b9cb547e2b24116541b969a74

https://github.com/pinterest/gestalt/pull/1301/files#r535544373

ohariz:
Screen Shot 2020-12-08 at 12 27 33 PM

packages/gestalt/src/Tag.js
```diff
@@ -0,0 +1,117 @@
...
+          {errorMessage && (
+            <Icon
+              accessibilityLabel={errorMessage}
+              color={fgColor}
+              icon="workflow-status-problem"
+              size={12}
+            />
+          )}
+        </Box>
+        <div className={typographyStyles.truncate} title={text}>
```
ohariz, line 85: I had to import these directly because `truncate` doesn't work with `inline`

docs/src/Tag.doc.js
```diff
@@ -0,0 +1,123 @@
...
+        type: '({| event: SyntheticMouseEvent<> |}) => void',
+        description:
+          'Callback fired when the tag is removed. Should handle state updates to stop rendering the component. Required unless the tag is in a disabled state.',
+      },
+      {
+        name: 'removeIconAccessibilityLabel',
+        type: 'string',
+        description:
+          'Accessibility label for the icon button to remove the tag. Required unless the tag is in a disabled state.',
+      },
```
ayeshakmaz, line 44: Could we update to something like "Accessibility label for the icon button to remove the tag, ideally something like "Remove [Tag Name] Tag". Required unless the tag is in a disabled state. "

Could you also update the doc examples to be more specific, like "Remove New Tag" instead of just "remove"?

ayeshakmaz (changes requested):

AlbertCarreras (approved):

ayeshakmaz (approved):
Looks great, thanks for adding this!!

ohariz:
Thanks for reviewing!

PR #1304, opened 2020-12-05
Callout: Responsiveness updates

norrise120:
This PR updates Callout to be consistent with Upsell by: 
1. Expanding the button link when on small display
2. Centering text when on small display
3. Shifting icon from left of text to above of text on small display
4. Moving primary button to top when on small display

Before:
Screen Shot 2020-12-04 at 5 31 27 PM

After:
Screen Shot 2020-12-04 at 5 31 41 PM

## Test Plan
Updated snapshot tests, verified that component looks as expected.

### DIFF
--- a/packages/gestalt/src/Callout.js
+++ b/packages/gestalt/src/Callout.js
@@ -225,7 +225,7 @@ export default function Callout({
 
 Callout.propTypes = {
   // $FlowFixMe[signature-verification-failure] flow 0.135.0 upgrade
-  dismissButton: PropTypes.shape({
+  dismissButton: PropTypes.exact({
     accessibilityLabel: PropTypes.string.isRequired,
     onDismiss: PropTypes.func.isRequired,
   }),
--- a/packages/gestalt/src/Callout.js
+++ b/packages/gestalt/src/Callout.js
@@ -145,8 +145,8 @@ export default function Callout({
           <Box
             marginBottom={4}
             marginTop={0}
-            mdMarginBottom="auto"
-            mdMarginTop="auto"
+            smMarginBottom="auto"
+            smMarginTop="auto"
           >
             <Icon
               accessibilityLabel={iconAccessibilityLabel}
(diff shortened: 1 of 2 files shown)


ashleyseto (approved):

packages/gestalt/src/Callout.js
```diff
@@ -193,7 +225,7 @@ export default function Callout({
 
 Callout.propTypes = {
   // $FlowFixMe[signature-verification-failure] flow 0.135.0 upgrade
-  dismissButton: PropTypes.exact({
+  dismissButton: PropTypes.shape({
```
ayeshakmaz, line 228: Curious why you needed to update from exact to shape?

ayeshakmaz (commented):
Looks great, thank you for this!

packages/gestalt/src/Callout.js
norrise120: I actually did this because @AlbertCarreras asked me to make the same change on Upsell. I'm not sure the reasoning behind it though

https://github.com/pinterest/gestalt/pull/1298#discussion_r535920164

AlbertCarreras: mea culpa. Indeed, .shape is far more used than .exact so I thought for consistency. however, in many cases we should use .exact rather than .shape. 

@norrise120 , I apologize, We should leave `.exact`.

norrise120: Updated this and the other PR to be `.exact`. Thank you both!!

ayeshakmaz (approved):
Looks great! 🎉

packages/gestalt/src/Callout.js
ayeshakmaz: Awesome, thank you!

AlbertCarreras:
Hey Erica, on md we should center the icon. Now it's going to the top. 
image

PR #1305, opened 2020-12-05
Dropdown: New component to display actions or selectable options

ayeshakmaz:
New cmp: Dropdown

A dropdown displays a list of actions or options. It is triggered when a user interacts with a button, textfield or other controls.

Screen Shot 2020-12-05 at 2 05 36 AM

## Test Plan

ToDo:

- [x] Add tests
- [x] Add props table in docs
- [X] Refactor common code w/ Typeahead
- [ ] Fix keyboard nav bug on Links (to be done in separate PR)


ashleyseto:
Hey. So excited to see this! 

I wasn't able to Design QA the whole thing but I took some gifs of a few things.

----

We seem to have some jumping when hovering over the dropdown
dropdown jump

----

I wasn't able to QA this part because there seems to be an error.
dropdown disappear

----

I think we need to reserve the space for the checkmark to not cause the lines of text to move.
reserve space

ashleyseto (approved):

packages/gestalt/src/Dropdown.js
```diff
@@ -0,0 +1,247 @@
...
+  subtext?: string,
+|};
+type Props = {|
+  anchor?: ?HTMLElement,
+  children: Node,
+  headerContent?: Node,
+  idealDirection?: 'up' | 'right' | 'down' | 'left',
+  onDismiss: () => void,
+  onSelect?: ({|
+    event: SyntheticKeyboardEvent<HTMLElement> | {| keyCode: number |},
```
AlbertCarreras, line 24: should this also include `SyntheticMouseEvent<HTMLElement>`?

AlbertCarreras:
```
  onSelect: AbstractEventHandler<
    SyntheticKeyboardEvent <HTMLInputElement> | SyntheticMouseEvent<HTMLElement> | {| keyCode: number |} ,
    {| item: ?OptionObject |}
  >,

```

packages/gestalt/src/Dropdown.js
```diff
@@ -0,0 +1,247 @@
...
+   * used for keyboard navigation,
+   * we must clone the item and inject the index prop
+   */
+  const renderChildrenWithIndex = () => {
+    let numItemsRendered = 0;
+    const items = [];
+
+    const dropdownChildrenArray = React.Children.toArray(children);
+
+    dropdownChildrenArray.forEach((child, index) => {
```
dangerismycat, line 171: `reduce` could be useful here as well

packages/gestalt/src/Dropdown.jsdom.test.js
```diff
@@ -0,0 +1,61 @@
+// @flow strict
+import React from 'react';
+import renderer from 'react-test-renderer';
+import Dropdown from './Dropdown.js';
+
+const handleSelectMock = jest.fn();
+const setOpenMock = jest.fn();
+const element = document.createElement('button');
+describe('Dropdown', () => {
```
dangerismycat, line 9: We should probably also test the user interaction, keyboard nav, etc

packages/gestalt/src/DropdownItem.js
```diff
@@ -0,0 +1,93 @@
...
+import MenuOption from './MenuOption.js';
+import DropdownContext from './DropdownContextProvider.js';
+
+type OptionObject = {|
+  label: string,
+  value: string,
+  subtext?: string,
+|};
+
+type OptionProps = {|
```
dangerismycat, line 13: nit: just Props, nothing to differentiate from here

packages/gestalt/src/DropdownItem.js
```diff
@@ -0,0 +1,93 @@
...
+  isExternal?: boolean,
+  /* the option with value and label */
+  option: OptionObject,
+  selected?: OptionObject | Array<OptionObject> | null,
+  handleSelect?: ({|
+    item: OptionObject,
+    event: SyntheticFocusEvent<HTMLInputElement>,
+  |}) => void,
+  url?: string,
+  // Private prop, not listed in props table
```
dangerismycat, line 25: Since comments easily go stale or lose the lines they're commenting on, consider splitting this out to PublicProps and PrivateProps and then spreading those together to compose Props

packages/gestalt/src/DropdownItem.jsdom.test.js
```diff
@@ -0,0 +1,60 @@
...
+  test('adds badge and external icon', () => {
+    const instance = create(
+      <Dropdown.Item
+        handleSelect={handleSelectMock}
+        badgeText="New"
+        isExternal
+        option={{ value: 'item 4', label: 'Item 4' }}
+      />
+    ).root;
+    expect(instance.findByType(Icon).props.icon).toBe('arrow-up-right');
```
dangerismycat, line 47: This should also check that the New badge is present?

packages/gestalt/src/DropdownItem.jsdom.test.js
```diff
@@ -0,0 +1,60 @@
...
+import { create } from 'react-test-renderer';
+
+import { render } from '@testing-library/react';
+import Dropdown from './Dropdown.js';
+import Icon from './Icon.js';
+
+describe('DropdownItem', () => {
+  const handleSelectMock = jest.fn();
+  test('calls handleSelect when Item clicked', () => {
+    const { getByText } = render(
```
dangerismycat, line 12: The current best practice is to not use the queries from `render`, but from `screen`: https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#not-using-screen

packages/gestalt/src/DropdownItem.jsdom.test.js
```diff
@@ -0,0 +1,60 @@
+// @flow strict
+import React from 'react';
+import { create } from 'react-test-renderer';
```
dangerismycat, line 3: Is there a reason to use both testing libraries here?

packages/gestalt/src/DropdownSection.js
```diff
@@ -0,0 +1,32 @@
+// @flow strict
+import React, { type Node } from 'react';
+import PropTypes from 'prop-types';
+import Box from './Box.js';
+import Text from './Text.js';
+import styles from './Dropdown.css';
+
+type SectionProps = {|
```
dangerismycat, line 8: nit: just Props

packages/gestalt/src/MenuOption.js
```diff
@@ -0,0 +1,229 @@
...
+import focusStyles from './Focus.css';
+import useFocusVisible from './useFocusVisible.js';
+
+type OptionObject = {|
+  label: string,
+  value: string,
+  subtext?: string,
+|};
+
+type OptionProps = {|
```
dangerismycat, line 22: nit: just Props

packages/gestalt/src/MenuOption.js
```diff
@@ -0,0 +1,229 @@
...
+  option,
+  role,
+  selected,
+  setHoveredItem,
+  setOptionRef,
+  shouldTruncate = false,
+  textWeight = 'normal',
+  url,
+}: OptionProps): Node {
+  let foundAMatch = [];
```
dangerismycat, line 58: Consider something like `const matches = (Array.isArray(selected) ? selected : []).filter(({ value }) => value === option.value)`

packages/gestalt/src/MenuOption.js
```diff
@@ -0,0 +1,229 @@
...
+    focusStyles.hideOutline,
+    {
+      [focusStyles.accessibilityOutline]: isFocusVisible,
+      [styles.fullWidth]: true,
+      [styles.pointer]: true,
+    }
+  );
+
+  // Default option color
+  let optionStateColor = 'transparent';
```
dangerismycat, line 87: Consider using a ternary so this can be `const`

packages/gestalt/src/MenuOption.js
```diff
@@ -0,0 +1,229 @@
...
+
+  // Set color on item hover
+  if (index === hoveredItem) optionStateColor = 'lightGray';
+
+  const menuOptionContents = (
+    <Flex>
+      <Flex flex="grow" direction="column">
+        <Flex alignItems="center">
+          {children || (
```
dangerismycat, line 96: nit: nullish coalescing

packages/gestalt/src/MenuOption.js
```diff
@@ -0,0 +1,229 @@
...
+    item: OptionObject,
+    event: SyntheticFocusEvent<HTMLInputElement>,
+  |}) => void,
+  hoveredItem: ?number,
+  isExternal?: boolean,
+  role?: 'option' | 'menuitem',
+  setHoveredItem: (number) => void,
+  setOptionRef: (?HTMLElement) => void,
+  shouldTruncate?: boolean,
+  textWeight?: 'bold' | 'normal',
```
dangerismycat, line 38: We want to avoid duplicating type definitions, so for these types that are passed directly to other components, we should import the type definition from that component (see the shared types between Box and Flex for examples, though the types can remain defined in Text rather than a separate types file)

packages/gestalt/src/MenuOption.js
```diff
@@ -0,0 +1,229 @@
...
+                truncate={shouldTruncate}
+                weight={textWeight}
+                color="darkGray"
+                inline
+              >
+                {option?.label}
+              </Text>
+              {badgeText && (
+                <Box marginStart={2} marginTop={1}>
+                  <Box display="visuallyHidden">{`, `}</Box>
```
dangerismycat, line 108: I assume this helps with accessibility? What does this do? (It might be worth commenting in the code for documentation)

packages/gestalt/src/Typeahead.js
```diff
@@ -9,15 +9,17 @@ import React, {
...
 import TypeaheadInputField from './TypeaheadInputField.js';
-import TypeaheadOption from './TypeaheadOption.js';
+import MenuOption from './MenuOption.js';
 import Box from './Box.js';
 import Text from './Text.js';
 import Flyout from './Flyout.js';
 import Layer from './Layer.js';
+import handleContainerScrolling from './utils/keyboardNavigation.js';
 
 type OptionObject = {|
```
dangerismycat, line 19: This type should be deduped and only defined once

packages/gestalt/src/Typeahead.js
```diff
@@ -339,6 +303,8 @@ TypeaheadWithForwardRef.propTypes = {
     PropTypes.exact({
       label: PropTypes.string.isRequired,
       value: PropTypes.string.isRequired,
+      // eslint-disable-next-line react/no-unused-prop-types
```
dangerismycat, line 306: Deduping this type may obviate the need for this suppression as well

packages/gestalt/src/utils/keyboardNavigation.js
```diff
@@ -0,0 +1,41 @@
+// @flow strict
+const handleContainerScrolling = (
+  direction: -1 | 0 | 1,
```
dangerismycat, line 3: It looks like these could also use the up/down/enter common type

docs/src/Dropdown.doc.js
```diff
@@ -0,0 +1,628 @@
...
+    name="Accessibility"
+  />
+);
+
+card(
+  <Example
+    id="default"
+    name="Basic Example"
+    description={`
+      A Dropdown can be attached to a Button, IconButton, or Textfield (via [Typeahead](/Typeahead)) using the \`anchor\` prop and is composed of multiple Dropdown Items.
```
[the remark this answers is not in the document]
ayeshakmaz: 😭 noooooo

docs/src/Flyout.doc.js
```diff
@@ -253,8 +253,8 @@ card(
     description={`
     We recommend passing in the following ARIA attributes to the anchor element:
 
-    * \`aria-haspopup\` lets the screenreader know that there is a flyout linked to the tigger.
```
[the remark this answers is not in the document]
ayeshakmaz: 😂

docs/src/components/Example.js
```diff
@@ -25,6 +26,7 @@ const Example = ({
   name,
   headingSize,
   showHeading,
+  showCode = true,
```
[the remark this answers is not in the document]
ayeshakmaz: Our new doc style has this option in order to present users with an immediate example, before getting into the nitty gritty details (help designers see a visual of the component right away without having to scroll past a bunch of props tables and docs first)

packages/gestalt/src/Dropdown.js
```diff
@@ -0,0 +1,247 @@
...
+        items.push(React.cloneElement(child, { index }));
+        numItemsRendered += 1;
+      }
+    });
+
+    return items;
+  };
+
+  return (
+    <Layer>
```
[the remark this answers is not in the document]
ayeshakmaz: Ahh yes, probably

packages/gestalt/src/Dropdown.js
```diff
@@ -0,0 +1,247 @@
...
+  const handleKeyDown = (event) => {
+    const KEYS = {
+      UP: -1,
+      DOWN: 1,
+      ENTER: 0,
+    };
+    // Up Arrow
+    if (event.keyCode === 38) {
+      handleKeyNavigation(event, KEYS.UP);
+      // $FlowFixMe[prop-missing]
```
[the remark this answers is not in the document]
ayeshakmaz: Yeah, I've been having a lot of issues getting these events typed properly because of Controller's acceptance of just `{| keyCode: number |})` for its event

packages/gestalt/src/DropdownItem.jsdom.test.js
ayeshakmaz: Nope, just updated to only use one

packages/gestalt/src/MenuOption.js
ayeshakmaz: It prevents a run on sentence b/w the item label and the badge text (will add a comment)

packages/gestalt/src/Typeahead.js
ayeshakmaz: Sadly didn't seem to help

docs/src/Dropdown.doc.js
```diff
@@ -0,0 +1,628 @@
...
+        type: 'string',
+        description:
+          "When supplied, will display a Badge next to the item's label.",
+        href: 'default',
+      },
+      {
+        name: 'children',
+        type: 'React.Node',
+        description:
+          'If needed, users can supply custom content to each Dropdown Item. This can be useful when extra functionality is needed beyond a basic Link, such as supporting React Router.',
```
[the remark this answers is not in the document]
ayeshakmaz: We'll be supporting account rep out of the box with V2, so we'd rather people not create their own right now

packages/gestalt/src/MenuOption.js
dangerismycat: Oh interesting! That seems like a good trick to document in an a11y best practices doc somewhere

docs/src/Typeahead.doc.js
```diff
@@ -192,6 +192,52 @@ function Example(props) {
...
+
+  const handleOnChange = ({ value }) => {
+    setItem(value);
+  };
+
+  const handleSelect = ({item}) => {
+    setSelected(item);
+  };
+
+  const label = "Selected Item: " + (selected && selected.label || '');
```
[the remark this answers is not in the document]
AlbertCarreras: I think optional chaining doesn't work in the Docs.

packages/gestalt/src/Controller.js
```diff
@@ -127,10 +131,11 @@ export default class Controller extends Component<Props, State> {
   }
 
   handleKeyDown: (event: {| keyCode: number |}) => void = (event) => {
-    const { onDismiss } = this.props;
+    const { handleKeyDown, onDismiss } = this.props;
     if (event.keyCode === ESCAPE) {
       onDismiss();
     }
+    if (handleKeyDown) handleKeyDown(event);
```
AlbertCarreras, line 138: I dont' think we support this format in Pinboard, any thought?

image

packages/gestalt/src/Dropdown.flowtest.js
```diff
@@ -0,0 +1,12 @@
+// @flow strict
+import React from 'react';
+import Dropdown from './Dropdown.js';
+
```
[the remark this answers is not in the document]
AlbertCarreras: Yeah, we usually don't test much with these tests. Agree we could add more test cases

packages/gestalt/src/Controller.js
dangerismycat: We don't, and I don't particularly like it, but I didn't comment on it as that's not anything specific to this PR. That's something we should change via the lint rule and autofix / codemod.

docs/src/Dropdown.doc.js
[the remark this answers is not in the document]
dangerismycat: Ah ok, I was thinking of what we'd need to replace the existing DropdownMenu usages in Pinboard but it sounds like the main navbar will need to wait

docs/src/Typeahead.doc.js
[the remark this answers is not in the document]
dangerismycat: Really? That's…odd. Does it run through a different build process? Perhaps we need to update that, if so

docs/src/components/Example.js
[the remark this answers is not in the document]
dangerismycat: I would think we'd still want to allow developers to see how that example was created though. We could establish a pattern of always repeating the main example as the first one below the props table, and perhaps enforce that…somehow 😅 

Then that main example could be as you mention, but we'd never completely hide things either.

packages/gestalt/src/Dropdown.flowtest.js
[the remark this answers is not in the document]
dangerismycat: Ah, we should up that bar then. If the pitch is that the components are fully tested and rock-solid, then we need to make sure that's true. More, better tests in Gestalt so we can be safer with fewer tests in Pinboard.

packages/gestalt/src/Dropdown.js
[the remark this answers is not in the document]
dangerismycat: Hm…we should probably discuss more with Christian and Jenny in January. I think we need a better story for our event typing.

packages/gestalt/src/Dropdown.js
```diff
@@ -0,0 +1,247 @@
...
+  children,
+  headerContent,
+  idealDirection = 'down',
+  onDismiss,
+  onSelect,
+}: Props): Node {
+  const getFlattenedChildren = () => {
+    const dropdownChildrenArray = React.Children.toArray(children);
+
+    const items = dropdownChildrenArray.map((child) => {
```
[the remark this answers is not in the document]
AlbertCarreras: good one.

packages/gestalt/src/Dropdown.js
```diff
@@ -0,0 +1,247 @@
...
+      if (child.type.displayName === 'DropdownItem') {
+        return child;
+      }
+      return null;
+    });
+
+    return items.flat();
+  };
+
+  const availableOptions = getFlattenedChildren();
```
AlbertCarreras, line 56: `const availableOptions=Children.toArray(children).reduceChildren.toArray(children).reduce...`

I wonder if it's worth adding a tiny comment explaining what this is doing... maybe not.

packages/gestalt/src/DropdownContextProvider.js
```diff
@@ -0,0 +1,20 @@
+// @flow strict
+import { createContext, type Context } from 'react';
```
AlbertCarreras, line 2: could we use `docs/src/components/createHydra.js` pattern here?

packages/gestalt/src/Dropdown.js
dangerismycat: The comment seems unnecessary as long as the code is still organized as the getFlattenedChildren function. Variable naming ftw! 🙂

packages/gestalt/src/DropdownContextProvider.js
dangerismycat: Potentially. What would that gain us here?

packages/gestalt/src/MenuOption.js
ayeshakmaz: This is actually done by eslint, maybe something we need to discuss as a team?

dangerismycat: Ah, sounds good. I'm not sure if a lint rule exists for this yet (a quick search came up empty), but it's a good candidate for that I think.

docs/src/Dropdown.doc.js
```diff
@@ -0,0 +1,628 @@
...
+        type:
+          'Object: {label: string, value: string, subtext: string} || Array<{label: string, value: string, subtext: string}>',
+        description:
+          'Either the selected item info or an array of selected items, used to determine when the "selected" icon appears on an item',
+        href: 'default',
+      },
+      {
+        name: 'handleSelect',
+        type:
+          '({ event: SyntheticInputEvent<>, item: {label: string, value: string, subtext: string} }) => void',
```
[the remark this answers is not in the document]
dangerismycat: Can this type be closed?

docs/src/Dropdown.doc.js
```diff
@@ -0,0 +1,642 @@
...
+
+card(
+  <PropTable
+    Component={Dropdown.Section}
+    name="Dropdown.Section"
+    id="Dropdown.Section"
+    props={[
+      {
+        name: 'children',
+        type: 'React.Node',
```
dangerismycat, line 215: This should probably also be a ChildrenArray of DropdownItems, right?

docs/src/Dropdown.doc.js
```diff
@@ -0,0 +1,642 @@
...
+        const [selected, setSelected] = React.useState(null);
+        const anchorRef = React.useRef(null);
+        const handleSelect = ({item}) => {
+          setSelected(item);
+        };
+
+        return (
+          <Box display="flex" justifyContent="center">
+            <Button
+              accessibilityHaspopup
```
dangerismycat, line 275: For future versions: is there a way to better encapsulate this behavior? Perhaps the dev has to use a "Dropdown.Button" component that automatically adds these props?

packages/gestalt/src/Dropdown.flowtest.js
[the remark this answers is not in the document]
dangerismycat: I still think we should expand these tests, but that can happen in future work

packages/gestalt/src/TypeaheadInputField.js
```diff
@@ -113,15 +113,15 @@ const TypeaheadInputFieldWithForwardRef: React$AbstractComponent<
       ENTER: 0,
     };
     // Up Arrow
```
dangerismycat, line 115: nit: you can remove all these comments since the imported keyCodes communicate effectively

packages/gestalt/src/Dropdown.js
```diff
@@ -0,0 +1,253 @@
...
+          });
+      }
+    }
+
+    // Scrolling
+    handleContainerScrolling(direction, containerRef, selectedElement);
+  };
+
+  const handleKeyDown = (event) => {
+    // Up Arrow
```
dangerismycat, line 119: nit: these comments can be removed

dangerismycat (approved):
Nice!! A handful more nits but I think this is pretty much good to go 🎉

packages/gestalt/src/TypeaheadInputField.js
ayeshakmaz: good call

docs/src/Dropdown.doc.js
```diff
@@ -0,0 +1,642 @@
...
+        required: true,
+        type:
+          'ChildrenArray<Element<typeof DropdownItem | typeof DropdownSection>>',
+        description:
+          'Should consist of Dropdown.Item and/or Dropdown.Section components',
+        href: 'default',
+      },
+      {
+        name: 'headerContent',
+        type: 'React.Node',
```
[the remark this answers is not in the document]
ayeshakmaz: I think we're actually ok with allowing anything here, since there are already varying uses of this header in the product that we want to account for.

docs/src/Dropdown.doc.js
ayeshakmaz: I was thinking about this too - maybe a wrapper component, like you said Dropdown.Anchor or Dropdown.Trigger, that injects those props automatically? I'll add it to the ticket for the next round of updates

packages/gestalt/src/Dropdown.js
```diff
@@ -0,0 +1,239 @@
...
+        type: { displayName },
+      } = currentChild;
+      if (currentItemChildren && displayName === 'DropdownSection') {
+        return [...accumulatedChildren, ...currentItemChildren];
+      }
+      if (displayName === 'DropdownItem') {
+        return [...accumulatedChildren, currentChild];
+      }
+      // eslint-disable-next-line no-console
+      console.warn(
```
AlbertCarreras, line 64: I wonder if we could throw an Error instead. warning vs error?

ayeshakmaz: Once we get a chance to chat with the others about typing this cmp (the children prop is not yet properly typed b/c of errors that come up), this will go away anyway and we'll rely on that

packages/gestalt/src/DropdownContextProvider.js
AlbertCarreras: Standardization across Gestalt?

packages/gestalt/src/DropdownItem.js
```diff
@@ -0,0 +1,92 @@
...
+}: Props): Node {
+  return (
+    <DropdownContext.Consumer>
+      {({ hoveredItem, setHoveredItem, setOptionRef }) => (
+        <MenuOption
+          key={`${option.value + index}`}
+          badgeText={badgeText}
+          handleSelect={handleSelect}
+          hoveredItem={hoveredItem}
+          index={index}
```
AlbertCarreras, line 47: should we create a specific id like `dropdown-${index}`? Could other potential single index ids confuse the aria-activedescendant?

AlbertCarreras:
The indentation in the examples is pretty big:

image
vs 
image

docs/src/Dropdown.doc.js
```diff
@@ -0,0 +1,662 @@
...
+                anchor={anchorRef.current}
+                onDismiss={() => {setOpen(false)}}
+              >
+                <Dropdown.Item
+                  isExternal
+                  option={{ value: 'item 1', label: 'Custom link 1' }}
+                  handleSelect={handleSelect}
+                  selected={selected}
+                >
+                  <Box width={"100%"}>
```
AlbertCarreras, line 609: `<Box width={"100%"}>
`
>> 
`<Box width="100%">
`

docs/src/Dropdown.doc.js
```diff
@@ -0,0 +1,662 @@
...
+              ref={anchorRef}
+            />
+            {open && (
+              <Dropdown
+                anchor={anchorRef.current}
+                onDismiss={() => {setOpen(false)}}
+              >
+                <Dropdown.Item
+                  isExternal
+                  option={{ value: 'item 1', label: 'Custom link 1' }}
```
AlbertCarreras, line 605: Can we replace single with double quotes cross docs?

packages/gestalt/src/DropdownContextProvider.js
dangerismycat: Makes sense. I don't think this is specific to this PR (and can be easily refactored in the future), so let's discuss in our next meeting?

packages/gestalt/src/DropdownItem.js
```diff
@@ -0,0 +1,92 @@
...
+  isExternal,
+  option,
+  selected,
+  href,
+}: Props): Node {
+  return (
+    <DropdownContext.Consumer>
+      {({ hoveredItem, setHoveredItem, setOptionRef }) => (
+        <MenuOption
+          key={`${option.value + index}`}
```
dangerismycat, line 43: Ah, just noticed this: we should avoid using the `+` for string concatenation, as it can lead to subtle and unintuitive bugs. There's an eslint rule for this; I'll add it to our meeting topics this week.

AlbertCarreras (approved):

packages/gestalt/src/DropdownItem.js
ayeshakmaz: Updated menu option to use specific ids, and got rid of aria-activedescendant since it's actually not doing anything (we're focusing each item instead)

PR #1306, opened 2020-12-07
TextField: add tags prop

ohariz:
Adding the `tags` prop to `Textfield`.
[Tag component](https://github.com/pinterest/gestalt/pull/1301)
[Doc](https://paper.dropbox.com/doc/Gestalt-Tag--BAbKmQvyCBritpD1GMhuN4pfAg-uceyUciJMgYCjtMRZkmwr)
[Figma](https://www.figma.com/file/Dp6tZGQigleqTA8LGCtcdj/Tags)

Looking for feedback on the approach before polishing this PR and making the same changes to `Textarea` and `Typeahead`

## Test Plan

Unit test, manual testing for no tags, single tag, multiple tags and long tags to span multiple lines
Screen Shot 2020-12-09 at 4 36 31 PM
Screen Shot 2020-12-09 at 4 36 42 PM
Screen Shot 2020-12-09 at 4 36 56 PM
Screen Shot 2020-12-09 at 4 37 13 PM


packages/gestalt/src/TextField.js
```diff
@@ -40,6 +42,7 @@ type Props = {|
     value: string,
   |}) => void,
   placeholder?: string,
+  tags?: Array<Node>,
```
AlbertCarreras, line 45: could this be `Array<Element<typeof Tag>>`

ohariz: oh nice, I couldn't figure out how to do this

ayeshakmaz:
Do tags ever need to be more than one word? It seems in this implementation they can only by one continuous word, but some of the designs suggest stuff like "soup recipes" will be an option

ohariz:
@ayeshakmaz there's nothing intrinsically limiting tags to a single word. The code that's splitting tags on spaces is only in the example. I figured it made more sense to handle the splitting in the `onChange` function to give a bit more freedom over how to split tags, and let the user of the component manage it themselves.

ayeshakmaz:
@ohariz Makes sense! I think it'd be great to add all that info to the example then in the description, pointing out that users can implement as needed, and that on space is just an example : )

docs/src/TextField.doc.js
```diff
@@ -199,6 +205,53 @@ function Example(props) {
...
+        }}
+        tags={tags.map((tag, idx) => (
+          <Tag
+            key={tag}
+            onRemove={() => {
+              const newTags = [...tags];
+              newTags.splice(idx, 1);
+              setTags([...newTags]);
+            }}
+            removeIconAccessibilityLabel="Remove"
```
AlbertCarreras, line 242: `            removeIconAccessibilityLabel={`Remove ${tag} tag`}
`

docs/src/TextField.doc.js
```diff
@@ -199,6 +205,53 @@ function Example(props) {
...
+card(
+  <Example
+    id="tagsExample"
+    name="Example: Tags"
+    description={`You can include \`Tag\` elements in the input using the \`tags\` prop.`}
+    defaultCode={`
+function Example(props) {
+  const [value, setValue] = React.useState('');
+  const [tags, setTags] = React.useState(['a@pinterest.com', 'b@pinterest.com']);
+  return (
```
AlbertCarreras, line 217: can we add a line here?

docs/src/TextField.doc.js
```diff
@@ -199,6 +205,53 @@ function Example(props) {
...
+    <Box padding={2} color="white">
+      <TextField
+        id="emails"
+        label="Emails"
+        onChange={({value}) => {
+          const chunks = value.split(/[ ,;]+/);
```
AlbertCarreras, line 223: can we rename to tagUnit or tagInput or tagValue or tagText?

chunk reminds me of https://i5.walmartimages.com/asr/d80579d1-3a30-49f6-8fa4-d71b5bde72a0.6881a73c35b1fe1f768d972510cbd6a8.jpeg and makes me get hungry

docs/src/TextField.doc.js
```diff
@@ -199,6 +205,53 @@ function Example(props) {
...
+          if (chunks.length > 1) {
+            setTags([...tags, ...chunks.slice(0, -1)]);
+          }
+          setValue(chunks[chunks.length - 1]);
+        }}
+        onKeyDown={({event: {keyCode, target: {selectionEnd}}}) => {
```
AlbertCarreras, line 229: can we add spaces here? 

`        onKeyDown={({ event: { keyCode, target: { selectionEnd }}}) => {
`

docs/src/TextField.doc.js
```diff
@@ -199,6 +205,53 @@ function Example(props) {
...
+          if (keyCode === 8 /* Backspace */ && selectionEnd === 0) {
```
AlbertCarreras, line 230: I don't understand this, what does it do?

AlbertCarreras:
Autocomplete keeps poping up with 'off'

image

docs/src/TextField.doc.js
```diff
@@ -199,6 +205,53 @@ function Example(props) {
...
+    <Box padding={2} color="white">
+      <TextField
+        id="emails"
+        label="Emails"
+        onChange={({value}) => {
+          const chunks = value.split(/[ ,;]+/);
+          if (chunks.length > 1) {
+            setTags([...tags, ...chunks.slice(0, -1)]);
+          }
+          setValue(chunks[chunks.length - 1]);
```
AlbertCarreras, line 227: to simplify, this could just be 

`          setValue(chunks[1]);
`

docs/src/TextField.doc.js
```diff
@@ -199,6 +205,53 @@ function Example(props) {
...
+  const [tags, setTags] = React.useState(['a@pinterest.com', 'b@pinterest.com']);
+  return (
+    <Box padding={2} color="white">
+      <TextField
+        id="emails"
+        label="Emails"
+        onChange={({value}) => {
+          const chunks = value.split(/[ ,;]+/);
+          if (chunks.length > 1) {
+            setTags([...tags, ...chunks.slice(0, -1)]);
```
AlbertCarreras, line 225: to simplify a little 

```
            setTags([...tags, ...chunks.splice(0, chunks.length-1)]);
          }
          setValue(chunks);

```

packages/gestalt/src/TextField.js
```diff
@@ -97,35 +102,56 @@ const TextFieldWithForwardRef: React$AbstractComponent<
     styles.textField,
     formElement.base,
     disabled ? formElement.disabled : formElement.enabled,
-    hasError || errorMessage ? formElement.errored : formElement.normal,
-    size === 'md' ? layout.medium : layout.large
+    (hasError || errorMessage) && !focused
```
AlbertCarreras, line 105: what made you add `&& !focused`

packages/gestalt/src/TextField.js
```diff
@@ -97,35 +102,56 @@ const TextFieldWithForwardRef: React$AbstractComponent<
...
+      ? formElement.errored
+      : formElement.normal,
+    size === 'md' ? layout.medium : layout.large,
+    {
```
AlbertCarreras, line 109: could we gate this? 

```
   tags
      ? {
          [focusStyles.accessibilityOutlineFocus]: focused,
          [styles.textFieldWrapper]: true,
        }
      : {}

```

AlbertCarreras:
LGTM, @ayeshakmaz can you check a11y? I tried to keyboard navigate and only was able to find the remove  buttons, but I probably was doing something wrong.

AlbertCarreras:
Disabled on disabled looks incorrect compared to Figma:

image

image

docs/src/TextField.doc.js
ohariz: This handles backspaces. If the cursor is in the beginning of the field and you press backspace, it removes the last tag. Should I add a comment?

ohariz:
Updated:
Screen Shot 2020-12-10 at 10 22 51 AM

ohariz:
I'm trying to figure out what's happening with `autoComplete` now

docs/src/TextField.doc.js
```diff
@@ -199,6 +205,57 @@ function Example(props) {
...
+    defaultCode={`
+function Example(props) {
+  const [value, setValue] = React.useState('');
+  const [tags, setTags] = React.useState(['a@pinterest.com', 'b@pinterest.com']);
+
+  return (
+    <Box padding={2} color="white">
+      <TextField
+        autoComplete="off"
+        id="tags"
```
ohariz, line 224: I had to change the `id` value to get `autoComplete="off"` to work. I'm not sure what's happening here tbh, but I don't think it's related to my changes.

ohariz: Looks like making this set of changes to a clean branch will break autocomplete:
```
--- a/docs/src/TextField.doc.js
+++ b/docs/src/TextField.doc.js
@@ -116,6 +116,7 @@ function Example(props) {
   const [value, setValue] = React.useState('')
   return (
     <TextField
+      autoComplete="off"
       id="email"
       onChange={({value}) => setValue(value)}
       placeholder="Add email"
@@ -129,6 +130,32 @@ function Example(props) {
   />
 );

+card(
+  <Example
+    id="basicExample2"
+    name="Example2"
+    description={`
+    A \`TextField\` will expand to fill the width of the parent container.
+  `}
+    defaultCode={`
+function Example(props) {
+  return (
+    <TextField
+      autoComplete="off"
+      id="email2"
+      onChange={({value}) => setValue(value)}
+      placeholder="Add email"
+      label="Email"
+      value={value}
+      type="email"
+    />
+  );
+}
+`}
+  />
+);
+
 card(
   <Example
     id="disabledExample"
```
It looks like it's a Chrome issue, and I'm seeing "solutions" online where people use `autocomplete="none"` or some other invalid value to get around it ¯\\_(ツ)_/¯

docs/src/TextField.doc.js
AlbertCarreras: no, i should probably read this twice haha it was late tonight.

docs/src/TextField.doc.js
AlbertCarreras: gotcha. if it's not related to your change, ignore. we can address in follow up

docs/src/TextField.doc.js
```diff
@@ -199,6 +205,57 @@ function Example(props) {
   />
 );
 
+card(
+  <Example
+    id="tagsExample"
+    name="Example: Tags"
+    description={`
+    You can include \`Tag\` elements in the input using the \`tags\` prop.
+    Note that the \`TextField\` component does not internally manage tags. That should be handled in the application state through the component's event callbacks. This example creates new tags by splitting the input on spaces, commas, and semicolons, and removes them on backspaces.`}
```
ayeshakmaz, line 214:
```suggestion
    You can include [Tag](#/Tag) elements in the input using the \`tags\` prop.
    
    Note that the \`TextField\` component does not internally manage tags. That should be handled in the application state through the component's event callbacks. This example creates new tags by splitting the input on spaces, commas, and semicolons, and removes them on backspaces.`}
```

ayeshakmaz: Just a small tweak to link to the Tag page

ayeshakmaz (approved):
Looks great! One docs change, but otherwise good to go. Thanks for adding this!

PR #1308, opened 2020-12-09
Callout/Upsell: Implemented new useResponsiveMinWidth hook and utils

AlbertCarreras:
Implemented `useResponsiveMinWidth` which uses `window.matchMedia` to return the size of the screen for each breakpoint size. This new hook implementation allows implementing responsiveness to any component props.

Having duplicated components in the `render()`  on tests is confusing for developers. Test could require changes as a result of having to duplicate one same component using `display='none'` & `smDisplay='none'` to  be able to achieve some responsiveness behaviors.

For example, we had this in `Callout` and `Upsell`

```
{/* We repeat this code block to ensure that text is 
              centered for our smaller displays and left aligned 
              for larger displays */}
              <Box smDisplay="none">
                {title && (
                  <Box marginBottom={2}>
                    <Heading align="center" size="sm">
                      {title}
                    </Heading>
                  </Box>
                )}
                <Text align="center">{message}</Text>
              </Box>
              <Box smDisplay="block" display="none">
                {title && (
                  <Box marginBottom={2}>
                    <Heading size="sm">{title}</Heading>
                  </Box>
                )}
                <Text>{message}</Text>
              </Box>
            </Box>

```
Tests using `render()` could show the duplication and that solution would force engineers to know the under the hood behavior of these two components.  

With `useResponsiveMinWidth`, we use the `matchMedia` approach which was already being used for `useColorScheme` and `useReducedMotion`

Built-in  `Box` responsive props such as display, margin, etc are still valid in most cases, but to facilitate making components more responsive in the future those won’t be sufficient for many cases as we see with `Callout` / `Upsell`.

The original  `Box` responsive props aren’t super intuitive at the beginning. The way this hook is implemented is slightly more intuitive. In this case, the hook is returning the actual screen breakpoint range. Therefore, to check for size matches we have to use logic like:
`responsiveMinWidth === 'xs'` ,
`responsiveMinWidth !== 'xs'`  or
`['md', 'lg].includes(responsiveMinWidth)`

This allows responsiveness to any other component props.

### DIFF
--- a/packages/gestalt/src/Callout.js
+++ b/packages/gestalt/src/Callout.js
@@ -11,6 +11,7 @@ import Text from './Text.js';
 import { useColorScheme } from './contexts/ColorScheme.js';
 import { type AbstractEventHandler } from './AbstractEventHandler.js';
 import styles from './Callout.css';
+import useResponsiveMinWidth from './useResponsiveMinWidth.js';
 
 type LinkData = {|
   accessibilityLabel?: string,
@@ -109,6 +110,7 @@ export default function Callout({
   // that all text is readable.
   const { name } = useColorScheme();
   const isDarkMode = name === 'darkMode';
+  const responsiveMinWidth = useResponsiveMinWidth();
 
   return (
     <Box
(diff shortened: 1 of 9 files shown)


packages/gestalt/src/useResponsiveMinWidth.js
```diff
@@ -0,0 +1,66 @@
...
+import {
+  addListener,
+  removeListener,
+  supportsMatchMedia,
+} from './utils/matchMedia.js';
+
+type MinWidthType = 'lg' | 'md' | 'sm' | 'xs';
+
+export default function useResponsiveMinWidth(): ?MinWidthType {
+  const breakpoints = useMemo(
```
jennyscript, line 12: Seems like there's no need to `useMemo` these instead of defining outside the hook. That would also let you not `useCallback` getMinWidth as well

AlbertCarreras: DONE

packages/gestalt/src/Callout.js
```diff
@@ -164,27 +166,19 @@ export default function Callout({
...
+                <Box marginBottom={2}>
+                  <Heading
+                    align={responsiveMinWidth === 'xs' ? 'center' : undefined}
+                    size="sm"
+                  >
+                    {title}
+                  </Heading>
+                </Box>
+              )}
+              <Text align={responsiveMinWidth === 'xs' ? 'center' : undefined}>
```
norrise120, line 179: Should we also be doing something like this below for the secondary button where we also use display="none"?

```
          {secondaryLink && responsiveMinWidth !== 'xs' && (
             <CalloutLink type="secondary" data={secondaryLink} />
          )}
          {primaryLink && <CalloutLink type="primary" data={primaryLink} />}
          {secondaryLink && responsiveMinWidth === 'xs' && (
              <CalloutLink
                type="secondary"
                data={secondaryLink}
                stacked={!!secondaryLink}
              />
          )}
```

AlbertCarreras: Good point. DONE

packages/gestalt/src/useResponsiveMinWidth.js
```diff
@@ -0,0 +1,69 @@
...
+
+type MinWidthType = 'lg' | 'md' | 'sm' | 'xs';
+
+const breakpoints = {
+  sm: '(min-width: 576px)',
+  md: '(min-width: 768px)',
+  lg: '(min-width: 1312px)',
+};
+
+export default function useResponsiveMinWidth(): ?MinWidthType {
```
ayeshakmaz, line 13: would "getResponsiveMinWidth" make more sense than "useResponsive..."?

packages/gestalt/src/useResponsiveMinWidth.jsdom.test.js
```diff
@@ -0,0 +1,89 @@
...
+import useResponsiveMinWidth from './useResponsiveMinWidth.js';
+
+const mediaqueryDefaults = {
+  matches: false,
+  onchange: null,
+  addEventListener: jest.fn(),
+  removeEventListener: jest.fn(),
+};
+
+describe('useReducedMotion', () => {
```
ayeshakmaz, line 12: I think the name here needs to be updated

ayeshakmaz (changes requested):

packages/gestalt/src/useResponsiveMinWidth.js
AlbertCarreras: good point... it's intended to simulate the hook naming pattern. 

useColorScheme
useReducedMotion
useFocusVisible
useDebouncedCallback
etc

I'm not exposing it yet, so we can discuss this on the next meeting.
I think for now works well. Maybe we can discuss further the naming all the different hooks, utils we are building.

packages/gestalt/src/useResponsiveMinWidth.jsdom.test.js
AlbertCarreras: good catch

ayeshakmaz (approved):
Looks good to me 👍

PR #1309, opened 2020-12-09
Modal: Update backdrop wash to 80% opacity

ayeshakmaz:
Update Modal's Backdrop wash to 80% opacity black

- added new var   --g-colorTransparentGray800: rgba(0, 0, 0, 0.8);
- Also changed the Mask wash to be 4% opacity to be consistent with mobile

Before:
Screen Shot 2020-12-09 at 2 24 25 PM

After:

Screen Shot 2020-12-09 at 2 24 34 PM

## Test Plan
Manually verified visually

### DIFF
--- a/packages/gestalt/src/Backdrop.css
+++ b/packages/gestalt/src/Backdrop.css
@@ -2,7 +2,7 @@
   0% {
     opacity: 0;
   }
-  
+
   100% {
     opacity: 0.9;
   }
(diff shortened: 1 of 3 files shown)


ajmoo:
@ayeshakmaz looks like the background is set to the right color, but the whole thing is set to 90% opacity.

ajmoo (approved):
this looks good to me

AlbertCarreras (approved):

PR #1311, opened 2020-12-15
Textfield: update tags wrapping behavior

ohariz:
Updating the wrapping behavior with tags to match designs

Details: adding an invisible "measuring" div to implement flex wrapping behavior which is not supported by inputs, and having the actual input track it with absolute positioning

## Test Plan

tf-tags

(the choppiness is just from the recording being in gif format)


packages/gestalt/src/TextField.css
```diff
@@ -9,17 +9,26 @@
...
+.unstyledTextField {
+  composes: absolute left0 top0 right0 bottom0 from "./Layout.css";
+  width: 100%;
+}
+
 .unstyledTextField:focus {
   outline: 0;
 }
 
+.measure {
```
ayeshakmaz, line 27: Only thing I'm not totally loving is this name "measure", as I don't think it does a great job explaining the intent behind this invisible div. Maybe `textFieldSpacer` or `textFieldWrapper` or something similar will help future devs better understand the purpose

packages/gestalt/src/TextField.js
```diff
@@ -149,7 +149,12 @@ const TextFieldWithForwardRef: React$AbstractComponent<
               {tag}
             </Box>
           ))}
-          {inputElement}
+          <Box flex="grow" marginEnd={2} maxWidth="100%" position="relative">
+            <div aria-hidden className={styles.measure}>
+              {value}
+            </div>
```
ayeshakmaz, line 155: Also probably worth adding a comment here to explain this div's purpose for future users (the details in this PR description are nice and clear)

ayeshakmaz (changes requested):

packages/gestalt/src/TextField.css
```diff
@@ -9,17 +9,26 @@
   color: var(--g-colorGray200);
 }
 
-.unstyledTextField {
+.unstyledTextField, .measure {
```
ayeshakmaz, line 12: Ah, just need to update the class name here too

packages/gestalt/src/TextField.js
```diff
@@ -149,7 +149,16 @@ const TextFieldWithForwardRef: React$AbstractComponent<
               {tag}
             </Box>
           ))}
-          {inputElement}
+          <Box flex="grow" marginEnd={2} maxWidth="100%" position="relative">
+            {/* This is an invisible spacer div which mirrors the input's
+             * content. We use it to implement the flex wrapping behavior
+             * which is not supported by inputs, by having the actual input
+             * track it with absolute positioning. */}
+            <div aria-hidden className={styles.textFieldSpacer}>
```
ayeshakmaz, line 157: 🎉

ayeshakmaz (changes requested):

packages/gestalt/src/TextField.css
ohariz: good catch

ayeshakmaz (approved):
Looks great, thank you!

PR #1312, opened 2020-12-15
Textfield: update tags props docs

ohariz:
Updating the docs for the tags example
- Link to Tag component
- Enter/return and whitespace characters now create a new tag in the example
- We now filter out empty tags

## Test Plan
Played around with the example, made sure the above works

### DIFF
--- a/docs/src/TextField.doc.js
+++ b/docs/src/TextField.doc.js
@@ -210,7 +210,8 @@ card(
     id="tagsExample"
     name="Example: Tags"
     description={`
-    You can include \`Tag\` elements in the input using the \`tags\` prop.
+    You can include [Tag](/Tag) elements in the input using the \`tags\` prop.
+
     Note that the \`TextField\` component does not internally manage tags. That should be handled in the application state through the component's event callbacks. This example creates new tags by splitting the input on spaces, commas, and semicolons, and removes them on backspaces.`}
     defaultCode={`
 function Example(props) {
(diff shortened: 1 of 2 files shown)


ashleyseto (approved):

docs/src/TextField.doc.js
```diff
@@ -210,7 +210,8 @@ card(
     id="tagsExample"
     name="Example: Tags"
     description={`
-    You can include \`Tag\` elements in the input using the \`tags\` prop.
+    You can include [Tag](/Tag) elements in the input using the \`tags\` prop.
+
     Note that the \`TextField\` component does not internally manage tags. That should be handled in the application state through the component's event callbacks. This example creates new tags by splitting the input on spaces, commas, and semicolons, and removes them on backspaces.`}
     defaultCode={`
```
ayeshakmaz, line 216: I think these docs need to update given the new functionality you added

ayeshakmaz: And maybe add a note/recommendation to remove empty tags

ayeshakmaz (changes requested):
Would also be great, in the Tag docs, to link to this Tags example "`The tag can appear within a [form field](/TextField#tagsExample) or outside of it.`"

docs/src/TextField.doc.js
```diff
@@ -224,15 +225,21 @@ function Example(props) {
         id="tags"
         label="Emails"
         onChange={({ value }) => {
-          const tagInput = value.split(/[ ,;]+/);
+          const tagInput = value.split(/[\\s,;]+/);
```
AlbertCarreras, line 228: not super opinionated on this but the example could read easier if we move out callbacks


image

docs/src/TextField.doc.js
```diff
@@ -210,43 +210,63 @@ card(
     id="tagsExample"
     name="Example: Tags"
     description={`
-    You can include \`Tag\` elements in the input using the \`tags\` prop.
-    Note that the \`TextField\` component does not internally manage tags. That should be handled in the application state through the component's event callbacks. This example creates new tags by splitting the input on spaces, commas, and semicolons, and removes them on backspaces.`}
+    You can include [Tag](/Tag) elements in the input using the \`tags\` prop.
+
+    Note that the \`TextField\` component does not internally manage tags. That should be handled in the application state through the component's event callbacks. We recommend creating new tags on enter key presses, and removing them on backspaces when the cursor is in the beginning of the field. We also recommend filtering out empty tags.
+
+    This example showcases the recommended behavior. In addition, it creates new tags by splitting the input on spaces, commas, semicolons.`}
```
ohariz, line 217: @ayeshakmaz feel free to recommend more changes, I'm not the best at writing these things

ayeshakmaz: Sounds great!

ayeshakmaz (approved):
Looks great!

PR #1314, opened 2020-12-17
Callout/Upsell: update primary/secondaryLink to be primary/secondaryAction + Codemod

norrise120:
The purpose of this PR is to update Callout so that primary/secondaryLink can be used as buttons in addition to links. This makes the href prop in primary/secondaryLinks to be optional. If supplied, the button will have a role="link" and role="button" if not. The prop names have also been updated from primary/secondaryLinks to primary/secondaryActions to be more indicative of the new functionality.

## Test Plan
Tests and docs have been updated!

Codemod has been added for the breaking changes!


packages/gestalt/src/Callout.js
```diff
@@ -84,15 +98,27 @@ const CalloutLink = ({
...
-        onClick={onClick}
-        role="link"
-        size="lg"
-        text={label}
-      />
+      {isLink && (
+        <Button
+          accessibilityLabel={accessibilityLabel}
+          color={color}
+          href={data.href}
```
norrise120, line 105: I'm getting a flow here that I'm not sure how to fix. `Cannot get data.href because property href is missing in ButtonData`. I thought adding in the `isLink` check would work, but it isn't 🤔

norrise120: Ohh I wonder if instead of adding `role` we could just make `href` optional? If given, href, assume role is a link and if not, assume it's a button.

AlbertCarreras: yeah, this is expected, you have to segment Flowtype in a particular, way I'm not sure we can accomplish that here...  i might patch and play around.

ayeshakmaz: I like the idea of using the presence of `href` instead of specifying a role **and** an href. (doing a similar thing in the new Dropdown) 
I think you could simplify even more as well:

<Button
          color={color}
          href={href ? href : undefined} //this may get linted to {href || undefined}
          onClick={onClick}
          role={href ? "link" : "button"}
          size="lg"
          text={label}
        />

packages/gestalt/src/Callout.js
```diff
@@ -222,14 +250,15 @@ Callout.propTypes = {
...
+  primaryButton: PropTypes.shape({
+    href: PropTypes.string,
     label: PropTypes.string.isRequired,
     onClick: PropTypes.func,
     accessibilityLabel: PropTypes.string,
+    role: PropTypes.oneOf(['button', 'link']),
   }),
   // $FlowFixMe[signature-verification-failure] flow 0.135.0 upgrade
-  secondaryLink: PropTypes.shape({
+  secondaryButton: PropTypes.shape({
```
ayeshakmaz, line 261: Would `primaryAction` and `secondaryAction` make it more flexible and clear vs. specifying either Button or Link?

norrise120: I really like primary/secondaryAction! Updated!

docs/src/Upsell.doc.js
```diff
@@ -154,7 +138,7 @@ card(
 <Upsell
   title="Stay healthy and safe"
   message="Check out our resources for adapting to these times."
-  primaryLink={{href: "https://pinterest.com", label:"Visit"}}
+  primaryAction={{href: "https://pinterest.com", label:"Visit"}}
```
AlbertCarreras, line 141: is this missing a role='link' ?

docs/src/Upsell.doc.js
```diff
@@ -134,7 +118,7 @@ card(
 <Upsell
   title="Give $30, get $60 in ads credit"
   message="Earn $60 of ads credit, and give $30 of ads credit to a friend"
-  primaryLink={{href: "https://pinterest.com", label:"Send invite"}}
+  primaryAction={{href: "https://pinterest.com", label:"Send invite"}}
```
AlbertCarreras, line 121: is this missing a role='link' ?

packages/gestalt/src/Callout.js
AlbertCarreras: yeah, I looked at it, and this looks ok. The API here is very reduced for Button so this approach makes sense. @ayeshakmaz 's implementation is great.

AlbertCarreras: One concern, is that Links on Callout now open on the same page. I wonder if we should set it to 'blank'. This will be fixed the `useNavigation` hook. any thoughts for the interim state?

norrise120: I'll set the target to blank! I noticed it during testing and it was a really bad experience

packages/gestalt-codemods/16.0.0/upsell-primary-secondary-action-replace-primary-secondary-link.js
```diff
@@ -0,0 +1,63 @@
...
+  let fileHasModifications = false;
+
+  src.find(j.ImportDeclaration).forEach((path) => {
+    const decl = path.node;
+    if (decl.source.value !== 'gestalt') {
+      return null;
+    }
+
+    localIdentifierName = decl.specifiers
+      .filter((node) => node.imported.name === 'Upsell')
```
AlbertCarreras, line 19: we could merge two codemods in one by doing

`.filter((node) => ['Upsell, 'Callout'].includes(node.imported.name))`

packages/gestalt/src/Callout.js
AlbertCarreras: > I like the idea of using the presence of `href` instead of specifying a role **and** an href. (doing a similar thing in the new Dropdown)
> I think you could simplify even more as well:
> 
> ```
> <Button
>           accessibilityLabel={accessibilityLabel}
>           color={color}
>           href={href ? href : undefined} //this may get linted to {href || undefined}
>           onClick={onClick}
>           role={href ? "link" : "button"}
>           size="lg"
>           text={label}
>         />
> ```

Flow probably doesn't allow this. 

Button is segmented so Button role='button' can't have href/target props.  The only way to do this woulde {...href ? { href } : {} } but this aint good for codemods, spreading on Gestalt components makes props opaque.

AlbertCarreras: > I'll set the target to blank! I noticed it during testing and it was a really bad experience

as per discussion, leaving this change to future PR to not change more behavior and break stuff, we should see use cases and decide if we should add target prop or not or default to blank.

AlbertCarreras (approved):

PR #1315, opened 2020-12-17
TextArea: Add support for Tags

ohariz:
Same as https://github.com/pinterest/gestalt/pull/1306 but for TextArea

[Tag component](https://github.com/pinterest/gestalt/pull/1301)
[Doc](https://paper.dropbox.com/doc/Gestalt-Tag--BAbKmQvyCBritpD1GMhuN4pfAg-uceyUciJMgYCjtMRZkmwr)
[Figma](https://www.figma.com/file/Dp6tZGQigleqTA8LGCtcdj/Tags)

## Test Plan

Screen Shot 2020-12-17 at 3 31 45 PM
Screen Shot 2020-12-17 at 3 31 56 PM
Screen Shot 2020-12-17 at 3 32 20 PM
(the weird scrollbar overflow issue is unrelated to these changes, and I couldn't find a quick fix)


packages/gestalt/src/TextArea.js
```diff
@@ -91,28 +99,69 @@ const TextAreaWithForwardRef: React$AbstractComponent<
...
+      placeholder={placeholder}
+      ref={ref}
+      rows={tags ? undefined : rows}
+      value={value}
+    />
   );
 
+  const tagsWrapperStyle = {
+    height: rows * ROW_HEIGHT + INPUT_PADDING_WITH_TAGS,
+  };
```
ohariz, line 134: I had to manually size the component based on number of rows, since the sizing is on the div. Not sure if this would be the preferred way of doing this

AlbertCarreras: Can we check with design what's the right height for this?

Default is 3 rows, but only two rows of tags are visible.

 image

AlbertCarreras: Also do we need placeholder if we have Tags? Maybe the example could remove the placeholder after Tags are created.

image

packages/gestalt/src/TextArea.js
```diff
@@ -73,7 +81,7 @@ const TextAreaWithForwardRef: React$AbstractComponent<
   };
 
   const handleFocus = (event: SyntheticFocusEvent<HTMLTextAreaElement>) => {
-    setFocused(false);
+    setFocused(true);
```
AlbertCarreras, line 84: what's the reasoning behind this change? just curious.

packages/gestalt/src/TextArea.css
```diff
@@ -10,3 +10,36 @@
 .textArea::placeholder {
   color: var(--g-colorGray200);
 }
+
+.unstyledTextArea, .textAreaSpacer {
+  composes: unstyled from "./FormElement.css";
+  composes: Text fontSize3 from "./Text.css";
+  line-height: 1.5;
```
AlbertCarreras, line 17: i wonder if there are implications with i18n. we shouldnt apply line-height  because some languages are super tall. check with thai etc.

AlbertCarreras: Changes not allowed
Do not use the following CSS style properties:
line-height: Property in CSS that controls the space between lines of text. Not defining a line-height allows the browser to determine line-height based on language which prevents localization issues.

packages/gestalt/src/TextArea.js
ayeshakmaz: @ashleyseto what do you think about the height and placeholder options here?

packages/gestalt/src/TextArea.css
```diff
@@ -10,3 +10,36 @@
...
+  resize: none;
+}
+
+.unstyledTextArea {
+  composes: absolute left0 top0 right0 bottom0 from "./Layout.css";
+  height: 100%;
+  padding: 6px 8px;
```
ayeshakmaz, line 24: We're actually in the process of making Spacing tokens, so in the meantime we should probably stick to our "boint" system (4px = 1 boint), like so:

`padding: calc(var(--g-boint) * 4)`

Also, 6px will not be part of the new system, can this go up to 8 or down to 4?

packages/gestalt/src/TextArea.css
```diff
@@ -10,3 +10,36 @@
...
+  width: 100%;
+}
+
+.unstyledTextArea:focus {
+  outline: 0;
+}
+
+.textAreaSpacer {
+  padding: 4px 0;
```
ayeshakmaz, line 33: Same here for "boints"

packages/gestalt/src/TextArea.css
```diff
@@ -10,3 +10,36 @@
...
+  word-break: break-word;
+}
+
+.textAreaWrapper {
+  align-content: flex-start;
+  border-radius: 8px;
+  display: flex;
+  flex-wrap: wrap;
+  overflow-y: auto;
+  padding-bottom: 4px;
```
ayeshakmaz, line 44: Boints here as well

packages/gestalt/src/TextField.css
```diff
@@ -14,6 +14,7 @@
   composes: Text fontSize3 from "./Text.css";
   flex: 1;
   margin-bottom: 4px;
+  padding: 4px 8px;
```
ayeshakmaz, line 17: Should also probably do boints here (sorry, I missed this in the last PRs!)

packages/gestalt/src/TextArea.js
ohariz: It was wrong :) `handleFocus` should set focus to `true`. `handleBlur` is correctly setting it to `false`. I assume it was a typo

packages/gestalt/src/TextArea.css
ohariz: This is to match the line height of `.textArea` on line 5

packages/gestalt/src/TextArea.css
ohariz: The 6px are there to align the text with the tags. I can't really do any centering hackery because the `<textarea>` needs to have a 100% height. This is what it would look like with 4px:
Screen Shot 2020-12-18 at 10 17 13 AM
Screen Shot 2020-12-18 at 10 17 22 AM
I can change it if @ashleyseto is ok with it

packages/gestalt/src/TextArea.js
ohariz: > Default is 3 rows, but only two rows of tags are visible.

it's 3 rows of text, which ends up being 2 rows of text 😕  Another option could be to have the rows as a min height, and just grow from there instead of scrolling

packages/gestalt/src/TextArea.css
AlbertCarreras: OK,  then this is a Gestalt issue ahahah no worries.

packages/gestalt/src/TextArea.js
ashleyseto: I think if someone wants to used text area, then ideally they're setting a height and then it expands from there so yeah a min-height makes sense.

I know also I added a placeholder to continue, I think I could update the designs so the placeholder doesn't keep appearing after 1 tag is added.

packages/gestalt/src/TextArea.css
ashleyseto: I think the issue right now is the text in the input is larger than the text in the tags. Is that correct? I'm not really sure how to resolve this right now. If we want to keep to 4px for now to not add odd spaces, we can. I think most of my spacing comments were around the internal part of the tag so that doesn't look weird.

docs/src/TextArea.doc.js
```diff
@@ -234,6 +240,71 @@ function Example(props) {
   />
 );
 
+card(
+  <Example
```
ayeshakmaz, line 244: One thing I noticed: in your Tyepahead example, after deleting a tag, focus is back on the input, but that doesn't seem to be the case here in Textarea. After deleting a tag in text area, focus is lost.

packages/gestalt/src/TextArea.test.js
```diff
@@ -60,4 +61,34 @@ describe('TextArea', () => {
     ).toJSON();
     expect(tree).toMatchSnapshot();
   });
+
+  it('TextArea with tags', () => {
```
ayeshakmaz, line 65: Maybe rename to "renders tags when supplied" so the full test name reads better

packages/gestalt/src/TextArea.js
AlbertCarreras: @ashleyseto @ohariz Just to wrap up this change: `rows` still sets a (min)height that doesn't match with the number of Tag rows, however, TextArea grows as new lines are added.  

image

docs/src/TextArea.doc.js
```diff
@@ -234,6 +240,71 @@ function Example(props) {
...
+        ...tags,
+        // Avoid creating a tag on content on the last line, and filter out
+        // empty tags
+        ...tagInput.splice(0, tagInput.length - 1).filter(val => val !== ''),
+      ]);
+    }
+    setValue(tagInput[tagInput.length - 1]);
+  }
+
+  const onKeyDownTagManagement = ({ event: { keyCode, target: { selectionEnd } } }) => {
```
AlbertCarreras, line 273: nit

 ```
 const onKeyDownTagManagement = ({
    event: {
      keyCode,
      target: { selectionEnd },
    },
  }) => {
```

docs/src/TextArea.doc.js
AlbertCarreras: Same behavior in `Textfield`.

packages/gestalt/src/TextArea.js
AlbertCarreras: Also, quick one. Could you specify in the Docs that the rows props `Number of text (not Tag) rows to display.`   
or something like that?

ashleyseto (approved):

ayeshakmaz (approved):
Looks great 🎉

PR #1316, opened 2020-12-17
Modal: Add subHeading and align props

norrise120:
This PR updates modal so that it can take a subHeading prop to add a subtitle under the heading and an align prop to justify the heading text left or center.

Screen Shot 2020-12-17 at 1 38 28 PM
Screen Shot 2020-12-17 at 1 38 21 PM

## Test Plan
All tests are passing and I manually checked the doc examples to make sure that no unintended changes occurred.

### DIFF
--- a/docs/src/Modal.doc.js
+++ b/docs/src/Modal.doc.js
@@ -27,6 +27,14 @@ card(
         href: 'accessibility',
       },
       {
+        name: 'align',
+        type: `"center" | "left"`,
+        defaultValue: 'center',
+        description:
+          'Use to specify the alignment of heading when supplied heading is a string',
+        href: 'headingOptions',
+      },
+      {
         name: 'children',
         type: 'React.Node',
       },
@@ -72,6 +80,12 @@ card(
         defaultValue: 'sm',
         description: `sm: 540px, md: 720px, lg: 900px`,
         href: 'sizesExample',
+      },
+      {
+        name: 'subHeading',
+        type: `string`,
+        required: false,
+        href: 'headingOptions',
       },
     ]}
   />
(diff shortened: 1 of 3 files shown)


ashleyseto (approved):

packages/gestalt/src/Modal.js
```diff
@@ -34,16 +37,29 @@ const SIZE_WIDTH_MAP = {
   lg: 900,
 };
 
-function Header({ heading }: {| heading: string | Node |}) {
+function Header({
+  align,
+  heading,
+  subHeading,
+}: {|
```
AlbertCarreras, line 44: can we try segmenting flow here

type Header = {| heading:  Node |} | {| align: 'left' | 'center',
  heading: string,
  subHeading?: string,|}

packages/gestalt/src/Modal.js
```diff
@@ -14,18 +14,21 @@ import Backdrop from './Backdrop.js';
 import focusStyles from './Focus.css';
 import Heading from './Heading.js';
 import StopScrollBehavior from './behaviors/StopScrollBehavior.js';
+import Text from './Text.js';
 import TrapFocusBehavior from './behaviors/TrapFocusBehavior.js';
 import modalStyles from './Modal.css';
 
 type Props = {|
   accessibilityModalLabel: string,
+  align?: 'left' | 'center',
```
AlbertCarreras, line 23: here something like

```
type Props = {|
  accessibilityModalLabel: string,
  children?: Node,
  footer?: Node,
  onDismiss: () => void,
  role?: 'alertdialog' | 'dialog',
  size?: 'sm' | 'md' | 'lg' | number,
  ...
    | {|
        heading?: Node,
      |}
    | {|
        heading?: string,
        subHeading?: string,
        align?: 'left' | 'center',
      |},
|};

```

docs/src/Modal.doc.js
```diff
@@ -73,6 +81,12 @@ card(
         description: `sm: 540px, md: 720px, lg: 900px`,
         href: 'sizesExample',
       },
+      {
+        name: 'subHeading',
+        type: `string`,
+        required: false,
+        href: 'headingOptions',
+      },
```
AlbertCarreras, line 89: can we add description here saying it only renders with Heading strings

docs/src/Modal.doc.js
```diff
@@ -26,6 +26,14 @@ card(
           'String that clients such as VoiceOver will read to describe the modal. Always localize the label.',
         href: 'accessibility',
       },
+      {
+        name: 'align',
+        type: `"center" | "left"`,
+        defaultValue: 'center',
+        description:
+          'Use to specify the alignment of heading when supplied heading is a string',
```
AlbertCarreras, line 34: 'Use to specify the alignment of heading & subHeading strings',

packages/gestalt/src/Modal.js
AlbertCarreras: hope this works, not quite sure ha

packages/gestalt/src/Modal.js
AlbertCarreras: hope this works too. it might require playing around a little.

norrise120: I got some flow errors doing this, but I was able to get it to work if I moved the string check to the component. Let me know if you think this is a better way!

AlbertCarreras: yeah makes sense!

AlbertCarreras (approved):

PR #1317, opened 2020-12-18
Typeahead: Add support for Tags

ohariz:
Same as https://github.com/pinterest/gestalt/pull/1315 but for Typeaheads

## Test Plan

Screen Shot 2020-12-18 at 3 56 25 PM
Screen Shot 2020-12-18 at 3 56 31 PM
Screen Shot 2020-12-18 at 3 56 39 PM
Screen Shot 2020-12-18 at 3 56 44 PM
Screen Shot 2020-12-18 at 3 57 03 PM
Screen Shot 2020-12-18 at 3 57 18 PM

### DIFF
--- a/docs/src/Typeahead.doc.js
+++ b/docs/src/Typeahead.doc.js
@@ -249,50 +249,60 @@ card(
     This example showcases the recommended behavior.`}
     defaultCode={`
 function Example(props) {
-  const options = [
-    { label: 'blue', value: 'blue' },
-    { label: 'green', value: 'green' },
-    { label: 'orange', value: 'orange' },
-    { label: 'purple', value: 'purple' },
-    { label: 'red', value: 'red' },
-    { label: 'yellow', value: 'yellow' },
-  ];
-
+  const colors = ['blue', 'green', 'orange', 'purple', 'red', 'yellow'];
+  const ref = React.useRef();
   const [selected, setSelected] = React.useState([]);
+
+  const setFocus = () => {
+    // Focus needs to happen after selection is complete
+    setTimeout(() => {
+      ref.current.focus();
+    }, 0);
+  };
 
   const onSelectTagManagement = ({ item: { value } }) => {
     if (!selected.includes(value)) {
       setSelected([...selected, value]);
     }
+    setFocus();
+  }
+
+  const onRemoveTagManagement = (value) => {
+    setSelected(selected.filter(color => color !== value));
+    setFocus();
   }
 
   const onKeyDownTagManagement = ({ event: { keyCode, target: { selectionEnd } } }) => {
     if (keyCode === 8 /* Backspace */ && selectionEnd === 0) {
       // Remove tag on backspace if the cursor is at the beginning of the field
       setSelected([...selected.slice(0, -1)]);
+      setFocus();
     }
   }
 
-  const renderedTags = options
-    .filter(({ value }) => selected.includes(value))
-    .map(({ label, value }) => (
+  const options = colors
+    .filter(color => !selected.includes(color))
+    .map(color => ({ label: color, value: color }));
+
+  const renderedTags = selected.map(color => (
       <Tag
-        key={value}
-        onRemove={() => setSelected({ ...selected, [value]: false })}
-        removeIconAccessibilityLabel={\`Remove \${label} tag\`}
-        text={label}
+        key={color}
+        onRemove={() => onRemoveTagManagement(color)}
+        removeIconAccessibilityLabel={\`Remove \${color} tag\`}
+        text={color}
       />
     ));
 
   return (
     <Typeahead
       key={JSON.stringify(selected) /* force update when the options change */}
-      label="Select your favorite colors"
+      ref={ref}
       id="favorite-colors"
+      label="Select your favorite colors"
       noResultText="No Results"
       onKeyDown={onKeyDownTagManagement}
       onSelect={onSelectTagManagement}
-      options={options.filter(({ value }) => !selected.includes(value))}
+      options={options}
       placeholder="Select colors"
       tags={renderedTags}
     />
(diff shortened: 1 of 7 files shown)


ashleyseto:
Hey. Seems like most of this is working correctly. A few things in regards to functionality I noticed.

- The tags seem to be auto-sorting once entered, I'd say they shouldn't be doing that and should appear in the order they were entered 
- The cursor should still be active after entering in one tag to allow users to keep entering tags.

Probably something to think about in a future version is the ability to add tags that aren't in the list but I don't think that's needed for now.

ashleyseto:
Thanks for the updates. One other thing I noticed now that I can continue entering tags is that the dropdown part is bound to where the tag is located. The dropdown itself should maintain the width of the input field rather than following the tag.

Screen Shot 2021-01-04 at 3 29 38 PM

ohariz:
I've been messing around with it, and it seems that the only way to get a full width dropdown is to change the anchoring of the flyout, given that we need relative positioning on the input's parent to give it the full-width wrapping behavior.

@ayeshakmaz @AlbertCarreras how important is it from an accessibility perspective to have the flyout anchored to the input element itself as opposed to a parent div?

ashleyseto:
> I've been messing around with it, and it seems that the only way to get a full width dropdown is to change the anchoring of the flyout, given that we need relative positioning on the input's parent to give it the full-width wrapping behavior.
> 
> @ayeshakmaz @AlbertCarreras how important is it from an accessibility perspective to have the flyout anchored to the input element itself as opposed to a parent div?

It'll be important from a UX standpoint that this behavior is correct. We can't have the amount of information you're seeing getting smaller and smaller.

Again I'd probably reference other examples out there. https://ant.design/components/tree-select/

docs/src/Typeahead.doc.js
```diff
@@ -203,25 +216,100 @@ function TypeaheadExample() {
...
+      />
     </Flex>
   );
 }`}
   />
 );
 
+card(
+  <Example
+    id="tagsExample"
```
ayeshakmaz, line 242: It seems that if you type something not included in the list of options, and then hit enter, it deletes everything you typed. I think instead it should leave what you typed, without adding it as a tag. Deleting everything typed could cause frustration is users simply made a typo and then have to re-type everything

ayeshakmaz (changes requested):

docs/src/Typeahead.doc.js
ohariz: That's just the current behavior of the Typeahead component, unrelated to tags 😕

packages/gestalt/src/TypeaheadInputField.css
```diff
@@ -27,4 +48,12 @@ html:not([dir="rtl"]) .icon {
 
 html[dir="rtl"] .icon {
   left: 0;
-}
\ No newline at end of file
+}
+
+html:not([dir="rtl"]) .inputWrapper,
+html[dir="rtl"] .inputWrapper {
+  /* Increase specificity for this class to override SearchField defaults */
```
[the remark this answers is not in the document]
ohariz: This styling in `SearchField.css`:
```
html:not([dir="rtl"]) .inputActive {
  padding: 8px 32px 8px 16px;
}

html[dir="rtl"] .inputActive {
  padding: 8px 16px 8px 32px;
}
 ```
 Because this has a higher [specificity](https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity) than just `.inputWrapper`, I had to add the `html` selector to actually be able to override it

AlbertCarreras:
@ashleyseto  @ohariz  Do we want to keep the placeholder here or remove it as in `TextArea`? 
image

ohariz:
Good call. I think it makes sense to remove it

docs/src/Typeahead.doc.js
ayeshakmaz: Ahh good catch, ignore that then!

packages/gestalt/src/TypeaheadInputField.css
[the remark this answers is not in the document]
ayeshakmaz: ahh, so it's just the bottom padding that needs to be overridden? In that case, could we separate this into two sections of CSS, one with just `.inputWrapper` specifying the display and flex-wrap, and then another with the higher specificity for the padding bottom?

docs/src/Typeahead.doc.js
ashleyseto: Yeah I think that's maybe where a bigger amount of work needs to be done around adding tags and validation around tags in a typeahead. I'm assuming that's a decent chunk of work?

ayeshakmaz: Could you elaborate @ashleyseto? Are you thinking that if people try to add a tag in typeahead that's not one of the options that we warn/alert the user it's not an option? Or would we have an option to add tags that aren't included in the options list? (or both 😛 )

ashleyseto: Oh I mean that we could have a select from list of items but an option to 'add your own' item to a list if this was a category of items. I don't have a use case for this yet that I know of so it isn't important.

The original thinking with the typeahead was to clear invalid items to avoid error states. In our case now though, we do offer an error tag that we could use if there's feedback that clearing is an issue.

ashleyseto:
> @ashleyseto @ohariz Do we want to keep the placeholder here or remove it as in `TextArea`?
> image

I think it makes sense to remove it unlike what I had in the original design because of this edge case.

packages/gestalt/src/TypeaheadInputField.css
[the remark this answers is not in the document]
ohariz: yeah, we can do that

ashleyseto (approved):

docs/src/Typeahead.doc.js
ayeshakmaz: Makes sense, we'll keep an eye out and see

ayeshakmaz (approved):
Looks good, great work!!

PR #1318, opened 2020-12-18
Modal: Fix Modal Flow type

AlbertCarreras:


jackhsu978:
fixed in https://github.com/pinterest/gestalt/pull/1322

PR #1319, opened 2020-12-21
Sheet: Add shouldCloseOnKeyEvent prop

jazomand:

Currently there is no way to disable the close key event for Sheet component. This PR allows the sheet component to have such behavior.

## Test Plan

* Is it tested? Yes
* Is it accessible? Yes
* Is it documented? Yes

### DIFF
--- a/docs/src/Sheet.doc.js
+++ b/docs/src/Sheet.doc.js
@@ -115,7 +115,7 @@ card(
         type: `boolean`,
         defaultValue: 'true',
         description: ['Determine if component should close on key event'],
-        href: 'refExample',
+        href: 'shouldCloseOnKeyEventExample',
       },
       {
         name: 'size',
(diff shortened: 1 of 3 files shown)


packages/gestalt/src/Sheet.js
```diff
@@ -51,6 +51,7 @@ type SheetMainProps = {|
   closeOnOutsideClick?: boolean,
   footer?: Node,
   onDismiss: () => void,
+  shouldCloseOnKeyEvent?: () => boolean,
```
ayeshakmaz, line 54: Does this need to be a function, or could it be a boolean, which defaults to true, and any additional functionality needed can be part of the `onDismiss` passed in?

ayeshakmaz:
Could you also describe when we would want to disable `ESC` from closing the sheet? Accessibility-wise, users will always expect the ESC key to close any type of [modal dialog](https://www.w3.org/TR/wai-aria-practices-1.1/examples/dialog-modal/dialog.html)

jazomand:
> Could you also describe when we would want to disable `ESC` from closing the sheet? Accessibility-wise, users will always expect the ESC key to close any type of [modal dialog](https://www.w3.org/TR/wai-aria-practices-1.1/examples/dialog-modal/dialog.html) (which includes Sheets)

Yes, the idea to disable the ESC key behavior is actually to fix some usability issues in CFPB (Create Flow Pin Builder). ESC key is needed to close inner/children components. When those are closed, ESC key behavior could be restored as usual.

packages/gestalt/src/Sheet.js
jazomand: I believe it will be better to use a function in this case, because the code is using "window.addEventListener" directly, without react synthetic events.

docs/src/Sheet.doc.js
```diff
@@ -698,6 +705,60 @@ function RefExample() {
   />
 );
 
+card(
+  <Example
+    id="shouldCloseOnKeyEventExample"
+    name="Example: shouldCloseOnKeyEvent"
+    description={`
+    A \`Sheet\` with shouldCloseOnKeyEvent
```
ayeshakmaz, line 713: Could you elaborate a bit more here as to when people would want this feature disabled, and maybe add a tooltip/flyout to the sheet to help demonstrate the behavior. 

Maybe something like "The `shouldCloseOnKeyEvent` may need to be disabled if the Sheet contains components that rely on the escape key themselves. When this property is disabled, using the escape key to close an inner component will not inadvertently close the Sheet. If all inner components are closed, this property should be restored to true to allow for the escape key to close the Sheet."

docs/src/Sheet.doc.js
```diff
@@ -698,6 +705,60 @@ function RefExample() {
...
+              onDismiss={() => setShouldShow(false)}
+              shouldCloseOnKeyEvent={shouldCloseOnKeyEvent}
+              size="md"
+            >
+              <Box color="white" minHeight={400} padding={8}>
+                <Box marginBottom={4}>
+                  <Heading size="md">Sheet</Heading>
+                </Box>
+                <Text>
+                  This sheet {shouldCloseOnKeyEvent ? "CAN" : "CAN'T"} be closed with ESC key.
```
ayeshakmaz, line 748: Might be great to add an inner component here that relies on esc key functionality

ayeshakmaz (changes requested):

docs/src/Sheet.doc.js
jazomand: Done

docs/src/Sheet.doc.js
jazomand: Done

ayeshakmaz:
Hey Antonio, just a heads up: the team wants to look into this a bit more before adding it, I'll update here with any questions/suggestions we might have

dangerismycat:
This hasn't seen any work in over a year, so closing for now. Feel free to re-open if this functionality is still needed and we can discuss possible solutions.

PR #1322, opened 2020-12-22
Modal: fix bad flow types

jackhsu978:
Fix flow errors for Modal introduced in https://github.com/pinterest/gestalt/pull/1316

### DIFF
--- a/packages/gestalt/src/Modal.js
+++ b/packages/gestalt/src/Modal.js
@@ -20,21 +20,15 @@ import modalStyles from './Modal.css';
 
 type Props = {|
   accessibilityModalLabel: string,
+  align?: 'left' | 'center',
   children?: Node,
   closeOnOutsideClick?: boolean,
   footer?: Node,
+  heading?: Node,
   onDismiss: () => void,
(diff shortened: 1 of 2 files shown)


packages/gestalt/src/Modal.js
```diff
@@ -20,21 +20,15 @@ import modalStyles from './Modal.css';
...
-  ...
-    | {|
-        heading?: Node,
-      |}
-    | {|
-        heading?: string,
-        subHeading?: string,
-        align?: 'left' | 'center',
-      |},
+  subHeading?: string,
```
jennyscript, line 31: I think the idea here is to only have the subHeading and align props if the heading is a string specifically. I see Flow is getting confused by Node including string, but is it possible to keep the original idea? Can Node be changed to something that's more specifically a React component so it doesn't include string?

jennyscript: Discussed and it looks not possible. This is fine since it's similar but less strict.

jennyscript (approved):

PR #1324, opened 2021-01-04
Layer: Update documentation examples around avoiding unwanted re-rendering

themindfuldev:
This diff is just improving documentation for Layer to explain and illustrate with an example the importance of avoiding re-renderings to prevent the content from being destroyed with an accidental parent re-render.

### DIFF
--- a/docs/src/Layer.doc.js
+++ b/docs/src/Layer.doc.js
@@ -79,8 +79,8 @@ function Example() {
   // gets passed down as a prop to <ResilientLayer>.
   // It's very important to pass down any function prop using useCallback,
   // otherwise, each parent re-render would recreate these functions, 
-  // which would force a children re-render as is receives different props.
-  // In our case, our children Layer would re-render and destroy its content.
+  // which would force a children re-render as it receives different props.
+  // In our case, our children Layer would re-render and destroy its contents.
   const onDismiss = React.useCallback(() => setShowLayer(false), []);
 
   return (
--- a/docs/src/Layer.doc.js
+++ b/docs/src/Layer.doc.js
@@ -45,6 +45,62 @@ card(
 card(
   <Example
     description="
+    Since Layer is rendered within a portal element being manually added to the DOM, it is supposed to be rendered and dismissed only once. 
+    Any re-renders triggered by the parent will effectively remove the portal element from the DOM, hence it won't be recreated and internal elements like Sheet may not reappear.
+    The example below demonstrates a technique to avoid the re-rendering of Layer by making sure parent re-renders don't force children re-renders. This is accomplished by memoizing locally-defined functions with React.useCallback() if they're passed down as props, and memoizing the entire component with React.memo().
+  "
+    name="Avoid Re-rendering"
+    defaultCode={`
+function Example() {
+  // We're using React.memo() here to prevent re-renders. This is similar to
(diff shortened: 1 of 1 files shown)


themindfuldev:
Docs are live at: https://deploy-preview-1324--gestalt.netlify.app/Layer

docs/src/Layer.doc.js
```diff
@@ -42,6 +42,62 @@ card(
   />
 );
 
+card(
+  <Example
+    description="
+    Since Layer is rendered within a portal element being manually added to the DOM, it is supposed to be rendered and dismissed only once. 
+    Any re-renders triggered by the parent will effectively remove the portal element from the DOM, hence it won't be re-created and internal elements like Sheet may not reappear.
+    The example below demonstrate a technique to avoid Layer re-rendering by making sure parent re-renders don't force re-renders by passing useCallback function props and preventing re-renders with memo().
```
dangerismycat, line 50: typo: demonstrates

Also: it's a bit of a run-on sentence, and could use some clarification. Consider breaking into smaller chunks, something like:

"The example below demonstrates a technique to avoid Layer re-rendering by making sure parent re-renders don't force children re-renders. This is accomplished by memoizing locally-defined functions with React.useCallback if they're passed down as props, and memoizing the entire component with React.memo()."

docs/src/Layer.doc.js
```diff
@@ -42,6 +42,62 @@ card(
...
+  "
+    name="Avoid Re-rendering"
+    defaultCode={`
+function Example() {
+  // We're using memo() here to prevent re-renders similarly to
```
dangerismycat, line 55: Another run-on that can be broken up:
"We're using memo() here to prevent re-renders. This is similar to shouldComponentUpdate(), but for functional components."

dangerismycat (commented):
This is awesome, thanks Tiago!

docs/src/Layer.doc.js
themindfuldev: Suggestion accepted, thank you!

docs/src/Layer.doc.js
themindfuldev: Suggestion accepted, thank you!

docs/src/Layer.doc.js
```diff
@@ -42,6 +42,62 @@ card(
...
+    );
+  });
+
+  const [showLayer, setShowLayer] = React.useState(false);
+
+  // We're using React.useCallback() here to memoize this function since it 
+  // gets passed down as a prop to <ResilientLayer>.
+  // It's very important to pass down any function prop using useCallback,
+  // otherwise, each parent re-render would recreate these functions, 
+  // which would force a children re-render as is receives different props.
```
dangerismycat, line 82: typo: "as _it_ receives"

docs/src/Layer.doc.js
```diff
@@ -42,6 +42,62 @@ card(
...
+  // In our case, our children Layer would re-render and destroy its content.
```
dangerismycat, line 83: typo: contents

dangerismycat (approved):

docs/src/Layer.doc.js
themindfuldev: 👍🏽

ayeshakmaz (approved):
This is great! Thanks for the addition

PR #1327, opened 2021-01-05
Layer: fix a bug where Layer ummounts children on rerender when zIndex changes

### LINKED ISSUE #1326: Layer unmounts upon zIndex changes
As pointed out by @chrislloyd and @jackhsu978 , this block of `useEffect` will make DOM manipulation upon `zIndex` updates, it would be best to split them up and also rely on shallow values (which `zIndex` is not because it's an `Indexable`):

```js
  useEffect(() => {
    setMounted(true);

    if (typeof document !== 'undefined' && document.createElement) {
      element.current = document.createElement('div');
      element.current.style.zIndex = zIndex ? `${zIndex.index()}` : '';
      element.current.className = zIndex ? styles.layer : '';
    }

    if (typeof document !== 'undefined' && document.body && element.current) {
      document.body.appendChild(element.current);
    }

    return () => {
      if (typeof document !== 'undefined' && document.body && element.current) {
        document.body.removeChild(element.current);
      }
    };
  }, [zIndex]);
```

A possible suggestion:
```js
useEffect(() => {
  // the same as whatever we have
}, []);

const z = zIndex?.index().toString() ?? '';

useEffect(() => {
  if (element.current) {
    element.current.style.zIndex = z;
    element.current.className = z !== '' ? styles.layer : '';
  }
}, [z]);
```

When testing, verify that before the fix, a re-render which re-creates a new instance of zIndex ends up removing the children element from the DOM, and after the fix this won't happen anymore.

jackhsu978:
This fixes #1326.

When we pass `zIndex` indexable object and the reference is changed, the cleanup routine of `useEffect` hook, which depends on `zIndex` would be fired and would replace the entire children tree with a new one. I believe this is an oversight when we added `zIndex` support in #1223. I am fixing this bug by:

1. make sure the `useEffect` hook which appends/removes a div element only runs on mount/unmount.
2. the zIndex is handled by a separate `useEffect` after the `useEffect` hook above so that it (1) initiailizes the zIndex after the div has been created on mount, and (2) updates zIndex-related attributes of the div element whenever the value of the zIndex prop changes.

I also removed the documentation where we encourage people to avoid rerendering. Avoid rerendering should only be a perf optimization and should not be used as a semantic guarentee.

### DIFF
--- a/packages/gestalt/src/Layer.js
+++ b/packages/gestalt/src/Layer.js
@@ -6,21 +6,20 @@ import styles from './Layer.css';
 
 export default function Layer({
   children,
-  zIndex,
+  zIndex: zIndexIndexable,
 }: {|
   children: Node,
   zIndex?: Indexable,
 |}): Portal | null {
   const [mounted, setMounted] = useState(false);
   const element = useRef<?HTMLDivElement>(null);
+  const zIndex: number | null = zIndexIndexable?.index() ?? null;
 
   useEffect(() => {
     setMounted(true);
 
     if (typeof document !== 'undefined' && document.createElement) {
       element.current = document.createElement('div');
-      element.current.style.zIndex = zIndex ? `${zIndex.index()}` : '';
-      element.current.className = zIndex ? styles.layer : '';
     }
 
     if (typeof document !== 'undefined' && document.body && element.current) {
@@ -32,6 +31,13 @@ export default function Layer({
         document.body.removeChild(element.current);
       }
     };
+  }, []);
+
+  useEffect(() => {
+    if (element.current) {
+      element.current.style.zIndex = zIndex === null ? '' : `${zIndex}`;
+      element.current.className = zIndex === null ? '' : styles.layer;
+    }
   }, [zIndex]);
 
   if (!mounted || !element.current) {
--- a/packages/gestalt/src/Layer.js
+++ b/packages/gestalt/src/Layer.js
@@ -13,7 +13,7 @@ export default function Layer({
 |}): Portal | null {
   const [mounted, setMounted] = useState(false);
   const element = useRef<?HTMLDivElement>(null);
-  const zIndex: number | null = zIndexIndexable?.index() ?? null;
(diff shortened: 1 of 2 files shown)


packages/gestalt/src/Layer.js
```diff
@@ -6,21 +6,20 @@ import styles from './Layer.css';
...
   children,
-  zIndex,
+  zIndex: zIndexIndexable,
 }: {|
   children: Node,
   zIndex?: Indexable,
 |}): Portal | null {
   const [mounted, setMounted] = useState(false);
   const element = useRef<?HTMLDivElement>(null);
+  const zIndex: number | null = zIndexIndexable?.index() ?? null;
```
chrislloyd, line 16:
```suggestion
  const zIndex = zIndexIndexable?.index();
```

packages/gestalt/src/Layer.js
```diff
@@ -32,6 +31,13 @@ export default function Layer({
         document.body.removeChild(element.current);
       }
     };
+  }, []);
+
+  useEffect(() => {
+    if (element.current) {
+      element.current.style.zIndex = zIndex === null ? '' : `${zIndex}`;
+      element.current.className = zIndex === null ? '' : styles.layer;
+    }
```
chrislloyd, line 40:
```suggestion
    if (!element.current) {
      return;
    }
    element.current.style.zIndex = zIndex === null ? '' : `${zIndex}`;
    element.current.className = zIndex === null ? '' : styles.layer;
```

jackhsu978: done and also changed to `null` check to `undefined` check.

packages/gestalt/src/Layer.js
jackhsu978: done

jackhsu978:
@chrislloyd I addressed the comments. can you please take another look?

chrislloyd (approved):
Love it! Thanks for the fix @jackhsu978

PR #1328, opened 2021-01-07
Callout: Fix vertical padding in md viewport when no actions present

mtanji:
This PR intends to fix a bogus vertical padding behavior of Callout component when there isn't action buttons and the message is long and its width transitions from something around 750px to 700px. Current behavior presents a bottom padding that is taller than the upper padding. That behavior can be confirmed by adjusting the window width in the sandbox link and the screenshot below.
https://codesandbox.io/s/error-example-forked-mg0ji?file=/example.js

Screen Shot 2021-01-07 at 7 57 58 PM

## Changes
The only change was to avoid rendering a Box without action buttons in the Callout component

## Test Plan
Manually tested the change by loading the Callout Gestalt Docs page and set a long message in a sample that has no action buttons, then narrowing window width and checking that the isn't a vertical padding issue anymore. See attached screen recording.
https://user-images.githubusercontent.com/3874746/103953970-f1f3f600-5121-11eb-99a4-b77955297389.mov

### DIFF
--- a/packages/gestalt/src/Callout.js
+++ b/packages/gestalt/src/Callout.js
@@ -193,21 +193,28 @@ export default function Callout({
             </Box>
           </Box>
         </Box>
-        <Box smDisplay="flex" marginStart="auto" smMarginEnd={4} smPaddingY={3}>
-          {secondaryAction && responsiveMinWidth !== 'xs' && (
-            <CalloutAction type="secondary" data={secondaryAction} />
-          )}
-          {primaryAction && (
-            <CalloutAction type="primary" data={primaryAction} />
-          )}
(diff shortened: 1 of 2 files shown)


ashleyseto:
Thanks for this fix! @AlbertCarreras or @ayeshakmaz can you review?

ayeshakmaz:
Looks good from a glance, @mtanji mind fixing the prettier issues and then I can review again? Thanks!

mtanji:
sure, doing it now

mtanji:
@ayeshakmaz , prettier issues were fixed. It's ok to see the Semver issue failing, right?

ayeshakmaz (approved):
Looks good, thank you!

PR #1330, opened 2021-01-08
Callout: Include Node type for Message property

simkessy:
Using a React Node in the message causes a type error: 

image

### DIFF
--- a/packages/gestalt/src/Callout.js
+++ b/packages/gestalt/src/Callout.js
@@ -31,7 +31,7 @@ type Props = {|
     onDismiss: () => void,
   |},
   iconAccessibilityLabel: string,
-  message: string,
+  message: string | Node,
   primaryAction?: ActionData,
   secondaryAction?: ActionData,
   type: 'error' | 'info' | 'warning',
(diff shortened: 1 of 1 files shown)


ashleyseto:
@simkessy 
We do not wish to change the message area to react node. Bold is our link style so these may be confused as links. I'd recommend removing the bold right now in that message and updating the design.

AlbertCarreras:
Hey Kessy, can we close this PR until the design for this is discussed?

simkessy:
> @simkessy
> We do not wish to change the message area to react node. Bold is our link style so these may be confused as links. I'd recommend removing the bold right now in that message and updating the design.

@ashleyseto  I need a react node to get the spacing / separation between the lines. A string alone won't allow us to do that.

ashleyseto:
> > @simkessy
> > We do not wish to change the message area to react node. Bold is our link style so these may be confused as links. I'd recommend removing the bold right now in that message and updating the design.
> 
> @ashleyseto I need a react node to get the spacing / separation between the lines. A string won't allow us to do that.

Hey, I would encourage you to ask your designer to sign up for design system office hours then. We should first start with the design and make a decision. The PR should come after design approval not the other way around. @anniemochi for the creators side of things.

Ideally we do not open this up to "node" that is too much freedom for this component. It may not be the right pattern here.

PR #1336, opened 2021-01-13
Module: Add Static version

ayeshakmaz:
Add new, Static version of Module
Screen Shot 2021-01-13 at 4 44 29 PM

## Test Plan

Added new snapshot tests

Jira: [PDS-2369](https://jira.pinadmin.com/browse/PDS-2369)


packages/gestalt/src/Module.js
```diff
@@ -1,9 +1,56 @@
...
+}: BaseProps): Node {
+  const MODULE_TYPE_ATTRIBUTES = {
+    info: {
+      icon,
+      color: 'darkGray',
+    },
+    error: {
+      icon: 'workflow-status-problem',
+      color: 'red',
+    },
```
ayeshakmaz, line 25: Would love to de-dup this (used here and in the Expandable version), but the use of `icon: icon` throws a kink in it. Welcome to any ideas (thinking maybe just a util function?)

docs/src/Module.doc.js
```diff
@@ -12,49 +12,201 @@ card(
...
-"
+      A Module is a container that holds content about one subject. Its contents can be visible at all times, or expand and collapse as individual modules or a group of modules.
+    "
   />
 );
 
 card(
   <PropTable
     Component={Module}
+    name="Non-Expandable Module"
```
dangerismycat, line 23: "Non-expandable" is a bit awkward…how would you feel about "static"?

docs/src/Module.doc.js
```diff
@@ -12,49 +12,201 @@ card(
...
       },
     ]}
   />
 );
 
 card(
   <Example
-    name="Example"
+    name="Non-Expandable"
+    description={`A Module is a container that can hold any content, and can optionally have a \`title\` that describes the content inside. The default, non-expandable Module is used to display information that should always be visible.`}
```
dangerismycat, line 115: nit: just use a string prop to obviate the need to escape characters

docs/src/Module.doc.js
```diff
@@ -12,49 +12,201 @@ card(
...
+  );
+}
+`}
+  />
+);
+
+card(
+  <Example
+    name="Non-Expandable - Icon"
+    description={`
```
dangerismycat, line 142: ditto here

docs/src/Module.doc.js
```diff
@@ -12,49 +12,201 @@ card(
...
+}
+`}
+  />
+);
+
+card(
+  <Example
+    name="Expandable"
+    id="expandable-default"
+    description={`Modules can also allow for expanding and collapsing content. The \`title\` is required and always present, the collapsed state shows optional \`summary\` content, and the expanded state shows any content desired.`}
```
dangerismycat, line 203: ditto here

docs/src/Module.doc.js
```diff
@@ -12,49 +12,201 @@ card(
...
+    </Box>
+  );
+}
+`}
+  />
+);
+
+card(
+  <Example
+    name="Non-Expandable - Error"
```
dangerismycat, line 169: This example is a little odd to play around with…errors on an input should disappear when corrected, but in this example we continue to show the error module state even if the user enters something into the input. Perhaps input validation isn't a great example? Do we have real designs we can model this example on?

Also: should the error examples have a description to indicate when they should be used?

docs/src/Module.doc.js
```diff
@@ -108,7 +262,8 @@ function ModuleExample2() {
 
 card(
   <Example
-    name="Example with icon"
+    name="Expandable - Icon"
+    description={`An Icon can be provided to be placed before the \`title\`. If the icon provides information about this module (locked, disabled, new, etc.), be sure to provide an \`iconAccessibilityLabel\`.`}
```
dangerismycat, line 266: ditto here

docs/src/Module.doc.js
```diff
@@ -134,9 +289,10 @@ function ModuleExample3() {
 
 card(
   <Example
-    name="Error state"
+    name="Expandable - Error"
```
dangerismycat, line 292: This example has the same issue. It's a little weird to play around with.

packages/gestalt/src/Module.js
```diff
@@ -1,9 +1,30 @@
 // @flow strict
-import { type Node } from 'react';
+import React, { type Node } from 'react';
+import Box from './Box.js';
+import { type BaseProps } from './moduleTypes.js';
+import { renderModuleTitle } from './moduleUtils.js';
 import ModuleExpandable from './ModuleExpandable.js';
 
-export default function Module(): Node {
-  return null;
```
dangerismycat, line 6: I didn't notice this before, but nice work anticipating this future PR! Makes this a whole lot easier.

packages/gestalt/src/moduleUtils.js
```diff
@@ -0,0 +1,40 @@
+// @flow strict
+import React, { type Node } from 'react';
+import Box from './Box.js';
+import Icon from './Icon.js';
+import Text from './Text.js';
+
+// eslint-disable-next-line import/prefer-default-export
+export function renderModuleTitle(
```
dangerismycat, line 8: This isn't really a util function — it's a component. Feel free to move it to something like ModuleTitle.js, and we'll consider options for organizing our file tree in the future as it gets more cluttered.

dangerismycat (commented):
This is looking great!

docs/src/Module.doc.js
ashleyseto: lol @ayeshakmaz and I were debating between the two words. I don't know if we should ask Kelly instead to see if she can weigh in on this quickly since this is harder to change.

docs/src/Module.doc.js
ashleyseto: Design guidelines still need to be written. There are no 'real examples' because no one has this component to use yet.

ashleyseto (approved):

docs/src/Module.doc.js
ayeshakmaz: This is actually done to create <code> blocks out of the prop names (string prop will just show the tics instead)

packages/gestalt/src/Module.js
ayeshakmaz: Not me, all props to @galaxy0101! And I agree, this was really great anticipation

docs/src/Module.doc.js
ayeshakmaz: I updated the examples to be fulling working, and we can add in more design guidance as we do the docs re-write (Ashley already has a v1 that will be iterated on)

packages/gestalt/src/moduleTypes.js
```diff
@@ -0,0 +1,41 @@
+// @flow strict
+import { type Node } from 'react';
+import icons from './icons/index.js';
+
+export type BaseProps = {|
+  id: string,
+  title?: string,
+  icon?: $Keys<typeof icons>,
+  iconAccessibilityLabel?: string,
+  type?: 'error' | 'info',
```
dangerismycat, line 10: This type should be deduped

dangerismycat (approved):
Looks great! Nice work! 🎉

dangerismycat (approved):

PR #1340, opened 2021-01-15
Callout, Upsell: Fix action href type in docs, dedupe type definitions

dangerismycat:
We were showing this type as required when it's actually optional. This PR updates the docs for both Callout and Upsell to reflect the correct type.

I also deduped the common types between the two components, putting them in a `commonTypes.js` file. That file has a lot of potential to become a "kitchen sink" type dumping ground, so we'll need to revisit the name in the future. That more correct name wasn't obvious to me now, so I went with the (temporary) generic name.

### DIFF
--- a/docs/src/Callout.doc.js
+++ b/docs/src/Callout.doc.js
@@ -54,7 +54,7 @@ card(
       {
         name: 'primaryAction',
         type:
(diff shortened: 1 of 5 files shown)


AlbertCarreras:
@dangerismycat can you check Upsell? it might have same issue

AlbertCarreras (approved):

AlbertCarreras:
@dangerismycat Upsell and Callout could potentially share an internal component, cos' they are siblings

PR #1344, opened 2021-01-19
Scrollcontainer (prototype version)

rootvar:
**WIP**

ScrollBox is a new component that can be used to "capture"<Layer />'s and mount React portals in nodes other than document.body.

The main unfinished change is altering logic in Contents.js to reference the correct node instead of window.


AlbertCarreras:
Work continues in https://github.com/pinterest/gestalt/pull/1348 commandeered by @AlbertCarreras

PR #1345, opened 2021-01-21
Docs: Add shared documentation components for docs redesign

ayeshakmaz:
Adds MainSection, MainSection.Subsection, and MainSection.Card components for docs redesign

Supports small and medium sizes (2 vs 3 across), and Do, Don't, or Info styles

**Dos and Don'ts:**
Screen Shot 2021-01-21 at 4 39 36 PM

**Variants Info:** 

Screen Shot 2021-01-21 at 4 39 25 PM

### DIFF
--- a/docs/src/Button.doc.js
+++ b/docs/src/Button.doc.js
@@ -245,31 +245,20 @@ card(
         type="do"
         description="Do place primary buttons to the right or bottom of other button styles."
         defaultCode={`
-      <ButtonGroup>
-        <Button text="Button" inline color="gray"/>
-        <Button text="Button" inline color="red"/>
-      </ButtonGroup>
+  <ButtonGroup>
+    <Button text="Button" inline color="gray"/>
+    <Button text="Button" inline color="red"/>
+  </ButtonGroup>
     `}
       />
       <MainSection.Card
         description="Don't place more than one primary button per surface."
         type="don't"
         defaultCode={`
-      <ButtonGroup>
-        <Button text="Button" inline color="red"/>
-        <Button text="Button" inline color="red"/>
-      </ButtonGroup>
-    `}
-      />
-      <MainSection.Card
-        description="Don't place more than one primary button per surface."
-        type="info"
-        title="Info Example"
-        defaultCode={`
-      <ButtonGroup>
-        <Button text="Button" inline color="red"/>
-        <Button text="Button" inline color="red"/>
-      </ButtonGroup>
+  <ButtonGroup>
+    <Button text="Button" inline color="red"/>
+    <Button text="Button" inline color="red"/>
+  </ButtonGroup>
     `}
       />
     </MainSection.Subsection>
@@ -308,7 +297,7 @@ function Example() {
         <Tooltip text="Submit button">
           <Button
             type="submit"
-            name="satisfaction-questionaire"
+            name="satisfaction-questionnaire"
             text="Submit your response"
             inline
             disabled={disabled}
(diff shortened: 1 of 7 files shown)


ashleyseto:
Is the current PR for example purposes of the doc component only? I want to make sure we're not shipping the updated button docs yet.

@anniemochi probably to take a pass.

Visually looks good to me. Are we thinking of making the containers more responsive?

Another thing to note which is why I was asking about if we're shipping the docs, we do want to pair the code with the new style to the examples so things don't repeat. If we need to look this over more on the design side, let me know.
Screen Shot 2021-01-21 at 9 24 12 AM

I think the containers though work well on this page for doing the examples. The docs will look so much cleaner!

docs/src/Button.doc.js
```diff
@@ -266,7 +297,7 @@ function Example() {
         <Tooltip text="Submit button">
           <Button
             type="submit"
-            name="satisfaction-questionaire"
+            name="satisfaction-questionnaire"
```
dangerismycat, line 300: 😊

docs/src/components/MainSectionCard.js
```diff
@@ -0,0 +1,97 @@
...
+import * as gestalt from 'gestalt'; // eslint-disable-line import/no-namespace
+import DatePicker from 'gestalt-datepicker';
+import React, { type Node } from 'react';
+import { LiveProvider, LiveError, LivePreview } from 'react-live';
+import ExampleCode from './ExampleCode.js';
+import theme from './atomDark.js';
+import Markdown from './Markdown.js';
+
+type Props = {|
+  defaultCode: string,
```
dangerismycat, line 12: nit: unless there's another obvious ordering scheme (like with the card size name map), I'm a big fan of alphabetizing lists of props. we could also add a lint rule for this (we tried that before in Pinboard, but it was…controversial 😂 )

docs/src/components/MainSectionCard.js
```diff
@@ -0,0 +1,97 @@
...
+  const borderStyle =
+    type !== 'info' ? `3px solid ${TYPE_TO_COLOR[type]}` : undefined;
+  return (
+    <Box
+      width={CARD_SIZE_NAME_TO_PIXEL[cardSize]}
+      marginTop={2}
+      marginBottom={8}
+    >
+      <LiveProvider code={code} scope={scope} theme={theme}>
+        <Box>
```
dangerismycat, line 52: do we need this Box, or would a Fragment work? (I have no idea what LiveProvider expects)

docs/src/components/MainSectionCard.js
```diff
@@ -0,0 +1,97 @@
...
+            position="relative"
+            rounding={2}
+          >
+            <LivePreview />
+          </Box>
+        </Box>
+        {showCode && cardSize !== 'sm' && (
+          <ExampleCode code={code} name={title || ''} />
+        )}
+        <Box paddingX={2}>
```
dangerismycat, line 70: nit: I have no idea how to lint for this (I don't think it's possible given the inherent subjectivity), but feel free to use semantic whitespace to break up chunks of markup for readability

docs/src/components/MainSectionSubsection.js
```diff
@@ -0,0 +1,36 @@
...
+  children: Node,
+|};
+
+const MainSectionSubsection = ({
+  children,
+  description,
+  title,
+}: Props): Node => {
+  return (
+    <Box>
```
dangerismycat, line 18: Fragment nit here too

dangerismycat (commented):
From a code perspective, this looks great. I got nothing but nits. I think once we nail down the exact experience we're going for, this is good to go.

docs/src/components/MainSectionCard.js
AlbertCarreras: In Pinboard maybe it was controversial, it'd be great for Gestalt.

docs/src/components/MainSectionCard.js
AlbertCarreras: Probably <> </> works
In Gestalt, we havent implemented the Frament rule that exists in Pinboard, so we still allow <></>

dangerismycat: That's something I'll address when I audit the lint rule differences between Gestalt and Pinboard

ayeshakmaz:
Thanks for the reviews @dangerismycat and @AlbertCarreras, made those updates and hid the Button changes for now 0 let me know if it's good to go :)

docs/src/Button.doc.js
```diff
@@ -5,6 +5,8 @@ import PropTable from './components/PropTable.js';
 import Example from './components/Example.js';
 import Combination from './components/Combination.js';
 import PageHeader from './components/PageHeader.js';
+// To be used when converting Button docs
```
dangerismycat, line 8: Totally understand the desire to not lose this work, but rather than leave the commented-out code here, let's revert the changes on this branch and commit them on a separate branch instead. That way the codebase remains clean in the interim, and the changes in this PR are more clear.

dangerismycat (approved):
Looks great!

PR #1346, opened 2021-01-21
[IN PROGRESS] Upsell: Add form options 

norrise120:
Currently in progress! I would like feedback on my approach for structuring the form prop and adding it to the code. Once I get some feedback here, I'll make design updates and add tests!

### DIFF
--- a/packages/gestalt/src/Upsell.js
+++ b/packages/gestalt/src/Upsell.js
@@ -10,6 +10,7 @@ import IconButton from './IconButton.js';
 import Image from './Image.js';
 import Mask from './Mask.js';
 import Text from './Text.js';
+import TextField from './TextField.js';
 import styles from './Upsell.css';
 import useResponsiveMinWidth from './useResponsiveMinWidth.js';
 import {
@@ -19,8 +20,19 @@ import {
   type DismissButtonType,
 } from './commonTypes.js';
 
+type TextFieldProps = {|
+  id: string,
+  onChange: ({|
+    event: SyntheticInputEvent<HTMLInputElement>,
+    value: string,
+  |}) => void,
+  placeholder?: string,
+  value?: string,
+|};
+
 type Props = {|
   dismissButton?: DismissButtonType,
+  formFields?: $ReadOnlyArray<TextFieldProps>,
   imageData?: {|
     component: Element<typeof Image | typeof Icon>,
     mask?: {|
@@ -30,6 +42,7 @@ type Props = {|
     width?: number,
   |},
   message: string,
+  onFormSubmit?: ActionDataType,
   primaryAction?: ActionDataType,
   secondaryAction?: ActionDataType,
   title?: string,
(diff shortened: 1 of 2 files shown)


packages/gestalt/src/Upsell.js
```diff
@@ -19,8 +20,23 @@ import {
   type DismissButtonType,
 } from './commonTypes.js';
 
+type TextFieldProps = {|
```
norrise120, line 23: Looking for feedback on this and the added UpsellForm!

AlbertCarreras:
HIgh-level discussion here: https://www.loom.com/share/381989997428459e9b260c8aa2805394 @norrise120

dangerismycat:
Totally agree with @AlbertCarreras that primary/secondary don't make sense in this context — he's spot-on about the lack of hierarchy. I'm not sold on the need for something like `Upsell.FormField` though…I _think_ I'd be fine with a `formFields` array prop and an `onFormSubmit` function prop. Sub-components are typically used where more flexibility is needed in the implementation, but we actually don't need much flexibility here: we'll show some form fields (I assume there should be a limit on how many), and we'll show a Submit button next to them. (Would there ever need to be a Cancel button as well?)

What do you think, @norrise120 ?

norrise120:
Thanks for the recommendations! I think you both make very great points. I'll play around with the code a bit and see what works best!

> we actually don't need much flexibility here: we'll show some form fields (I assume there should be a limit on how many), and we'll show a Submit button next to them. (Would there ever need to be a Cancel button as well?)

That's absolutely correct! I think we want to be super limited here in what we allow. Design allows for one or two forms and then one button. Designs currently do not include a cancel button and I don't think at this time we would want to add that functionality.

dangerismycat:
> Design allows for one or two forms and then one button. Designs currently do not include a cancel button and I don't think at this time we would want to add that functionality.

Ok, makes sense. I get the inclination to use an object instead of an array then, though I still agree that the primary/secondary hierarchy doesn't fit. I think we're probably fine using an array that we'll only use the first two items of, provided we communicate that in the docs. I'm not sure if it's possible to lock down the max number of items in an array via Flow, but if nothing else we can add a lint rule in Pinboard to alert devs that they're not going to see all their form fields if they had more than two.

AlbertCarreras:
> Totally agree with @AlbertCarreras that primary/secondary don't make sense in this context — he's spot-on about the lack of hierarchy. I'm not sold on the need for something like `Upsell.FormField` though…I _think_ I'd be fine with a `formFields` array prop and an `onFormSubmit` function prop. Sub-components are typically used where more flexibility is needed in the implementation, but we actually don't need much flexibility here: we'll show some form fields (I assume there should be a limit on how many), and we'll show a Submit button next to them. (Would there ever need to be a Cancel button as well?)
> 
> What do you think, @norrise120 ?

I agree it might be overkilling it. I still think we should internally discuss when we prefer one vs the other

image
vs 
image

norrise120:
@AlbertCarreras @dangerismycat Updated per our discussions! Let me know if this structure looks better or if we should go the Upsell.FormField route! If we're good with this approach, I can work on matching to design and fixing responsiveness.

norrise120:
Because we decided to use a different approach, I'm closing this PR in favor of https://github.com/pinterest/gestalt/pull/1396

PR #1347, opened 2021-01-21
Docs: component source link tweaks

dangerismycat:
A couple of small tweaks to the source link on component pages:
- Update copy to "View source on Github" to make the link more prominent and clear
- Add `target="blank"` to open the target in a new tab

### DIFF
--- a/docs/src/components/PageHeader.js
+++ b/docs/src/components/PageHeader.js
@@ -1,6 +1,6 @@
 // @flow strict
 import React, { type Node } from 'react';
-import { Badge, Box, Text, Tooltip, Heading, Link } from 'gestalt';
+import { Badge, Box, Flex, Heading, Link, Text, Tooltip } from 'gestalt';
 import Markdown from './Markdown.js';
 
 type Props = {|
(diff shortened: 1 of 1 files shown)


docs/src/components/PageHeader.js
```diff
@@ -47,8 +47,8 @@ export default function ComponentHeader({
         </Heading>
         {showSourceLink && (
           <Text color="gray">
-            <Link href={githubUrl(fileName || name)} inline>
-              Source
+            <Link href={githubUrl(fileName || name)} inline target="blank">
+              View source on Github
```
ayeshakmaz, line 51: Should be GitHub

AlbertCarreras: I'm just wondering if spacing out a little bit more the link could mprove the visibility. I'd like to hear @ashleyseto opinion, but I think many people skip this line due to being shadowed by the big header...

ashleyseto: I'm not sure exactly how line height would translate but putting in 4px of spacing in between the title and link seems to balance it out better.

docs/src/components/PageHeader.js
```diff
@@ -12,10 +12,9 @@ type Props = {|
 |};
 
 const gestaltPath = (component) => {
-  if (component === 'DatePicker') {
-    return `packages/gestalt-datepicker/src/${component}.js`;
-  }
-  return `packages/gestalt/src/${component}.js`;
+  const packageName =
+    component === 'DatePicker' ? 'gestalt-datepicker' : 'gestalt';
+  return `packages/${packageName}/src/${component}.js`;
```
dangerismycat, line 17: Unrelated, just DRYing out this logic a bit

ayeshakmaz (approved):
LGTM 👍

PR #1350, opened 2021-01-25
Typeahead: add zIndex prop to support component in Modals with zIndex

AlbertCarreras:
Typeahead didn't include zIndex prop to support component within Modals with existing zIndex. Other components as Dropdown do include zIndex to support these cases.

This PR implements zIndex prop in Typeahead and adds an example to the Docs

image

Without the zIndex, the Typeahead is overlayed by the Modal zIndex.

image

NOTE: This fix allows Typeahead in Modals with zIndex but scrollable Modals will present Flyout positioning issue.  To be fixed with the upcoming ScrollableBox component.

### DIFF
--- a/docs/src/Typeahead.doc.js
+++ b/docs/src/Typeahead.doc.js
@@ -106,6 +106,13 @@ card(
           'Forward the ref to the underlying component container element',
         href: 'refExample',
       },
+      {
+        name: 'zIndex',
+        type: 'interface Indexable { index(): number; }',
+        description:
+          'An object representing the z-index value of the Typeahead list. Only use when Typeahead is used within a parent component that has a z-index set.',
+        href: 'zIndex',
+      },
     ]}
   />
 );
@@ -249,7 +256,7 @@ function Example(props) {
 card(
   <Example
     id="defaultItemExample2"
-    name="Ref Example"
+    name="With Ref"
     defaultCode={`
 function TypeaheadExample() {
   const ref = React.useRef();
@@ -281,7 +288,7 @@ function TypeaheadExample() {
 card(
   <Example
     id="tagsExample"
-    name="Example: Tags"
+    name="With Tags"
     description={`
     You can include [Tag](/Tag) elements in the input using the \`tags\` prop.
 
@@ -350,6 +357,73 @@ function Example(props) {
   );
 }
 `}
+  />
+);
+
+card(
+  <Example
+    id="zIndex"
+    name="With ZIndex"
+    description={`
+    When Typeahead is used within a parent component that has a z-index set, a z-index will also need to be set on the Typeahead. Otherwise the Typeahead will render behind the parent component in the stacking context.`}
+    defaultCode={`
+function Example(props) {
+  const [open, setOpen] = React.useState(false);
+  const [option, setOption] = React.useState();
+
+  const HEADER_ZINDEX = new FixedZIndex(10);
+  const MODAL_ZINDEX = new CompositeZIndex([HEADER_ZINDEX]);
+  const TYPEAHEAD_ZINDEX = new CompositeZIndex([MODAL_ZINDEX]);
+
+  return (
+    <>
+      <Button
+        inline
+        text="Report a stolen account"
+        onClick={() => setOpen(true)}
+      />
+      {open && (
+        <Layer zIndex={MODAL_ZINDEX}>
+          <Modal
+            accessibilityModalLabel="Select the account being reported"
+            onDismiss={() => setOpen(false)}
+            size="sm"
+            heading="Report a stolen account"
+            footer={
+              <Button
+                onClick={() => setOpen(false)}
+                text="Submit"
+                color="red"
+                size="lg"
+                type="submit"
+              />
+            }
+          >
+            <Box padding={8}>
+              <Typeahead
+                zIndex={TYPEAHEAD_ZINDEX}
+                label="Select the account being reported"
+                id="reported_account"
+                noResultText="No Results"
+                options={[
+                  {
+                    label: "basic_user@email.com",
+                    value: "basic_user@email.com"
+                  },
+                  {
+                    label: "business_user@email.com",
+                    value: "business_user@email.com"
+                  },
+                ]}
+                placeholder="Select an account"
+              />
+            </Box>
+          </Modal>
+        </Layer>
+      )}
+    </>
+  );
+}`}
   />
 );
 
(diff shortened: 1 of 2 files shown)


docs/src/Typeahead.doc.js
```diff
@@ -353,4 +360,65 @@ function Example(props) {
   />
 );
 
+card(
+  <Example
+    id="zIndex"
+    name="With ZIndex"
+    description={`
+    Modals within Layer components with zIndexs will overlay any children component. Use zIndex prop in Typeahead to bring the Typeahead list to the top of the zIndex stack.`}
```
dangerismycat, line 368: I think more verbosity could communicate a bit better here. Perhaps something like
"When Typeahead is used within a parent component that has a z-index set, a z-index will also need to be set on the Typeahead. Otherwise the Typeahead will render behind the parent component in the stacking context."

docs/src/Typeahead.doc.js
```diff
@@ -353,4 +360,65 @@ function Example(props) {
...
+    defaultCode={`
+function Example(props) {
+  const [open, setOpen] = React.useState(false);
+  const HEADER_ZINDEX = new FixedZIndex(10);
+  const MODAL_ZINDEX = new CompositeZIndex([HEADER_ZINDEX]);
+  const TYPEAHEAD_ZINDEX = new CompositeZIndex([MODAL_ZINDEX]);
```
dangerismycat, line 374: Should the array passed in here also include HEADER_ZINDEX? Though trivial here, if that array is intended to include all z-indices that TYPEAHEAD_ZINDEX should be above, it seems like we should include the header one as well.

docs/src/Typeahead.doc.js
```diff
@@ -353,4 +360,65 @@ function Example(props) {
...
+        <Button inline text="Report an stolen account" onClick={() => setOpen(true)} />
+        {open && (
+          <Layer zIndex={MODAL_ZINDEX}>
+            <Modal
+              accessibilityModalLabel="Select the account being reported"
+              onDismiss={() => setOpen(false)}
+              size="sm"
+              heading="Report an stolen account"
+              footer={
+                <Box paddingX={1} paddingY={1}>
```
dangerismycat, line 390: nit: `padding={1}`

packages/gestalt/src/Typeahead.js
```diff
@@ -327,6 +330,8 @@ TypeaheadWithForwardRef.propTypes = {
   size: PropTypes.oneOf(['md', 'lg']),
   tags: PropTypes.arrayOf(PropTypes.node),
   value: PropTypes.string,
+  // eslint-disable-next-line react/forbid-prop-types
+  zIndex: PropTypes.any,
```
dangerismycat, line 334: Hm…we don't have a better PropType for this? Perhaps we could add something like
```js
export const UnsafeZIndexPropType = PropTypes.any;
```
in `/src/zIndex.js`, then import that where needed?

docs/src/Typeahead.doc.js
ayeshakmaz: +1 ^

docs/src/Typeahead.doc.js
```diff
@@ -353,4 +360,65 @@ function Example(props) {
...
+  const HEADER_ZINDEX = new FixedZIndex(10);
+  const MODAL_ZINDEX = new CompositeZIndex([HEADER_ZINDEX]);
+  const TYPEAHEAD_ZINDEX = new CompositeZIndex([MODAL_ZINDEX]);
+
+  const [option, setOption] = React.useState();
+
+  return (
+    <Box marginLeft={-1} marginRight={-1}>
+      <Box padding={1}>
+        <Button inline text="Report an stolen account" onClick={() => setOpen(true)} />
```
ayeshakmaz, line 381: Nit: should be "Report a stolen account"

docs/src/Typeahead.doc.js
```diff
@@ -353,4 +360,65 @@ function Example(props) {
...
+        {open && (
+          <Layer zIndex={MODAL_ZINDEX}>
+            <Modal
+              accessibilityModalLabel="Select the account being reported"
+              onDismiss={() => setOpen(false)}
+              size="sm"
+              heading="Report an stolen account"
```
ayeshakmaz, line 388: Same here

docs/src/Typeahead.doc.js
AlbertCarreras: DONE. Thanks, much betta.

docs/src/Typeahead.doc.js
```diff
@@ -106,6 +106,13 @@ card(
           'Forward the ref to the underlying component container element',
         href: 'refExample',
       },
+      {
+        name: 'zIndex',
+        type: 'interface Indexable { index(): number; }',
+        description:
+          'An object representing the zIndex value of the Typeahead list. Only use when Typeahead is overlayed by parent Layer components with zIndex.',
+        href: 'zIndex',
```
ayeshakmaz, line 114: I would update this as well. Maybe: 'An object representing the z-index value of the Typeahead list. Only use when Typeahead is used within a parent component that has a z-index set'

docs/src/Typeahead.doc.js
AlbertCarreras: In this case, it's not necessary as compositeZIndex is adding +1 under the hood, and we have them linked

HEADER_ZINDEX > MODAL_ZINDEX > TYPEAHEAD_ZINDEX

My assumption is that there could be cases where we want to get overlay the highest of different Zindexes and provide that flexibility to the engineer.

I don't think it's the case here.

docs/src/Typeahead.doc.js
AlbertCarreras: nice, I copy pasted this from somewhere else.

AlbertCarreras: DONE

packages/gestalt/src/Typeahead.js
AlbertCarreras: Not at the moment, all proptypes are .any, we can discuss this. out-of-scope for this PR.

docs/src/Typeahead.doc.js
AlbertCarreras: Thanks :) DONE

docs/src/Typeahead.doc.js
AlbertCarreras: DONE

docs/src/Typeahead.doc.js
AlbertCarreras: DONE

dangerismycat (approved):

PR #1353, opened 2021-01-28
Add eslint plugin for gestalt

jennyscript:
This moves over eslint rules we formerly enforced in one repo using gestalt to a new eslint plugin package `eslint-plugin-gestalt` that will be published as an npm package along with `gestalt` and `gestalt-datepicker` through the scripts. This will enable other projects using gestalt to also use these eslint rules if they choose.

### gestalt/button-icon-restrictions

Require specific props when using an icon with Button. Gestalt is more permissive than PDS recommends in adding icons to Buttons. Buttons using iconEnd must use:

- icon "arrow-down"
- color "white"
- size "lg"

### gestalt/no-box-marginleft-marginright

Disallow marginLeft/marginRight on Box. In order to have consistent usage of marginLeft/marginRight on Box in production, we update all of them to marginStart/marginEnd

### gestalt/no-dangerous-style-duplicates

Prevent using dangerouslySetInlineStyle on Box for props that are already directly implemented. Box supports some props already that are not widely known and instead are being implemented with dangerouslySetInlineStyle. This linter checks for usage of already available props as dangerous styles and suggests the alternative.

### gestalt/no-medium-formfields

Disallow medium form fields. In order to have consistent form fields in production, we update all of their sizes to large and disallow medium.

### gestalt/no-role-link-components

Do not allow `role='link'` on `Button`, `TapArea`, and `IconButton` in cases where an alternative with additional functionality must be used instead such as for use with a routing library

### gestalt/prefer-box

Prevent using inline styles on divs that could be gestalt Box props. We prefer using gestalt Box over divs with inline styling to get styling consistency across the app and shared css classes. This linter checks for usage of inline styling that is available as Box props.

## Test Plan

I did a yarn link to repo previously using the eslint rules internally and confirmed the new plugin works as expected.

Screen Shot 2021-01-28 at 11 07 51 AM


.github/workflows/release.yml
```diff
@@ -87,5 +87,5 @@ jobs:
           yarn publish --registry=https://registry.npmjs.org --no-git-tag-version --new-version ${{ steps.release.outputs.VERSION }}
           cd ../gestalt-datepicker
           yarn publish --registry=https://registry.npmjs.org --no-git-tag-version --new-version ${{ steps.release.outputs.VERSION }}
-
-
+          cd ../eslint-plugin-gestalt
+          yarn publish --registry=https://registry.npmjs.org --no-git-tag-version --new-version ${{ steps.release.outputs.VERSION }}
```
christianvuerings, line 91: Very curious wether this will work since we haven't yet released this package. Would be good to keep an eye out as soon as we merge this PR.

packages/eslint-plugin-gestalt/README.md
```diff
@@ -0,0 +1,89 @@
+# eslint-plugin-gestalt
```
christianvuerings, line 1: @jennyscript love this README - should we also include something on our docs site about this package?

.eslintignore
```diff
@@ -4,3 +4,4 @@ flow-typed
 node_modules
 packages/gestalt/dist
 packages/gestalt-datepicker/dist
+packages/eslint-plugin-gestalt
```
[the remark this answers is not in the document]
jennyscript: A lot of the eslint rules don't make sense in here. We have to write the plugin as commonJs and be sure not to use flow. Maybe I should actually set up different eslint rules to enforce on this dir?

.github/workflows/release.yml
jennyscript: Asked about that on slack! I was wondering if I needed to do the first publish manually. It'd be great if this just worked though

christianvuerings: @jennyscript yes - ideally we would use the @pinterest user on npm. Example repo: https://www.npmjs.com/package/@pinterest/graphql-lint-rules and namespace our packages that way.

packages/eslint-plugin-gestalt/README.md
jennyscript: I added a new docs section "Eslint Plugin" with similar info to the readme

.eslintignore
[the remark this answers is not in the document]
jennyscript: I updated this to build with rollup so it can use es6 and flow

christianvuerings (approved):

PR #1355, opened 2021-01-28
Internal: match Prettier config to Pinboard

dangerismycat:
To make the Gestalt codebase seem more familiar to other Pinterest employees, this PR syncs our Prettier config to that in Pinboard.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -61,7 +61,7 @@
     "no-return-await": "off",
     "prettier/prettier": [
       "error",
-      { "trailingComma": "es5", "singleQuote": true }
+      { "singleQuote": true }
     ],
     "react-hooks/exhaustive-deps": "error",
     "react-hooks/rules-of-hooks": "error",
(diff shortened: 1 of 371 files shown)


.eslintrc.json
```diff
@@ -61,7 +61,7 @@
     "no-return-await": "off",
     "prettier/prettier": [
       "error",
-      { "trailingComma": "es5", "singleQuote": true }
```
dangerismycat, line 64: Removing this to prevent the autofixer from fighting itself

.prettierrc.json
```diff
@@ -2,7 +2,15 @@
...
-      "files": "*.js"
+      "files": "*.js",
+      "options": {
+        "arrowParens": "always",
+        "jsxBracketSameLine": false,
+        "printWidth": 100,
+        "quoteProps": "preserve",
+        "tabWidth": 2,
+        "trailingComma": "all"
+      }
```
dangerismycat, line 13: This is the main change. The rest is all prettifying.

ayeshakmaz (approved):
I didn't look at _every_ file, but it looks good! Thanks for updating this!

PR #1356, opened 2021-01-29
Internal: fix package.json version in other packages

christianvuerings:
Depends on #1353 

## Issue

I noticed the version numbers of the `package.json` files got out of sync:
* `gestalt` is at `16.8.4`: https://github.com/pinterest/gestalt/blob/fd691067668b9fb8e8c47b555ef3f25573afcac8/packages/gestalt/package.json#L3
* `gestalt-datepicker` is at `5.0.1`: https://github.com/pinterest/gestalt/blob/fd691067668b9fb8e8c47b555ef3f25573afcac8/packages/gestalt-datepicker/package.json#L3

## Test Plan

After merging this PR, the release step ensures that the `package.json` versions are in sync.

### DIFF
--- a/scripts/releaseSteps.js
+++ b/scripts/releaseSteps.js
@@ -8,7 +8,12 @@ const fsPromises = require('fs').promises;
 const core = require('@actions/core');
 const { getOctokit, context } = require('@actions/github');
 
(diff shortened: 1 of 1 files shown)


dangerismycat (approved):
Thanks for catching this!

PR #1357, opened 2021-01-29
Gestalt: ScrollableContainer - new component with built-in scrollability logic to allow anchored-based components get correctly positioned inside scrolling containers

AlbertCarreras:
# Problem

Layer uses React.Portal to place components outside the DOM hierarchy of the parent component. Layer appends a portal to document.body. 

However, components that depend on anchor components for placement such as 1. Tooltip, 2. Flyout, 3. Dropdown, and 4. Typeahead fail to get correctly positioned and follow anchor within scrolling containers. Under the hood, all of them use the same components for placement and positioning (Layer, Controller, Contents) which 1. append a portal to document.body and 2. take window as positioning reference.

Issue Example: https://codesandbox.io/s/bugissue-oo6vg?file=/example.js
BEFORE/AFTER : https://www.loom.com/share/71f3e523b18e440c8e54a7275550201f
## Gestalt Components involved
- Tooltip
- Flyout
- Dropdown
- Typeahead
- Layer
- Controller (internal)
- Contents (internal)

# Solution

## Solution parts
- ScrollableBoxStore Context: provide 2 callback function to ScrollableContainer and a store of Scrollable Boxes refs to be used by the smart Layer

- ScrollableContainer: a simple <div> container that registers its node ref into the ScrollableBoxStore so that
    - 1. Layer will get matched to and append a portal (containerNode instead of document.body), 
    - 2. Controller and Contents can use the container node to measure positioning (instead of window).

- Smart Layer: a Layer that places a temporary <div> and traverses all its parent nodes until matching a registered parent Scrollable Box. If no Scrollable Box is matched, Layer will remove the temp div and append a portal to document.body. If match is found, Layer will append portal to containerNode.

- Smart Controller & Contents: both components use same smart logic as in Layer to replace window with containerNode in all the positioning calculations for flyouts and carets. 

## Proposal: step by step logic

1. Flyout/Tooltip/Dropwdown/Typeahead are children of scrolling containers (regular Box,  Modal, etc,)
2. Add ScrollableContainer
    1. add optional `overflow='auto'|'scroll'|'scrollX'|'scrollY'` (default: `auto`)
    2. add optional `height='___px'/{___}/'___%'.`  (default: `100%`)
        1. Provider is automatically added to DOM wrapping <div>
        2. Layer detects `ScrollableBoxStore` context and triggers logic to match registered ScrollableContainer
        3. Layer appends portal to registered ScrollableContainer
        4. Flyout/Tooltip/Dropdown/Typeahead conditionally render through portal
        5. Controller/Content detect `ScrollableBoxStore` context and triggers logic to match registered ScrollableContainer
        6. Controller/Content use registered ScrollableContainer to calculate flyout & caret positioning

```
    <ScrollableContainer overflow='scroll' height={100}>
      <Box>
        <Button  ref={ref}/>
        <Layer>
          <Flyout anchor={ref.current} />
        </Layer>
      </Box>
    </ScrollableContainer>

```

## Test Plan
* Is it tested? YES
* Is it accessible? 
* Is it documented? YES
* Have you involved other stakeholders (such as a Pinterest Designer)? YES


packages/gestalt/src/ScrollableContainer.css
```diff
@@ -0,0 +1,3 @@
+.customClass {
```
dangerismycat, line 1: What is this class used for? Is it being used?

packages/gestalt/src/__snapshots__/Controller.jsdom.test.js.snap
```diff
@@ -6,9 +6,9 @@ exports[`Controller renders 1`] = `
     className="container"
     style={
       Object {
-        "left": -10,
```
AlbertCarreras, line 9: These changes are due to implementing `skipNoEdgeCondition`

packages/gestalt/src/__snapshots__/Dropdown.jsdom.test.js.snap
```diff
@@ -9,7 +9,7 @@ exports[`Dropdown renders a menu of 6 items 1`] = `
     <div>
       <div
         class="container"
-        style="stroke: #efefef; visibility: visible; top: -8px; left: -10px;"
```
AlbertCarreras, line 12: These changes are due to implementing `skipNoEdgeCondition`

packages/gestalt/src/ScrollableContainer.js
```diff
@@ -0,0 +1,67 @@
+/**
+ * ScrollableContainer is an optional wrapper component that ensures that flyout-based
```
dangerismycat, line 2: "anchor-based"?

packages/gestalt/src/ScrollableContainer.js
```diff
@@ -0,0 +1,67 @@
...
+function ScrollableContainer({ children, height = '100%', overflow = 'auto' }: Props): Node {
+  const { addRef } = useScrollableContainer();
+  const anchorRef = useRef<?HTMLDivElement>(null);
+
+  useEffect(() => {
+    if (anchorRef.current) {
+      const ref = anchorRef.current;
+      addRef(ref);
+    }
+    return undefined;
```
dangerismycat, line 43: Is this return statement doing anything?

packages/gestalt/src/ScrollableContainer.js
```diff
@@ -0,0 +1,67 @@
...
+    <ScrollableContainer {...passthroughProps} />
+  </ScrollableContainerProvider>
+);
+
+ScrollableContainerWithProvider.displayName = ScrollableContainer;
+
+export default ScrollableContainerWithProvider;
+
+ScrollableContainer.propTypes = {
+  children: PropTypes.node,
```
dangerismycat, line 64: `.isRequired`

packages/gestalt/src/ScrollableContainer.jsdom.test.js
```diff
@@ -0,0 +1,41 @@
...
+      const { addRef, scrollableContainerRef } = useScrollableContainer();
+      return addRef && scrollableContainerRef ? <Box ref={ref} /> : null;
+    };
+
+    render(
+      <ScrollableContainer>
+        <TestBox />
+      </ScrollableContainer>,
+    );
+    expect(ref.current instanceof HTMLDivElement).toEqual(true);
```
dangerismycat, line 22: This does test that the ref points at a div, but it doesn't actually test that it's the div we specified. Perhaps give that Box an id and look for it here?

packages/gestalt/src/ScrollableContainer.jsdom.test.js
```diff
@@ -0,0 +1,41 @@
...
+  });
+
+  it('stores ScrollableContainer node in scrollableContainerRef in context', () => {
```
dangerismycat, line 25: This seems identical to the last test?

packages/gestalt/src/__snapshots__/Controller.jsdom.test.js.snap
```diff
@@ -6,9 +6,9 @@ exports[`Controller renders 1`] = `
...
+        "left": 4,
```
dangerismycat, line 9: Are these snapshot changes expected?

packages/gestalt/src/__snapshots__/Dropdown.jsdom.test.js.snap
```diff
@@ -9,7 +9,7 @@ exports[`Dropdown renders a menu of 6 items 1`] = `
...
+        style="stroke: #efefef; visibility: visible; top: 8px; left: -10px;"
```
dangerismycat, line 12: ditto here?

packages/gestalt/src/__snapshots__/Flyout.jsdom.test.js.snap
```diff
@@ -10,7 +10,7 @@ exports[`Flyout renders 1`] = `
       Object {
         "left": -10,
         "stroke": "#efefef",
-        "top": NaN,
+        "top": 0,
```
dangerismycat, line 13: ditto here, though this almost certainly seems like an improvement 😅

packages/gestalt/src/ScrollableContainer.test.js
```diff
@@ -0,0 +1,25 @@
...
+    expect(tree).toMatchSnapshot();
+  });
+
+  it('ScrollableContainer correctly sets height and overflow props', () => {
+    const tree = create(
+      <ScrollableContainer height={200} overflow="scroll">
+        <Box />
+      </ScrollableContainer>,
+    ).toJSON();
+    expect(tree).toMatchSnapshot();
```
dangerismycat, line 23: Rather than just relying on the snapshot, consider looking specifically at the classes and styles being applied

packages/gestalt/src/boxTypes.js
```diff
@@ -82,6 +82,7 @@ export type Position = 'static' | 'absolute' | 'relative' | 'fixed';
 export type Right = boolean;
 export type Role = string;
 export type Rounding = 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 'circle' | 'pill';
+export type ScrollableOverflow = 'scroll' | 'scrollX' | 'scrollY' | 'auto';
```
dangerismycat, line 85: It doesn't look like this type is used in Box, so this file is an odd location for it. It doesn't look like this is used anywhere outside of ScrollableContainer, so this should just be defined there?

packages/gestalt/src/contexts/ScrollableContainer.js
```diff
@@ -0,0 +1,53 @@
...
+
+function useScrollableContainer(): ScrollableContainerContextType {
+  const scrollableContainerContext = useContext(ScrollableContainerContext);
+  return scrollableContainerContext;
+}
+
+export { ScrollableContainerProvider, useScrollableContainer };
+
+ScrollableContainerProvider.propTypes = {
+  children: PropTypes.node,
```
dangerismycat, line 52: `.isRequired`

packages/gestalt/src/contexts/ScrollableContainer.js
```diff
@@ -0,0 +1,53 @@
...
+
+  return <Provider value={scrollableContainerContext}>{children}</Provider>;
+}
+
+function useScrollableContainer(): ScrollableContainerContextType {
+  const scrollableContainerContext = useContext(ScrollableContainerContext);
+  return scrollableContainerContext;
+}
+
+export { ScrollableContainerProvider, useScrollableContainer };
```
dangerismycat, line 49: Is this syntax gaining us anything over just exporting the functions where defined?

packages/gestalt/src/utils/positioningUtils.js
```diff
@@ -0,0 +1,397 @@
...
+  flyoutSize,
+  idealDirection,
+  triggerRect,
+  windowSize,
+}: {|
+  flyoutSize: Dimensions,
+  idealDirection: FlyoutDir,
+  triggerRect: ClientRect,
+  windowSize: Window,
+|}): 'down' | 'left' | 'right' | 'up' {
```
dangerismycat, line 133: `MainDirections`

dangerismycat (commented):
This is looking great! Thanks for adding all the comments, especially on the legacy code. That knowledge capture is really important.

docs/src/ScrollableContainer.doc.js
```diff
@@ -0,0 +1,465 @@
...
+import PropTable from './components/PropTable.js';
+
+const cards: Array<Node> = [];
+const card = (c) => cards.push(c);
+
+card(
+  <PageHeader
+    name="ScrollableContainer"
+    fileName="ScrollableContainer"
+    description="ScrollableContainer is a simple container with scroll that allows flyout-based components, such as Flyout, Tooltip, Dropdown, and Typeahead, to get correctly positioned inside scrolling containers. To achieve this, under the hood, ScrollableContainer implements a context provider so that children components can access the parent ScrollableContainer node ref rather than the browsers' window and use it for positioning."
```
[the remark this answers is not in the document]
AlbertCarreras: DONE

docs/src/ScrollableContainer.doc.js
```diff
@@ -0,0 +1,465 @@
...
+                  marginTop={-1}
+                  marginBottom={-1}
+                  display="flex"
+                  wrap
+                >
+                  <Box paddingX={1} paddingY={1}>
+                    <Button text="Cancel" size="lg" inline />
+                  </Box>
+                  <Box paddingX={1} paddingY={1}>
+                    <Button text="Save" color="red" size="lg" type="submit" />
```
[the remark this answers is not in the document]
AlbertCarreras: I know... I didn't want to implement all functionality, just the apperance... ha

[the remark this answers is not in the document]
AlbertCarreras: FIXED

docs/src/ScrollableContainer.doc.js
```diff
@@ -0,0 +1,465 @@
+// @flow strict
```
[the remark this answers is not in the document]
AlbertCarreras: FIXED

docs/src/ScrollableContainer.doc.js
```diff
@@ -0,0 +1,465 @@
...
+                  justifyContent="end"
+                  marginStart={-1}
+                  marginRight={-1}
+                  marginTop={-1}
+                  marginBottom={-1}
+                  display="flex"
+                  wrap
+                >
+                  <Box paddingX={1} paddingY={1}>
+                    <Button text="Cancel" size="lg" inline />
```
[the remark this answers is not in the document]
AlbertCarreras: FIXED

packages/gestalt/src/Contents.js
```diff
@@ -431,51 +162,51 @@ class WrappedContents extends Component<Props, State> {
...
+    const containerNode = getContainerNode({ scrollableContainerRef, initialPositionRef: anchor });
+    const containerBoundingClientRect = containerNode?.getBoundingClientRect();
+
+    // If there's a parent ScrollableContainer, replace window's dimensions and scroll with the ScrollableContainer node ones
     const windowSize = {
-      height: window.innerHeight,
-      width: window.innerWidth,
-      scrollX,
-      scrollY,
+      height: containerBoundingClientRect?.height || window.innerHeight,
```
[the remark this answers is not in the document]
AlbertCarreras: FIXED.

packages/gestalt/src/boxTypes.js
AlbertCarreras: DONE

packages/gestalt/src/contexts/ScrollableContainer.js
AlbertCarreras: DONE

packages/gestalt/src/contexts/ScrollableContainer.js
AlbertCarreras: DONE

packages/gestalt/src/utils/positioningUtils.js
AlbertCarreras: good catch. DONE

packages/gestalt/src/ScrollableContainer.test.js
AlbertCarreras: All tests rely on the snapshot... we might want to discuss this.

packages/gestalt/src/ScrollableContainer.css
AlbertCarreras: REMOVED

packages/gestalt/src/ScrollableContainer.js
AlbertCarreras: it was. not now.

REMOVED

packages/gestalt/src/__snapshots__/Controller.jsdom.test.js.snap
AlbertCarreras: They were, but you made me rethink about them. I just made a change that will prevent them, and make the implementation even safer in Pinboard. They are related to the new constant `skipNoEdgeCondition` that was needed for ScrollableContainer (flyout could not work as expected if placed in the top/left edge of the container without margins), but I made it conditional to having a ScrollableContainer as well.

packages/gestalt/src/ScrollableContainer.js
AlbertCarreras: ADDRESSED

docs/src/ScrollableContainer.doc.js
```diff
@@ -0,0 +1,593 @@
...
+
+card(
+  <PageHeader
+    name="ScrollableContainer"
+    description="ScrollableContainer is used with anchored, hovering components such Flyout, Tooltip, Dropdown, or Typeahead. This is needed for proper positioning when the anchor component could scroll within the viewport."
+  />,
+);
+
+card(
+  <MainSection name="Examples">
```
ayeshakmaz, line 18: I would move props above this, and rename this to Variants. I think for some of these more primitive components, we can skip the top-most example

docs/src/ScrollableContainer.doc.js
```diff
@@ -0,0 +1,593 @@
...
+    <MainSection.Subsection
+      description={`The following examples are all built using the \`ScrollableContainer\` component.`}
+    >
+      <MainSection.Card
+        title="Flyout within ScrollableContainer"
```
ayeshakmaz, line 23: After moving this to become variants, I would recommend putting the title-"Flyout within ScrollableContainer" on the Subsections, not the Card

ayeshakmaz: Same would apply for the rest as well - put titles on subsections, not the full-width card

docs/src/ScrollableContainer.doc.js
AlbertCarreras: Agreed.

AlbertCarreras: DONE

docs/src/ScrollableContainer.doc.js
AlbertCarreras: DONE

packages/gestalt/src/__snapshots__/Controller.jsdom.test.js.snap
AlbertCarreras: FIXED

packages/gestalt/src/__snapshots__/Dropdown.jsdom.test.js.snap
AlbertCarreras: FIXED

packages/gestalt/src/ScrollableContainer.jsdom.test.js
AlbertCarreras: kinda, merged into single test.

one is making sure children renders, the other one checks ref from context was received.

AlbertCarreras: ADDRESSED IN LAST COMMIT

packages/gestalt/src/ScrollableContainer.jsdom.test.js
AlbertCarreras: ADDRESSED IN LAST COMMIT

dangerismycat (approved):
This looks great!

PR #1358, opened 2021-01-29
Box: Add design guidelines and update documentation

ayeshakmaz:
Update Box docs to new style

## Test Plan

Lots of manual testing and feedback gathering


docs/src/Box.doc.js
```diff
@@ -1,24 +1,157 @@
...
 const cards: Array<Node> = [];
 const card = (c) => cards.push(c);
 
 card(
   <PageHeader
     name="Box"
-    description={`In the darkest night, Box will rise to bring the light. The Lloyd has spoken.
-
-&mdash; Anon _(Winning Box Haiku, 2017)_`}
+    description="Box is a low level component that can be used to build the foundation of pretty much anything. Using Box allows you to focus on the important content, without worrying about the pixel details."
```
AlbertCarreras, line 18: I agree the previous description was... a joke! I still feel we don't describe Box to its essence which is a div. 

I have mix feelings about: `Using Box allows you to focus on the important content, without worrying about the pixel details.` It's describing the intention, but not describing it.

docs/src/Flex.doc.js
```diff
@@ -57,7 +57,6 @@ card(
         description:
           'Establishes the main axis, thus defining the direction flex items are placed in the flex container.',
       },
-      { name: 'fit', type: 'boolean', defaultValue: false },
```
ayeshakmaz, line 60: Think this was accidentally listed in Flex

docs/src/components/CombinationNew.js
```diff
@@ -0,0 +1,66 @@
...
+    case 'boolean':
+      return (value && key).toString();
+    case 'string':
+      return `${key}=${JSON.stringify(value)}`;
+    default:
+      return `${key}={${JSON.stringify(value)}}`;
+  }
+};
+
+export default function CombinationNew({ children, shaded, ...props }: Props): Node {
```
ayeshakmaz, line 50: I don't have a great name for this component (eventually it will replace all uses of Combination)

docs/src/Box.doc.js
```diff
@@ -480,143 +540,235 @@ card(
...
+      </CombinationNew>
+    </MainSection.Subsection>
+    <MainSection.Subsection description="" title="Opacity">
+      <CombinationNew opacity={[0, 0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9, 1]}>
+        {(props) => <Box color="darkGray" width={60} height={60} {...props} />}
+      </CombinationNew>
+    </MainSection.Subsection>
+    <MainSection.Subsection
+      description={`
+    Box uses Flexbox to achieve layout. If you're new to flex layout, please read the excellent [CSS-Tricks guide to flexbox](https://css-tricks.com/snippets/css/a-guide-to-flexbox/).
```
dangerismycat, line 680: Similar question here about if we should include flex stuff in the Box docs or point people to Flex instead

docs/src/Box.doc.js
```diff
@@ -1,24 +1,154 @@
...
-
-&mdash; Anon _(Winning Box Haiku, 2017)_`}
+    description="Box is a low level component that can be used to build the foundation of pretty much anything. Using Box allows you to focus on the important content, without worrying about the pixel details."
   />,
 );
 
+card(
+  <MainSection name="Examples" showHeading={false}>
+    <MainSection.Subsection
+      description={`The following examples are all built using the \`Box\` component and its various properties. `}
```
dangerismycat, line 22: nit: s/properties/props (since `props` has a specific meaning in the context of React components)

docs/src/Box.doc.js
```diff
@@ -1,24 +1,154 @@
...
+  </Box>
+  <Box
+    alignItems="center"
+    direction="row"
+    display="flex"
+    marginStart={-1}
+    marginEnd={-1}
+  >
+    <Box paddingX={1}>
+      <Avatar name="chrislloyd" size="md" />
```
dangerismycat, line 60: nit: use a different name so this seems more intentional

docs/src/Box.doc.js
```diff
@@ -480,143 +540,235 @@ card(
...
-    Sets a border radius for the Box. Options are "circle" or "pill" for fully rounded corners or 0-8 representing the radius in 4px increments.
-  `}
-    id="rounding"
-    name="Rounding"
-  >
-    <Combination
-      id="rounding"
-      rounding={['pill', 'circle', 0, 1, 2, 3, 4, 5, 6, 7, 8]}
-      showHeading={false}
+    The SearchField uses absolute positioning to position the search Icon at the beginning of the input.
```
dangerismycat, line 743: Same question here: it seems like we're trying to teach people how to create their own custom versions of our components

docs/src/Box.doc.js
```diff
@@ -480,143 +540,235 @@ card(
...
-    name="Example: ref"
-    description={`
-    A \`Box\` with an anchor ref to a Flyout component
-  `}
-    defaultCode={`
-function ButtonFlyoutExample() {
+      <MainSection.Card
+        cardSize="lg"
+        defaultCode={`
+function BoxFlyoutExample() {
```
dangerismycat, line 771: nit: it'd be nice to be able to close the Flyout

docs/src/Box.doc.js
```diff
@@ -651,23 +803,38 @@ function ButtonFlyoutExample() {
...
-    description={`
-It's possible to use box with external elements using the css \`z-index\` property by capturing those values in controlled objects. The example below shows using a \`FixedZIndex\` for a value that comes from somewhere else, and a \`CompositeZIndex\` to layer the Box on top of it.
-  `}
-    id="zindex"
-    name="ZIndex"
-    defaultCode={`
+      />
+    </MainSection.Subsection>
+    <MainSection.Subsection
+      description={`It's possible to use Box with external elements using the css \`z-index\` property by capturing those values in controlled objects. The example below shows using a \`FixedZIndex\` for a value that comes from somewhere else, and a \`CompositeZIndex\` to layer the Box on top of it.`}
```
dangerismycat, line 809: We'll need to expand on this a lot more in the zIndex docs. I think our approach is pretty confusing.

docs/src/Box.doc.js
```diff
@@ -651,23 +803,38 @@ function ButtonFlyoutExample() {
...
+  </MainSection>,
+);
+
+card(
+  <MainSection
+    name="Related"
+    description={`
+      [Flex](/Flex)
+
+      - Use Flex and Flex.Item for flex layouts when you don't need the extra functionality in Box, and could benefit from the ability to supply a gap between items.
```
dangerismycat, line 832: I think this is underselling the usefulness of `gap`? Also we could mention the benefits of better separating layout from other concerns that Box addresses

docs/src/Box.doc.js
```diff
@@ -651,23 +803,38 @@ function ButtonFlyoutExample() {
...
+
+      [Container](/Container)
+
+      - Use Container to responsively layout content at different screen sizes.
```
dangerismycat, line 836: This doesn't really say much…it's fine for now, but we should expand on this when we get to the Container docs

docs/src/Box.doc.js
```diff
@@ -651,23 +803,38 @@ function ButtonFlyoutExample() {
...
 }
 `}
+      />
+    </MainSection.Subsection>
+  </MainSection>,
+);
+
+card(
+  <MainSection
+    name="Related"
```
dangerismycat, line 828: This should probably also include TapArea, Sticky, Card, Column, Layer, Letterbox, Mask

docs/src/Flex.doc.js
dangerismycat: From a previous version I think. Good catch!

docs/src/components/CombinationNew.js
```diff
@@ -0,0 +1,66 @@
+// @flow strict
+import { Flex } from 'gestalt';
+import React, { type Node } from 'react';
+import MainSectionCard from './MainSectionCard.js';
+
+type Props = {
```
dangerismycat, line 6: Can we close this type?

docs/src/components/CombinationNew.js
dangerismycat: this works for now! or something like CombinationV2 — but it really doesn't matter since we'll delete Combination and rename this component in the near future

docs/src/components/MainSection.js
```diff
@@ -5,14 +5,15 @@ import Card from './Card.js';
 import MainSectionSubsection from './MainSectionSubsection.js';
 
 type Props = {|
-  children: Node,
+  children?: Node,
```
dangerismycat, line 8: this component won't always have children?

docs/src/components/MainSectionCard.js
```diff
@@ -9,18 +9,20 @@ import theme from './atomDark.js';
 import Markdown from './Markdown.js';
 
 type Props = {|
-  cardSize?: 'sm' | 'md',
-  defaultCode: string,
+  cardSize?: 'sm' | 'md' | 'lg',
+  children?: Node,
```
dangerismycat, line 13: ditto here

dangerismycat (commented):
I can't believe you started with Box 😂 so much work! Nice job!

The layout is a bit confusing…I'm having a hard time telling which header/content goes with which example. I think we need to space things out a bit more.

We're also currently in a "worst of both worlds" with the props table positioning. It's not immediately visible at the top, but it's also not at the bottom, it's…somewhere in the middle. That is going to frustrate a lot of people.

Not sure if this is already planned for future work, but I think we can compress the combination examples a bit more. I get the need for whitespace, but these are displayed pretty inefficiently right now. (But the text doesn't overlap! 😍 )

docs/src/BOILERPLATE.doc.js
```diff
@@ -0,0 +1,189 @@
...
+          'maroon',
+          'watermelon',
+          'orange',
+          'transparent',
+          'transparentDarkGray',
+          'lightWash',
+          'darkWash',
+        ]}
+      >
+        {(props) => <Box width={60} height={60} rounding="circle" {...props} />}
```
AlbertCarreras, line 166: can we avoid spreading props? 

We don't allow it (Eslint rule) in Pinboard as a best practice. Codemods can't read them.

docs/src/Box.doc.js
```diff
@@ -1,24 +1,154 @@
...
+
+           In the double-sided example we use \`flex="grow"\` to mark a flex child as something that can expand. Try removing the \`grow\` property and see what happens.
+
+        `}
+        cardSize="lg"
+        defaultCode={`
+<Box column={10}>
+  <Box
+    alignItems="center"
+    direction="row"
```
AlbertCarreras, line 37: `direction="row"`
 is the default, can we remove?

docs/src/Box.doc.js
```diff
@@ -1,24 +1,154 @@
...
+    display="flex"
+    width={300}
+    marginStart={-1}
+    marginEnd={-1}
+    paddingY={2}
+  >
+    <Box paddingX={1}>
+      <Avatar name="chrislloyd" size="md" />
```
AlbertCarreras, line 45: Can we change example names, and maybe add another follow button as well?

AlbertCarreras:
image

docs/src/Box.doc.js
```diff
@@ -1,24 +1,154 @@
...
 
+card(
+  <MainSection name="Examples" showHeading={false}>
+    <MainSection.Subsection
+      description={`The following examples are all built using the \`Box\` component and its various properties. `}
+    >
+      <MainSection.Card
+        title="Media Object"
+        description={`
+           The [media object](http://www.stubbornella.org/content/2010/06/25/the-media-object-saves-hundreds-of-lines-of-code/) is a common pattern for displaying data. What's interesting about this example is the use of \`flex\` to align the items. If you try changing the size of the \`Avatar\` or the number of lines of \`Text\`, both will stay aligned because they are center aligned.
```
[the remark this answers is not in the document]
AlbertCarreras: Agree, Maybe a do/don't or best practice should be Box display flex vs Flex component
Maybe a note should be maybe pointing to the Flex component

docs/src/Box.doc.js
```diff
@@ -1,24 +1,154 @@
...
+
+           In the double-sided example we use \`flex="grow"\` to mark a flex child as something that can expand. Try removing the \`grow\` property and see what happens.
```
AlbertCarreras, line 29: This is a writer question... what's the tone of out text? 
`Try removing the \`grow\` property and see what happens.` We should discuss the tone of the Docs, is it a friendly voice that guide's you in the discovery of the Docs or just a standard Doc voice that simply describes?

I thnk it's important to get to the voice, maybe in a second round, but the sooner the less we'll have to change of wor already done.

docs/src/Box.doc.js
```diff
@@ -1,24 +1,154 @@
...
+      <Button text="Follow" size="sm" color="red" />
+    </Box>
+  </Box>
+</Box>
+`}
+      />
+      <MainSection.Card
+        cardSize="md"
+        title="Expandable Module"
+        description="Underneath the hood, [Expandable Module](/Module) uses Box to create the styling of the rounded edges and shadow, while also ensuring the title can grow while keeping the icon aligned to the end."
```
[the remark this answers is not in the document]
AlbertCarreras: +1 I wouldn't use Module here, we should do that in Module's docs.

docs/src/Box.doc.js
```diff
@@ -1,24 +1,154 @@
...
+            alt="Botanical art in coral and green"
+            src="https://i.ibb.co/7bQQYkX/stock2.jpg"
+            fit="contain"
+            naturalWidth={1}
+            naturalHeight={1}
+          />
+        </Box>
+      }
+      name="favorite art"
+      onChange={({ checked }) => {
```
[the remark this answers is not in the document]
AlbertCarreras: can we remove the console.log?  and just a () => {}

docs/src/Box.doc.js
```diff
@@ -1,24 +1,154 @@
...
+  <Box padding={2}>
+    <Checkbox
+      checked={false}
+      id="coral"
+      label="Coral"
+      subtext="Botanical art in coral and green"
+      image={
+        <Box height={100} width={80}>
+          <Image
+            alt="Botanical art in coral and green"
```
AlbertCarreras, line 108: can we remove the spaces in here that push the indentation? 
image

docs/src/Box.doc.js
```diff
@@ -1,24 +1,154 @@
...
+        icon: 'lock',
+      },
+    ]}
+  />
+`}
+      />
+      <MainSection.Card
+        cardSize="md"
+        title="Stacked Checkboxes"
+        description="In this example, Box is used to create spacing between [Checkbox](/Checkbox) items, as well as set the height of the image in each item."
```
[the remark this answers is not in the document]
AlbertCarreras: agree

docs/src/BOILERPLATE.doc.js
```diff
@@ -0,0 +1,189 @@
+// @flow strict
```
AlbertCarreras, line 1: Wjy don't we add this to? `scripts/templates/ComponentName.doc.js`

docs/src/components/CombinationNew.js
```diff
@@ -0,0 +1,66 @@
...
+  children: (Object, number) => Node, // flowlint-line unclear-type:off
```
AlbertCarreras, line 7: children: ({[string]: string}, number) can we try this?

docs/src/components/MainSectionSubsection.js
```diff
@@ -10,15 +11,37 @@ type Props = {|
...
+          }}
+          id={slugifiedId}
+          data-anchor
+          paddingY={2}
+        >
+          <Flex alignItems="baseline" gap={2}>
+            <Heading size="sm">{title}</Heading>
+            <IconButton
+              dangerouslySetSvgPath={{
+                __path:
```
AlbertCarreras, line 32: are we reusing this link icon somewhere else? could we create a const and export if so? Maybe it's just used here...

AlbertCarreras (approved):
There are a few nits you can address before merging, then there are other major comments re Box content & related or even some design issues that I think are out of scope for this PR.

docs/src/Box.doc.js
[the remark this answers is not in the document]
ayeshakmaz: Oo I like that idea!

docs/src/Box.doc.js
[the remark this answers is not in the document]
ayeshakmaz: Maybe we need a ticket in Jira to audit components and convert them to using Flex?

docs/src/Box.doc.js
ayeshakmaz: I think we should limit Related to 3 or 4 options, so we don't totally overwhelm people. It should be the 3-4 closest matches, so maybe TapArea and Sticky? Both things that are more utility components, and not presentational

docs/src/components/CombinationNew.js
ayeshakmaz: Given how this component works, I don't actually think we can. The extra props added get used to render the combinations

docs/src/components/MainSection.js
ayeshakmaz: No, sometimes subsections are only titles and descriptions

docs/src/components/MainSectionCard.js
ayeshakmaz: Same here, sometimes you can render easily with defaultCode, other times you need to pass in children (CombinationNew)

AlbertCarreras:
New comments:

- Props table: props not properly linked: alignItems, column, dangerouslySetInlineStyle (maybe linked to best practices? dangerlusly vs div? I know do/don't mentions it, maybe an example could help make it more explicit), justifyContent, margin (to auto margin?), 
-for those Flex-related props, should we link them to 'Related'?

General observations:
- fit doesn't have any description not example.
- role doesn't have any description not example
- userSelect could have an example, it's pretty abstract as per the description
- we have responsive 1. columns, 2. margin/marginStart/marginEnd/marginTop/marginBottom,  3. padding/paddingX/paddingY. However, there's only one variant for responsive padding? Should we document all cases?

docs/src/Box.doc.js
```diff
@@ -225,390 +217,418 @@ card(
...
-      'transparent',
-      'transparentDarkGray',
-      'lightWash',
-      'darkWash',
-    ]}
-  >
-    {(props) => <Box width={60} height={60} rounding="circle" {...props} />}
-  </Combination>,
-);
+</Box>
```
AlbertCarreras, line 587: kill line 586

docs/src/Box.doc.js
```diff
@@ -638,23 +658,47 @@ function ButtonFlyoutExample() {
...
+
+      - Use Flex and Flex.Item for flex layouts when you want to separate layout concerns from the extra functionality in Box. Flex is also extremely beneficial when trying to arrange items with consistent, repetitive gutters using the \`gap\` property.
+
+      [Container](/Container)
+
+      - Use Container to responsively layout content at different screen sizes.
+
+      [TapArea](/TapArea)
+
+      - If a tap target is needed in order to click on a portion of the page, use TapArea, since onClick is not supported on Box.
```
AlbertCarreras, line 695: onCLick should be type font?

docs/src/Flex.doc.js
```diff
@@ -13,7 +14,7 @@ card(
   <PageHeader
     name="Flex"
     description={`
-      Flex is a layout component with a very limited subset of the props available to Box and a few special props of its own.
+      Flex is a layout component with a very limited subset of the props available to [Box](/Box) and a few special props of its own.
```
AlbertCarreras, line 17: @dangerismycat the only few special props is gap, right? should we just be direct with that?

And then say somethinglike 

Use this component for flex layouts, especially when even spacing between elements is desired by using the 'gap' property.

docs/src/Flex.doc.js
```diff
@@ -182,30 +183,27 @@ card(
 );
 
 card(
-  <Combination
-    description={`
-    Flex is strictly for flex layouts. If you're new to flex layout, please read the excellent [CSS-Tricks guide to flexbox](https://css-tricks.com/snippets/css/a-guide-to-flexbox/).
+  <MainSection name="Variants">
```
AlbertCarreras, line 186: It looks great! thanks

docs/src/Flex.doc.js
dangerismycat: Yeah, that's pretty much it. And that copy sounds good to me 👍

docs/src/components/MainSectionSubsection.js
ayeshakmaz: Will address this separately

docs/src/Flex.doc.js
ayeshakmaz: Done :)

docs/src/Box.doc.js
```diff
@@ -225,390 +247,487 @@ card(
...
-  <Box paddingX={1}>
-    <Text weight="bold">Chris Lloyd</Text>
-    <Text>joined 2 years ago</Text>
+  <MainSection name="Best practices">
+    <MainSection.Subsection>
+      <MainSection.Card
+        cardSize="md"
+        type="do"
+        description={`
+        Use Box as a building block when creating other components or layouts that do not rely on Flexbox. The included properties should cover any variations needed to create a diverse range of options.
```
dangerismycat, line 256: nit: "flexbox" shouldn't be capitalized, per MDN: https://developer.mozilla.org/en-US/docs/Learn/CSS/CSS_layout/Flexbox#introducing_a_simple_example

docs/src/Box.doc.js
```diff
@@ -225,390 +247,487 @@ card(
...
+    </Box>
   </Box>
 </Box>
 `}
-  />,
+      />
+      <MainSection.Card
+        cardSize="md"
+        type="don't"
+        description={`Avoid using arbitrary \`<div>\` elements. Instead, when building a component, prioritize using Box. If you need to set a custom style, you can do so using the \`dangerouslySetInlineStyle\` prop. However, this should be avoided whenever possible by utilizing the other props provided in Box.`}
```
dangerismycat, line 357: You may also want to mention / link to our lint rule for this

docs/src/Box.doc.js
dangerismycat: it would still be nice to be able to close the Flyout 😅

docs/src/Box.doc.js
```diff
@@ -638,23 +757,51 @@ function ButtonFlyoutExample() {
...
+      title="Z-Index"
+    >
+      <MainSection.Card
+        cardSize="lg"
+        defaultCode={`
 function Example() {
   const HEADER_ZINDEX = new FixedZIndex(100);
   const zIndex = new CompositeZIndex([HEADER_ZINDEX]);
-  return <Box color="blue" width={60} height={60} zIndex={zIndex} />
+  return <Box color="midnight" width={60} height={60} zIndex={zIndex} />
```
dangerismycat, line 772: I think this would be better demonstrated with two (or more) Boxes with various z-indices? A single Box doesn't convey the layering aspect

docs/src/Box.doc.js
dangerismycat: I agree in general but think it's fine to break that rule for Box, since it's so foundational. I don't think it's really clear when a user would want Card, Column, or Letterbox

(this can def be addressed in a future PR)

docs/src/Flex.doc.js
dangerismycat, line 17: I think this is still underselling the usefulness of `gap`, but we can revisit later

docs/src/components/MainSectionSubsection.js
```diff
@@ -35,16 +40,19 @@ const MainSectionSubsection = ({ children, description, title }: Props): Node =>
...
           </Flex>
         </Box>
       )}
-      {description && (
-        <Box width="80%" marginTop={-3}>
-          <Markdown text={description} />
-        </Box>
-      )}
+
+      <Box width="80%" marginTop={-2} marginBottom={6}>
```
dangerismycat, line 52: Do we want to render this empty Box if there is no description? I would think "no" in general

ayeshakmaz: We actually do in this case to account for a funky spacing thing

docs/src/Flex.doc.js
ayeshakmaz: Yeah since this is the Flex docs, we can address it when we re-write these

docs/src/Box.doc.js
ayeshakmaz: good call

docs/src/components/MainSectionSubsection.js
dangerismycat: That doesn't sound like the right solution. Can that be solved by using flex properties on the parent/siblings?

docs/src/Box.doc.js
```diff
@@ -1,20 +1,18 @@
...
 
 const cards: Array<Node> = [];
 const card = (c) => cards.push(c);
 
 card(
   <PageHeader
     name="Box"
-    description={`In the darkest night, Box will rise to bring the light. The Lloyd has spoken.
-&mdash; Anon _(Winning Box Haiku, 2017)_`}
+    description="Box is a component primitive that can be used to build the foundation of pretty much any other component. Using Box allows users to focus on the content, while keeping the pixel details, like spacing, borders, or colors, automatically consistent with the rest of Gestalt."
```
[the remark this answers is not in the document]
ayeshakmaz: Love it!

docs/src/Box.doc.js
ayeshakmaz: Done :D

docs/src/components/MainSectionSubsection.js
ayeshakmaz: Wrapped it in a Box instead since Flex doesn't have margin

PR #1359, opened 2021-02-01
Docs: Update shared components for new Docs design

ayeshakmaz:
PR to update the shared components used for the new design

Also updates the template for docs pages


docs/src/components/CombinationNew.js
```diff
@@ -0,0 +1,66 @@
+// @flow strict
+import { Flex } from 'gestalt';
+import React, { type Node } from 'react';
+import MainSectionCard from './MainSectionCard.js';
+
+type Props = {
+  children: (Object, number) => Node, // flowlint-line unclear-type:off
+  shaded?: boolean,
+  ...
```
dangerismycat, line 9: I have some ideas for how we could close this type and tighten up the passed-through types, but that shouldn't be a blocker here. This is totally fine for now.

docs/src/components/CombinationNew.js
```diff
@@ -0,0 +1,66 @@
+// @flow strict
+import { Flex } from 'gestalt';
+import React, { type Node } from 'react';
+import MainSectionCard from './MainSectionCard.js';
+
+type Props = {
+  children: (Object, number) => Node, // flowlint-line unclear-type:off
```
dangerismycat, line 7: We can improve this type (`Object` is to be avoided, and will be removed in future Flow versions) but that can be handled repo-wide in a separate PR, not important here

docs/src/components/MainSectionSubsection.js
```diff
@@ -10,15 +11,37 @@ type Props = {|
...
+          id={slugifiedId}
+          data-anchor
+          paddingY={2}
+        >
+          <Flex alignItems="baseline" gap={2}>
+            <Heading size="sm">{title}</Heading>
+            <IconButton
+              dangerouslySetSvgPath={{
+                __path:
+                  'M21.001 7.241l-4.053 4.052-1.06-1.06.672-.672a1.5 1.5 0 10-2.121-2.121l-.671.672-1.061-1.06L16.759 3l4.242 4.241zm-9.708 9.708l-4.052 4.052-4.242-4.241 4.053-4.053 1.059 1.06-.671.672a1.5 1.5 0 002.121 2.121l.671-.672 1.061 1.061zM14.639.879l-4.053 4.052a3 3 0 000 4.242l1.061 1.06-1.415 1.414-1.06-1.061a3 3 0 00-4.241 0L.879 14.638a2.998 2.998 0 000 4.242l4.241 4.242a3 3 0 004.241 0l4.053-4.052a3 3 0 000-4.242l-1.06-1.061 1.414-1.413 1.06 1.06a3 3 0 004.241 0l4.052-4.052a2.998 2.998 0 000-4.242L18.88.879a2.997 2.997 0 00-4.241 0z',
```
dangerismycat, line 33: In the future we can revisit this and possibly add this icon to our library, I think this is fine for now

AlbertCarreras (approved):

scripts/templates/ComponentName.doc.js
```diff
@@ -1,35 +1,167 @@
...
-import Example from './components/Example.js';
+import CombinationNew from './components/CombinationNew.js';
 import PageHeader from './components/PageHeader.js';
+import MainSection from './components/MainSection.js';
 
 const cards: Array<Node> = [];
 const card = (c) => cards.push(c);
 
-card(<PageHeader name="ComponentName" description="ComponentName allows users to ..." />);
+card(<PageHeader name="Component Name" description="Brief description of this cmp" />);
```
dangerismycat, line 12: nit: s/cmp/component for more clear communication (here and elsewhere)

scripts/templates/ComponentName.doc.js
```diff
@@ -1,35 +1,167 @@
...
+        description="Description about what you should Do"
+        defaultCode={`
+Code for this example goes here
+`}
+      />
+      <MainSection.Card
+        cardSize="md"
+        type="don't"
+        description={`
+          What not to do goes \`here\`. Using backticks instead of quotes allows you to use [Markdown]("https://www.markdownguide.org/")
```
dangerismycat, line 62: Can you add this note about backticks to all fields where it's relevant? Or we could add a note at the top of the file if it's going to be too repetitive.

EDIT: oh, I think you did already. Does it also need to be added to the "do" description above?

docs/src/components/CombinationNew.js
dangerismycat: Although a small change that would improve readability here would be to name the arguments (and we might as well fix the props type enough to obviate the Flow suppression):
```js
children: (props: { [key: string]: any, ... }, index?: number) => Node,
```

dangerismycat (commented):
This looks awesome!

scripts/templates/ComponentName.doc.js
ayeshakmaz: Yes, good catch!

dangerismycat (approved):
Looks great!! :shipit:

PR #1360, opened 2021-02-02
Docs: copyPaste functionality implemented into linking icon

AlbertCarreras:
image
Linking icon scrolls to top, hashes url in browser, and (NEW!) now it also copies new hashed url for quick sharing.

### DIFF
--- a/docs/src/components/Card.js
+++ b/docs/src/components/Card.js
@@ -15,6 +15,22 @@ type Props = {|
   showHeading?: boolean,
 |};
 
+export const copyToClipboard = (hash: string): boolean => {
+  if (!navigator.clipboard) {
+    // Clipboard API not available
+    return false;
+  }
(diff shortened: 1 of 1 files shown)


docs/src/components/Card.js
```diff
@@ -15,6 +15,22 @@ type Props = {|
   showHeading?: boolean,
 |};
 
+export const copyPaste = (hash: string): boolean => {
```
dangerismycat, line 18: There isn't any pasting happening, so a better name would probably be something like "copyToClipboard"?

dangerismycat (approved):

docs/src/components/Card.js
AlbertCarreras: indeed... DONE

dangerismycat (approved):

PR #1361, opened 2021-02-03
Internal: Add lint rules to enforce fragment style and use of keys

dangerismycat:
To reduce context switching with Pinboard (and allow for one style even where keys are needed), this PR adds a lint rule to enforce the longhand (`<Fragment>`) syntax for React fragments. It also includes a lint rule to ensure we're using keys where needed.

For reference:
https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-fragments.md
https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-key.md

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -67,6 +67,8 @@
     "react-hooks/rules-of-hooks": "error",
     "react/destructuring-assignment": "off",
     "react/jsx-filename-extension": "off",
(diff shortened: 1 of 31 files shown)


AlbertCarreras (approved):
LGTM, thanks for taking this one.

PR #1362, opened 2021-02-04
Box: fix margin-family props not overriding default 'auto' by @media  with boints

AlbertCarreras:
Bug: On Gestalt <Box /> component, the media queries for marginStart and marginEnd doesn't work if the default value is "auto"

Reproduce:
1. Create a <Box /> component with marginStart and/or marginEnd with value "auto"
2. Create a breakpoint for the properties, eg: mdMarginStart and/or mdMarginEnd with any value
3. Observe that the default value of "auto" never gets overwritten by the media query.

Expected Result:
The value "auto" should be overwritten by the value on the breakpoint.

Research:
- It didn't work for any other margin-family props except margin itself (due to duplication in union).
- It didn't work on `margin` when removing duplication: removing marginStart/marginEnd from 
`const lgMargin: MarginFunctorType = union(lgMarginTop, lgMarginBottom, lgMarginLeft, lgMarginRight); [packages/gestalt/src/boxTransforms.js]
` 
- This only happened with overriding `auto` with `boints`, not overriding `boints`with `boints`

Before: 
image

After: 
image


packages/gestalt/src/boxTransforms.js
```diff
@@ -212,44 +212,23 @@ const smMarginTop: MarginFunctorType = transformNumberOrPassthrough('smMarginTop
...
 const lgMarginTop: MarginFunctorType = transformNumberOrPassthrough('lgMarginTop');
 const lgMarginRight: MarginFunctorType = transformNumberOrPassthrough('lgMarginRight');
 const lgMarginBottom: MarginFunctorType = transformNumberOrPassthrough('lgMarginBottom');
 const lgMarginLeft: MarginFunctorType = transformNumberOrPassthrough('lgMarginLeft');
-const lgMargin: MarginFunctorType = union(
-  lgMarginTop,
-  lgMarginBottom,
-  lgMarginLeft,
-  lgMarginRight,
-  lgMarginStart,
```
AlbertCarreras, line 250: should `margin` prop respond like `marginStart/End` or `marginLeft/Right`? 

It doesn't matter if both are the same!!! Redundant code.

dangerismycat:
I'm working on a PR right now that deletes marginLeft/Right, so I think we can leave those changes out of this PR.

AlbertCarreras:
@dangerismycat Can we merge this and then rebase and replace right/left with start/end in the unions or margin unions? 

That, or fix bug on your same PR abandoning this one.

dangerismycat (approved):

PR #1363, opened 2021-02-04
Box: Remove marginLeft and marginRight props

dangerismycat:
These props have been recommended against in Pinboard since we introduced the RTL-friendly marginStart and marginEnd alternatives. Rather than rely on a lint rule, this PR removes the props altogether. This seems like a better DX, as IDEs will no longer suggest those props, and the dev can always use dangerouslySetInlineStyle if they really, really need a specific direction.

## Test Plan

This PR replaces 99% of marginLeft/marginRight strings in the repo; the remaining ones are in the lint rule against them, and the docs for that lint rule. All existing usages (both in src code and docs) should be properly updated.


AlbertCarreras:
@dangerismycat This looks great, thanks for this!  I didn't find any `marginLeft`/`marginRight` in Pinboard so it should be easy to be bumped. 

How do we wanna coordinate this PR with my PR? Wanna merge mine first? Then this one? The bug first, then the prop deprecation?

packages/gestalt/src/Box.js
```diff
@@ -183,7 +183,19 @@ type Props = {
 // back to the underlying `<div/>`. That's generally useful, but we'd also like
 // to strip out a few naughty properties that break style encapsulation
 // (className, style) or accessibility (onClick).
```
AlbertCarreras, line 185: can we update the comment on why margins are also disallowed?

packages/eslint-plugin-gestalt/src/no-box-marginleft-marginright.js
```diff
@@ -54,7 +54,7 @@ const rule = {
         if (isMarginLeftRightAttribute) {
```
AlbertCarreras, line 54: could we upgrade to be something like

```
['marginLeft',
  'smMarginLeft',
  'mdMarginLeft',
  'lgMarginLeft',
  'marginRight',
  'smMarginRight',
  'mdMarginRight',
  'lgMarginRight',].includes(
            value && value.name && value.name.name)
```

packages/gestalt-codemods/next/box-replace-marginleft-marginright.js
```diff
@@ -0,0 +1,63 @@
...
+      if (!localIdentifierName.includes(node.openingElement.name.name)) {
+        return null;
+      }
+
+      const attrs = node.openingElement.attributes;
+
+      const newAppendAttr = [];
+      const newAttrs = attrs
+        .map((attr) => {
+          if (attr?.name?.name && attr.name.name === 'marginLeft') {
```
AlbertCarreras, line 42: can we include responsive variations?

AlbertCarreras: This would not remove 
```
  'smMarginLeft',
  'mdMarginLeft',
  'lgMarginLeft',
  'smMarginRight',
  'mdMarginRight',
  'lgMarginRight',
```

AlbertCarreras: It's a codemod, so I wouldn't go very sophisticated, copy-paste `if{...}` for each case?

packages/gestalt-codemods/next/__testfixtures__/box-replace-marginleft-marginright.output.js
```diff
@@ -0,0 +1,7 @@
+// @flow strict
+import React from 'react';
+import { Box } from 'gestalt';
+
+export default function TestBox() {
+  return <Box marginStart={4} marginEnd={4} />;
```
AlbertCarreras, line 6: I would include responsive ones. We don't have any case in Pinboard, but as most of these Codemods... we must for our  due diligence!

docs/src/components/Card.js
```diff
@@ -77,7 +77,7 @@ export default function Card({
           </Box>
         </Heading>
       )}
-      <Box marginLeft={-2} marginRight={-2} display="flex" direction={stacked ? 'column' : 'row'}>
+      <Box marginStart={-2} marginEnd={-2} display="flex" direction={stacked ? 'column' : 'row'}>
```
AlbertCarreras, line 80: thanks for changing these ones!

AlbertCarreras:
snapshots need a refresh, there will be many updates in Pinboard as well!

packages/gestalt-codemods/next/box-replace-marginleft-marginright.js
dangerismycat: Psh, not very sophisticated! No way! 😅

AlbertCarreras (approved):

PR #1364, opened 2021-02-04
Provider/Link/Button/IconButton/TapArea: implement OnNavigation context for advanced link navigation

AlbertCarreras:
# Problem
Gestalt components do not support link navigation different that a regular `<a>`. Therefore, custom components were needed to wrap Gestalt components and modify their default behavior through the `onClick` event handler. The amount of Gestalt components using Links is increasing from more foundational (Link, Button, IconButton, TapArea) to more composed ones (Callout, Upsell, Dropdown, ActivationCard)

A new solution was needed to prevent the need to wrap components to add external logic, and instead, just internally configure the components so that engineers don't need to worry about custom solutions.

# Solution: A context provider for external link navigation logic and components with built-in management.

Components with link functionality use simple `<a>` tags. In order to replace the default link functionality with more complex ones (ex. withRouter or spam checking), `onNavigation` provides an interface to implement external logic into the `onClick` event handler in links.

`onNavigation` is a high-order function. If passed into the provider, consumer components (ex. Link, Button, IconButton, and TapArea) call the `onNavigation` function with 2 named parameters (`href` and `onNavigationOptions`) that returns a function that gets called inside the 'onClick' event handler. `onNavigationOptions` is an object that acts as a flexible API for your onNavigation external logic.`,

The examples had 4 rellevant parts: a Provider, an `onNavigation` high-order function, consumer components (Link, Button, IconButton, TapArea, Callout, Upsell, ActivationCard, Dropdown), and `onNavigationOptions` props. 

The top Provider will pass to the consumer components the `onNavigation` function which will make accessible a final function that will be called during the `onClick` event handler.  This provider implements high-level contexts meant to cover most of the apps tree rather than isolated branches.

The `onNavigation` function can contain complex logic including React hooks to perform side effects. In this case, we are using `onNavigation` to disable the link, show an alert message, and open a different url in a new window. 

Finally, the `onNavigationOptions` props provides a flexible API. In this case, `navigationMode` allows to disable/enable the `onNavigation` logic. Other uses could be accessing `event.stopPropagation`


docs/src/Provider.doc.js
```diff
@@ -1,16 +1,16 @@
...
+import MainSection from './components/MainSection.js';
 
 const cards: Array<Node> = [];
 const card = (c) => cards.push(c);
 
 card(
   <PageHeader
     name="Provider"
-    description="An app may optionally have a `Provider` to set up context for components further down the tree. The first usecase is setting the color scheme, but other uses such as right to left support will be added in the future."
+    description="An app may optionally have a `Provider` to set up context for components further down the tree. This provider implements high-level contexts meant to cover most of the apps tree rather than isolated branches."
```
dangerismycat, line 13: typo: "app tree"

docs/src/Provider.doc.js
```diff
@@ -19,65 +19,179 @@ card(
     props={[
       {
         name: 'colorScheme',
-        type: `"light" | "dark" | "userPreference"`,
+        type: `'light' | 'dark' | 'userPreference'`,
         defaultValue: 'light',
-        description:
-          'The color scheme for components inside the provider. Specify "userPreference" to use "prefers-color-scheme" media query.',
-        href: 'colorScheme',
+        description: `The color scheme for components inside the provider. Specify 'userPreference' to use 'prefers-color-scheme' media query.`,
```
dangerismycat, line 24: "Use `userPreference` to allow the end user to specify via their browser settings, using the `prefers-color-scheme` media query."

docs/src/Provider.doc.js
```diff
@@ -19,65 +19,179 @@ card(
...
+        href: 'Color-scheme',
       },
       {
         name: 'id',
         type: 'string',
         description:
-          'An optional id for your provider. If not passed in, settings will be applied as globally as possible (example: it sets color scheme variables at :root).',
+          'An optional id for your color scheme provider. If not passed in, settings will be applied as globally as possible (example: it sets color scheme variables at :root).',
```
dangerismycat, line 31: "example: setting color scheme…"

docs/src/Provider.doc.js
```diff
@@ -19,65 +19,179 @@ card(
...
-      label: "User Preference"
+      value: 'userPreference',
+      label: 'User Preference'
     }
   ];
   return (
-    <Provider colorScheme={scheme} id="docsExample">
-      <Box color="white" padding={2}>
+    <Provider colorScheme={scheme} id='docsExample'>
+      <Box color='white' padding={2}>
```
dangerismycat, line 75: These should remain double quotes

docs/src/Provider.doc.js
```diff
@@ -19,65 +19,179 @@ card(
...
       </Box>
     </Provider>
   );
 }`}
-  />,
+      />
+    </MainSection.Subsection>
+    <MainSection.Subsection
+      title="OnNavigation"
+      description="This example illustrates the implemention of `onNavigation` context so that the link functionality of Gestalt components can be controlled externally. This example had 4 rellevant parts: a Provider, an `onNavigation` high-order function, consumer components (Link, Button, IconButton, TapArea), and `onNavigationOptions` props. The top Provider will pass to the consumer components the `onNavigation` function which will make accessible a final function that will be called during the `onClick` event handler. The `onNavigation` function can contain complex logic including React hooks to perform side effects. In this case, we are using `onNavigation` to disable the link, show an alert message, and open a different url in a new window. Finally, the `onNavigationOptions` props provides a flexible API. In this case, `navigationMode` allows to disable/enable the `onNavigation` logic. Other uses could be accessing `event.stopPropagation`."
```
dangerismycat, line 97: Consider breaking this up into two chunks for readability. Also perhaps copy-paste to a word processor for spellcheck and grammar check, then paste back here

docs/src/Provider.doc.js
```diff
@@ -19,65 +19,179 @@ card(
...
+    href: 'https://pinterest.com',
+    onNavigationOptions: {
+      navigationMode: clientOnNavigationMode ? 'client' : 'server'
+    },
+    target: 'blank',
+  }
+
+  return (
+    <Provider onNavigation={onNavigation}>
+      <Flex direction='column' gap={2}>
```
dangerismycat, line 129: These should all use double quotes as well

packages/gestalt/src/Button.js
```diff
@@ -290,6 +293,8 @@ ButtonWithForwardRef.propTypes = {
   inline: PropTypes.bool,
   name: PropTypes.string,
   onClick: PropTypes.func,
+  // eslint-disable-next-line react/forbid-prop-types
+  onNavigationOptions: PropTypes.object,
```
dangerismycat, line 297: It looks like we can type this more specifically? At the very least we can define the type once (where the flow type is defined) and only suppress once, there

packages/gestalt/src/InternalLink.js
```diff
@@ -148,6 +151,17 @@ const InternalLinkWithForwardRef: AbstractComponent<Props, HTMLAnchorElement> =
       : {},
   );
 
+  // useOnNavigation is only accessible with Gestalt Provider
+  // and when onNavigation prop is passed to it
+  const { onNavigation } = useOnNavigation() || {};
```
dangerismycat, line 156: nit: nullish coalescing

packages/gestalt/src/InternalLink.js
```diff
@@ -148,6 +151,17 @@ const InternalLinkWithForwardRef: AbstractComponent<Props, HTMLAnchorElement> =
...
+
+  const onNavigationClick =
+    onNavigation &&
+    onNavigation({
```
dangerismycat, line 160: optional chaining works for function calls as well

packages/gestalt/src/Link.js
```diff
@@ -96,6 +99,17 @@ const LinkWithForwardRef: AbstractComponent<Props, HTMLAnchorElement> = forwardR
     },
   );
 
+  // useOnNavigation is only accessible with Gestalt Provider
+  // and when onNavigation prop is passed to it
+  const { onNavigation } = useOnNavigation() || {};
```
dangerismycat, line 104: ditto here

packages/gestalt/src/contexts/OnNavigation.js
```diff
@@ -0,0 +1,39 @@
...
+}
+
+function useOnNavigation(): OnNavigationType | void {
+  return useContext(OnNavigation);
+}
+
+export { OnNavigationProvider, useOnNavigation };
+
+OnNavigationProvider.propTypes = {
+  children: PropTypes.node,
```
dangerismycat, line 37: `.isRequired`

packages/gestalt/src/Link.js
```diff
@@ -96,6 +99,17 @@ const LinkWithForwardRef: AbstractComponent<Props, HTMLAnchorElement> = forwardR
...
+
+  const onNavigationClick =
```
dangerismycat, line 106: If the usage pattern is to get the onNavigation function and then call that with the href and onNavigationOptions, why not pass those to the custom hook?
```js
const onNavigationClick = useOnNavigation({ href, onNavigationOptions });
```
It'd be nice to adjust the return of useOnNavigation a bit as well, both to return the function directly (instead of needing to destructure it), and by returning a no-op instead of undefined, so later truthy checks aren't needed before calling the function.

dangerismycat (commented):
This is looking great! There are some usability tweaks that I think could make it even better 🙂

docs/src/Provider.doc.js
AlbertCarreras: DONE

docs/src/Provider.doc.js
AlbertCarreras: DONE

docs/src/Provider.doc.js
AlbertCarreras: DONE

docs/src/Provider.doc.js
AlbertCarreras: DONE

docs/src/Provider.doc.js
AlbertCarreras: DONE

docs/src/Provider.doc.js
AlbertCarreras: DONE

packages/gestalt/src/Button.js
AlbertCarreras: DONE

docs/src/Provider.doc.js
dangerismycat: I'm still seeing single quotes?

packages/gestalt/src/contexts/OnNavigation.js
```diff
@@ -0,0 +1,54 @@
+// @flow strict
+import React, { useContext, createContext, type Context, type Element, type Node } from 'react';
+import PropTypes from 'prop-types';
+
+type EventHandlerType = ({|
+  +event: SyntheticEvent<>,
+|}) => void;
+export type onNavigationOptionsType = {|
```
dangerismycat, line 8: nit: types are typically PascalCase, so `OnNavigationOptionsType`

packages/gestalt/src/contexts/OnNavigation.js
```diff
@@ -0,0 +1,54 @@
...
+  [string]: Node | EventHandlerType,
+|};
+
+export const onNavigationOptionsPropType: React$PropType$Primitive<{ +[string]: mixed }> =
```
dangerismycat, line 12: You should be able to use the flow type defined above here

packages/gestalt/src/contexts/OnNavigation.js
```diff
@@ -0,0 +1,54 @@
...
+function useOnNavigation({
+  href,
+  onNavigationOptions,
+}: {|
+  href: string,
+  onNavigationOptions?: onNavigationOptionsType,
+|}): * {
+  const { onNavigation } = useContext(OnNavigation) ?? {};
+
+  const noop = () => {};
```
dangerismycat, line 42: nit: define this outside the function so we're not creating a new no-op function every time

packages/gestalt/src/contexts/OnNavigation.js
```diff
@@ -0,0 +1,54 @@
...
+  return <Provider value={onNavigation ? { onNavigation } : undefined}>{children}</Provider>;
+}
+
+function useOnNavigation({
+  href,
+  onNavigationOptions,
+}: {|
+  href: string,
+  onNavigationOptions?: onNavigationOptionsType,
+|}): * {
```
dangerismycat, line 39: I think we know enough about the return type to type this more specifically than `*`?

packages/gestalt/src/contexts/OnNavigation.js
```diff
@@ -0,0 +1,54 @@
...
+  const { onNavigation } = useContext(OnNavigation) ?? {};
+
+  const noop = () => {};
+
+  const onNavigationClick = onNavigation?.({ href, onNavigationOptions }) ?? noop;
+
+  return onNavigationClick;
```
dangerismycat, line 46: Yeah! This is so much nicer. 🙂

packages/gestalt/src/InternalLink.js
AlbertCarreras: DONE

packages/gestalt/src/Link.js
AlbertCarreras: DONE

packages/gestalt/src/contexts/OnNavigation.js
AlbertCarreras: DONE

packages/gestalt/src/Link.js
AlbertCarreras: DONE

packages/gestalt/src/InternalLink.js
AlbertCarreras: DONE

packages/gestalt/src/contexts/OnNavigation.js
AlbertCarreras: I already tried... It breaks later

packages/gestalt/src/contexts/OnNavigation.js
AlbertCarreras: This was a temp fix. FIXED

packages/gestalt/src/contexts/OnNavigation.js
AlbertCarreras: DONE

packages/gestalt/src/contexts/OnNavigation.js
AlbertCarreras: DONE

docs/src/Provider.doc.js
AlbertCarreras: You better get drunk and see double! Come down to Florida! it's always 5 in here.

packages/gestalt/src/contexts/OnNavigation.js
AlbertCarreras: Maybe we can review types, later on, it works for now.

packages/gestalt/src/contexts/OnNavigation.js
AlbertCarreras: Agree.

docs/src/Provider.doc.js
```diff
@@ -77,7 +90,297 @@ function Example(props) {
...
   );
 }`}
-  />,
+      />
+    </MainSection.Subsection>
+    <MainSection.Subsection
+      title="On Navigation"
+      description="This example illustrates the implementation of `onNavigation` context to control the link functionality of Gestalt components externally. This example has 4 relevant parts: a Provider, an `onNavigation` high-order function, consumer components (Link, Button, IconButton, TapArea), and `onNavigationOptions` props. The top Provider passes the `onNavigation` function to consumer components. Then, `onNavigation` returns a function that gets called during the `onClick` event handler. The `onNavigation` function can contain complex logic including React hooks to perform side effects. In this case, `onNavigation` is used to disable the link, show an alert message, and open a different URL in a new window. Finally, the `onNavigationOptions` prop provides a flexible API. In this case, `navigationMode` allows to disable/enable the `onNavigation` logic. Other uses could be accessing `event.stopPropagation`."
+    >
+      <MainSection.Card
```
ayeshakmaz, line 99: I think these Cards could benefit from title's to help differentiate each and the differences b/w them

AlbertCarreras: I tried but didn't see any title.. retrying...

docs/src/Provider.doc.js
```diff
@@ -77,7 +90,297 @@ function Example(props) {
...
+
+  return (
+    <Provider onNavigation={onNavigation}>
+      <Flex direction="column" gap={2}>
+        <Flex direction="row" gap={2}>
+          <Switch
+            onChange={() => setClientOnNavigationMode(!clientOnNavigationMode)}
+            id="disable-buttons"
+            switched={clientOnNavigationMode}
+          />
```
ayeshakmaz, line 135: Could we use a radio button group in these examples instead? I'm not sure the Switch is very readable here

docs/src/Dropdown.doc.js
```diff
@@ -207,6 +207,14 @@ card(
           'When supplied, wraps the item in a Link, and directs users to the url when item is selected.',
         href: 'default',
       },
+      {
+        name: 'onNavigationOptions',
+        type: '({ [string]: Node | ({| +event: SyntheticEvent<> |}) => void }) => void',
+        description: [
+          'Pass custom props to onNavigation.',
+          'onNavigationOptions works in pair with a Provider. See Provider for examples.',
```
ayeshakmaz, line 215: maybe "works in conjunction with a Provider." Might also be helpful to expand the description a bit on what exactly should be passed to onNavigationOptions, I'm having trouble understanding the type

ayeshakmaz: (would apply to all the components :/ )

packages/gestalt/src/DropdownItem.js
```diff
@@ -55,6 +61,7 @@ export default function DropdownItem({
           shouldTruncate
           textWeight="bold"
           href={href}
+          onNavigationOptions={onNavigationOptions || {}}
```
dangerismycat, line 64: nit: nullish coalescing

docs/src/Provider.doc.js
```diff
@@ -77,7 +90,297 @@ function Example(props) {
     </Provider>
   );
 }`}
-  />,
+      />
+    </MainSection.Subsection>
+    <MainSection.Subsection
+      title="On Navigation"
```
ayeshakmaz, line 96: Maybe "Navigation context" instead?

ayeshakmaz: Or "Custom navigation context"?

dangerismycat (approved):

docs/src/Provider.doc.js
```diff
@@ -77,7 +90,297 @@ function Example(props) {
...
+      description="This example illustrates the implementation of `onNavigation` context to control the link functionality of Gestalt components externally. This example has 4 relevant parts: a Provider, an `onNavigation` high-order function, consumer components (Link, Button, IconButton, TapArea), and `onNavigationOptions` props. The top Provider passes the `onNavigation` function to consumer components. Then, `onNavigation` returns a function that gets called during the `onClick` event handler. The `onNavigation` function can contain complex logic including React hooks to perform side effects. In this case, `onNavigation` is used to disable the link, show an alert message, and open a different URL in a new window. Finally, the `onNavigationOptions` prop provides a flexible API. In this case, `navigationMode` allows to disable/enable the `onNavigation` logic. Other uses could be accessing `event.stopPropagation`."
```
ayeshakmaz, line 97: I find this sentence a bit confusing "In this case, `navigationMode` allows to disable/enable the `onNavigation` logic. "

But don't have a better suggestion : (

ayeshakmaz: Maybe @dangerismycat does?

docs/src/Provider.doc.js
AlbertCarreras: they are below the code gotcha

AlbertCarreras: DONE

packages/gestalt/src/DropdownItem.js
AlbertCarreras: DONE

docs/src/Provider.doc.js
AlbertCarreras: DONE

docs/src/Provider.doc.js
AlbertCarreras: DONE. REDID THE NAMING. IT WAS INCORRECT

docs/src/Provider.doc.js
AlbertCarreras: DONISH

PR #1368, opened 2021-02-05
Internal: disable Mergify

dangerismycat:
1) We're not all that enamored of auto-merges right now
2) If we change our minds about that, we'd like to try the native GitHub auto-merge functionality instead, since that is now a thing


AlbertCarreras (approved):

chrislloyd:
@dangerismycat I'd love more context around this ("not enamored" isn't that specific). Github's auto-merge functionality is quite different from Mergify.

dangerismycat:
> @dangerismycat I'd love more context around this ("not enamored" isn't that specific). Github's auto-merge functionality is quite different from Mergify.

We had a few instances of PRs being merged before expected, as some of us were using the "approve with some nits" strategy to speed up velocity. That led to the semver checking becoming the gating factor, which doesn't make any sense. So we decided to disable auto-merge as it was actually having negative impacts on our velocity.

chrislloyd:
Agree that the semver check being a manual step isn't ideal either.

I'm a little confused how "approve with some nits" was slower - are the nits optional?

dangerismycat:
It's not necessarily that the nits are optional, but that they're not big enough to be blockers — but are things that should be addressed. Removing auto-merge allows us to "approve with nits" such that the author can address the nits and merge when they're done, instead of needing to ping reviewers again. Auto-merge was merging before those nits could be addressed, necessitating a follow-up PR that both wastes time and doesn't fully encapsulate work in a single PR.

PR #1373, opened 2021-02-09
Docs: Update cypress accessibility testing, remove disabled contrast rules

ayeshakmaz:
- Remove disabled contrast checking in most doc pages
- Stop checking accessibility of LiveEditor (does not play well with the styling of LiveEditor that can't be changed)

## Test Plan

Ensuring test suite passes


docs/src/Checkbox.doc.js
```diff
@@ -287,6 +287,7 @@ card(
     description="Here is an example of a checkbox showing an error message."
     id="hasError"
     name="Example: Error state"
+    skipContrastCheck
```
ayeshakmaz, line 290: This is a fix available until the docs are converted to the new doc components - therefore, not documenting to avoid it being abused in the short term

docs/src/components/Example.js
```diff
@@ -14,6 +16,7 @@ type Props = {|
   name: string,
   showHeading?: boolean,
   showCode?: boolean,
+  skipContrastCheck?: boolean,
```
ayeshakmaz, line 19: Temporary fix until docs are converted to new components - not documenting to prevent use in the short term

docs/src/components/MainSectionCard.js
```diff
@@ -26,14 +26,14 @@ const CARD_SIZE_NAME_TO_PIXEL = {
 };
 
 const TYPE_TO_COLOR = {
-  do: 'green',
+  do: 'pine',
```
ayeshakmaz, line 29: Our green does not pass contrast requirements, switched to Pine

docs/src/components/Markdown.css
```diff
@@ -4,6 +4,19 @@
...
+  background-color: var(--g-colorGray0);
+}
+
+.Markdown p {
+  background-color: var(--g-colorGray0);
+  margin: 0;
+}
+
+.Markdown p + p {
+  margin-top: 14px;
```
ayeshakmaz, line 17: Adjustments to avoid negative margins being used

AlbertCarreras:
Should we include `scripts/templates/accessibility_ComponentName_spec.js`?

AlbertCarreras (approved):

PR #1377, opened 2021-02-09
Provider: onNavigationOptions proptype fix

AlbertCarreras:

### DIFF
--- a/packages/gestalt/src/contexts/OnNavigation.js
+++ b/packages/gestalt/src/contexts/OnNavigation.js
@@ -10,7 +10,8 @@ export type OnNavigationOptionsType = {|
   +[string]: Node | EventHandlerType,
 |};
 
-export const OnNavigationOptionsPropType: React$PropType$Primitive<{| +[string]: * |}> =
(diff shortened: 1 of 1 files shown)


packages/gestalt/src/commonTypes.js
```diff
@@ -24,6 +21,9 @@ export type DismissButtonType = {|
   onDismiss: () => void,
 |};
 
+const OnNavigationOptionsPropType: React$PropType$Primitive<OnNavigationOptionsType> =
```
dangerismycat, line 24: I don't think we want to define this type here; it makes much more sense in `onNavigation.js`

packages/gestalt/src/commonTypes.js
```diff
@@ -24,6 +21,9 @@ export type DismissButtonType = {|
...
+  PropTypes.any;
```
dangerismycat, line 25: We definitely want to avoid this if at all possible. Ignoring that Flow error, what would you otherwise type this as?

dangerismycat (approved):

PR #1379, opened 2021-02-10
Docs: Adjust spacing for large examples with titles

ayeshakmaz:
Move title and description above large Cards

(also fixes a code editor bug where the cursor wasn't visible)

### DIFF
--- a/docs/src/Provider.doc.js
+++ b/docs/src/Provider.doc.js
@@ -311,6 +311,7 @@ function OnNavigation() {
       />
       <MainSection.Card
         title="Example above: Dropdown"
+        description="kdshf iodfusf sjfh dsenbfr dkfuys adfejfha fdjkhfusiadf sjdhfasd "
         cardSize="lg"
         defaultCode={`
 function OnNavigation() {
--- a/docs/src/Provider.doc.js
+++ b/docs/src/Provider.doc.js
@@ -111,7 +111,7 @@ function Example(props) {
       `}
     >
       <MainSection.Card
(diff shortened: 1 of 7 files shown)


AlbertCarreras (approved):

docs/src/components/MainSectionCard.js
```diff
@@ -76,8 +76,33 @@ const MainSectionCard = ({
...
+      marginTop={borderStyle ? 4 : 3}
+      marginBottom={cardSize === 'lg' ? 4 : 0}
+      dangerouslySetInlineStyle={{
+        __style: { borderTop: borderStyle },
+      }}
+    >
+      {(title || type !== 'info') && (
+        <Box paddingY={1}>
+          <Text weight="bold" color={TYPE_TO_COLOR[type]}>
+            {cardTitle || type.charAt(0).toUpperCase() + type.slice(1)}
```
dangerismycat, line 90: Don't we have a util for capitalization now?

docs/src/Provider.doc.js
```diff
@@ -111,7 +111,7 @@ function Example(props) {
       `}
     >
       <MainSection.Card
-        title="Examples above from left to right: (1) Link, (2) Button, (3) IconButton, (4) TapArea"
+        title="Examples from start to end: Link, Button, IconButton, TapArea"
```
dangerismycat, line 114: Good catch!

dangerismycat (approved):

docs/src/components/MainSectionCard.js
ayeshakmaz: We didn't, but I made a new file for us to use w/i the Docs components

PR #1381, opened 2021-02-10
[Eslint] Add border to prefer-box eslint rule

jennyscript:
Shares logic with the existing `no-dangerous-style-duplicates` to check for divs that could be a `Box`.

## Test Plan

New unit test added to check rule works as expected

### DIFF
--- a/packages/eslint-plugin-gestalt/src/validators.js
+++ b/packages/eslint-plugin-gestalt/src/validators.js
@@ -97,3 +97,26 @@ export const validateBorderRadius = (value: string): number | string => {
   }
   return roundingLookup[value];
 };
+
+export const validateBorder = (value: string): ?string => {
+  // If the value is a string:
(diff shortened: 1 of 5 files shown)


packages/eslint-plugin-gestalt/src/validators.js
```diff
@@ -97,3 +97,26 @@ export const validateBorderRadius = (value: string): number | string => {
   }
   return roundingLookup[value];
 };
+
+export const validateBorder = (value: string): ?string => {
+  // If the value is a string:
+  // 1) convert everything to lowerCase (css is case-insensitive)
+  // 2) sort the values since some found uses have the wrong order
+  const cleanValue =
+    value && value.toLowerCase ? value.toLowerCase().split(' ').sort().join(' ') : value;
```
dangerismycat, line 106: this is really smart! 👍

dangerismycat (approved):
thanks for the update, Jenny!

PR #1382, opened 2021-02-11
Docs: Update Prop Table to support Markdown for descriptions

ayeshakmaz:
Change prop table descriptions to use Markdown (allows for links)

Before: 
Screen Shot 2021-02-10 at 6 05 00 PM

After:
Screen Shot 2021-02-10 at 1 20 59 PM
Screen Shot 2021-02-10 at 6 04 49 PM

### DIFF
--- a/docs/src/TapArea.doc.js
+++ b/docs/src/TapArea.doc.js
@@ -203,8 +203,8 @@ card(
         defaultValue: 'button',
         description: [
           `Select a TapArea variant:`,
-          `- 'button': Use for TapArea to act like buttons. The TapArea is rendered as a '<div>'.`,
-          `- 'link': Use for TapArea to act like links. The button is rendered as an '<a>'.`,
+          `- 'button': Use for TapArea to act like buttons. The TapArea is rendered as a \`<div>\`.`,
+          `- 'link': Use for TapArea to act like links. The button is rendered as an \`<a>\`.`,
           `Required with role=link.`,
         ],
         href: 'roles',
--- a/docs/src/components/Markdown.js
+++ b/docs/src/components/Markdown.js
@@ -8,6 +8,7 @@ import 'highlight.js/styles/a11y-light.css';
 import sidebarIndex from './sidebarIndex.js';
 
 type Props = {|
+  color?: string,
   text: string,
   type?: string,
 |};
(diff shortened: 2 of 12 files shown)


AlbertCarreras:
Can you add a screenshot to the description to see the before/after?

docs/src/components/PropTable.js
```diff
@@ -201,7 +200,11 @@ export default function PropTable({
                             )}
                           </Td>
                           <Td colspan={1} color="gray">
-                            {Array.isArray(description) ? Description(description) : description}
+                            {Array.isArray(description) ? (
+                              Description(description)
```
AlbertCarreras, line 204: ONe all docs migrated, Arrays wont be needed right?

docs/src/components/Markdown.js
```diff
@@ -65,7 +66,7 @@ const formatComponentName = (listitem) => {
   return listitem;
 };
 
-export default function Markdown({ text, type }: Props): Node {
+export default function Markdown({ color, text, type }: Props): Node {
```
AlbertCarreras, line 69: just a thought.. would `textColor` be more explicit? I wasn't sure if color was background etc

docs/src/TapArea.doc.js
```diff
@@ -203,8 +203,8 @@ card(
         defaultValue: 'button',
         description: [
           `Select a TapArea variant:`,
-          `- 'button': Use for TapArea to act like buttons. The TapArea is rendered as a '<div>'.`,
-          `- 'link': Use for TapArea to act like links. The button is rendered as an '<a>'.`,
+          `- 'button': Use for TapArea to act like buttons. The TapArea is rendered as a \`<div>\`.`,
+          `- 'link': Use for TapArea to act like links. The button is rendered as an \`<a>\`.`,
```
AlbertCarreras, line 207: thanks for correcting all these one.

AlbertCarreras:
I'd r+ but see all the red lights

docs/src/components/Markdown.js
ayeshakmaz: mmm good call, will update

docs/src/components/PropTable.js
ayeshakmaz: yup! we can eventually remove that

AlbertCarreras (approved):

PR #1384, opened 2021-02-11
Callout, Upsell, ActivationCard: implemented missing target, ref props for link functionality

AlbertCarreras:
## Problem
- Link, Button, IconButton, and TapArea all had the same link API: `href`, `target`, `rel`, `onNavigationProps`
- Callout, Upsell, ActivationCard only had `href` and `onNavigationProps`

## PR
- This PR standardizes the link API by adding `target`, `rel` to Callout, Upsell, ActivationCard
- Updates Docs

- This change would also help the implementation of `OnNavigation` context as `target` is one of the parameters passed to the OnNavigation prop.

### DIFF
--- a/packages/gestalt/src/ActivationCard.js
+++ b/packages/gestalt/src/ActivationCard.js
@@ -26,6 +26,8 @@ type LinkData = {|
     | SyntheticKeyboardEvent<HTMLButtonElement>,
   >,
   onNavigationOptions?: OnNavigationOptionsType,
+  rel?: 'none' | 'nofollow',
+  target?: null | 'self' | 'blank',
 |};
 
 type Props = {|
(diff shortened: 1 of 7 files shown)


packages/gestalt/src/ActivationCard.js
```diff
@@ -26,6 +26,8 @@ type LinkData = {|
     | SyntheticKeyboardEvent<HTMLButtonElement>,
   >,
   onNavigationOptions?: OnNavigationOptionsType,
+  rel?: 'none' | 'nofollow',
```
dangerismycat, line 29: At some point (can be in a future PR) we should de-dupe these types

packages/gestalt/src/ActivationCard.js
```diff
@@ -241,6 +245,8 @@ ActivationCard.propTypes = {
     onClick: PropTypes.func,
     accessibilityLabel: PropTypes.string,
     onNavigationOptions: OnNavigationOptionsPropType,
+    rel: PropTypes.oneOf(['none', 'nofollow']),
```
dangerismycat, line 248: ditto here

dangerismycat (approved):

PR #1386, opened 2021-02-11
ScrollableContainer: add Box to Related in Docs

AlbertCarreras:

### DIFF
--- a/docs/src/ScrollableContainer.doc.js
+++ b/docs/src/ScrollableContainer.doc.js
@@ -627,6 +627,12 @@ card(
   <MainSection
     name="Related"
     description={`
+
+[Box](/Box)
+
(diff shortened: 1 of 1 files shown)


docs/src/ScrollableContainer.doc.js
```diff
@@ -627,6 +627,12 @@ card(
   <MainSection
     name="Related"
     description={`
+
+[Box](/Box)
+
+Box's overflow prop specifies what should happen if content overflows an element's box. Do not replace Box with ScrollableContainer if you only need a scroll bar. ScrollableContainer is only needed when anchored components such as [Tooltip](/Tooltip), [Flyout](/Flyout), [Typeahead](/Typeahead) and [Dropdown](/Dropdown) are used within a container that could possibly scroll.
```
ayeshakmaz, line 633: Maybe:

Box's overflow prop specifies what should happen if the content is larger than the bounding box. Box should not be replaced with ScrollableContainer if the goal is simply to allow the Box to scroll when content overflows. ScrollableContainer is only needed when anchored components, such as [Tooltip](/Tooltip), [Flyout](/Flyout), [Typeahead](/Typeahead) and [Dropdown](/Dropdown), are used within a container that could potentially scroll.

ayeshakmaz (approved):
🥳

PR #1388, opened 2021-02-16
Internal: Update to React 17

jennyscript:
This updates React-related dependencies to version 17.x and also allows these versions as a peerDependency. This should be safe because react is set up as an external in the build and the main repo using gestalt has already been on React 17 for a few weeks without issue.

## Test Plan
Confirmed gestalt is working properly in the main repo

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -13,13 +13,13 @@
   "dependencies": {
     "date-fns": "^2.16.1",
     "gestalt": ">0.0.0",
-    "gestalt-datepicker":  ">0.0.0",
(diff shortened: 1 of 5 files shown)


dangerismycat (approved):
Awesome, thanks Jenny! Though an internal upgrade, does this call for a major semver bump?

jennyscript:
I **think** it doesn't have to be a Major upgrade for two reasons, but could be convinced otherwise:

- React 16 is still allowed as a peerDependency
- Since react is specified as external in rollup, gestalt will continue using whatever version of React the app installing it uses

PR #1390, opened 2021-02-17
Typeahead: Bugfix to ensure the current value of `options` is used, not just the initial

dangerismycat:
As described in [this bug](https://jira.pinadmin.com/browse/BUG-122028), Typeahead doesn't update when `options` changes. Because it was storing the initial value of `options` in a ref (which is persisted over renders), the component doesn't respond to any changes to that prop.

This PR removes that ref and uses the `options` value directly instead.

### DIFF
--- a/packages/gestalt/src/Typeahead.js
+++ b/packages/gestalt/src/Typeahead.js
@@ -72,9 +72,6 @@ const TypeaheadWithForwardRef: React$AbstractComponent<Props, HTMLInputElement> 
     zIndex,
   } = props;
 
-  // Store original data
-  // const dataRef = useRef(options);
-
   // Parent ref for positioning
   const wrapperRef = useRef(null);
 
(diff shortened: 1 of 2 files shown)


packages/gestalt/src/Typeahead.js
```diff
@@ -72,24 +72,18 @@ const TypeaheadWithForwardRef: React$AbstractComponent<Props, HTMLInputElement>
     zIndex,
   } = props;
 
-  // Store original data
-  const dataRef = useRef(options);
```
dangerismycat, line 76: This was the source of the bug, since the ref was initiated with the initial `options` value but then persisted, so would never update.

packages/gestalt/src/Typeahead.js
```diff
@@ -135,21 +127,20 @@ const TypeaheadWithForwardRef: React$AbstractComponent<Props, HTMLInputElement>
   };
 
   // Handler for when text is typed
-  // eslint-disable-next-line no-shadow
```
dangerismycat, line 138: Removed this suppression, aliased arg and updated where used

ayeshakmaz (approved):
Looks good to me from here, definitely worth extensive pinboard testing

packages/gestalt/src/Typeahead.js
AlbertCarreras: nice

PR #1391, opened 2021-02-18
Docs: Make sure right sidebar doesn't extend over footer

dangerismycat:
The recent right sidebar scrolling bugfix has a bug, where the sidebar is overlaid on the footer. This PR adds a bottom margin to the right sidebar to account for the footer height.

https://user-images.githubusercontent.com/12059539/108290663-bb0ff780-7145-11eb-951e-7d12d16bef00.mov


ayeshakmaz (approved):
Thanks for fixing this!

PR #1393, opened 2021-02-22
[WIP] DatePicker: Add timezone prop to support UTC dates

kexin-zhang:
Currently, the DatePicker only handles dates that are in the local timezone. This PR adds a `timezone` prop with a `utc` option to handle cases where the input dates are in UTC

## Test Plan

* Added a unit test
* Tested manually
* Added an example in the documentation

### DIFF
--- a/docs/src/DatePicker.doc.js
+++ b/docs/src/DatePicker.doc.js
@@ -180,6 +180,14 @@ card(
         href: 'rangePicker',
       },
       {
+        name: 'timezone',
+        type: `'local' | 'utc'`,
+        defaultValue: 'local',
+        description:
+          "Timezone that input dates are in. While `local` uses the local timezone provided by the user's browser, `utc` provides the equivalent Coordinated Universal Time or UTC date, the primary time standard by which the world regulates clocks and time.",
+        href: 'utcTimezone',
+      },
+      {
         name: 'value',
         type: 'Date',
         description: 'Pre-selected date value.',
@@ -423,6 +431,29 @@ function DatePickerExample() {
   )
 }
 `}
+  />,
+);
+
+card(
+  <Example
+    id="utcTimezone"
+    description="Use dates in UTC timezone"
+    name="Example: UTC Dates"
+    defaultCode={`
+    function DatePickerExample() {
+      const [date, setDate] = React.useState(new Date(Date.UTC(2021, 1, 15)));
+
+      return (
+        <DatePicker
+          id="example-utc"
+          label="Select a date"
+          value={date}
+          onChange={({value}) => setDate(value)}
+          timezone="utc"
+        />
+      )
+    }
+    `}
   />,
 );
 
--- a/packages/gestalt-datepicker/src/DatePicker.js
+++ b/packages/gestalt-datepicker/src/DatePicker.js
@@ -17,6 +17,7 @@ import styles from './DatePicker.css';
 import dateFormat from './dateFormat.js';
 import { LocaleDataPropTypes, type LocaleData } from './LocaleDataTypes.js';
 
+type Timezone = 'utc' | 'local';
 type Props = {|
   disabled?: boolean,
   errorMessage?: string,
@@ -38,8 +39,25 @@ type Props = {|
   rangeEndDate?: Date,
   rangeSelector?: 'start' | 'end',
   rangeStartDate?: Date,
+  timezone?: Timezone,
   value?: Date,
 |};
+
+export const getDisplayDate = (date: Date, timezone: Timezone): Date => {
+  if (timezone === 'utc') {
+    return new Date(date.getUTCFullYear(), date.getUTCMonth(), date.getUTCDate());
+  }
+
+  return date;
+};
+
+export const getUpdatedDate = (date: Date, timezone: Timezone): Date => {
+  if (timezone === 'utc') {
+    return new Date(Date.UTC(date.getFullYear(), date.getMonth(), date.getDate()));
+  }
+
+  return date;
+};
 
 const DatePickerWithForwardRef: React$AbstractComponent<Props, HTMLDivElement> = forwardRef<
   Props,
@@ -63,6 +81,7 @@ const DatePickerWithForwardRef: React$AbstractComponent<Props, HTMLDivElement> =
     rangeEndDate,
     rangeSelector,
     rangeStartDate,
+    timezone = 'local',
     value: dateValue,
   } = props;
 
@@ -71,7 +90,9 @@ const DatePickerWithForwardRef: React$AbstractComponent<Props, HTMLDivElement> =
 
   // We keep month in state to trigger a re-render when month changes since height will vary by where days fall
   // in the month and we need to keep the flyout pointed at the input correctly
-  const [selected, setSelected] = useState<?Date>(dateValue);
+  const [selected, setSelected] = useState<?Date>(
+    dateValue ? getDisplayDate(dateValue, timezone) : null,
+  );
   const [, setMonth] = useState<?number>();
   const [format, setFormat] = useState<?string>();
   const [updatedLocale, setUpdatedLocale] = useState<?string>();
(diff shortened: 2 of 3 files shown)


docs/src/DatePicker.doc.js
```diff
@@ -179,6 +179,12 @@ card(
           'Required for date range selection. Pass a ref object to DatePicker to autofocus on the unselected date range field.',
         href: 'rangePicker',
       },
+      {
+        name: 'timezone',
```
AlbertCarreras, line 183: hey can we set the default value here?

packages/gestalt-datepicker/src/DatePicker.js
```diff
@@ -38,9 +39,24 @@ type Props = {|
...
+    return date;
+  }
+  return new Date(date.getUTCFullYear(), date.getUTCMonth(), date.getUTCDate());
+};
+
+export const getUpdatedDate = (date: Date, timezone: Timezone): Date => {
+  if (timezone === 'local') {
+    return date;
+  }
+  return new Date(Date.UTC(date.getFullYear(), date.getMonth(), date.getDate()));
```
AlbertCarreras, line 57: can we invert this? 

```
if (timezone === 'utc') {
  return new Date(Date.UTC(date.getFullYear(), date.getMonth(), date.getDate()))
}

return date;

```

I'm just thinking if we pass for whatever reason the incorrect prop value 'locol' rather than 'local' (which it should be caught by Flow, but in case we don't have Flow coverage!) then we would be applying 'utc' instead. So I would set the conditional logic to 'utc'

packages/gestalt-datepicker/src/DatePicker.js
```diff
@@ -38,9 +39,24 @@ type Props = {|
   rangeEndDate?: Date,
   rangeSelector?: 'start' | 'end',
   rangeStartDate?: Date,
+  timezone?: Timezone,
   value?: Date,
 |};
 
+export const getDisplayDate = (date: Date, timezone: Timezone): Date => {
+  if (timezone === 'local') {
```
AlbertCarreras, line 47: same here; see below

packages/gestalt-datepicker/src/DatePicker.js
```diff
@@ -109,6 +128,9 @@ const DatePickerWithForwardRef: React$AbstractComponent<Props, HTMLDivElement> =
     }
   };
 
+  const min = rangeSelector === 'end' ? maxDate : rangeEndDate || maxDate;
```
AlbertCarreras, line 131: min / max are interchanged.

packages/gestalt-datepicker/src/DatePicker.js
```diff
@@ -124,20 +146,20 @@ const DatePickerWithForwardRef: React$AbstractComponent<Props, HTMLDivElement> =
...
+        endDate={rangeEndDate && getDisplayDate(rangeEndDate, timezone)}
+        excludeDates={excludeDates && excludeDates.map((date) => getDisplayDate(date, timezone))}
         highlightDates={initRangeHighlight ? [initRangeHighlight] : []}
         id={id}
-        includeDates={includeDates && [...includeDates]}
+        includeDates={includeDates && includeDates.map((date) => getDisplayDate(date, timezone))}
         locale={updatedLocale}
-        maxDate={rangeSelector === 'end' ? maxDate : rangeEndDate || maxDate}
-        minDate={rangeSelector === 'start' ? minDate : rangeStartDate || minDate}
+        maxDate={min && getDisplayDate(min, timezone)}
```
AlbertCarreras, line 155: min/max are interchanged

packages/gestalt-datepicker/src/DatePicker.js
```diff
@@ -124,20 +146,20 @@ const DatePickerWithForwardRef: React$AbstractComponent<Props, HTMLDivElement> =
         dateFormat={format}
         dayClassName={() => classnames(styles['react-datepicker__days'])}
         disabled={disabled}
-        endDate={rangeEndDate}
-        excludeDates={excludeDates && [...excludeDates]}
+        endDate={rangeEndDate && getDisplayDate(rangeEndDate, timezone)}
+        excludeDates={excludeDates && excludeDates.map((date) => getDisplayDate(date, timezone))}
         highlightDates={initRangeHighlight ? [initRangeHighlight] : []}
```
AlbertCarreras, line 151: `initRangeHighlight` this is a date as well

AlbertCarreras: line 101 you might had to add function as well

`setInitRangeHighlight(rangeStartDate || rangeEndDate);`

packages/gestalt-datepicker/src/DatePicker.jsdom.test.js
```diff
@@ -105,4 +105,29 @@ describe('DatePicker', () => {
       'react-datepicker__day react-datepicker__day--013 react-datepicker__day--selected',
     );
   });
+
+  test('accepts utc timezone dates', () => {
+    const initialUTCDate = new Date(Date.UTC(2018, 2, 24));
```
AlbertCarreras, line 110: can we refine this test to get a non-UTC date and expect a UTC to be displayed instead. It should show a change in day date, something like 


`    const initialUTCDate = new Date((2018, 2, 24)); // You might need to set time maybe?
`


`    fireEvent.focus(screen.getByDisplayValue('03/25/2018')); // we could see the effect taking place
`

packages/gestalt-datepicker/src/DatePicker.jsdom.test.js
```diff
@@ -105,4 +105,29 @@ describe('DatePicker', () => {
...
+    fireEvent.focus(screen.getByDisplayValue('03/24/2018'));
+
+    expect(screen.queryByText('March 2018')).toBeInTheDocument();
+
+    const selectedDay = screen.getByText('14');
+
+    fireEvent.click(selectedDay);
+
+    expect(mockOnChange).toHaveBeenCalledWith(expect.objectContaining({ value: expectedNewDate }));
+  });
```
AlbertCarreras, line 132: Can you test both 

`getDisplayDate`
`getUpdatedDate`?

Like unit test for both functions, for different case scenarios.

packages/gestalt-datepicker/src/DatePicker.jsdom.test.js
AlbertCarreras: Could also test `getDisplayDate`
`getUpdatedDate` using  `hasBeenCalledWith`

you might have to mock them. 

This test might be a little more tricky, I can help with it. 

Timezones are really important so we want to make sure the testing and functionality is robust.

AlbertCarreras:
This is looking good! Thank you so much @kexin-zhang!

packages/gestalt-datepicker/src/DatePicker.js
```diff
@@ -38,9 +39,26 @@ type Props = {|
   rangeEndDate?: Date,
   rangeSelector?: 'start' | 'end',
   rangeStartDate?: Date,
+  timezone?: Timezone,
   value?: Date,
 |};
 
+export const getDisplayDate = (date: Date, timezone: Timezone): Date => {
```
AlbertCarreras, line 46: i have question regarding `getDisplayDate()`

packages/gestalt-datepicker/src/DatePicker.js
```diff
@@ -38,9 +39,26 @@ type Props = {|
...
+  if (timezone === 'utc') {
+    return new Date(date.getUTCFullYear(), date.getUTCMonth(), date.getUTCDate());
```
AlbertCarreras, line 48: can we add here a comment?

// react-datepicker only supports local timezones. When 'value' prop is a UTC date, it must be turned into local  timezone before passing it to react-datepicker

kexin-zhang:
closing as discussed

PR #1394, opened 2021-02-22
ScrollBoundaryContainer: ScrollableContainer renamed to ScrollBoundaryContainer

AlbertCarreras:
`ScrollableContainer`: Rename to `ScrollBoundaryContainer`

`Box` should not be replaced with `ScrollableContainer` if the goal is simply to allow `Box` to scroll when content overflows.  However, `ScrollableContainer`'s component name leads to confusion. To prevent it, it's renamed to  `ScrollBoundaryContainer` with the intention of focusing on the intention of the container to set the scroll boundary. This name is inspired by [Popper.js ](https://popper.js.org/docs/v2/utils/detect-overflow/#boundary)where `boundary` is the prop to determines the area that the element will be checked for overflow relative to.


AlbertCarreras:
image

packages/gestalt-codemods/18.0.0/__testfixtures__/scrollableContainer-rename-scrollBoundaryContainer.input.js
```diff
@@ -0,0 +1,11 @@
+// @flow strict
+import React from 'react';
+import { ScrollableContainer, Box } from 'gestalt';
```
christianvuerings, line 3: @AlbertCarreras might be good to also add a test for aliased imports:
```
import { ScrollableContainer as CustomContainer, Box } from 'gestalt';
```

dangerismycat (approved):

packages/gestalt-codemods/18.0.0/__testfixtures__/scrollableContainer-rename-scrollBoundaryContainer.input.js
AlbertCarreras: DONE

PR #1395, opened 2021-02-23
Table: Add ability to specify sticky columns

ayeshakmaz:
Adds ability to specify sticky column(s)

## Test Plan

Added some snapshot tests


packages/gestalt/src/TableContextProvider.js
```diff
@@ -0,0 +1,14 @@
...
+type TableContextType = {|
+  stickyColumn?: number,
+|};
+
+const initialState = {
+  stickyColumn: -1,
+};
+
+const TableContext: Context<TableContextType> = createContext<TableContextType>(initialState);
+
```
AlbertCarreras, line 13: can we export here the Provider rather than the context?

AlbertCarreras: can we export useContext as well, useTableContext()

packages/gestalt/src/TableHeaderCell.js
```diff
@@ -1,20 +1,35 @@
...
+        left: shouldBeSticky ? previousTotalWidth : undefined,
+        right: shouldBeSticky ? previousTotalWidth : undefined,
+      }}
+    >
       {children}
     </th>
   );
 }
+
+TableHeaderCell.displayName = 'TableHeaderCell';
```
AlbertCarreras, line 35: I don't think we need this displayNames. If needed, 

I think we do:

TableHeaderCell.displayName = 'TableHeaderCell';

export default TableHeaderCell;

AlbertCarreras:
Looking good!

docs/src/Table.doc.js
```diff
@@ -259,6 +259,362 @@ card(
   />,
 );
 
+card(
+  <Example
+    id="stickyColumn"
+    name="Example: Sticky Column"
+    description="Try scrolling horizontally to see the first column remain in place."
+    defaultCode={`
+<Box width={"50%"} overflow="auto">
```
dangerismycat, line 268: nit: use string props directly, no need for the curlies (here and throughout)

docs/src/Table.doc.js
```diff
@@ -259,6 +259,362 @@ card(
...
+<Table maxHeight={200} stickyColumns={1}>
```
dangerismycat, line 269: These code examples are pretty hefty; can you break things up with some whitespace for readability? Since they start collapsed, I'm not really concerned with taking up more real estate

docs/src/Table.doc.js
```diff
@@ -259,6 +259,362 @@ card(
...
+</Box>
+
+`}
+  />,
+);
+
+card(
+  <Example
+    id="stickyColumn3"
+    name="Example: Sticky header and sticky columns"
```
dangerismycat, line 491: This UI looks so good!! Nice work!

docs/src/Table.doc.js
```diff
@@ -563,6 +920,212 @@ card(
...
+              accessibilityCollapseLabel="Collapse"
+              id="row1"
+              onExpand={() => setTextShown(!textShown)}
+              expandedContents={
+                <Box maxWidth={236} padding={2} column={12}>
+                  <Card
+                    image={
+                      <Avatar
+                        name="tony avatar"
+                        src="https://i.ibb.co/8948ym5/avenge.png"
```
dangerismycat, line 983: These examples are great, but are there copyright issues with using these images? Or is this considered fair use?

packages/gestalt/src/Table.js
```diff
@@ -10,23 +11,61 @@ import TableHeaderCell from './TableHeaderCell.js';
...
   children: Node,
   borderStyle?: 'sm' | 'none',
   maxHeight?: number | string,
+  stickyColumns?: number,
 |};
 
 export default function Table(props: Props): Node {
-  const { borderStyle, children, maxHeight } = props;
+  const { borderStyle, children, maxHeight, stickyColumns } = props;
+  const [showShadowScroll, setShowShadowScroll] = useState(null);
```
dangerismycat, line 25: I don't think Flow will be able to properly infer the type here, so you may want to add it. Looks like it's an optional enum? `?('left' | 'right')`

packages/gestalt/src/Table.js
```diff
@@ -10,23 +11,61 @@ import TableHeaderCell from './TableHeaderCell.js';
...
+      return;
+    }
+    if (target.scrollLeft > 0) {
+      setShowShadowScroll('right');
+    } else if (target.scrollLeft < 0) {
+      setShowShadowScroll('left');
+    } else {
+      setShowShadowScroll(null);
+    }
+  }, []);
```
dangerismycat, line 40: This seems like it might need `contentRef` as a dependency? It looks like it could have a stale value in the closure.

packages/gestalt/src/TableContextProvider.js
dangerismycat: +1 – it's a bit unexpected that a file named for the Provider in fact exports the Context

packages/gestalt/src/TableHeaderCell.js
dangerismycat: I'm _pretty sure_ this doesn't matter, but worth testing out.

packages/gestalt/src/TableRow.js
```diff
@@ -1,10 +1,41 @@
...
-  return <tr>{props.children}</tr>;
+  const { stickyColumns } = useContext(TableContext);
+  const rowRef = React.useRef();
+  const [columnWidths, setColumnWidths] = useState([]);
+
+  useEffect(() => {
+    if (rowRef && rowRef.current && stickyColumns) {
+      const colWidths = [];
+      const tableRowChildrenArray = [...rowRef.current.children];
+      tableRowChildrenArray.forEach((child) => {
```
dangerismycat, line 18: This should use `.map` instead:
```js
const colWidths = [...rowRef.current.children].map((item) => item.clientWidth);
```

packages/gestalt/src/TableRow.js
```diff
@@ -1,10 +1,41 @@
...
+import TableContext from './TableContextProvider.js';
 
 type Props = {|
   children: Node,
 |};
 
 export default function TableRow(props: Props): Node {
-  return <tr>{props.children}</tr>;
+  const { stickyColumns } = useContext(TableContext);
+  const rowRef = React.useRef();
```
dangerismycat, line 11: nit: import useRef explicitly to match surrounding style

packages/gestalt/src/TableRow.js
```diff
@@ -1,10 +1,41 @@
...
+      tableRowChildrenArray.forEach((child) => {
+        colWidths.push(child.clientWidth);
+      });
+      setColumnWidths(colWidths);
+    }
+  }, [rowRef, stickyColumns]);
+
+  const renderCellsWithIndex = () => {
+    const cells = [];
+    const tableRowChildrenArray = Children.toArray(props.children);
```
dangerismycat, line 27: This can be simplified with `Children.map`:
```js
const renderCellsWithIndex = () => Children.map(props.children, (item) => {
  ...
  return cloneElement(...);
});
```

Another idea to prevent needing this separate faux-render method would be to extract the map function to a helper, so the component return is something like
```js
  return <tr ref={rowRef}>{stickyColumns > 0 ? Children.map(props.children, renderCellsWithIndex) : props.children}</tr>;
```

packages/gestalt/src/TableRowExpandable.js
```diff
@@ -35,6 +44,9 @@ export default function TableRowExpandable(props: Props): Node {
   const [expanded, setExpanded] = useState(false);
   const hoverStyle = props.hoverStyle || 'gray';
   const cs = hoverStyle === 'gray' ? cx(styles.hoverShadeGray) : null;
+  const { stickyColumns } = useContext(TableContext);
+  const rowRef = React.useRef();
```
dangerismycat, line 48: same nit here about importing directly

packages/gestalt/src/TableRowExpandable.js
```diff
@@ -35,6 +44,9 @@ export default function TableRowExpandable(props: Props): Node {
...
+  const [columnWidths, setColumnWidths] = useState([]);
 
   const handleButtonClick = ({ event }) => {
     setExpanded(!expanded);
```
dangerismycat, line 52: This always makes me nervous…we should use functional updates for this sort of toggling to prevent using stale values: https://reactjs.org/docs/hooks-reference.html#functional-updates

packages/gestalt/src/TableRowExpandable.js
```diff
@@ -43,10 +55,38 @@ export default function TableRowExpandable(props: Props): Node {
     }
   };
 
+  useEffect(() => {
+    if (rowRef && rowRef.current && stickyColumns) {
+      const colWidths = [];
+      const tableRowChildrenArray = [...rowRef.current.children];
+      tableRowChildrenArray.forEach((child) => {
```
dangerismycat, line 62: Same feedback here re: `.map`

packages/gestalt/src/TableRowExpandable.js
```diff
@@ -43,10 +55,38 @@ export default function TableRowExpandable(props: Props): Node {
...
+      });
+      setColumnWidths(colWidths);
+    }
+  }, [rowRef, stickyColumns]);
+
+  const renderCellsWithIndex = () => {
+    const cells = [];
+    const tableRowChildrenArray = Children.toArray(props.children);
+
+    tableRowChildrenArray.forEach((child, index) => {
```
dangerismycat, line 73: same feedback here re: `Children.map`

dangerismycat (commented):
This looks awesome!! The UI is 100% there. 🙂

packages/gestalt/src/TableRowExpandable.js
AlbertCarreras: Thanks for pointing to this... I feel we do that across many cmps.

packages/gestalt/src/Table.js
AlbertCarreras: image

packages/gestalt/src/Table.js
AlbertCarreras: I think refs aren't required. eslintreact-hooks/exhaustive-deps would complain otherwise.

AlbertCarreras: image

docs/src/Table.doc.js
```diff
@@ -259,6 +259,362 @@ card(
   />,
 );
 
+card(
+  <Example
+    id="stickyColumn"
+    name="Example: Sticky Column"
```
AlbertCarreras, line 265: I';d reduce the container width to make the table scroll horizontally. Otherwise the window must be resized down

image

AlbertCarreras:
I really value your work on this shadow... but I must admit it doesn't feel good to me. I feel like a think border line could bring me more peace. haha

packages/gestalt/src/Table.js
```diff
@@ -10,23 +11,61 @@ import TableHeaderCell from './TableHeaderCell.js';
...
+
+  useEffect(() => {
+    const target = contentRef.current;
+    target?.addEventListener('scroll', updateShadows);
+    return () => {
+      target?.removeEventListener('scroll', updateShadows);
+    };
+  }, [updateShadows]);
+
+  useEffect(() => {
```
AlbertCarreras, line 50: do we need this useEffect? 

can we merge wth previous?

packages/gestalt/src/Table.js
```diff
@@ -10,23 +11,61 @@ import TableHeaderCell from './TableHeaderCell.js';
...
+    styles.table,
+    showShadowScroll === 'right' && styles.horizontalScrollRight,
+    showShadowScroll === 'left' && styles.horizontalScrollLeft,
+  );
   return (
     <Box
       overflow="auto"
       {...(borderStyle === 'sm' ? { borderStyle: 'sm', rounding: 1 } : {})}
       maxHeight={maxHeight}
+      ref={contentRef}
```
AlbertCarreras, line 64: very stupid nit,,, what about `tableRef`?

packages/gestalt/src/TableContextProvider.js
AlbertCarreras: I would rename to `TableContext`

and then export both Provider and hook.

packages/gestalt/src/TableRow.js
```diff
@@ -1,10 +1,41 @@
...
+import React, { Children, type Node, cloneElement, useContext, useEffect, useState } from 'react';
+import TableContext from './TableContextProvider.js';
 
 type Props = {|
   children: Node,
 |};
 
 export default function TableRow(props: Props): Node {
-  return <tr>{props.children}</tr>;
+  const { stickyColumns } = useContext(TableContext);
```
AlbertCarreras, line 10: import { useTableContext } from './TableContext.js';
then

  const { stickyColumns } = `useTableContext()`;

packages/gestalt/src/TableRow.js
```diff
@@ -1,10 +1,41 @@
...
+  const rowRef = React.useRef();
+  const [columnWidths, setColumnWidths] = useState([]);
+
+  useEffect(() => {
+    if (rowRef && rowRef.current && stickyColumns) {
```
AlbertCarreras, line 15: rowRef && rowRef.current >>  rowRef?.current

packages/gestalt/src/TableRow.js
AlbertCarreras: +1

packages/gestalt/src/TableRow.js
```diff
@@ -1,10 +1,41 @@
...
+      const colWidths = [];
+      const tableRowChildrenArray = [...rowRef.current.children];
+      tableRowChildrenArray.forEach((child) => {
+        colWidths.push(child.clientWidth);
+      });
+      setColumnWidths(colWidths);
+    }
+  }, [rowRef, stickyColumns]);
```
AlbertCarreras, line 23: is needed rowRef? rowRef.current aint needed. I'm not sure about this, just asking a question to discuss and learn.

packages/gestalt/src/TableRowExpandable.js
AlbertCarreras: I wonder if some of this repeated logic could be extracted out into a helper file.

AlbertCarreras:
Looking good. Some Flow issues here and there!

packages/gestalt/src/TableRowExpandable.js
ayeshakmaz: Since it's unrelated to these changes, I'll make a separate ticket to look at this across our cmps

packages/gestalt/src/TableRow.js
ayeshakmaz: It isn't actually, this was left over from a previous attempt, good catch!

packages/gestalt/src/contexts/TableContext.js
```diff
@@ -0,0 +1,28 @@
+// @flow strict
```
AlbertCarreras, line 1:
```
// @flow strict
import React, { createContext, useContext, type Context, type Element, type Node } from 'react';

type TableContextType = ?{|
  stickyColumns: number,
|};

type Props = {|
  children: Node,
  stickyColumns: number,
|};

const TableContext: Context<TableContextType | void> = createContext<TableContextType | void>();

const { Provider } = TableContext;

function TableContextProvider({ children, stickyColumns }: Props): Element<typeof Provider> {
  return <Provider value={{ stickyColumns }}>{children}</Provider>;
}

function useTableContext(): TableContextType {
  return useContext(TableContext);
}

export { TableContextProvider, useTableContext };
```

AlbertCarreras: Usee this, there are 4 flow errors, but they should be easy to fix.

docs/src/Table.doc.js
AlbertCarreras: That be a good PR stunt for our gestalt open source library! :)

packages/gestalt/src/Table.js
AlbertCarreras: nit, I think ryan was suggesting here `useState<'left' | 'right'>()`

packages/gestalt/src/TableRow.js
AlbertCarreras: https://github.com/facebook/react/issues/14387#issuecomment-503616820

AlbertCarreras (approved):
Awesome job! Sticky power!

PR #1396, opened 2021-02-23
Upsell: Add Upsell.Form component

norrise120:
This adds in the new Upsell.Form component to be used with Upsell.

Screen Shot 2021-02-23 at 9 33 47 PM

## Test Plan
New unit tests have been added in addition to testing that the form behaves as expected through manual testing.


docs/src/Upsell.doc.js
```diff
@@ -251,4 +251,38 @@ function Example(props) {
   />,
 );
 
+card(
+  <Example
+    name="Upsell with Form"
+    description={`
+      Form doc stuff
```
dangerismycat, line 258: 😂

docs/src/Upsell.doc.js
```diff
@@ -251,4 +251,38 @@ function Example(props) {
...
+  dismissButton={{
+    accessibilityLabel: 'Dismiss banner',
+    onDismiss: ()=>{},
+  }}
+  imageData={{
+    component: <Icon icon="pinterest" accessibilityLabel="Pin" color="darkGray" size={32}/>
+  }}
+>
+  <Upsell.Form
+  onSubmit={()=>{}}
```
dangerismycat, line 274: nit: we should probably preventDefault here to keep the page from reloading on click

packages/gestalt/src/Upsell.js
```diff
@@ -154,15 +157,21 @@ export default function Upsell({
...
-          {secondaryAction && responsiveMinWidth !== 'xs' && (
-            <UpsellAction type="secondary" data={secondaryAction} />
-          )}
-          {primaryAction && <UpsellAction type="primary" data={primaryAction} />}
-          {secondaryAction && responsiveMinWidth === 'xs' && (
-            <UpsellAction type="secondary" data={secondaryAction} stacked={!!secondaryAction} />
-          )}
-        </Box>
+        {children ? (
+          <Box smDisplay="flex" marginStart="auto" smMarginEnd={4} smPaddingY={3}>
```
dangerismycat, line 161: This Box is identical, so the ternary should be moved to inside of it to prevent the repetition (use a Fragment in the !children path)

packages/gestalt/src/UpsellForm.js
```diff
@@ -0,0 +1,45 @@
...
+  onSubmit,
+  submitButtonText,
+  submitButtonAccessibilityLabel,
+  submitButtonDisabled,
+}: Props): Node {
+  const responsiveMinWidth = useResponsiveMinWidth();
+
+  return (
+    <form onSubmit={onSubmit}>
+      <Flex gap={2} direction={responsiveMinWidth === 'xs' ? 'column' : 'row'}>
```
dangerismycat, line 32: I think we probably want the children to wrap?

packages/gestalt/src/UpsellForm.css.flow
```diff
@@ -0,0 +1,5 @@
+// @flow strict
```
dangerismycat, line 1: this file can be deleted

packages/gestalt/src/Upsell.js
```diff
@@ -154,15 +157,21 @@ export default function Upsell({
...
-        <Box smDisplay="flex" marginStart="auto" smMarginEnd={4} smPaddingY={3}>
-          {secondaryAction && responsiveMinWidth !== 'xs' && (
-            <UpsellAction type="secondary" data={secondaryAction} />
-          )}
-          {primaryAction && <UpsellAction type="primary" data={primaryAction} />}
-          {secondaryAction && responsiveMinWidth === 'xs' && (
-            <UpsellAction type="secondary" data={secondaryAction} stacked={!!secondaryAction} />
-          )}
-        </Box>
+        {children ? (
```
dangerismycat, line 160: Hm…this is a bit of a challenge. We need to convey to the user that they can provide children or primary/secondaryActions, but not both. How do you think we should handle that?

dangerismycat (commented):
This is looking great! There's one outstanding API issue we need to resolve (and a couple of implementation tweaks), but this is almost there. Nice work!

docs/src/Upsell.doc.js
dangerismycat: @ayeshakmaz will revisit this next week, but suggested this as a placeholder: "Upsell.Form can be used to include form fields and a submit button within Upsell."

packages/gestalt/src/Upsell.js
norrise120: Is this a case where thorough documentation is enough?

I feel like we had a similar case with Modal where there was a certain combination of props you could have (like `x` or `y and z` but not `x and y and z`). I think @AlbertCarreras and I tried using flow to type it so that it would only allow the right combinations but if I remember correctly that caused issues when we went to pinboard and had to be removed.

docs/src/Upsell.doc.js
```diff
@@ -251,4 +257,79 @@ function Example(props) {
   />,
 );
 
+card(
```
norrise120, line 260: Updated the doc using Table as an example! Let me know if there are any changes that should be made!

packages/gestalt/src/Upsell.js
```diff
@@ -154,15 +157,21 @@ export default function Upsell({
...
+          <Box smDisplay="flex" width="100%" justifyContent="end" smMarginEnd={4} smPaddingY={3}>
```
norrise120, line 161: @dangerismycat Updated per your offline advice but my TextField is still too short 🤔
Screen Shot 2021-02-23 at 7 00 03 PM

cypress/integration/accessibility_UpsellForm_spec.js
```diff
@@ -0,0 +1,10 @@
+describe('UpsellForm Accessibility check', () => {
```
norrise120, line 1: Is this needed? It's failing the integration check and when looking through the code it looks like similar components (i.e. Table.Cell) don't have these checks

packages/gestalt/src/Upsell.js
dangerismycat: Yep, there's not a good way to prevent this with Flow that I'm aware of. I think more likely we'll need to add a lint rule to guard against invalid prop combinations. That can be a follow-on PR though…if you're also interested in working on that, check out some of the existing lint rules in `packages/eslint-plugin-gestalt/src` for inspiration 🙂

cypress/integration/accessibility_UpsellForm_spec.js
dangerismycat: @ayeshakmaz Do you think this can be disabled for now, or do you see any obvious issues?

packages/gestalt/src/Upsell.js
dangerismycat: I believe you'll also need to set `width="100%"` on the `<form>` and then within UpsellForm, wrap `children` in `<Flex.Item flex="grow">`.

@christianvuerings is there a better/cleaner solution we're missing here?

cypress/integration/accessibility_UpsellForm_spec.js
ayeshakmaz: These should only be created at the page level (Upsell), so you can delete this file : ) 

It's probably something we should note when using the script to generate the files

docs/src/Upsell.doc.js
dangerismycat: Awesome! It looks like the props in the table need to be updated, though 🙂

ayeshakmaz:
@norrise120 For the failing cypress test, you can disable that check in the Upsell_spec file for now (it's unrelated to your addition):

cy.configureAxe({
      rules: [
        {
          id: 'scrollable-region-focusable',
          enabled: false,
        },
      ],
    });

packages/gestalt/src/Upsell.js
norrise120: Sweet thanks! I'll address that in a follow up! I tried adding it in this PR but it looks like its going to take me a little bit more time to make sure I'm properly finding the children of Upsell when a form is being used.

AlbertCarreras:
Erica, thanks so much for all this work. Looking great. 

I found that missing top margin above the form field. I'd also double-check with design about the field width.

image
image

docs/src/Upsell.doc.js
```diff
@@ -251,4 +257,83 @@ function Example(props) {
...
+      dismissButton={{
+        accessibilityLabel: 'Dismiss banner',
+        onDismiss: ()=>{},
+      }}
+      imageData={{
+        component: <Icon icon="pinterest" accessibilityLabel="Pin" color="darkGray" size={32}/>
+      }}
+    >
+      <Upsell.Form
+      onSubmit={(event) => {event.preventDefault();}}
```
AlbertCarreras, line 321: super nit

indentation here for the props

norrise120:
> Erica, thanks so much for all this work. Looking great.
> 
> I found that missing top margin above the form field. I'd also double-check with design about the button width.
> 
> image
> image

Confirmed with design that button should expand in the smallest breakpoint in #gestalt-design!

packages/gestalt/src/Upsell.js
AlbertCarreras: I'll be just dead code. `children` has priority over actions. We could throw errors but those only show on runtime. 

I remember the Flow issue we discussed (not exactly remember which component was and the case). Ideally yes, we would segment Upsell in 2, so when children are present we need to also include the submit button props. 

If we were to create `Upsell.primaryAction` &  `Upsell.secondaryAction`, this would put us in the same situation, how to communicate which props go with different use cases. 

Another brainstorm idea could be adding a prop variant = "form" | "actions" , and these props could help segment Flow.

I think for now doc + eslint work. At somepoint we'd have to discuss the official way of segmenting.

packages/gestalt/src/UpsellForm.js
```diff
@@ -0,0 +1,47 @@
...
+  return (
+    <form onSubmit={onSubmit} style={{ width: '100%' }}>
+      <Flex gap={2} direction={responsiveMinWidth === 'xs' ? 'column' : 'row'} wrap>
+        <Flex.Item flex="grow">{children}</Flex.Item>
+        <Flex.Item>
+          <Button
+            accessibilityLabel={submitButtonAccessibilityLabel}
+            color="red"
+            disabled={submitButtonDisabled}
+            inline={responsiveMinWidth !== 'xs'}
```
AlbertCarreras, line 39: I haven't played with the code, but what if we inline by default? It might prevent the 100%-width button in xs breakpoint.

packages/gestalt/src/UpsellForm.js
```diff
@@ -0,0 +1,47 @@
...
+  submitButtonText,
+  submitButtonAccessibilityLabel,
+  submitButtonDisabled,
+}: Props): Node {
+  const responsiveMinWidth = useResponsiveMinWidth();
+
+  return (
+    <form onSubmit={onSubmit} style={{ width: '100%' }}>
+      <Flex gap={2} direction={responsiveMinWidth === 'xs' ? 'column' : 'row'} wrap>
+        <Flex.Item flex="grow">{children}</Flex.Item>
```
AlbertCarreras, line 33: we don't really need it cos it doesn't work in column, `flex={responsiveMinWidth === 'xs'  ? 'shrink' : 'grow'}`

packages/gestalt/src/UpsellForm.js
norrise120: Screen Shot 2021-02-25 at 9 56 06 AM

Definitely needed!

packages/gestalt/src/UpsellForm.js
```diff
@@ -0,0 +1,47 @@
...
+  onSubmit,
+  submitButtonText,
+  submitButtonAccessibilityLabel,
+  submitButtonDisabled,
+}: Props): Node {
+  const responsiveMinWidth = useResponsiveMinWidth();
+
+  return (
+    <form onSubmit={onSubmit} style={{ width: '100%' }}>
+      <Flex gap={2} direction={responsiveMinWidth === 'xs' ? 'column' : 'row'} wrap>
```
AlbertCarreras, line 32: align="center" if we set inline by default, see comment below.

packages/gestalt/src/UpsellForm.js
AlbertCarreras: https://codesandbox.io/s/example-applying-flex-properties-to-children-forked-rpik6?file=/example.js

AlbertCarreras: ignore if we go with full width buttons.

norrise120: We actually do want to expand the button, so resolving this conversation!

AlbertCarreras:
LGTM!

AlbertCarreras:
Also, how do we manage long buttons? 

image

ashleyseto:
Looking good so far. One comment on the form placement. We do want to reserve the left rail and line up all of the content so the form should start where the text is currently starting when at our largest breakpoints.

ashleyseto:
> Also, how do we manage long buttons?
> 
> image
> 
> image

Should we handle breakpoints within this form component separately or now? I don't expect CTAs to get this long. If we want to handle this now, it will take some more work to define that behavior. I'd probably be following some of the form template work we already did on gestalt where things start to go to the next line when it makes sense.

norrise120:
> Looking good so far. One comment on the form placement. We do want to reserve the left rail and line up all of the content so the form should start where the text is currently starting when at our largest breakpoints.
> 
> Screen Shot 2021-02-25 at 10 20 31 AM

@ashleyseto Quick question on what we should do when the icon/image is not as tall as the rest of the content? The way we have it now, the icon/image is centered with the text content. Including the form, this is what it would look like with image and with icon. I just want to make sure that looks right to you or if we need to use special logic for this case! (Ignore the length of the form, I'll fix so that it extends to the end like the designs)

image
Screen Shot 2021-02-25 at 11 46 59 AM


icon
Screen Shot 2021-02-25 at 11 47 53 AM

norrise120:
ohh centering with all that content ends up looking kind of wonky when we have an icon and are in the medium breakpoint--so I'm not quite sure if this is what we want. 

Screen Shot 2021-02-25 at 12 54 13 PM

norrise120:
Opted for aligning items at start with forms (keeping center when there are no forms). Let me know what you think, @ashleyseto!
https://deploy-preview-1396--gestalt.netlify.app/Upsell#formExample

Screen Shot 2021-02-25 at 4 19 38 PM

Screen Shot 2021-02-25 at 4 19 31 PM

ashleyseto (approved):

AlbertCarreras:
https://www.loom.com/share/27f5e737df74449a84fc628ad3671f78

AlbertCarreras:
We should add some top margin when it slides down. Maybe in English is short but in other languages it might get larger and in reduced screens, it definitely slides under.
image

AlbertCarreras:
> https://www.loom.com/share/27f5e737df74449a84fc628ad3671f78

Hey @norrise120 , I check with @ashleyseto  about the alignment of the form.  So in cases like this we should align top the entire title/description/form.

AlbertCarreras:
> We should add some top margin when it slides down. Maybe in English is short but in other languages it might get larger and in reduced screens, it definitely slides under.
> image

image
Nice, thanks!

AlbertCarreras:
> https://www.loom.com/share/27f5e737df74449a84fc628ad3671f78

As per Ashley's feedback re small fields right before breakpoint >> Leave as it is.

docs/src/Upsell.doc.js
```diff
@@ -251,4 +257,149 @@ function Example(props) {
...
+    Component={Upsell?.Form}
+    props={[
+      {
+        name: 'children',
+        type: 'React.Node',
+      },
+      {
+        name: 'onSubmit',
+        type:
+          '({ event: SyntheticMouseEvent<HTMLButtonElement> | SyntheticKeyboardEvent<HTMLButtonElement> | SyntheticMouseEvent<HTMLAnchorElement> | SyntheticKeyboardEvent<HTMLAnchorElement> }) => void',
```
AlbertCarreras, line 280: quick one `({| event: Sy` exact type

packages/gestalt/src/index.js
```diff
@@ -122,6 +123,7 @@ export {
   Tooltip,
   Typeahead,
   Upsell,
+  UpsellForm,
```
AlbertCarreras, line 126: I don't think we need this one here, as we access it via 

```
import { Upsell } from 'gestalt'; 
<Upsell.Form>
```

AlbertCarreras:
LGTM!

AlbertCarreras (approved):

docs/src/Upsell.doc.js
```diff
@@ -251,4 +257,149 @@ function Example(props) {
...
+        onDismiss: ()=>{},
+      }}
+      imageData={{
+        component: <Icon icon="pinterest" accessibilityLabel="Pin" color="darkGray" size={32}/>
+      }}
+    >
+      <Upsell.Form
+        onSubmit={(event) => {event.preventDefault();}}
+        submitButtonText="Submit"
+        submitButtonAccessibilityLabel="Submit button"
```
ayeshakmaz, line 323: We should avoid ever putting the word "button" in a label for a button (becomes repetitive for a screen reader)

Maybe `submitButtonAccessibilityLabel="Submit name for ads credit"`

docs/src/Upsell.doc.js
```diff
@@ -251,4 +257,149 @@ function Example(props) {
...
+            src="https://i.ibb.co/7bQQYkX/stock2.jpg"
+          />,
+          mask: {rounding: 4},
+        width: 128,
+      }}
+    >
+      <Upsell.Form
+        onSubmit={(event) => {event.preventDefault();}}
+        submitButtonText="Contact me"
+        submitButtonAccessibilityLabel="Submit button"
```
ayeshakmaz, line 371: Maybe "Submit info for contact"

docs/src/Upsell.doc.js
```diff
@@ -251,4 +257,149 @@ function Example(props) {
...
+            <TextField
+              id="name"
+              onChange={({ value }) => setNameValue(value)}
+              placeholder="Name"
+              value={nameValue}
+            />
+          </Box>
+          <Box flex="grow" smMarginStart={1} marginStart={0}>
+            <TextField
+              id="email"
```
ayeshakmaz, line 390: Probably worth adding type="email"

ayeshakmaz (commented):
Few tweaks for a11y, but it looks great - thank you for adding this!

docs/src/Upsell.doc.js
```diff
@@ -27,6 +29,10 @@ card(
         ],
         href: '',
       },
+      {
+        name: 'children',
+        type: 'React.Node',
```
dangerismycat, line 34: Let's type this more specifically to `typeof UpsellForm`

docs/src/Upsell.doc.js
```diff
@@ -251,4 +257,149 @@ function Example(props) {
...
+);
+
+card(
+  <PropTable
+    name="Upsell.Form"
+    id="Upsell.Form"
+    Component={Upsell?.Form}
+    props={[
+      {
+        name: 'children',
```
dangerismycat, line 274: This should be required, since the component is useless without children

dangerismycat (commented):
This looks awesome! Nice work, @norrise120. 😊 Let's resolve these last two type issues and this is good to go.

dangerismycat (approved):

ayeshakmaz (approved):
🚢

PR #1397, opened 2021-02-24
ESLint: Add rule to check for unnecessary boxShadow

ayeshakmaz:
Check for any boxShadows set on Box via dangerouslySetInlineStyle that could use borderStyle="shadow" instead

## Test Plan

Added test

### DIFF
--- a/packages/eslint-plugin-gestalt/src/no-dangerous-style-duplicates.test.js
+++ b/packages/eslint-plugin-gestalt/src/no-dangerous-style-duplicates.test.js
@@ -47,6 +47,13 @@ const invalidBorder = readFileSync(
   path.resolve(__dirname, './__fixtures__/no-dangerous-style-duplicates/invalid/invalid-border.js'),
   'utf-8',
 );
+const invalidBoxShadow = readFileSync(
+  path.resolve(
+    __dirname,
+    './__fixtures__/no-dangerous-style-duplicates/invalid/invalid-boxShadow.js',
+  ),
+  'utf-8',
+);
 const invalidBorderRadius = readFileSync(
   path.resolve(
     __dirname,
@@ -209,7 +216,7 @@ ruleTester.run('no-dangerous-style-duplicates', rule, {
         {
           message:
             'Un-needed Box dangerous styles found. https://gestalt.netlify.app/gestalt/#/Box\n' +
-            '  Use prop `borderSize="lg"` instead',
+            '  Use prop `borderStyle="lg"` instead',
         },
       ],
     },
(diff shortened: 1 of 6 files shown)


packages/eslint-plugin-gestalt/src/no-dangerous-style-duplicates.test.js
```diff
@@ -389,5 +396,16 @@ ruleTester.run('no-dangerous-style-duplicates', rule, {
         },
       ],
     },
+    {
+      code: invalidBoxShadow,
+      parserOptions,
+      errors: [
+        {
+          message:
+            'Un-needed Box dangerous styles found. https://gestalt.netlify.app/gestalt/#/Box\n' +
```
dangerismycat, line 405: I think that URL needs to be updated to remove the `/gestalt/#`

dangerismycat: Actually, might as well go right to the business: https://gestalt.netlify.app/Box#Borders

packages/eslint-plugin-gestalt/src/validators.js
```diff
@@ -109,14 +109,47 @@ export const validateBorder = (value: string): ?string => {
...
+};
+
+export const validateBoxShadow = (value: string): ?string => {
+  // If the value is a string:
+  // 1) strip out the rgba portion
+  // 2) convert the pixel portion to only numbers
+  // 3) If both pieces match, recommend borderStyle="shadow"
+
+  const rgbaRegex =
+    '/rgba\\(\\s*(-?\\d+|-?\\d*\\.\\d+(?=%))(%?)\\s*,\\s*(-?\\d+|-?\\d*\\.\\d+(?=%))(\\2)\\s*,\\s*(-?\\d+|-?\\d*\\.\\d+(?=%))(\\2)\\s*,\\s*(-?\\d+|-?\\d*.\\d+)\\s*\\)/g';
```
dangerismycat, line 131: :dragon-faint:

packages/eslint-plugin-gestalt/src/validators.js
```diff
@@ -109,14 +109,47 @@ export const validateBorder = (value: string): ?string => {
...
+
+  let rgbaMatch = false;
+  let pixelsMatch = false;
+  if (
+    cleanRgbaPortion &&
+    (cleanRgbaPortion === 'rgba(0,0,0,0.1)' || cleanRgbaPortion === 'rgba(0,0,0,.1)')
+  ) {
+    rgbaMatch = true;
+  }
+  if (cleanPixelPortion === '008' || cleanPixelPortion === '0080') {
```
dangerismycat, line 147: nit: I'm a fan of inverting multiple checks of the same variable to something like `['008', '0080'].includes(cleanPixelPortion)`. It's a toss-up on two comparisons, but helps with readability with 3+

dangerismycat (approved):
Awesome!

packages/eslint-plugin-gestalt/src/validators.js
ayeshakmaz: Yeah I definitely did not make this 😂 thank you internet

PR #1398, opened 2021-02-27
Provider/Link/Button/IconButton/TapArea/+composed components: refactor link logic to support custom navigation within onClick prop in consumers + Codemode

AlbertCarreras:
FOLLOW UP ON Provider/Link/Button/IconButton/TapArea: implement OnNavigation context for advanced link navigation (#1364)

NOTE: Not using `noop = () => ({})` purposely as these can't take events.

# Problem
Gestalt components do not support link navigation different than a regular `<a>`. Therefore, custom components were needed to wrap Gestalt components and modify their default behavior through the `onClick` event handler. The amount of Gestalt components using Links is increasing from more foundational (Link, Button, IconButton, TapArea) to more composed ones (Callout, Upsell, Dropdown, ActivationCard)

A new solution was needed to prevent the need to wrap components to add external logic, and instead, just internally configure the components so that engineers don't need to worry about custom solutions.

# FOLLOW UP. Solution: A context provider for external link logic passed to consumer components and consumer components with optional custom navigation.

Components with link functionality use simple `<a>` tags. In order to replace the default link functionality with more complex ones (ex. withRouter or spam checking), `onNavigation` provides an interface to implement external logic into the `onClick` event handler in links.

`onNavigation` is a high-order function. It takes named arguments: `href`, `onClick` and `target`.  If passed to Provider, it gets passed down to consumer components where it's executed. Then, `onNavigation` returns a function that gets called during the `onClick` event handler

The `onNavigation` function can contain complex logic, including [React hooks](https://reactjs.org/docs/hooks-reference.html), to perform side effects. It must be a hook function, `useOnNavigation()` to allow React hooks.

The returned event handler function from  `onNavigation` uses the event access to [preventDefault()](https://developer.mozilla.org/en-US/docs/Web/API/Event/preventDefault). It could also be used to [stopPropagation()](https://developer.mozilla.org/en-US/docs/Web/API/Event/stopPropagation).

Each consumer component has an `onClick ` prop.  Provider can be disabled calling `disableOnNavigation()` inside `onClick`, a callback passed along with the `onClick` event.  After disabling Provider,  we can insert custom navigation behavior in a component level.

# PROS
- We can inject custom link logic to consumer without Provider
- Provider’s onNavigation logic can be disabled / overridden
- API is flexible AND more concise than previous `onNavigationOptions`
- We only need to document the `useOnNavigation()` hook in the codebase (the custom Pinteres-specific logic layer)
- `useOnNavigation()` or `useCustomOnNavigation()` can be reusable hooks across codebase

## Codemod

Detects
```
   <Link onNavigateOptions/>
   <Button onNavigateOptions/>
   <IconButton onNavigateOptions/>
   <TapArea onNavigateOptions/>
   <Dropdown onNavigateOptions/>
   <Callout primaryAction secondaryAction/>
   <Upsell primaryAction secondaryAction/>
   <ActivationCard link/>

``` 
Console logs each component for a manual update.
image

Run
`yarn codemod --parser=flow -t=packages/gestalt-codemods/20.0.0/detect-onNavigateOptions-to-manually-update.js relative/path/to/your/code
`

### DIFF
--- a/docs/src/ActivationCard.doc.js
+++ b/docs/src/ActivationCard.doc.js
@@ -3,6 +3,8 @@ import React, { type Node } from 'react';
 import PropTable from './components/PropTable.js';
 import Example from './components/Example.js';
 import PageHeader from './components/PageHeader.js';
+import MainSection from './components/MainSection.js';
+import { customNavigationDescription } from './components/docsUtils.js';
 
 const cards: Array<Node> = [];
 const card = (c) => cards.push(c);
@@ -42,18 +44,15 @@ card(
       {
         name: 'link',
         type:
-          '{| accessibilityLabel?: string , href: string, label: string, onClick?: ({ event: SyntheticMouseEvent<HTMLAnchorElement> | SyntheticKeyboardEvent<HTMLAnchorElement | SyntheticMouseEvent<HTMLButtonElement> | SyntheticKeyboardEvent<HTMLButtonElement> }) => void |}, onNavigationOptions: ({ [string]: Node | ({| +event: SyntheticEvent<> |}) => void }) => void, rel: "none" | "nofollow", target: "null" | "self" | "blank" |}',
+          '{| accessibilityLabel?: string , href: string, label: string, onClick?: ({ event: SyntheticMouseEvent<HTMLAnchorElement> | SyntheticKeyboardEvent<HTMLAnchorElement | SyntheticMouseEvent<HTMLButtonElement> | SyntheticKeyboardEvent<HTMLButtonElement> }, {| disableOnNavigation?: () => void |}) => void |}',
         required: false,
         defaultValue: null,
         description: [
           'Link-role button to render inside the activation card as a call-to-action to the user.',
           '- label: Text to render inside the button to convey the function and purpose of the button. The button text has a fixed size.',
           '- accessibilityLabel: Supply a short, descriptive label for screen-readers to replace button texts that do not provide sufficient context about the button component behavior. Texts like `Click Here,` `Follow,` or `Read More` can be confusing when a screen reader reads them out of context. In those cases, we must pass an alternative text to replace the button text.',
-          '- onClick: Callback fired when the button component is clicked (pressed and released) with a mouse or keyboard.',
-          `- onNavigationOptions: onNavigationOptions works in conjunction with a Provider. Pass custom props to onNavigation. See Provider for examples. onNavigation's type is flexible. Each key's value is a React.Node or an event handler function.`,
-          'Accessibility: `accessibilityLabel` populates aria-label. Screen readers read the `accessibilityLabel` prop, if present, instead of the button `text`.',
+          '- onClick: Callback fired when the button component is clicked (pressed and released) with a mouse or keyboard. See [custom navigation](#Custom-navigation) variant for examples.',
         ],
-        href: '',
       },
       {
         name: 'status',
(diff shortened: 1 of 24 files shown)


docs/src/ActivationCard.doc.js
```diff
@@ -42,18 +43,17 @@ card(
...
+          '{| accessibilityLabel?: string , href: string, label: string, onClick?: ({ event: SyntheticMouseEvent<HTMLAnchorElement> | SyntheticKeyboardEvent<HTMLAnchorElement | SyntheticMouseEvent<HTMLButtonElement> | SyntheticKeyboardEvent<HTMLButtonElement> }) => void |}, customOnNavigation: "disabled" | ({| href: string, onClick?: ({| event: SyntheticMouseEvent<HTMLAnchorElement> | SyntheticKeyboardEvent<HTMLAnchorElement> |}) => void,  target?: null | "self" | "blank" |}) => ({| event: SyntheticMouseEvent<HTMLAnchorElement> | SyntheticKeyboardEvent<HTMLAnchorElement> |}) => void',
         required: false,
         defaultValue: null,
         description: [
           'Link-role button to render inside the activation card as a call-to-action to the user.',
-          '- label: Text to render inside the button to convey the function and purpose of the button. The button text has a fixed size.',
+          '- label: Text to render inside the button to convey the function and purpose of the button. The button text has a fixed size',
```
dangerismycat, line 51: nit: this period should stay

docs/src/ActivationCard.doc.js
```diff
@@ -42,18 +43,17 @@ card(
...
           '- accessibilityLabel: Supply a short, descriptive label for screen-readers to replace button texts that do not provide sufficient context about the button component behavior. Texts like `Click Here,` `Follow,` or `Read More` can be confusing when a screen reader reads them out of context. In those cases, we must pass an alternative text to replace the button text.',
           '- onClick: Callback fired when the button component is clicked (pressed and released) with a mouse or keyboard.',
-          `- onNavigationOptions: onNavigationOptions works in conjunction with a Provider. Pass custom props to onNavigation. See Provider for examples. onNavigation's type is flexible. Each key's value is a React.Node or an event handler function.`,
-          'Accessibility: `accessibilityLabel` populates aria-label. Screen readers read the `accessibilityLabel` prop, if present, instead of the button `text`.',
+          '- customOnNavigation: Provides custom control on the link functionality. Overrides or disables onNavigation logic set by Provider. See [custom navigation](#Custom-navigation) variant for examples.',
```
dangerismycat, line 54: I'm not totally sold on this name. Perhaps `onNavigationOverride`?

docs/src/ActivationCard.doc.js
```diff
@@ -158,4 +158,136 @@ card(
...
 );
 
+card(
+  <MainSection name="Variants">
+    <MainSection.Subsection
+      title="Custom navigation"
+      description={`
+These examples illustrate a custom navigation implementation to externally control the link functionality within ActivationCard.
+
+If passed to Provider's \`onNavigation\` prop, \`useCustomOnNavigationProvider\`, a high-order function, is passed down to ActivationCard where it's executed. Then, \`onNavigation\` returns a function that gets called during the \`onClick\` event handler.
```
dangerismycat, line 168: nit: "higher-order" (here and throughout)

dangerismycat (approved):
LGTM, but I'm interested in others' reviews as well. This is really impressive work!

docs/src/ActivationCard.doc.js
jennyscript: Maybe just `onNavigation` is fine for the name? This is a function that gets called when you navigate.

docs/src/Callout.doc.js
```diff
@@ -465,6 +465,125 @@ function Example(props) {
...
+The \`useCustomOnNavigationProvider\` function can contain complex logic, including [React hooks](https://reactjs.org/docs/hooks-reference.html), to perform side effects. It takes named arguments: \`href\`, \`onClick\` and \`target\`.
+
+In the examples below, \`useCustomOnNavigationProvider\` executes the following actions:
+- Disable the default link behavior
+- Show an alert message
+- Open a different URL in a new window
+
+Finally, \`useCustomOnNavigationLink\` gets passed to ActivationCard using the \`customOnNavigation\` prop. It has the same structure as \`useCustomOnNavigationProvider\`. Callout's \`customOnNavigation\` prop also takes "disabled" to disable the Provider's \`onNavigation\` prop logic and restore the default link behaviour.
+
+The returned \`onNavigationClick\` function inside both hook functions uses the event access to [preventDefault()](https://developer.mozilla.org/en-US/docs/Web/API/Event/preventDefault). It could also be used to [stopPropagation()](https://developer.mozilla.org/en-US/docs/Web/API/Event/stopPropagation).
```
jennyscript, line 484: I see all the examples here use the same logic for examples. This is fine but is there a way we could share this text a bit to make sure it stays up to date without needing a manual update to every place it's used?

docs/src/Provider.doc.js
```diff
@@ -34,11 +34,11 @@ card(
...
         description: [
-          `Components with link functionality use simple \`<a>\` tags. In order to replace the default link functionality with more complex ones (ex. withRouter or spam checking), onNavigation provides an interface to implement external logic into the 'onClick' event handler in links.`,
-          `onNavigation is a high-order function. If passed into the provider, consumer components (ex. Link, Button, IconButton and TapArea) call the onNavigation function with three named parameters ('href', 'onNavigationOptions' and 'target') that returns a function that gets called inside the 'onClick' event handler. `,
-          `'onNavigationOptions' is an object that acts as a flexible API for your onNavigation external logic. The onNavigationOptions's type is flexible. Each key's value is a React.Node or an event handler function.`,
+          "Components with links use simple `<a>` tags. In order to replace the default link behavior with custom ones (e.g. withRouter), `onNavigation` provides an interface to pass external logic into the 'onClick' event handler in links.",
```
jennyscript, line 39: I might call out the `react-router` library specifically here instead of `withRouter`

packages/gestalt/src/Link.js
```diff
@@ -126,7 +139,10 @@ const LinkWithForwardRef: AbstractComponent<Props, HTMLAnchorElement> = forwardR
         }
       }}
       onClick={(event) => {
-        onNavigation({ event });
+        if (onNavigation) {
```
AlbertCarreras, line 142: if (onNavigation) {
onNavigation({ event });
        }

if (!onNavigation && onClick) {
    onClick({ event });
}

docs/src/ActivationCard.doc.js
AlbertCarreras: OUTDATED

docs/src/ActivationCard.doc.js
AlbertCarreras: DONE

docs/src/Provider.doc.js
AlbertCarreras: DONE

packages/gestalt/src/Link.js
AlbertCarreras: OUTDATED

docs/src/ActivationCard.doc.js
AlbertCarreras: DONE

docs/src/Callout.doc.js
AlbertCarreras: DONE

jennyscript (approved):

PR #1400, opened 2021-02-28
SearchField: Add onKeyDown prop 

keyworks:
I was trying to implement search results dropdown UI for the SearchField gestalt component. But I ran into a roadblock figuring out how to implement keyboard navigation in the search results dropdown since there is no way to handle keyboard events in SearchField. I concluded that this was not possible without adding a `onKeyDown` prop to SearchField. 

Alternatively, I could have used TextField component which has `onKeyDown` prop but I would lose some of the UI benefits of using SearchField such as the icons.

I also considered using Typeahead component but I need to customize the search results UI to more than just rows of text. 

Here is a screenshot of what I'm trying to build:
Screen Shot 2021-02-28 at 12 13 48 AM

I plan to handle ArrowUp, ArrowDown, and Enter keyboard events to implement keyboard navigation on this dropdown. 

## Test Plan

- Unit tests
- Manual verification that the callback works.

### DIFF
--- a/docs/src/SearchField.doc.js
+++ b/docs/src/SearchField.doc.js
@@ -51,6 +51,7 @@ card(
       {
         name: 'onKeyDown',
         type: '({ event: SyntheticKeyboardEvent<HTMLInputElement>, value: string }) => void',
+        description: 'Callback for key stroke events.',
       },
       {
         name: 'placeholder',
--- a/docs/src/SearchField.doc.js
+++ b/docs/src/SearchField.doc.js
@@ -49,6 +49,10 @@ card(
       }) => void`,
       },
       {
+        name: 'onKeyDown',
+        type: '({ event: SyntheticKeyboardEvent<HTMLInputElement>, value: string }) => void',
+      },
+      {
         name: 'placeholder',
         type: 'string',
       },
--- a/packages/gestalt/src/SearchField.js
+++ b/packages/gestalt/src/SearchField.js
@@ -188,6 +188,7 @@ SearchFieldWithForwardRef.propTypes = {
   onBlur: PropTypes.func,
   onChange: PropTypes.func.isRequired,
   onFocus: PropTypes.func,
+  onKeyDown: PropTypes.func,
   placeholder: PropTypes.string,
   size: PropTypes.oneOf(['md', 'lg']),
   value: PropTypes.string,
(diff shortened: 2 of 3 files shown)


ayeshakmaz:
Hey @keyworks, thanks for taking a look at this. I've got a few questions:

1) Are you planning to attach the Dropdown to SearchField for the actual search options? Or something custom using Flyout? The Dropdown has keyboard nav built in, but might need some adjustments in order to work with the SearchField

2) If not planning to use Dropdown, is the addition of onKeyDown to handle the faux-focus of the items in the menu? Or for triggering the menu to open/close?

keyworks:
> Are you planning to attach the Dropdown to SearchField for the actual search options? Or something custom using Flyout? The Dropdown has keyboard nav built in, but might need some adjustments in order to work with the SearchField

Oh I totally missed the Dropdown component. I'll take a look and close this if it works for me.

> If not planning to use Dropdown, is the addition of onKeyDown to handle the faux-focus of the items in the menu? Or for triggering the menu to open/close?

I'm using onKeyDown to handle faux-focus of items in the menu. I basically ripped the same onKeyDown handler code from Typeahead.

keyworks:
Hi @ayeshakmaz,  I tried using the Dropdown component and it won't work for my use case.

I need to be able to continue typing as the search results update live in the Dropdown however the Dropdown component seems to force focus on the Dropdown.Item once it appears which interrupts the typing. 
Screen Shot 2021-03-06 at 2 57 19 PM

Could we reconsider this PR given that TextField already supports onKeyDown? Or is there a way to control focus on Dropdown?

ayeshakmaz:
Thanks for checking that @keyworks, I had a feeling the focus would be an issue. 

In the future, I think we'll be supporting this out of the box for something like SearchField, but I think for now adding onKeyDown makes sense

docs/src/SearchField.doc.js
```diff
@@ -48,6 +48,10 @@ card(
         syntheticEvent: SyntheticEvent<HTMLInputElement>
       }) => void`,
       },
+      {
+        name: 'onKeyDown',
+        type: '({ event: SyntheticKeyboardEvent<HTMLInputElement>, value: string }) => void',
```
ayeshakmaz, line 53: Could you add a description fo the prop here?

keyworks: Done

ayeshakmaz (approved):
LGTM 👍

PR #1401, opened 2021-03-01
Icon: clean up icon svg files

AlbertCarreras:
This PR standarizes all Icon svg files to have same format.

```
<svg width="24" height="24" xmlns="http://www.w3.org/2000/svg"> // for visualizing raw svg file
    <path d="XXXXXXX"/> // path.d used by Icon cmp
</svg>
```

On Icon we only use the property `path.d`. 

image
image

Do not add property fill, e.g. `fill="#767676"`. 

For consistency, we don't add the property as it does nothing.  We override the color in the Gestalt Icon component (Gestalt only uses the path in from the SVG.); therefore, " fill" and related would only get used when looking at the raw icon files.


christianvuerings (approved):

christianvuerings:
@AlbertCarreras would be great if we had a check in CI which verified new icons were in this format. Otherwise we'll have to do more of this cleanup in the future. (can be a follow up PR)

AlbertCarreras:
I'll look into that.

PR #1403, opened 2021-03-01
Docs: fix Box z-index bug

dangerismycat:
We have a subtle bug on the new Box docs page, where the code example covers the header:

https://user-images.githubusercontent.com/12059539/109564651-381c6480-7a96-11eb-93bc-262c8e1e5e57.mov

Obviously not a show-stopper, but a little annoying, right?

While reading a blog post, [this section](https://www.joshwcomeau.com/css/stacking-contexts/#airtight-abstractions-with-isolation) jumped out at me. A possible solution for this bug? As it turns out: yes!

https://user-images.githubusercontent.com/12059539/109564848-79ad0f80-7a96-11eb-90d4-24556dae01f1.mov

### DIFF
--- a/docs/src/Box.doc.js
+++ b/docs/src/Box.doc.js
@@ -881,7 +881,13 @@ function Example() {
   const HEADER_ZINDEX = new FixedZIndex(100);
   const zIndex = new CompositeZIndex([HEADER_ZINDEX]);
   return (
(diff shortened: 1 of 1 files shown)


AlbertCarreras:
LGTM. If you want to document the new CSS property, you can add and merge it.

AlbertCarreras (approved):

PR #1405, opened 2021-03-02
Docs: Add Feedback component

ayeshakmaz:
New FeedbackComponent to be used at the top of each newly-converted docs page - Callout button links out to the Survey URL through Google Forms

Screen Shot 2021-03-02 at 9 59 04 AM

### DIFF
--- a/docs/src/components/FeedbackCallout.js
+++ b/docs/src/components/FeedbackCallout.js
@@ -1,0 +1,81 @@
+// @flow strict
+import { Box, Button, ButtonGroup, Callout, Layer, Modal } from 'gestalt';
+import React from 'react';
+
+const FeedbackCallout = () => {
+  const [showModal, setShowModal] = React.useState(false);
+
+  return (
+    <Box marginLeft={-1} marginRight={-1}>
+      <Callout
+        type="info"
+        iconAccessibilityLabel="Info icon"
+        title="We'd love your feedback!"
+        message="Feedbackkk!"
+        primaryAction={{
+          label: 'Give feedback',
+          onClick: () => {
+            setShowModal(!showModal);
+          },
+        }}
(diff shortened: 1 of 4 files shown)


docs/src/components/FeedbackCallout.js
```diff
@@ -0,0 +1,72 @@
...
+            accessibilityLabel: 'Dismiss banner',
+            onDismiss: () => {
+              setShowCallout(!showCallout);
+            },
+          }}
+        />
+      )}
+      {showModal && (
+        <Layer>
+          <Modal
```
AlbertCarreras, line 33: we are losing shadows

image

docs/src/components/FeedbackCallout.js
```diff
@@ -0,0 +1,72 @@
+// @flow strict
+import { Box, Button, Callout, Flex, Layer, Modal } from 'gestalt';
+import React, { type Node } from 'react';
+
+const FeedbackCallout = (): Node => {
```
AlbertCarreras, line 5: can we export default here?

docs/src/components/FeedbackCallout.js
AlbertCarreras: Not an expert but remember @christianvuerings mentioning iframes (related to [security](https://www.ostraining.com/blog/webdesign/against-using-iframes/#:~:text=Iframes%20Bring%20Security%20Risks,phishing%20your%20users'%20personal%20data.&text=A%20malicious%20user%20can%20hijack%20your%20users'%20keystrokes.)?) issues and suggesting using an external link

ayeshakmaz: We ended up switching to just a link, so no iframe :D

dangerismycat (approved):

PR #1408, opened 2021-03-03
Docs: fixes in links

AlbertCarreras:


ayeshakmaz (approved):
Thank you for catching this!!

PR #1409, opened 2021-03-04
SelectList: Update docs and add Best Practices

ponori:
cc @dangerismycat

### DIFF
--- a/docs/src/SelectList.doc.js
+++ b/docs/src/SelectList.doc.js
@@ -28,6 +28,7 @@ card(
           {label:'San Francisco', value: 'sf'},
           {label:'Seattle', value: 'seattle'},
         ]}
+        size='lg'
       />
     `}
   />,
@@ -123,6 +124,7 @@ card(
     {label:'Seattle', value: 'seattle'},
   ]}
   label='City'
+  size='lg'
 />`}
       />
       <MainSection.Card
@@ -141,6 +143,7 @@ card(
     {label:'Seattle', value: 'seattle'},
   ]}
   label='City'
+  size='lg'
 />`}
       />
     </MainSection.Subsection>
@@ -161,6 +164,7 @@ card(
     {label:'Ryan', value: 'ryan'},
   ]}
   label='People who are awesome'
+  size='lg'
 />`}
       />
       <MainSection.Card
@@ -176,6 +180,7 @@ card(
     {label:'Absolutely not', value: 'noagain'},
   ]}
   label='Is this a good use of a SelectList?'
+  size='lg'
 />`}
       />
     </MainSection.Subsection>
@@ -194,6 +199,7 @@ card(
       {label:'Filter value 2', value: 'fv12'},
     ]}
     placeholder='Filter 1'
+    size='lg'
   />
   <SelectList
     id="selectlistexample"
@@ -203,6 +209,7 @@ card(
       {label:'Filter value 2', value: 'fv22'},
     ]}
     placeholder='Filter 2'
+    size='lg'
   />
 </Flex>`}
       />
@@ -220,12 +227,14 @@ card(
       {label:'Filter value 2', value: 'fv12'},
     ]}
     placeholder='SelectList'
+    size='lg'
   />
   <Button
     accessibilityControls="basic-dropdown-example"
     accessibilityHaspopup
     iconEnd="arrow-down"
     text="Dropdown"
+    size='lg'
     inline
   />
 </Flex>`}
@@ -301,6 +310,7 @@ card(
         ]}
         label='Country'
         helperText='Pick your favorite country (the one with the best fries)'
+        size='lg'
       />`}
       />
     </MainSection.Subsection>
@@ -324,6 +334,7 @@ card(
           {label:'Seattle', value: 'seattle'},
         ]}
         label='City'
+        size='lg'
       />`}
       />
       <MainSection.Card
@@ -341,6 +352,7 @@ card(
         ]}
         label='City'
         focus
+        size='lg'
       />`}
       />
       <MainSection.Card
@@ -360,6 +372,7 @@ card(
         value='sailing'
         label='Hobby'
         errorMessage='Sailing is not a hobby, it is a passion'
+        size='lg'
       />`}
       />
       <MainSection.Card
@@ -377,6 +390,7 @@ card(
         ]}
         label='City'
         disabled={true}
+        size='lg'
       />`}
       />
     </MainSection.Subsection>
--- a/docs/src/SelectList.doc.js
+++ b/docs/src/SelectList.doc.js
@@ -56,7 +56,7 @@ card(
         name: 'id',
         type: 'string',
         description:
-          'A unique identifier to connect the underlying `<select>` with the associated label',
+          'A unique identifier to connect the underlying `<select>` with the associated label.',
         required: true,
       },
       {
@@ -72,7 +72,7 @@ card(
       {
         name: 'onChange',
         type: '({| event: SyntheticInputEvent<>, value: string |}) => void',
-        description: 'Callback triggered when the user makes any changes',
+        description: 'Callback triggered when the user makes any changes.',
         required: true,
       },
       {
@@ -98,7 +98,7 @@ card(
       {
         name: 'value',
         type: 'string',
-        description: 'The currently-selected value',
+        description: 'The currently-selected value.',
       },
     ]}
   />,
--- a/docs/src/SelectList.doc.js
+++ b/docs/src/SelectList.doc.js
@@ -185,7 +185,7 @@ card(
         type="do"
         description="Keep the same type of selection for a group of items. An example of this might be a filter bar. If some items could use SelectList and some items need to use Dropdown, use Dropdown for all the items in the group."
         defaultCode={`
-<Flex>
+<Flex gap={2}>
   <SelectList
     id="selectlistexample"
     onChange={() => {}}
@@ -211,7 +211,7 @@ card(
         type="don't"
         description="Mix Dropdown and SelectList in a group of items."
         defaultCode={`
-<Flex>
+<Flex gap={2}>
   <SelectList
     id="selectlistexample"
     onChange={() => {}}
--- a/docs/src/SelectList.doc.js
+++ b/docs/src/SelectList.doc.js
@@ -229,28 +229,79 @@ card(
         type="don't"
         description="Mix Dropdown and SelectList in a group of items."
         defaultCode={`
-<Flex gap={2}>
-  <SelectList
-    id="selectlistexample"
-    onChange={() => {}}
-    options={[
-      {label:'Impressions', value: 'v1'},
-      {label:'Pin clicks', value: 'v2'},
-      {label:'Outbound clicks', value: 'v3'},
-      {label:'Saves', value: 'v4'},
-      {label:'Engagement rate', value: 'v5'},
-    ]}
-    size='lg'
-  />
-  <Button
-    accessibilityControls="basic-dropdown-example"
-    accessibilityHaspopup
-    iconEnd="arrow-down"
-    text="Menu"
-    size='lg'
-    inline
-  />
-</Flex>`}
+
+
+  function SubtextIconButtonFlyoutExample() {
+    const [open, setOpen] = React.useState(false);
+    const [selected, setSelected] = React.useState(null);
+    const anchorRef = React.useRef(null);
+    const handleSelect = ({ item }) => {
+      setSelected(item);
+    };
+
+    return (
+      <Flex gap={2}>
+      <SelectList
+        id="selectlistexample"
+        onChange={() => {}}
+        options={[
+          {label:'Impressions', value: 'v1'},
+          {label:'Pin clicks', value: 'v2'},
+          {label:'Outbound clicks', value: 'v3'},
+          {label:'Saves', value: 'v4'},
+          {label:'Engagement rate', value: 'v5'},
+        ]}
+        size='lg'
+      />
+      <Box display="flex" justifyContent="center">
+
+        <Button
+          accessibilityControls="header-dropdown-example"
+          accessibilityHaspopup
+          accessibilityExpanded={ open }
+          iconEnd="arrow-down"
+          text="Menu"
+          inline
+          selected={open}
+          icon="add"
+          size="lg"
+          onClick={ () => setOpen((prevVal) => !prevVal) }
+          ref={anchorRef}
+        />
+        {open && (
+          <Dropdown id="subtext-dropdown-example" anchor={anchorRef.current} onDismiss={() => {setOpen(false)}}>
+            <Dropdown.Item
+              handleSelect={handleSelect}
+              selected={selected}
+              option={{ value: "item 2", label: "Engagement" }}
+            />
+            <Dropdown.Item
+              handleSelect={handleSelect}
+              selected={selected}
+              option={{ value: "item 2", label: "Pin clicks" }}
+            />
+            <Dropdown.Item
+              handleSelect={handleSelect}
+              selected={selected}
+              option={{ value: "item 2", label: "Outbound clicks" }}
+            />
+            <Dropdown.Item
+              handleSelect={handleSelect}
+              selected={selected}
+              option={{ value: "item 2", label: "Saves" }}
+            />
+            <Dropdown.Item
+              handleSelect={handleSelect}
+              selected={selected}
+              option={{ value: "item 2", label: "Engagement rate" }}
+            />
+          </Dropdown>
+        )}
+      </Box>
+      </Flex>
+    );
+  }
+`}
       />
     </MainSection.Subsection>
   </MainSection>,
--- a/docs/src/SelectList.doc.js
+++ b/docs/src/SelectList.doc.js
@@ -9,7 +9,10 @@ const cards: Array<Node> = [];
 const card = (c) => cards.push(c);
 
 card(
-  <FeedbackCallout link="https://docs.google.com/forms/d/e/1FAIpQLSe7h8kVcD7QqvPvjkE8s8WvnuFfhYvAEQ6L7tZwPgHjJPAbSw/viewform?usp=pp_url&entry.847151274=SelectList" />,
+  <FeedbackCallout
+    link="https://docs.google.com/forms/d/e/1FAIpQLSe7h8kVcD7QqvPvjkE8s8WvnuFfhYvAEQ6L7tZwPgHjJPAbSw/viewform?usp=pp_url&entry.847151274=SelectList"
+    componentName="SelectList"
+  />,
 );
 
 card(
@@ -19,7 +22,7 @@ card(
     SelectList displays a list of actions or options using the browser’s native select.`}
     defaultCode={`
       <SelectList
-        id="selectlistexample"
+        id="selectlistexample1"
         onChange={() => {}}
         options={[
           {label:'Algeria', value: 'algeria'},
@@ -117,7 +120,7 @@ card(
         description="Use SelectList when the user needs to select from a simple list of items."
         defaultCode={`
 <SelectList
-  id="selectlistexample"
+  id="selectlistexample2"
   onChange={() => {}}
   options={[
     {label:'Algeria', value: 'algeria'},
@@ -137,7 +140,7 @@ card(
         description="Use SelectList when additional functionality such as subtext or images are needed. Use Dropdown instead."
         defaultCode={`
 <SelectList
-  id="selectlistexample"
+  id="selectlistexample3"
   onChange={() => {}}
   options={[
     {label:'Algeria', value: 'algeria'},
@@ -159,7 +162,7 @@ card(
         description="Order the list items in SelectList either alphabetically or by usage."
         defaultCode={`
 <SelectList
-  id="selectlistexample"
+  id="selectlistexample4"
   onChange={() => {}}
   options={[
     {label:'Algeria', value: 'algeria'},
@@ -179,7 +182,7 @@ card(
         description="Use SelectList if there are fewer than 4 items in the list and there is space to display all options. Use RadioButtons instead."
         defaultCode={`
 <SelectList
-  id="selectlistexample"
+  id="selectlistexample5"
   onChange={() => {}}
   options={[
     {label:'Male', value: 'male'},
@@ -199,7 +202,7 @@ card(
         defaultCode={`
 <Flex gap={2}>
   <SelectList
-    id="selectlistexample"
+    id="selectlistexample6"
     onChange={() => {}}
     options={[
       {label:'Impressions', value: 'v1'},
@@ -212,7 +215,7 @@ card(
     size='lg'
   />
   <SelectList
-    id="selectlistexample"
+    id="selectlistexample7"
     onChange={() => {}}
     options={[
       {label:'Engagement', value: 'v1'},
@@ -242,7 +245,7 @@ card(
     return (
       <Flex gap={2}>
       <SelectList
-        id="selectlistexample"
+        id="selectlistexample8"
         onChange={() => {}}
         options={[
           {label:'Impressions', value: 'v1'},
@@ -269,7 +272,7 @@ card(
           ref={anchorRef}
         />
         {open && (
-          <Dropdown id="subtext-dropdown-example" anchor={anchorRef.current} onDismiss={() => {setOpen(false)}}>
+          <Dropdown id="selectlistexample9" anchor={anchorRef.current} onDismiss={() => {setOpen(false)}}>
             <Dropdown.Item
               handleSelect={handleSelect}
               selected={selected}
@@ -326,7 +329,7 @@ card(
         description={`Use \`lg\` as the recommended size within Pinterest products.`}
         defaultCode={`
 <SelectList
-  id="selectlistexample"
+  id="selectlistexample10"
   onChange={() => {}}
   options={[
     {label:'Algeria', value: 'algeria'},
@@ -345,7 +348,7 @@ card(
         description={`Use \`md\` on denser surfaces, such as Business products or internal tools.`}
         defaultCode={`
   <SelectList
-    id="selectlistexample"
+    id="selectlistexample11"
     onChange={() => {}}
     options={[
       {label:'Algeria', value: 'algeria'},
@@ -367,7 +370,7 @@ card(
         cardSize="lg"
         defaultCode={`
 <SelectList
-  id="selectlistexample"
+  id="selectlistexample12"
   onChange={() => {}}
   options={[
     {label: 'ARS - Argentine peso', value: 'ars'},
@@ -421,7 +424,7 @@ function Example(props) {
   ];
   return (
     <SelectList
-      id="selectlistexample"
+      id="selectlistexample13"
       name="country"
       onChange={({ value }) => setCountry(value)}
       options={countryOptions}
@@ -440,7 +443,7 @@ function Example(props) {
         cardSize="lg"
         defaultCode={`
 <SelectList
-  id="selectlistexample"
+  id="selectlistexample14"
   onChange={() => {}}
   options={[
     {label:'Algeria', value: 'algeria'},
--- a/docs/src/SelectList.doc.js
+++ b/docs/src/SelectList.doc.js
@@ -248,7 +248,7 @@ card(
     <MainSection.Subsection
       title="Labels"
       description={`
-      SelectList comes with Label built-in: just use the \`label\` prop. Be sure to provide a unique id so the Label is associated with the correct SelectList.`}
+      SelectList comes with a built-in label: just use the \`label\` prop. Be sure to provide a unique id so the Label is associated with the correct SelectList.`}
     />
   </MainSection>,
 );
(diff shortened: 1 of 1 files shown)


docs/src/SelectList.doc.js
```diff
@@ -1,16 +1,37 @@
...
+);
+
 card(
   <PageHeader
     name="SelectList"
-    description={`Use a \`SelectList\` when you have four or more items you want a user to choose from.`}
+    description={`
+    SelectList displays a list of actions or options using the browser’s native select.`}
+    defaultCode={`
+      <Flex>
```
dangerismycat, line 21: Is this Flex needed? I think we can probably just use SelectList directly without needing to wrap it

docs/src/SelectList.doc.js
```diff
@@ -20,185 +41,361 @@ card(
...
         type: 'string',
-        description: 'More information about how to complete the form field',
-        href: 'helperText',
+        description: `Used to provide more information about the form field. Be sure to localize the text. See the [helper text](#Helper-text) variant to learn more.`,
       },
       {
         name: 'id',
         type: 'string',
+        description:
+          'A unique identifier to connect the underlying `<select>` with the associated label',
```
dangerismycat, line 61: nit (the sort that shouldn't be a blocker, but can be addressed if you're making other changes): there's some inconsistency here with finals periods. I believe we always want to include them, even if there's only one sentence

docs/src/SelectList.doc.js
```diff
@@ -20,185 +41,361 @@ card(
...
+        cardSize="md"
+        type="don't"
+        description="Use SelectList when additional functionality such as subtext or images are needed. Use Dropdown instead."
+        defaultCode={`
+<SelectList
+  id="selectlistexample"
+  onChange={() => {}}
+  options={[
+    {label:'List item 1', value: 'Value'},
+    {label:'I want to put an image here, but I can’t :( ', value: 'Value'},
```
dangerismycat, line 140: 😊

docs/src/SelectList.doc.js
```diff
@@ -20,185 +41,361 @@ card(
...
+      <MainSection.Card
+        cardSize="md"
+        type="do"
+        description="Order the list items in SelectList either alphabetically or by usage."
+        defaultCode={`
+<SelectList
+  id="selectlistexample"
+  onChange={() => {}}
+  options={[
+    {label:'Alberto', value: 'alberto'},
```
dangerismycat, line 159: That is a lot of A's 😅

docs/src/SelectList.doc.js
```diff
@@ -20,185 +41,361 @@ card(
...
+    {label:'Annie', value: 'annie'},
+    {label:'Ashley', value: 'ashley'},
+    {label:'Ayesha', value: 'ayesha'},
+    {label:'Ryan', value: 'ryan'},
```
dangerismycat, line 163: one of these is not like the other ones… 😂

docs/src/SelectList.doc.js
```diff
@@ -20,185 +41,361 @@ card(
...
+  ]}
+  label='People who are awesome'
```
dangerismycat, line 165: ❤️

docs/src/SelectList.doc.js
```diff
@@ -20,185 +41,361 @@ card(
...
+        cardSize="md"
+        type="don't"
+        description="Use SelectList if there are fewer than 4 items in the list and there is space to display all options. Use RadioButtons instead."
+        defaultCode={`
+<SelectList
+  id="selectlistexample"
+  onChange={() => {}}
+  options={[
+    {label:'No', value: 'no'},
+    {label:'Absolutely not', value: 'noagain'},
```
dangerismycat, line 178: 😂

docs/src/SelectList.doc.js
```diff
@@ -20,185 +41,361 @@ card(
...
+/>`}
+      />
+    </MainSection.Subsection>
+    <MainSection.Subsection>
+      <MainSection.Card
+        cardSize="md"
+        type="do"
+        description="Keep the same type of selection for a group of items. An example of this might be a filter bar. If some items could use SelectList and some items need to use Dropdown, use Dropdown for all the items in the group."
+        defaultCode={`
+<Flex>
```
dangerismycat, line 190: design nit: I'd throw a `gap` of 1-2 in here

docs/src/SelectList.doc.js
```diff
@@ -20,185 +41,361 @@ card(
...
+  <SelectList
+    id="selectlistexample"
+    onChange={() => {}}
+    options={[
+      {label:'Filter value 1', value: 'fv11'},
+      {label:'Filter value 2', value: 'fv12'},
+    ]}
+    placeholder='SelectList'
+  />
+  <Button
```
dangerismycat, line 226: This should probably include an actual Dropdown

docs/src/SelectList.doc.js
```diff
@@ -20,185 +41,361 @@ card(
...
-        placeholder="Select country"
-        label="Country"
-        helperText="Pick your favourite country (the one with the best fries)"
-        value={country}
+  <MainSection name="Variants">
+    <MainSection.Subsection title="Size">
+      <MainSection.Card
+        cardSize="md"
+        title="Large"
+        description={`Use \`lg\` as the default size within Pinterest products.`}
```
dangerismycat, line 255: Oh, I totally missed that `'md'` is the actual component default…whoops. 😬 Add that to the list of defaults we need to update…

In the meantime, we should probably add `size="lg"` to all examples in the docs, aside from the ones that should specifically be `'md'`

docs/src/SelectList.doc.js
```diff
@@ -20,185 +41,361 @@ card(
...
+        defaultCode={`<SelectList
```
dangerismycat, line 256: If you look at the code in the example sandbox, the formatting is a bit funky. I think you need to start with a line break (so `<SelectList` is on the next line with no spaces preceding it), then space the remaining lines accordingly. (It looks like this is the case with all the examples that show a code sandbox)

docs/src/SelectList.doc.js
```diff
@@ -20,185 +41,361 @@ card(
...
+        onChange={() => {}}
+        options={[
+          {label:'Austin', value: 'austin'},
+          {label:'Los Angeles', value: 'la'},
+          {label:'New York', value: 'nyc'},
+          {label:'San Francisco', value: 'sf'},
+          {label:'Seattle', value: 'seattle'},
+        ]}
+        label='City'
+        focus
```
dangerismycat, line 345: `focus` isn't a prop on SelectList. I'm not quite sure how to trigger this state without messing with a11y for the user…I'll bring this up in the channel

dangerismycat (commented):
This looks awesome!! Not a whole lot that needs to change. Nice work! 🙂

docs/src/SelectList.doc.js
```diff
@@ -1,16 +1,37 @@
...
+card(
+  <FeedbackCallout link="https://docs.google.com/forms/d/e/1FAIpQLSe7h8kVcD7QqvPvjkE8s8WvnuFfhYvAEQ6L7tZwPgHjJPAbSw/viewform?usp=pp_url&entry.847151274=SelectList" />,
+);
+
 card(
   <PageHeader
     name="SelectList"
-    description={`Use a \`SelectList\` when you have four or more items you want a user to choose from.`}
+    description={`
+    SelectList displays a list of actions or options using the browser’s native select.`}
```
AlbertCarreras, line 19: what would be an example of an action?

docs/src/SelectList.doc.js
```diff
@@ -20,185 +41,361 @@ card(
...
         name: 'name',
         type: 'string',
-        href: 'basicExample',
+        description: 'Used to specify the name of the control.',
       },
       {
         name: 'onChange',
-        type: '({ event: SyntheticInputEvent<>, value: string }) => void',
+        type: '({| event: SyntheticInputEvent<>, value: string |}) => void',
+        description: 'Callback triggered when the user makes any changes',
```
AlbertCarreras, line 77: `any changes` seems like there are many options, but it's just `selects a new option`. We can't delete the option nor search for it.

docs/src/SelectList.doc.js
```diff
@@ -20,185 +41,361 @@ card(
...
         required: true,
-        href: 'basicExample',
       },
       {
         name: 'options',
-        type: 'Array<{ label: string, value: string, disabled?: boolean }>',
+        type: 'Array<{| label: string, value: string, disabled?: boolean |}>',
+        description:
+          'The options displayed in the dropdown list. Note that ``disabled`` here is used to disable a single option. Be sure to localize the label.',
```
AlbertCarreras, line 84: should we add an example in Variant >> Disabled?

docs/src/SelectList.doc.js
```diff
@@ -20,185 +41,361 @@ card(
...
       {
         name: 'label',
         type: 'string',
+        description: 'The label shown above the input. Be sure to localize the label.',
       },
       {
         name: 'name',
         type: 'string',
-        href: 'basicExample',
+        description: 'Used to specify the name of the control.',
```
AlbertCarreras, line 72: I had no context on this attribute, googled and found this

```
The name attribute specifies the name for a drop-down list.

The name attribute is used to reference elements in a JavaScript, or to reference form data after a form is submitted.

```

Any of the examples use it, so I wonder how we could explain this better. When to use it vs not.

docs/src/SelectList.doc.js
```diff
@@ -20,185 +41,361 @@ card(
...
+        description:
+          'The options displayed in the dropdown list. Note that ``disabled`` here is used to disable a single option. Be sure to localize the label.',
         required: true,
-        href: 'basicExample',
       },
       {
         name: 'placeholder',
         type: 'string',
-        href: 'basicExample',
+        description:
```
AlbertCarreras, line 90: I think the placeholder is broken. If passed, options can't be selected. Maybe it requires a specific implementation like value?

AlbertCarreras: If no variant is added for this, it's hard to catch issues.

docs/src/SelectList.doc.js
```diff
@@ -20,185 +41,361 @@ card(
...
-        description: 'md: 40px, lg: 48px',
+        description: `md: 40px, lg: 48px. See the [size](#Size) variant to learn more.`,
         defaultValue: 'md',
       },
       {
         name: 'value',
         type: 'string',
-        description: 'Value that is selected.',
-        href: 'basicExample',
+        description: 'The currently-selected value',
```
AlbertCarreras, line 103: value is a _prepopulated_ value and it requires state to make the form component controlled. 

passing just value makes the options not selectable, need to use onChange to change the value's value. See error state example to see the case.

i'd be worth an example of the implementation.

docs/src/SelectList.doc.js
```diff
@@ -20,185 +41,361 @@ card(
...
+    {label:'Austin', value: 'austin'},
+    {label:'Los Angeles', value: 'la'},
+    {label:'New York', value: 'nyc'},
+    {label:'San Francisco', value: 'sf'},
+    {label:'Seattle', value: 'seattle'},
+  ]}
+  label='City'
+/>`}
+      />
+      <MainSection.Card
```
AlbertCarreras, line 130: I know Do/Dont go in pairs, but this seems more like a Related > Dropdown. It's not possible to add an image to SelectList therefore it's an impossible Don't. I would keep this in Related and find a more relevant Don't

docs/src/SelectList.doc.js
```diff
@@ -20,185 +41,361 @@ card(
...
+  ]}
+  label=''
+/>`}
+      />
+    </MainSection.Subsection>
+    <MainSection.Subsection>
+      <MainSection.Card
+        cardSize="md"
+        type="do"
+        description="Order the list items in SelectList either alphabetically or by usage."
```
AlbertCarreras, line 153: I'd stick to the direction of the guidelines of finding Pinterest-relevant / Pinterest-inspired examples. Here a simple country list would do the job.

docs/src/SelectList.doc.js
```diff
@@ -20,185 +41,361 @@ card(
...
+        defaultCode={`
+<SelectList
```
AlbertCarreras, line 155: Same here, I'd find a good case of checkboxes in Pinboard, and translate it into SelectList or a common example of misused SelectList

For example, we could do gender here.
image

image

docs/src/SelectList.doc.js
```diff
@@ -20,185 +41,361 @@ card(
...
+    {label:'Ayesha', value: 'ayesha'},
+    {label:'Ryan', value: 'ryan'},
+  ]}
+  label='People who are awesome'
+/>`}
+      />
+      <MainSection.Card
+        cardSize="md"
+        type="don't"
+        description="Use SelectList if there are fewer than 4 items in the list and there is space to display all options. Use RadioButtons instead."
```
AlbertCarreras, line 171: Same here, I'd find a good case of checkboxes in Pinboard, and translate it into SelectList or a common example of misused SelectList

For example, we could do gender here.
image

image

docs/src/SelectList.doc.js
```diff
@@ -20,185 +41,361 @@ card(
...
+      />
+    </MainSection.Subsection>
+    <MainSection.Subsection>
+      <MainSection.Card
+        cardSize="md"
+        type="do"
+        description="Keep the same type of selection for a group of items. An example of this might be a filter bar. If some items could use SelectList and some items need to use Dropdown, use Dropdown for all the items in the group."
+        defaultCode={`
+<Flex>
+  <SelectList
```
AlbertCarreras, line 191: I'd find a Pinterest-inspired example

image

image

docs/src/SelectList.doc.js
```diff
@@ -20,185 +41,361 @@ card(
...
+    </MainSection.Subsection>
+  </MainSection>,
+);
+
+card(
+  <MainSection name="Accessibility">
+    <MainSection.Subsection
+      title="Labels"
+      description={`
+      SelectList comes with Label built-in: just use the \`label\` prop. Be sure to provide a unique id so the Label is associated with the correct SelectList.`}
```
AlbertCarreras, line 244: Label isn't a component so I think I'd lower case it. 
`id` is  prop so I'd format it \`id\`

docs/src/SelectList.doc.js
AlbertCarreras: +1 good catch.

docs/src/SelectList.doc.js
```diff
@@ -20,185 +41,361 @@ card(
...
+          {label:'New York', value: 'nyc'},
+          {label:'San Francisco', value: 'sf'},
+          {label:'Seattle', value: 'seattle'},
+        ]}
+        size='md'
+      />`}
+      />
+    </MainSection.Subsection>
+    <MainSection.Subsection
+      title="Helper text"
```
AlbertCarreras, line 288: I'd find a more Pinterest-relevant example for hypertext. I think we would not add a helper text to select a country.

AlbertCarreras: image

AlbertCarreras: This one might be a hard one to find, so maybe more Pinterest-inspired.

docs/src/SelectList.doc.js
```diff
@@ -20,185 +41,361 @@ card(
...
 );
 
+card(
+  <MainSection name="Related">
+    <MainSection.Subsection
+      description={`
+**[Dropdown](/Dropdown)**
+If additional functionality is needed in the menu, such as subtext, headers or custom styling, use Dropdown.
+
+**[Typeahead](/Typeahead)**
```
AlbertCarreras, line 395: I would add Radiobutton,
If the selection list is 4 items or less, use Radiobutton.

Could Checkbox be another of these cases?

docs/src/SelectList.doc.js
ponori: Good call! Removing!

docs/src/SelectList.doc.js
ponori: Done!

docs/src/SelectList.doc.js
ponori: Added an example for no subtext as well.

docs/src/SelectList.doc.js
ponori: Done!

docs/src/SelectList.doc.js
ponori: Very, very nice catch!

docs/src/SelectList.doc.js
ponori: As per our conversation today, I removed the `disabled` and `focus` states and moved `errorMessage` to the variants seciton.

docs/src/SelectList.doc.js
ponori: Good question! @ashleyseto and @anniemochi, what's your guidance on 

> this?

docs/src/SelectList.doc.js
ponori: I'm going to backlog the disabled option example. Thanks!

docs/src/SelectList.doc.js
ponori: Love it. Just made that change.

docs/src/SelectList.doc.js
ponori: Love it - done!

docs/src/SelectList.doc.js
ponori: Done!

docs/src/SelectList.doc.js
ponori: Done!

docs/src/SelectList.doc.js
```diff
@@ -20,185 +41,361 @@ card(
...
+          {label:'Seattle', value: 'seattle'},
+        ]}
+        label='City'
+        focus
+      />`}
+      />
+      <MainSection.Card
+        cardSize="lg"
+        title="Error"
+        description={`Use \`errorMessage\` to pass in a description of the error in SelectList.`}
```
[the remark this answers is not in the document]
ponori: Perfect. Done!

docs/src/SelectList.doc.js
ponori: Great callout. I'll add!

docs/src/SelectList.doc.js
ponori: Done!

ponori:
OK, I _think_ I hit all the feedback. Let me know what you think.

docs/src/SelectList.doc.js
dangerismycat: I'm still not seeing a Dropdown here?

docs/src/SelectList.doc.js
dangerismycat: I don't believe it's broken, but placeholder and value must be used together as a controlled component.

docs/src/SelectList.doc.js
dangerismycat: Label is a component: https://gestalt.netlify.app/Label

And it's what we're using internally here

docs/src/SelectList.doc.js
```diff
@@ -1,16 +1,37 @@
...
     name="SelectList"
-    description={`Use a \`SelectList\` when you have four or more items you want a user to choose from.`}
+    description={`
+    SelectList displays a list of actions or options using the browser’s native select.`}
+    defaultCode={`
+      <SelectList
+        id="selectlistexample"
+        onChange={() => {}}
+        options={[
+          {label:'Alergia', value: 'alergia'},
```
dangerismycat, line 25: typo: Algeria (here and throughout, for both label and value

docs/src/SelectList.doc.js
ponori: Really? It's working on my latest build.

docs/src/SelectList.doc.js
ponori: Totally makes sense. Do you have a go-to example for this? I don't necessarily think I should be writing code for these kinds of examples.

docs/src/SelectList.doc.js
ponori: Oh, interesting. OK, so maybe revert back?

ponori: Actually, did some digging, we use `FormLabel` in this component - which I'm assuming is different from `Label`?

docs/src/SelectList.doc.js
ponori: Nice catch - fixed!

docs/src/SelectList.doc.js
ponori: Finally got to this one. Thanks for the patience.

docs/src/SelectList.doc.js
ponori: Any suggestions on how to document this? Is this something we could backlog?

docs/src/SelectList.doc.js
ponori: I think that's really good feedback. @ashleyseto, any suggestions on alternate examples?

docs/src/SelectList.doc.js
ponori: I don't think I have a good notion of how to describe this attribute. Looks like what we're using comes straight from MDN. I'd suggest we keep it for now.

docs/src/SelectList.doc.js
ponori: Sounds good - thanks!

ponori:
OK, I _think_ I covered all the feedback. Thanks again!

docs/src/SelectList.doc.js
dangerismycat: Now I see it! Nice!

docs/src/SelectList.doc.js
dangerismycat: > Looks like what we're using comes straight from MDN

Guilty as charged. 😅 TBH I don't fully understand why we need this prop either. I'm fine with leaving it as-is for now and updating in the future

docs/src/SelectList.doc.js
dangerismycat: We should probably add a variant that demonstrates how SelectList works as a controlled component. It'll need a parent component that maintains state, something roughly like
```
function Parent() {
const [value, setValue] = React.useState('');

return (
  <SelectList
    placeholder="foo"
    onChange={setValue}
    value={value}
  />
)
}
```

docs/src/SelectList.doc.js
dangerismycat: the placeholder variant covers this

docs/src/SelectList.doc.js
dangerismycat: this example makes sense to me?

docs/src/SelectList.doc.js
dangerismycat: FormLabel is a wrapper around Label, so this is still using Label under the hood (it wouldn't make sense to mention FormLabel, since that's a purely internal component)

ponori: OK, switching back. Thanks!

docs/src/SelectList.doc.js
ponori: I'd say let's hold for now. We can always improve down the road.

docs/src/SelectList.doc.js
ponori: OK, cool. I set up a Placeholder variant with onChange hooked up. I'm assuming I should migrate all the examples over to use `onChange` correct?

docs/src/SelectList.doc.js
ponori: Oh, no knock at all for using MDN! I think we go with this and re-evaluate if/when necessary.

docs/src/SelectList.doc.js
ponori: OK, few things:
* I added a variant for onChange, but the card title enforces sentence case, so the title looks like Onchange. :( 
* I am not at all equipped to write a description for this example. :D
* Does `onChange` belong in the "Variant" category? Seems more functional, but I'll defer to y'all.

dangerismycat: This looks great, as does the placeholder example. However, I think we could combine the two for a "Controlled Component" variant, since they're showing basically the same thing. That gives us a good place to call out that using `placeholder` and `value` require the component to be controlled, and both props can link to this variant from the props table.

docs/src/SelectList.doc.js
dangerismycat: We don't need to — we do offer the possibility that SelectList can be used within an uncontrolled form, basically something like
```js
<form onSubmit={handleSubmit}>
  <SelectList />
</form>
```
I'm personally not a fan of uncontrolled forms, but some people use them. We could possibly take a stronger opinion there in the future.

docs/src/SelectList.doc.js
dangerismycat: Looks great, though agreed with Alberto that we also need backticks around `id` to call out that it's a prop

ponori: Ah, I missed piece of feedback - sorry and thanks!

docs/src/SelectList.doc.js
ponori: Ah, perfect - good call. OK, just made that change.

docs/src/SelectList.doc.js
ponori: 👍

docs/src/SelectList.doc.js
```diff
@@ -20,185 +41,442 @@ card(
...
+  ]}
+  label='Default currency'
+  helperText='Product prices in your data source without an ISO currency code will default to this currency'
+  size='lg'
+/>`}
+      />
+    </MainSection.Subsection>
+    <MainSection.Subsection
+      title="Controlled component"
+      description="The controlled variant allows form data to be handled by the component. You will ned to use this pattern in situations where you want to explicitly set `value` or `placeholder` props."
```
dangerismycat, line 389: "SelectList must be used as a controlled component when the `placeholder` or `value` props are needed. When used in this manner, `onChange` and `value` are required, while `placeholder` is optional."

docs/src/SelectList.doc.js
```diff
@@ -20,185 +41,442 @@ card(
...
-    id="helperText"
-    name="Example: Helper Text"
-    description={`Whenever you want to provide more information about a form field, you should use \`helperText\`.`}
-    defaultCode={`
+  <MainSection name="Variants">
+    <MainSection.Subsection title="Size">
+      <MainSection.Card
+        cardSize="md"
+        title="Large"
+        description={`Use \`lg\` as the default size within Pinterest products.`}
```
dangerismycat, line 326: We should probably avoid the word "default" here since `'lg'` isn't actually the default value for the prop. Something like "`'lg'` is the recommended size for all Pinterest products"

dangerismycat (approved):
Looks great! A couple of minor copy updates needed, otherwise this is good to go!

docs/src/SelectList.doc.js
ponori: Great suggestion. Thank you!

docs/src/SelectList.doc.js
ponori: I think we should loop back to this as the recommendation is different between Pinner and M10n. Mind if I backlog this as something for us to revisit as an improvement?

ponori: And yeah, I think the existing language is deceptive as well, but my thinking is to just keep it as-is and have Ashley/Annie re-evaluate in a future sprint.

dangerismycat: Right, it's more about not using the word "default" since that has specific meaning within React, therefore leading to unexpected behavior and UI bugs. Anything that avoids using that word is fine by me.

ponori: Yeah, fair enough. I kept the original sentence and switched out `default` for `standard` - just to keep the changes minimal. Thanks Ryan!

docs/src/SelectList.doc.js
```diff
@@ -1,16 +1,37 @@
...
 import PropTable from './components/PropTable.js';
 import PageHeader from './components/PageHeader.js';
+import MainSection from './components/MainSection.js';
+import FeedbackCallout from './components/FeedbackCallout.js';
 
 const cards: Array<Node> = [];
 const card = (c) => cards.push(c);
 
+card(
+  <FeedbackCallout link="https://docs.google.com/forms/d/e/1FAIpQLSe7h8kVcD7QqvPvjkE8s8WvnuFfhYvAEQ6L7tZwPgHjJPAbSw/viewform?usp=pp_url&entry.847151274=SelectList" />,
```
ayeshakmaz, line 12: This just needs to be updated to `componentName="SelectList"` (no more link prop) (and add it to the survey)

docs/src/SelectList.doc.js
```diff
@@ -20,185 +40,456 @@ card(
...
         name: 'disabled',
         type: 'boolean',
+        description: `Used to disable the entire SelectList.`,
         defaultValue: 'false',
       },
       {
         name: 'errorMessage',
         type: 'string',
-        href: 'exampleWithError',
+        description: `Used to communicate error information to the user. Be sure to localize the text. See the [error message](##Error-message) variant to learn more.`,
```
ayeshakmaz, line 49: Extra # here, breaking the link

ayeshakmaz (approved):
:shipit: looks great!!

PR #1410, opened 2021-03-05
Module: Fix border-radius + misc cleanup

dangerismycat:
Per https://jira.pinadmin.com/browse/PDS-2601, Module was using the wrong border-radius. This PR updates that value. It also includes misc cleanup to simplify markup, lint/prettify the docs examples code, etc.

## PR Changes

- Border radius to 16px

DESIGN | BEFORE | AFTER
------------- | ------------ | -------------
image | image | image
image | image | image

- Refactor Module and subModules props so shared ones are in separate files but non-shared ones stay in the same file. Alphabetize.
- Refactor Module and subModules to replace Flex with Box when possible
- Refactor Module.Expandable/ExpandableItem to replace misused Fragments with Box
- Docs: cleanup for better reading and prop alphabetization.

### DIFF
--- a/docs/src/Module.doc.js
+++ b/docs/src/Module.doc.js
@@ -132,15 +132,11 @@ card(
 function ModuleExample() {
   return (
     <Flex direction="column" gap={2} maxWidth={800}>
-      <Module
-        id="ModuleExample - default - 1"
-        >
+      <Module id="ModuleExample - default - 1">
         <Text size="md">This is example content.</Text>
       </Module>
-      <Module
-        id="ModuleExample - default - 2"
-        title="Title"
-        >
+
+      <Module id="ModuleExample - default - 2" title="Title">
         <Text size="md">This is example content.</Text>
       </Module>
     </Flex>
@@ -162,12 +158,12 @@ card(
     defaultCode={`
 function ModuleExample() {
   return (
-    <Box maxWidth={800} padding={2} column={12}>
+    <Box column={12} maxWidth={800} padding={2}>
       <Module
-        id="ModuleExample - icon"
-        title="Title"
         icon="lock"
         iconAccessibilityLabel="Module Locked - check permission settings"
+        id="ModuleExample - icon"
+        title="Title"
         >
         <Text size="md">This is example content.</Text>
       </Module>
@@ -185,23 +181,25 @@ card(
     defaultCode={`
 function ModuleExample() {
   const [value, setValue] = React.useState('');
+
   return (
-    <Box maxWidth={800} padding={2} column={12}>
+    <Box column={12} maxWidth={800} padding={2}>
       <Module
         id="ModuleExample - error"
         title="Personal Info"
         type={!value ? "error" : "info"}
-        >
-        <Box marginBottom={4}>
+      >
+        <Flex direction="column" gap={4}>
           <Text size="md">This is example content.</Text>
-        </Box>
-        <TextField
-          id="first-name"
-          errorMessage={!value ? "This field can't be blank!" : null}
-          onChange={({ value }) => setValue(value)}
-          label="Enter Your Name"
-          value={value}
-        />
+
+          <TextField
+            errorMessage={!value ? "This field can't be blank!" : null}
+            id="first-name"
+            label="Enter Your Name"
+            onChange={({ value }) => setValue(value)}
+            value={value}
+          />
+        </Flex>
       </Module>
     </Box>
   );
@@ -218,16 +216,16 @@ card(
     defaultCode={`
 function ModuleExample1() {
   return (
-    <Box maxWidth={800} padding={2} column={12} >
+    <Box column={12} maxWidth={800} padding={2}>
       <Module.Expandable
-        id="ModuleExample - default"
         accessibilityExpandLabel="Expand the module"
         accessibilityCollapseLabel="Collapse the module"
+        id="ModuleExample - default"
         items={[
           {
-            title: 'Title',
-            summary: ['summary1', 'summary2', 'summary3'],
             children: <Text size="md">Children1</Text>,
+            summary: ['summary1', 'summary2', 'summary3'],
+            title: 'Title',
           }]}>
       </Module.Expandable>
     </Box>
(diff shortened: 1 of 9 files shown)


docs/src/Module.doc.js
```diff
@@ -132,15 +132,11 @@ card(
 function ModuleExample() {
   return (
     <Flex direction="column" gap={2} maxWidth={800}>
-      <Module
-        id="ModuleExample - default - 1"
-        >
+      <Module id="ModuleExample - default - 1">
```
dangerismycat, line 135: nothing changed, just condensing

docs/src/Module.doc.js
```diff
@@ -162,12 +158,12 @@ card(
     defaultCode={`
 function ModuleExample() {
   return (
-    <Box maxWidth={800} padding={2} column={12}>
+    <Box column={12} maxWidth={800} padding={2}>
```
dangerismycat, line 161: alphabetizin' (here and throughout, there are a lot of these)

docs/src/Module.doc.js
```diff
@@ -352,16 +351,18 @@ function ModuleExample5() {
       'second-1': 1,
   }
   return (
-    <Box maxWidth={800} padding={2} column={12}>
-      <Flex direction='column' gap={3}>
-        <Box padding={1} borderStyle='sm'>
-          <Text>Step 1</Text>
+    <Box column={12} maxWidth={800} padding={2}>
+      <Flex direction="column" gap={4}>
+        <Flex direction="column" gap={2}>
```
dangerismycat, line 356: I reworked this a bit to both simplify the existing markup, and also prettify the example UI a bit. The border was weird, spacing far too tight, etc

packages/gestalt/src/ModuleExpandable.js
```diff
@@ -1,53 +1,63 @@
...
   useEffect(() => {
-    setExpandedId(Number.isFinite(expandedIndex) ? expandedIndex : null);
+    setExpandedId(getExpandedId(expandedIndex));
   }, [expandedIndex, setExpandedId]);
 
   return (
-    <Box rounding={2} borderStyle="shadow">
-      {items.map(({ icon, iconAccessibilityLabel, title, type, summary, children }, index) => (
-        <React.Fragment key={index}>
-          <Box>
```
dangerismycat, line 28: This Box was previously needed because of leaky styling in ModuleExpandableItem — see that component for the fix

packages/gestalt/src/ModuleExpandableItem.js
```diff
@@ -1,83 +1,88 @@
...
+  id,
   isCollapsed,
   onModuleClicked,
+  summary,
+  title,
   type = 'info',
-  children,
 }: Props): Node {
   return (
-    <Fragment>
```
dangerismycat, line 29: A component whose top-level export is a Fragment — aside from combos like Button + Dropdown — is a bit of a code smell. In this case, it meant that this component _had_ to be wrapped in a Box where used, which is leaky.

packages/gestalt/src/ModuleExpandableItem.js
```diff
@@ -1,83 +1,88 @@
...
+        <TapArea
+          accessibilityControls={id}
+          accessibilityExpanded={!isCollapsed}
+          accessibilityLabel={isCollapsed ? accessibilityExpandLabel : accessibilityCollapseLabel}
+          onTap={() => {
+            onModuleClicked(!isCollapsed);
+          }}
+        >
+          <Flex>
+            <Box alignItems="baseline" display="flex" flex="grow" marginEnd={6}>
```
dangerismycat, line 45: You'd think this would be easily replaced with Flex by moving the `marginEnd` to `gap`on the parent, but it's not quite that simple. I didn't feel like spending the time to debug it now, since it doesn't really matter and this PR is already pretty bloated 😅

packages/gestalt/src/moduleTypes.js
```diff
@@ -2,44 +2,18 @@
 import { type Node } from 'react';
```
dangerismycat, line 2: I reworked this file a bit, colocating types with their components when possible (for easier reference) and restricting the types in this file to only the ones shared across components

packages/gestalt/src/Module.js
```diff
@@ -1,33 +1,39 @@
...
+  id,
   title,
   type = 'info',
-}: BaseProps): Node {
+}: Props): Node {
   return (
-    <Box id={id} rounding={2} borderStyle="shadow" direction="column">
-      {title && (
-        <Box padding={6} display="flex">
+    <Box borderStyle="shadow" id={id} padding={6} rounding={4}>
```
AlbertCarreras, line 23: Main change

dangerismycat: Yep, and the same on ModuleExpandable. Sorry, meant to call that out 😅

AlbertCarreras:
I'm gonna try review this, second attempt. Yesterday my brain got scared ha For PRs like this, I would use commits to create "stacked diffs", where each commit has a different work scope.

packages/gestalt/src/ModuleExpandableItem.js
AlbertCarreras: Nice catch and good to know.

packages/gestalt/src/ModuleExpandableItem.js
```diff
@@ -1,83 +1,88 @@
...
+                <ModuleTitle
+                  icon={icon}
+                  iconAccessibilityLabel={iconAccessibilityLabel}
+                  title={title}
+                  type={type}
+                />
               </Box>
-            )}
-          </Box>
-          <Box id={id}>
```
AlbertCarreras, line 60: good catch

packages/gestalt/src/ModuleTitle.js
```diff
@@ -21,20 +29,18 @@ export default function ModuleTitle({
       color: 'red',
     },
   };
+
+  const { color, icon: iconName } = MODULE_TYPE_ATTRIBUTES[type];
```
AlbertCarreras, line 33: nice

AlbertCarreras (approved):
LGTM, but you snaps are failing

PR #1411, opened 2021-03-05
Popover: Rename Flyout to Popover + Codemod

AlbertCarreras:


docs/src/Popover.doc.js
```diff
@@ -10,8 +10,8 @@ const card = (c) => cards.push(c);
 
 card(
   <PageHeader
-    name="Flyout"
-    description="Flyouts are similar to modals. They’re an alternative when there is less content to display,
+    name="Popover"
+    description="Popovers are similar to modals. They’re an alternative when there is less content to display,
```
dangerismycat, line 14: We should probably add a temporary note here that the component was formerly known as Flyout, something like "Popover (formerly known as Flyouts) are similar to modals."

packages/gestalt-codemods/20.0.0/__testfixtures__/flyoutAlias-rename-popover.output.js
```diff
@@ -0,0 +1,13 @@
+// @flow strict
+import React from 'react';
+import { Box, Popover } from 'gestalt';
```
dangerismycat, line 3: I don't think we want to change whatever renaming people have been doing. I think this codemod should only update the import, but not alter the renaming aspect

AlbertCarreras: I agree. That's how all renaming codemods we've been doing work though. Row/Stack, ScrollableContainer, GestaltProvider, etc.  I tried trying to keep the alias but I couldn't figure out. I'd try again. If it's too cumbersome, I think we just need to grep for `Flyout as` and see the use cases.

docs/src/Popover.doc.js
AlbertCarreras: Got it.

packages/gestalt-codemods/20.0.0/__testfixtures__/flyoutAlias-rename-popover.output.js
AlbertCarreras: DONE, I kept aliases.

docs/src/Popover.doc.js
AlbertCarreras: DONE

dangerismycat (approved):
Nice work figuring out the alias issue!

PR #1412, opened 2021-03-05
Typeahead: Add `disabled` prop and highlight selected value on activation

simkessy:
Add the ability to disable the typeahead component 

image

When user clicks a typeahead with a value, highlight the value so you can start typing right away 
image

### DIFF
--- a/docs/src/Typeahead.doc.js
+++ b/docs/src/Typeahead.doc.js
@@ -450,4 +450,61 @@ function Example(props) {
   />,
 );
 
+card(
+  <Example
+    id="programaticallySetExample"
+    name="Set Programatically"
+    defaultCode={`
+function Example(props) {
+  const [item, setItem] = React.useState('');
+  const [selected, setSelected] = React.useState(null);
+
+  const options = Array.from(Array(20).keys()).map((item) => ({
+    value: 'Value-' + (item + 1),
+    label: 'Label-' + (item + 1),
+  }));
+
+  const handleOnChange = ({ value }) => {
+    setItem(value);
+  };
+
+  const handleSelect = ({ item }) => {
+    setSelected(item);
+  };
+
+  if (selected !== null) console.log(selected.value);
+
+  const label = 'Selected Item: ' + ((selected && selected.label) || '');
+
+  return (
+    <React.Fragment>
+      <Box marginBottom={4}>
+        <Text>Selected Item: {(selected && selected.label) || ''}</Text>
+      </Box>
+
+      <Typeahead
+        label="Typeahead Example 1"
+        id="Typeahead-example"
+        noResultText="No Results"
+        options={options}
+        placeholder="Select a Label"
+        onChange={handleOnChange}
+        onSelect={handleSelect}
+      />
+      <Typeahead
+        label="Typeahead Example 2"
+        id="Typeahead-example"
+        noResultText="No Results"
+        options={options}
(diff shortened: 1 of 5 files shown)


packages/gestalt/src/Typeahead.js
```diff
@@ -54,8 +55,9 @@ type Props = {|
 const TypeaheadWithForwardRef: React$AbstractComponent<Props, HTMLInputElement> = forwardRef<
   Props,
   HTMLInputElement,
->(function Typeahead(props, ref): Node {
-  const {
+>(function Typeahead(
+  {
```
AlbertCarreras, line 59: can we keep as it was? ForwardRefs are hard to read and all other components keep the same structure of the destructuring outside. Just for consistency across cmps.

packages/gestalt/src/TypeaheadInputField.js
```diff
@@ -47,6 +49,7 @@ const TypeaheadInputFieldWithForwardRef: React$AbstractComponent<
 > = forwardRef<Props, HTMLInputElement>(function InputField(props, ref): Node {
   const {
     id,
+    disabled = false,
```
AlbertCarreras, line 52: technically the value was set up on the props or default value on the parent?

AlbertCarreras:
LGTM, just a few nits.

docs/src/Typeahead.doc.js
```diff
@@ -282,6 +288,26 @@ function TypeaheadExample() {
...
+  <Example
+    id="disabledExample"
+    name="Disabled"
+    defaultCode={`
+function TypeaheadExample() {
+  return (
+      <Typeahead
+        disabled
+        label="This typeahead is disabled"
+        id="favorite-shape"
```
AlbertCarreras, line 301: change the id cos it's triggering the previous example @simkessy

docs/src/Typeahead.doc.js
```diff
@@ -282,6 +288,26 @@ function TypeaheadExample() {
   />,
 );
 
+card(
+  <Example
+    id="disabledExample"
+    name="Disabled"
+    defaultCode={`
+function TypeaheadExample() {
```
dangerismycat, line 296: A nice additional to this example would be a button that controls the disabled state, so the user can play with it and see how that state is controlled programmatically

ashleyseto:
If we do need a disabled option, we do need this to work with all Typeaheads including the tag typeahead. Can someone please confirm we're including this?

docs/src/Typeahead.doc.js
ashleyseto: I'd hold off on this until we're doing more complex interactive examples.

AlbertCarreras:
@ashleyseto @simkessy 

Are we supposed to capture values when disabled? 

image

If so, should the clean button appear? 

A general issue is that we can still hover the mouse and see the hand pointer. WIth Textfield and TextArea there is not pointer on disabled.

For tags, Tags are still dismissible on disable.

image

simkessy:
> If we do need a disabled option, we do need this to work with all Typeaheads including the tag typeahead. Can someone please confirm we're including this?

Who can confirm this?

dangerismycat:
@simkessy Thanks for the PR! We've decided to freeze work on Typeahead for the moment, as it needs more comprehensive design planning. We'll take this need into consideration in that work.

PR #1413, opened 2021-03-05
FeedbackCallout: Simplify API

dangerismycat:
These are all using the same URL, with the only change being the component name query param. This PR makes FeedbackCallout simpler to use and eliminates possible copy/paste errors with the link.


ayeshakmaz (approved):
Thanks!

PR #1415, opened 2021-03-08
Popover: fix Codemod 20.0.0

AlbertCarreras:


packages/gestalt-codemods/20.0.0/flyout-rename-popover.js
```diff
@@ -65,10 +68,10 @@ export default function transformer(file, api) {
       }
 
       // Sort attributes alphabetically
-      node.openingElement.attributes.sort((a, b) => a.name.name.localeCompare(b.name.name));
+      // node.openingElement.attributes.sort((a, b) => a.name.name.localeCompare(b.name.name));
```
dangerismycat, line 71: This should be deleted if it's not needed

dangerismycat (approved):

PR #1418, opened 2021-03-09
Image: add elementtiming attribute for profiling

jackhsu978:
This PR adds the `elementTiming` attribute to `` tag rendered by the `<Image>` component so that it is possible to use the [Element Timing API](https://developer.mozilla.org/en-US/docs/Web/API/Element_timing_API) to get the image timings such as [`loadTime`](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceElementTiming/loadTime) and [`renderTime`](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceElementTiming/renderTime)

Note that while the actual HTML attribute is all lowercased, the `elementTiming` prop is camelCased since it is an established Gestalt convention.

Screen Shot 2021-03-09 at 12 11 25 PM

### DIFF
--- a/docs/src/Image.doc.js
+++ b/docs/src/Image.doc.js
@@ -39,6 +39,12 @@ card(
         href: 'placeholders',
       },
       {
+        name: 'elementtiming',
+        type: 'string',
+        description:
+          'HTML attribute For performance profiling. See https://developer.mozilla.org/en-US/docs/Web/API/Element_timing_API',
+      },
+      {
         name: 'fit',
         type: `"cover" | "contain" | "none"`,
         defaultValue: 'none',
--- a/docs/src/Image.doc.js
+++ b/docs/src/Image.doc.js
@@ -41,7 +41,7 @@ card(
       {
         name: 'elementTiming',
         type: 'string',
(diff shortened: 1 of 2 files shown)


chrislloyd (approved):

docs/src/Image.doc.js
```diff
@@ -38,6 +38,12 @@ card(
         required: true,
         href: 'placeholders',
       },
+      {
+        name: 'elementTiming',
+        type: 'string',
+        description:
+          'HTML attribute For performance profiling. See https://developer.mozilla.org/en-US/docs/Web/API/Element_timing_API',
```
ayeshakmaz, line 45: Do these need to be unique, specific strings to ensure accurate testing? Would be great to clarify that here

jackhsu978: @ayeshakmaz I want to keep the documentation as light as possible because this is for a very specific use case that requires reading the element timing API documentation

packages/gestalt/src/Image.js
```diff
@@ -132,6 +135,7 @@ export default class Image extends PureComponent<Props> {
         <img
           alt={alt}
           className={styles.img}
+          elementtiming={elementTiming}
```
christianvuerings, line 138: Might be good calling out that this will only happen when `fit` is not set to `cover` or `contain`

See `shouldScaleImage`

jackhsu978: updated. thanks!

PR #1422, opened 2021-03-11
Internal: Run codemod to remove unneeded React imports

dangerismycat:
As a follow-on to [upgrading our JSX transform](https://github.com/pinterest/gestalt/pull/1420), this PR runs the official codemod (`npx react-codemod update-react-imports`) to update our imports.

Unfortunately I couldn't find any prettier or eslint rules related to import styles, so we now have some cases of less-than-ideal code style, e.g.
```js
import React, { type Node, useState } from 'react';
```
is now
```js
import type { Node } from 'react';
```
Hopefully we can find a prettier/eslint rule in the future to correct this, but it shouldn't be a blocker for now.


AlbertCarreras (changes requested):
Tests are failing

PR #1423, opened 2021-03-12
PageHeader: Add component and documentation

ayeshakmaz:
New PageHeader cmp

## Test Plan

Added some snapshot tests


packages/gestalt/src/PageHeader.css
```diff
@@ -0,0 +1,3 @@
+.customClass {
```
dangerismycat, line 1: These two files aren't needed and should be deleted

packages/gestalt/src/PageHeader.js
```diff
@@ -0,0 +1,72 @@
...
+  };
+  return (
+    <Flex flex="grow" justifyContent="center" maxWidth="100%">
+      <Flex justifyContent="between" alignItems="center" width={maxWidth}>
+        <Flex direction="column" gap={subtext ? 2 : 0} minWidth={0}>
+          <Box marginEnd={4}>
+            <Heading truncate>{title}</Heading>
+          </Box>
+          {subtext && (
+            <Box marginEnd={4}>
```
dangerismycat, line 58: Rather than needing two Boxes with the same marginEnd, move this conditional within the Box wrapping `title`

packages/gestalt/src/PageHeader.js
```diff
@@ -0,0 +1,72 @@
...
+export default function PageHeader({
+  maxWidth = '100%',
+  primaryAction,
+  secondaryAction,
+  subtext,
+  title,
+}: Props): Node {
+  const renderAction = (action?: Node): Node => {
+    // This is technically all temporary, until we create a proper Menu component
+    // Then, we could type check using Flow for Buttons, Links, or Menus
```
dangerismycat, line 26: That's true, though we can use Flow right now to check for `IconButton | Button | Link | Tooltip` and obviate all this logic

packages/gestalt/src/PageHeader.js
```diff
@@ -0,0 +1,72 @@
...
+|};
+
+export default function PageHeader({
+  maxWidth = '100%',
+  primaryAction,
+  secondaryAction,
+  subtext,
+  title,
+}: Props): Node {
+  const renderAction = (action?: Node): Node => {
```
dangerismycat, line 24: We typically want to avoid sub-render methods like this. If something is separate enough to necessitate a method, it should be a separate component.

packages/gestalt/src/PageHeader.js
```diff
@@ -0,0 +1,72 @@
...
+        ) {
+          allowedChildFound = true;
+        }
+      });
+    }
+    if (!allowedChildFound) {
+      throw new Error('PageHeader actions must be of type Button, IconButton, Link, or Tooltip.');
+    }
+
+    return allowedChildFound ? action : <div />;
```
dangerismycat, line 48: I would think we'd want to return null  here if the action is invalid? We don't want random divs for no reason

docs/src/PageHeader.doc.js
```diff
@@ -0,0 +1,645 @@
...
+card(
+  <PropTable
+    Component={PageHeader}
+    props={[
+      {
+        name: 'title',
+        type: 'string',
+        required: true,
+        description:
+          'Page title, will always be an level 1 heading. Content should be [Localized](/Localized).',
```
dangerismycat, line 123: typo: "a level 1 heading"

docs/src/PageHeader.doc.js
```diff
@@ -0,0 +1,645 @@
...
+      },
+      {
+        name: 'maxWidth',
+        type: `number | string`,
+        description: `Use numbers for pixels: maxWidth={100} and strings for percentages: maxWidth="100%."`,
+      },
+      {
+        name: 'primaryAction',
+        type: `React Node`,
```
dangerismycat, line 132: These should be `React.Node`, though I think we should type more specifically for the valid components we accept. It's not a fair API to have secretly valid types that aren't reflected in the docs.

docs/src/PageHeader.doc.js
```diff
@@ -0,0 +1,645 @@
...
+        name: 'title',
+        type: 'string',
+        required: true,
+        description:
+          'Page title, will always be an level 1 heading. Content should be [Localized](/Localized).',
+      },
+      {
+        name: 'maxWidth',
+        type: `number | string`,
+        description: `Use numbers for pixels: maxWidth={100} and strings for percentages: maxWidth="100%."`,
```
dangerismycat, line 128: nit: the two examples should probably have backticks

docs/src/PageHeader.doc.js
```diff
@@ -0,0 +1,645 @@
...
+        defaultCode={`
+<PageHeader
+  title="Product groups"
+  subtext="S. E. All products USD"
+  primaryAction={<Button color="red" text="Create product group" />}
+  secondaryAction={<Button color="red" text="Promote" />}
+/>;
+        `}
+      />
+      <MainSection.Card
```
dangerismycat, line 280: visual bug on this one:
Screen Shot 2021-03-12 at 1 13 03 PM

docs/src/PageHeader.doc.js
```diff
@@ -0,0 +1,645 @@
...
+        defaultCode={`
+<PageHeader
+  title="Product groups"
+  subtext="Product groups are created in order to relate certain products together for tracking purposes"
+  primaryAction={<Button color="red" text="Create product group" />}
+  secondaryAction={<Button text="Promote" />}
+/>;
+        `}
+      />
+      <MainSection.Card
```
dangerismycat, line 295: It seems like a bug that "Settings" is center-aligned here?

docs/src/PageHeader.doc.js
```diff
@@ -0,0 +1,645 @@
...
+  <MainSection
+    name="Localization"
+    description={`Be sure to localize the \`title\`, \`subtext\` and actions within PageHeader.`}
+  />,
+);
+
+card(
+  <MainSection name="Variants">
+    <MainSection.Subsection
+      description={`PageHeader can support an optional \`primaryAction\`, which can be a [Button](Button), a [Link](/Link) or an [IconButton](/IconButton) with a [Tooltip](/Tooltip) and optional [Dropdown](/Dropdown).`}
```
dangerismycat, line 403: This sentence isn't clear — consider breaking it into two sentences for readability

dangerismycat, line 403: I think we can be more direct here: "PageHeader supports an optional…"

docs/src/PageHeader.doc.js
```diff
@@ -0,0 +1,645 @@
...
+      />
+    </Flex>
+  );
+}
+        `}
+      />
+    </MainSection.Subsection>
+
+    <MainSection.Subsection
+      description={`\`subtext\` should be used to add metadata about the content on the page, not to describe the page itself.`}
```
dangerismycat, line 589: This kinda reads more like a Do/Don't? It's fine here for now but a potential future improvement

docs/src/PageHeader.doc.js
```diff
@@ -0,0 +1,645 @@
...
+  title="Create product group"
+  subtext="2,131 products"
+  primaryAction={<Button color="red" text="Quick create" />}
+/>;
+        `}
+      />
+    </MainSection.Subsection>
+
+    <MainSection.Subsection
+      description={`\`maxWidth\` should be set when the content of the page is centered and has a maximum width itself.`}
```
dangerismycat, line 605: I'm not sure if this is visually communicating what we're trying to get across? Would a contrasting background help demonstrate the page max width?
Screen Shot 2021-03-12 at 1 18 56 PM

packages/gestalt/src/PageHeader.js
ayeshakmaz: doi, great call!

packages/gestalt/src/PageHeader.js
ayeshakmaz: But how do we do that and account for the React.Fragment needed when using Dropdowns?

ashleyseto:
For the page header, we're using heading md in the designs. Heading lg is very, very big.

We may want to do a background wash of grey just to be able to see the full header component. Right now it is 80px high in spec and 32px of padding left and right.

docs/src/PageHeader.doc.js
ayeshakmaz: It's also a Don't

docs/src/PageHeader.doc.js
```diff
@@ -0,0 +1,645 @@
...
+                    accessibilityExpanded={open}
+                    accessibilityLabel=""
+                    icon="ellipsis"
+                    iconColor="darkGray"
+                    selected={open}
+                    onClick={() => setOpen((prevVal) => !prevVal)}
+                    ref={anchorRef}
+                  />
+                </Tooltip>
+                {open && (
```
AlbertCarreras, line 49: Can we simplify this dropdown with 1-3 product-related actions instead?

packages/gestalt/src/PageHeader.flowtest.js
```diff
@@ -0,0 +1,8 @@
+// @flow strict
+import React from 'react';
+import PageHeader from './PageHeader.js';
+
+const Valid = <PageHeader />;
```
AlbertCarreras, line 5: how's valid if Title is missing?

docs/src/PageHeader.doc.js
AlbertCarreras: nit Content should be [localized]

packages/gestalt/src/PageHeader.js
```diff
@@ -0,0 +1,54 @@
...
+import Link from './Link.js';
+import Tooltip from './Tooltip.js';
+import Flex from './Flex.js';
+import Heading from './Heading.js';
+import Text from './Text.js';
+import { type Dimension } from './boxTypes.js';
+
+type Props = {|
+  title: string,
+  maxWidth?: Dimension,
```
AlbertCarreras, line 15: no action needed, but Dimension is just number or string, so we can pass anything else such as 'fit' etc.

docs/src/PageHeader.doc.js
```diff
@@ -0,0 +1,643 @@
...
+card(
+  <PropTable
+    Component={PageHeader}
+    props={[
+      {
+        name: 'title',
+        type: 'string',
+        required: true,
+        description:
+          'Page title, will always be a level 1 heading. Content should be [Localized](/Localized).',
```
AlbertCarreras, line 102: Content should be [localized]

docs/src/PageHeader.doc.js
```diff
@@ -0,0 +1,643 @@
...
+            );
+          } else {
+            setSelected((selected) => [...selected, item]);
+          }
+        };
+
+        return (
+          <PageHeader
+            title="Product groups"
+            subtext="S. E. All products USD"
```
AlbertCarreras, line 34: Could we add a subtext that is easier to understand as metadata? `S. E. All products USD` I have no idea what this means haha

docs/src/PageHeader.doc.js
```diff
@@ -0,0 +1,643 @@
...
+      },
+      {
+        name: 'maxWidth',
+        type: `number | string`,
+        description: `Use numbers for pixels: \`maxWidth={100}\` and strings for percentages: \`maxWidth="100%"\`.`,
+      },
+      {
+        name: 'primaryAction',
+        type: `Element<typeof Button | typeof IconButton | typeof Link | typeof Tooltip>`,
+        description: `The primary action of the page. Can be a [Button](/Button), [IconButton](/IconButton), [Link](/Link), [Tooltip](/Tooltip) surrounding an IconButton or a combination of an IconButton, Tooltip, and [Dropdown](/Dropdown).`,
```
AlbertCarreras, line 112: really nit, no oxford commas haha

docs/src/PageHeader.doc.js
```diff
@@ -0,0 +1,643 @@
...
+  }
+/>;
+        `}
+        shaded
+      />
+      <MainSection.Card
+        cardSize="lg"
+        type="do"
+        description={`
+        Plan for most PageHeaders to be full width. A \`maxWidth\` should only be supplied when the content of the page is center aligned.
```
AlbertCarreras, line 166: What's the difference in this example from the previous ones? I'm not sure I'm connecting the text with the example.

AlbertCarreras: I think I got it from looking at the Variant. It's a tricky one.

AlbertCarreras: nit comment: I'd replace texfield with name, address, etc.

docs/src/PageHeader.doc.js
```diff
@@ -0,0 +1,643 @@
...
+        shaded
+      />
+    </MainSection.Subsection>
+  </MainSection>,
+);
+
+card(
+  <MainSection
+    name="Accessibility"
+    description={`Be sure to follow any accessibility guidelines for the components used within PageHeader. The heading within PageHeader will be rendered as a level 1 heading, so ensure there are no other level 1 headings present on the page. `}
```
AlbertCarreras, line 368: Could we add here some Resource links? Probably not, just to make it a little bit more juicy.

docs/src/PageHeader.doc.js
```diff
@@ -0,0 +1,643 @@
...
+/>;
+  `}
+    />
+  </MainSection>,
+);
+
+card(
+  <MainSection name="Variants">
+    <MainSection.Subsection
+      description={`PageHeader supports an optional \`primaryAction\`. It can be a [Button](Button), a [Link](/Link) or an [IconButton](/IconButton) with a [Tooltip](/Tooltip) and optional [Dropdown](/Dropdown). Any Buttons or IconButtons should be \`size="lg"\`.`}
```
AlbertCarreras, line 394: Are we missing the IconButton with no Tooltip variant? 
image

AlbertCarreras: nit: no article "a"

docs/src/PageHeader.doc.js
AlbertCarreras: Maybe an example in combination with Heading 2, 3 could help here. Also, what about adding See Related to learn more about the Heading and HeaderPage. Or maybe we could add `Heading should be used to create level 2-6 headings on a page. If a level 1 heading is needed, use PageHeader` here as well as it's pretty connected.

docs/src/Box.doc.js
```diff
@@ -530,39 +530,17 @@ card(
...
-        title='borderStyle="shadow"'
-        defaultCode={`<Box
-        width={60}
-        height={60}
-        rounding="circle"
-        color="white"
-        borderStyle="shadow"
-      />`}
-      />
+      <CombinationNew borderStyle={['sm', 'lg', 'shadow']}>
```
AlbertCarreras, line 533: Nice add-on haha

docs/src/components/PageHeader.js
```diff
@@ -61,7 +63,12 @@ export default function ComponentHeader({
       {description && <Markdown text={description} />}
       {defaultCode && (
         <Box marginTop={8}>
-          <MainSection.Card cardSize="lg" showCode={false} defaultCode={defaultCode} />
+          <MainSection.Card
+            cardSize="lg"
+            showCode={false}
+            defaultCode={defaultCode}
+            shaded={shadedCodeExample}
```
AlbertCarreras, line 70: what is this doing?            `shaded={shadedCodeExample}`

packages/gestalt/src/PageHeader.flowtest.js
```diff
@@ -0,0 +1,8 @@
+// @flow strict
+import React from 'react';
+import PageHeader from './PageHeader.js';
+
+const Valid = <PageHeader title="Settings" />;
```
AlbertCarreras, line 5: how can it be Valid if there's no title prop?

image

packages/gestalt/src/PageHeader.js
```diff
@@ -0,0 +1,54 @@
...
+  primaryAction,
+  secondaryAction,
+  subtext,
+  title,
+}: Props): Node {
+  return (
+    <Box color="white" paddingX={8} width="100%">
+      <Flex flex="grow" justifyContent="center" maxWidth="100%">
+        <Flex justifyContent="between" alignItems="center" width={maxWidth}>
+          <Box marginEnd={4} minWidth={0} marginTop={2} marginBottom={2}>
```
AlbertCarreras, line 32: `minWidth={0}` ? what is it doing?

AlbertCarreras: just curious

packages/gestalt/src/PageHeader.js
```diff
@@ -0,0 +1,54 @@
...
+              <Box display="flex" alignItems="center" marginTop={4} marginBottom={4} height="48px">
+                {primaryAction}
+              </Box>
+            </Flex>
+          )}
+        </Flex>
+      </Flex>
+    </Box>
+  );
+}
```
AlbertCarreras, line 54: This might be an unnecessary edge case, but should we consider a minWidth for the Title? 
image

docs/src/PageHeader.doc.js
```diff
@@ -0,0 +1,643 @@
...
+      <PageHeader
+        title="Product groups"
+        subtext="S. E. All products USD"
+        primaryAction={<Button color="red" size="lg" text="Create product group" />}
+        secondaryAction={<Button text="Promote" size="lg"/>}
+      />
+      <Divider />
+      <PageHeader
+        title="Custom reports"
+        primaryAction={<Button color="red" text="Create new report" />}
```
AlbertCarreras, line 523: `size ="lg"`

packages/gestalt/src/PageHeader.js
```diff
@@ -0,0 +1,54 @@
...
+            {subtext && (
+              <Box marginTop={2}>
+                <Text truncate>{subtext}</Text>
+              </Box>
+            )}
+          </Box>
+          {primaryAction && (
+            <Flex gap={2} flex="none" alignItems="center">
+              {secondaryAction}
+              <Box display="flex" alignItems="center" marginTop={4} marginBottom={4} height="48px">
```
AlbertCarreras, line 45: maybe unnecessary, but could we add a comment explaining why height="48px"? My assumption is that we want to keep the same height for link texts. etc

packages/gestalt/src/PageHeader.js
```diff
@@ -0,0 +1,54 @@
...
+            </Heading>
+            {subtext && (
+              <Box marginTop={2}>
+                <Text truncate>{subtext}</Text>
+              </Box>
+            )}
+          </Box>
+          {primaryAction && (
+            <Flex gap={2} flex="none" alignItems="center">
+              {secondaryAction}
```
AlbertCarreras, line 44: why is this one not inside Box?

docs/src/PageHeader.doc.js
ayeshakmaz: In this case it's the name of the Product Group. I think this is a real example from business, @ashleyseto is there something else we could put here?

packages/gestalt/src/PageHeader.test.js
```diff
@@ -0,0 +1,16 @@
...
+  it('renders', () => {
+    const tree = create(<PageHeader title="Settings" />).toJSON();
+    expect(tree).toMatchSnapshot();
+  });
+
+  it('renders subtext', () => {
+    const tree = create(<PageHeader title="Settings" subtext="5 followers" />).toJSON();
+    expect(tree).toMatchSnapshot();
+  });
+});
```
AlbertCarreras, line 16: I guess this is enough. Considering there are so many combinations of primary/secondary actions.

docs/src/components/PageHeader.js
ayeshakmaz: allows the page header example to have a shaded background

packages/gestalt/src/PageHeader.js
ayeshakmaz: It's needed to ensure that truncation happens with Flex https://css-tricks.com/using-flexbox-and-text-ellipsis-together/

packages/gestalt/src/PageHeader.js
ayeshakmaz: @ashleyseto do we want a min-width for the title?

packages/gestalt/src/PageHeader.js
ayeshakmaz: Would be redundant, since we can't have a secondary without a primary, putting it on primary is enough to set the height across since it's all flex'd

AlbertCarreras (approved):
LGTM, minor comments but good to go when you feel it's ready!

docs/src/components/PageHeader.js
AlbertCarreras: Gotcha, this 
image

packages/gestalt/src/PageHeader.js
AlbertCarreras: Nice trick. If it's a well-known thing I guess ain't needed but maybe adding  a comment to it might help to document the intention of `minWidth={0}` and the expected behavior.

ashleyseto (approved):

PR #1428, opened 2021-03-12
ScrollBoundaryContainer: fix positioning within when no available spa…

AlbertCarreras:
…ce condition is met

Popovers within a ScrollBoundaryContainer have much less space around to position themselves. This PR fixes Popover's positioning within ScrollBoundaryContainer (Including Modals and Sheet) but limiting the positioning directions to just "Up" and "Down" if no available space is available. 

Be limiting to "Up" and "Down", the Popover won't overflow to the sides ("left", "right") which could lead to hiding the Popover's overflow "up" or "down". Therefore, the Popover will make the container overflow with a scroll bar.

To see bug in action: 
image
https://www.loom.com/share/d464133f3dbd40f08921259e9af95c9e

To see fix in action: https://deploy-preview-1421--gestalt.netlify.app/Popover#With-Layer

BEFORE / AFTER video: https://www.loom.com/share/867634ec3c5547628bbd9c8926626b4d?from_recorder=1

### DIFF
--- a/packages/gestalt/src/utils/positioningUtils.js
+++ b/packages/gestalt/src/utils/positioningUtils.js
@@ -26,6 +26,11 @@ const SPACES_INDEX_MAP = {
   '3': 'left',
 };
 
+const NO_SPACE_CONDITION_INDEX_MAP = {
+  '0': 'up',
+  '1': 'down',
+};
+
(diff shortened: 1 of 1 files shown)


dangerismycat (approved):
LGTM!

PR #1430, opened 2021-03-16
Tooltip: Only delay visibility on interaction if `link` present

dangerismycat:
[bugfix](https://jira.pinadmin.com/browse/BUG-122739)

We currently delay the fade-out on mouseLeave/blur to allow the user to interact with the Tooltip content. However, this is only necessary if a `link` is present, and the delay leads to a confusing UI when multiple tooltips are used in a row. Long-term, we should create a more robust solution to allow user interaction with the tooltip content while not needing a delay. As a short-term fix, this PR only applies the delay if the `link` prop is used.


ayeshakmaz (approved):
LGTM 👍

PR #1439, opened 2021-03-25
TapArea: Add onMouseDown and onMouseUp handlers

dangerismycat:
I need these handlers for the Tabs redesign (and they're generally useful).

Note that these are specifically for mouse events only, hence using `onMouseDown` instead of something like `onTapStart`. The Tabs design (and likely others) calls for different interaction behavior between desktop and mobile, so it's not a valid assumption to lump `onMouseDown` and `onTouchStart` behavior together. If we need `onTouchStart` behavior in the future, we can add that as well (probably also with `onTouchMove` and `onTouchCancel`).


docs/src/TapArea.doc.js
```diff
@@ -130,14 +130,29 @@ card(
...
+        description: ['Callback fired when a click event ends.'],
+      },
       {
         name: 'onMouseEnter',
         type:
           '({ event: SyntheticMouseEvent<HTMLDivElement> | SyntheticMouseEvent<HTMLAnchorElement>, {| disableOnNavigation?: () => void |}> }) => void',
         required: false,
         defaultValue: null,
         description: ['Callback fired when a mouse pointer moves onto a TapArea component.'],
-        href: 'basicExample',
```
dangerismycat, line 140: This example doesn't exist, so this link is dead

packages/gestalt/src/InternalLink.js
```diff
@@ -166,47 +166,31 @@ const InternalLinkWithForwardRef: AbstractComponent<Props, HTMLAnchorElement> =
       id={id}
       onContextMenu={isTapArea ? (event) => event.preventDefault() : null}
       onBlur={(event) => {
-        if (onBlur) {
-          onBlur({ event });
-        }
+        onBlur?.({ event });
```
dangerismycat, line 169: cleaned up all these if statements with optional chaining

AlbertCarreras (approved):

PR #1440, opened 2021-03-25
Tabs: Add experimental redesign

dangerismycat:
[Design doc](https://www.figma.com/file/vlihOKdfamsqUQhd9ybkQv/Tab-component-redesign?node-id=30%3A2387)

This PR adds a new tab variant for the Tabs component, accessible by passing a new undocumented prop, `_dangerouslyUseV2`. The rest of the component API remains the same, with one exception: the `indicator` field of each tab object can now also accept a number. Though not used initially, this will be used in future experiments as a replacement for the notification dot. (For now we need to support both; we _may_ be able to deprecate the dot variant in the future.)

### Test Plan

Add `_dangerouslyUseV2={true}` to the docs example to see the new UI. Note that we're dropping support for the `'md'` size, so both examples will be identical. Try passing a number instead of `'dot'` as the `indicator` for a tab. Also be sure to open DevTools and view as mobile to observe the different touch interactions.

### DIFF
--- a/packages/gestalt/src/Tabs.js
+++ b/packages/gestalt/src/Tabs.js
@@ -4,48 +4,103 @@ import PropTypes from 'prop-types';
 import Box from './Box.js';
 import Flex from './Flex.js';
 import Link from './Link.js';
+import TapArea from './TapArea.js';
 import Text from './Text.js';
 import { type AbstractEventHandler } from './AbstractEventHandler.js';
 
 type OnChangeHandler = AbstractEventHandler<
-  SyntheticMouseEvent<HTMLAnchorElement> | SyntheticKeyboardEvent<HTMLAnchorElement>,
+  | SyntheticMouseEvent<HTMLAnchorElement>
+  | SyntheticKeyboardEvent<HTMLAnchorElement>
+  | SyntheticMouseEvent<HTMLDivElement>
+  | SyntheticKeyboardEvent<HTMLDivElement>,
   {| +activeTabIndex: number |},
 >;
 
-function Circle() {
+function Dot() {
   return (
     <Box
       color="red"
       dangerouslySetInlineStyle={{ __style: { marginTop: '1px' } }}
       height={6}
-      marginStart={2}
       rounding="circle"
       width={6}
     />
   );
 }
 
+const UNDERLINE_HEIGHT = 3;
+
+function Underline() {
+  return (
+    <Box
+      color="darkGray"
+      dangerouslySetInlineStyle={{
+        __style: {
+          borderRadius: 1.5,
+        },
+      }}
+      height={UNDERLINE_HEIGHT}
+      width="100%"
+    />
+  );
+}
+
+const COUNT_HEIGHT_PX = 16;
+
+function Count({ count }: {| count: number |}) {
+  const displayCount = count < 100 ? `${count}` : '99+';
+
+  return (
+    <Box
+      color="red"
+      dangerouslySetInlineStyle={{
+        __style: {
+          padding: `0 ${displayCount.length > 1 ? 3 : 0}px`,
+        },
+      }}
+      height={COUNT_HEIGHT_PX}
+      minWidth={COUNT_HEIGHT_PX}
+      rounding="pill"
+    >
+      <Box
+        dangerouslySetInlineStyle={{
+          __style: { padding: '0 0 1px 1px' },
+        }}
+      >
+        <Text align="center" color="white" size="sm" weight="bold">
+          {displayCount}
+        </Text>
+      </Box>
+    </Box>
+  );
+}
+
+type TabType = {|
+  href: string,
+  id?: string,
+  indicator?: 'dot' | number,
+  text: Node,
+|};
+
 function Tab({
-  children,
-  size,
   href,
-  indicator,
   id,
   index,
+  indicator,
   isActive,
   onChange,
+  size,
+  text,
 }: {|
-  children: Node,
-  size: 'md' | 'lg',
-  isActive: boolean,
-  href: string,
-  indicator?: 'dot',
+  ...TabType,
   index: number,
-  id?: string,
+  isActive: boolean,
   onChange: OnChangeHandler,
+  size: 'md' | 'lg',
 |}) {
   const [hovered, setHovered] = useState(false);
   const [focused, setFocused] = useState(false);
+
   return (
     <Box position={focused ? 'relative' : undefined}>
       <Link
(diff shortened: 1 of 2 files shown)


packages/gestalt/src/Tabs.js
```diff
@@ -4,48 +4,103 @@ import PropTypes from 'prop-types';
...
 >;
 
-function Circle() {
+function Dot() {
   return (
     <Box
       color="red"
       dangerouslySetInlineStyle={{ __style: { marginTop: '1px' } }}
       height={6}
-      marginStart={2}
```
dangerismycat, line 21: replaced at callsites with `gap`

packages/gestalt/src/Tabs.js
```diff
@@ -4,48 +4,103 @@ import PropTypes from 'prop-types';
...
+      height={UNDERLINE_HEIGHT}
+      width="100%"
+    />
+  );
+}
+
+const COUNT_HEIGHT_PX = 16;
+
+function Count({ count }: {| count: number |}) {
+  const displayCount = count < 100 ? `${count}` : '99+';
```
dangerismycat, line 51: In addition to truncating values over 100, we need to know the number of digits we're displaying, so stringifying here

packages/gestalt/src/Tabs.js
```diff
@@ -4,48 +4,103 @@ import PropTypes from 'prop-types';
...
+      >
+        <Text align="center" color="white" size="sm" weight="bold">
+          {displayCount}
+        </Text>
+      </Box>
+    </Box>
+  );
+}
+
+type TabType = {|
```
dangerismycat, line 78: DRY

packages/gestalt/src/Tabs.js
```diff
@@ -73,45 +128,137 @@ function Tab({
           rounding="pill"
           userSelect="none"
         >
-          <Text color={isActive ? 'white' : 'darkGray'} weight="bold" overflow="noWrap">
-            {children}
-          </Text>
-          {indicator === 'dot' && <Circle />}
+          <Flex alignItems="center" gap={2} justifyContent="center">
```
dangerismycat, line 131: only changing to allow removal of margin on Dot

packages/gestalt/src/Tabs.js
```diff
@@ -73,45 +128,137 @@ function Tab({
...
+function TabV2({
+  href,
+  indicator,
+  id,
+  index,
+  isActive,
+  onChange,
+  // No longer supported, will be removed when this variant ships
+  // eslint-disable-next-line no-unused-vars
+  size,
```
dangerismycat, line 156: Can't remove this yet because of Flow, but will as part of shipping the new variant

packages/gestalt/src/Tabs.js
```diff
@@ -73,45 +128,137 @@ function Tab({
...
+            userSelect="none"
+          >
+            <Flex alignItems="center" gap={2} justifyContent="center">
+              <Text color="darkGray" overflow="noWrap" weight="bold">
+                {text}
+              </Text>
+
+              {indicator === 'dot' && <Dot />}
+              {/* Flow is dumb and doesn't realize Number.isFinite will return false for a string or undefined */}
+              {typeof indicator === 'number' && Number.isFinite(indicator) && (
```
dangerismycat, line 206: `typeof` here is very 🙄

packages/gestalt/src/Tabs.js
```diff
@@ -73,45 +128,137 @@ function Tab({
...
+  size = 'md',
+  tabs,
+  wrap,
+  _dangerouslyUseV2,
+}: Props): Node {
+  const TabComponent = _dangerouslyUseV2 ? TabV2 : Tab;
+
   return (
-    <Flex alignItems="center" justifyContent="start" wrap={wrap}>
+    <Flex alignItems="center" gap={_dangerouslyUseV2 ? 4 : 0} justifyContent="start" wrap={wrap}>
```
dangerismycat, line 249: handling all spacing between tabs here so we're not leaking that spacing to the Tabs component as a whole, allowing for tighter layouts on either side of the tabs

basically making it `TAB space TAB space TAB` instead of `space TAB space TAB space TAB space`

packages/gestalt/src/Tabs.js
```diff
@@ -73,45 +128,137 @@ function Tab({
...
+
+  let color = 'white';
+  if (pressed) {
+    color = 'lightWash';
+  } else if ((hovered || focused) && !isActive) {
+    color = 'lightGray';
+  }
+
+  return (
+    <Box id={id} paddingY={3}>
```
christianvuerings, line 177: @dangerismycat In gestalt we tried to avoid extra padding around components. It make the components less composable, e.g. when `<Tabs />` is part of a `<Flex />` with a `gap`?

Totally fine in the docs example to wrap the `<Tabs />` in a `<Box paddingY={3} />` but I would avoid making the outside padding part of the component.

image

packages/gestalt/src/Tabs.js
christianvuerings: @dangerismycat does it work with `Number.isInteger()`? Or do we also need to support floats?

packages/gestalt/src/Tabs.js
dangerismycat: Makes sense. That vertical padding was part of the design — though the way this is implemented, there might be too much padding if the tabs wrap. Asking @anniemochi on Slack now

packages/gestalt/src/Tabs.js
dangerismycat: It doesn't unfortunately. We don't need to support floats, `.isFinite` is just my go-to for determining if a value is a number (and not NaN). I can update to `.isInteger` if you think that's the better choice.

packages/gestalt/src/Tabs.js
dangerismycat: Yep, this implementation is approved by Annie (as opposed to setting the vertical padding on the Tabs component as a whole, so that the rows would be closer vertically)

packages/gestalt/src/Tabs.js
```diff
@@ -4,48 +4,103 @@ import PropTypes from 'prop-types';
...
+      minWidth={COUNT_HEIGHT_PX}
+      rounding="pill"
+    >
+      <Box
+        dangerouslySetInlineStyle={{
+          __style: { padding: '0 0 1px 1px' },
+        }}
+      >
+        <Text align="center" color="white" size="sm" weight="bold">
+          {displayCount}
```
ayeshakmaz, line 71: with the addition of numbers here, I think we need an indicatorAccessibilityLabel prop to append here. The tricky thing is do we hide the number, and just use the indicatorAccessibilityLabel, or do we append indicatorAccessibilityLabel to the number

indicator={50}, indicatorAccessibilityLabel={"unread notifications"} vs. indicator={50}, indicatorAccessibilityLabel={"50 unread notifications"} 

Right now users just get "Boards for You 50", which is not helpful

ayeshakmaz: This would apply to the 'dot' version too. It could be a follow up, but we shouldn't forget about it

ayeshakmaz (approved):
One a11y comment, but it can be addressed in a follow up

PR #1441, opened 2021-03-26
Docs: Bump up maxWidth

ayeshakmaz:
Increases max width of our docs to our largest break point

Note: This PR makes the PageHeader docs un-ideal (but still useable) - can fix that in a follow up PR

### DIFF
--- a/docs/src/Box.doc.js
+++ b/docs/src/Box.doc.js
@@ -252,7 +252,7 @@ card(
 
 card(
   <MainSection name="Best practices">
-    <MainSection.Subsection>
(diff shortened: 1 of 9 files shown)


dangerismycat:
This is 😍 in general! Two concerns:
- Text blocks look a little funky, splitting into a left-aligned column for no apparent visual reason. Not sure what the appropriate fix is; we don't want full-width text and centering that column would be a bit odd.
- The three column sections could stand to be center-aligned, I think?

dangerismycat (approved):

Issue #649, opened 2020-02-05
Maximum callstack exceeded when two Modals are on the screen


jennyscript:
1) Modify the Modal docs so the first example looks like the code below
2) Open both Modals
3) Click anywhere inside the top-most modal
4) See the stacktrace in the console

```class Example extends React.Component {
  constructor(props) {
    super(props);
    this.handleToggleSmall = this._handleToggleSmall.bind(this);
    this.handleToggleMedium = this._handleToggleMedium.bind(this);
    this.state = {
      sm: false,
      md: false,
    };
  }

  _handleToggleSmall() {
    this.setState(prevState => ({ sm: !prevState.sm }));
  }

  _handleToggleMedium() {
    this.setState(prevState => ({ md: !prevState.md }));
  }

  render() {
    const { sm, md, lg } = this.state;
    return (
      <Box marginLeft={-1} marginRight={-1}>
        <Box padding={1}>
          <Button
            text="size='sm'"
            onClick={this.handleToggleSmall}
          />
          {sm && (
            <Layer>
              <Modal
                accessibilityCloseLabel="close"
                accessibilityModalLabel="View default padding and styling"
                heading="Small modal"
                onDismiss={this.handleToggleSmall}
                footer={<Heading size="sm">Footer</Heading>}
                size="sm"
              >
                <Box padding={2}>
                  <Heading size="sm">Children</Heading>
                  <Button
                    text="size='md'"
                    onClick={this.handleToggleMedium}
                  />
                </Box>
              </Modal>
            </Layer>
          )}
        </Box>
        {md && (
          <Layer>
            <Modal
              accessibilityCloseLabel="close"
              accessibilityModalLabel="View default padding and styling"
              heading="Medium modal"
              onDismiss={this.handleToggleMedium}
              footer={<Heading size="sm">Footer</Heading>}
              size="md"
            >
              <Box padding={2}>
                <Heading size="sm">Children</Heading>
              </Box>
            </Modal>
          </Layer>
        )}
      </Box>
    );
  }
}
```

simkessy:
Bump

jennyscript:
There appears to be a similar issue when there's an input (plain html input or gestalt component input) in a Modal. Clicking the input which would give it focus results in this stacktrace.

![image (3)](https://user-images.githubusercontent.com/1767822/102539169-2c59fe00-4062-11eb-9aca-86d08d90fa11.png)

hazem-a1:
[...]

PR #1442, opened 2021-03-26
Modal, Sheet: Fix infinite focus loop (#649)

leoschmitz:
The TrapFocusBehavior component listens to all document focus events. If
a target of a FocusEvent does not match any children of the trapped
component, the event is blocked and a new one is dispatched for the
first child of the trapped component. This works well for a single trap.

The infinite loop occurs when there is more than one TrapFocusBehaviour
trying to trap the focus for its first child. Calling .focus() for it
triggers a new FocusEvent, which is then captured by all Trap components,
calling .focus() again and filling all call stack up.

The solution is to check whether the event has been triggered by a trap
component or not. If it has, then let the event run its course. Since
the first child is actually the event target, and not the trap
component, Element.closest is used to identify if the target has a trap
ancestor.

In short,

* 'trap-focus' name added to identify TrapFocusBehavior components
* Element.closest() used to identify trapped components
* class component migrated to hook

Impact: both Modal and Sheet use TrapFocusBehavior.


dangerismycat:
Thanks for fixing this bug! We're reviewing this PR internally and we'll get back to you soon. 🙂

dangerismycat:
Hey @leoschmitz –

Thanks again for the PR. It looks like the stack overflow bug has been fixed! However, it also seems that focus is not being trapped anymore, which you can repro by opening a modal and using the keyboard to tab through the interactive elements. Focus should be confined to the open Modal, but as you can see in this video, we're able to tab through the rest of the elements as well.
https://user-images.githubusercontent.com/12059539/112914281-70708c00-90b0-11eb-9c2e-911099389eb2.mp4

dangerismycat:
Sounds good. Thanks again for your work on this!

leoschmitz:
Hey @dangerismycat,

It appears that identifying the event target as a child of the Trap component solves it. Problem is, since the focus goes to the first child of the trap component, I had to search the ancestors up to see if the event came from a Trap component. I did this with the help of a new 'trap-focus' attribute in the TrapFocusBehavior div. 

I don't know if adding a name there is acceptable. What do you think?

dangerismycat:
That seems to do the trick! The name seems fine to me, though I'm interested in opinions from @christianvuerings or @jennyscript.

dangerismycat:
@leoschmitz Thanks again for this work, and our apologies for the lack of action on it. We're taking this issue off the back burner and addressing it again soon — I may end up re-opening and merging this PR, or perhaps adapting it if needed. 🙂

dangerismycat:
@leoschmitz I finally had time to look through this, and it looks great! Seems to work perfectly to me. I'm re-opening this PR and will merge it soon. Thank you again for this work, and many apologies for the extended delays! 🙏

dangerismycat (approved):

PR #1444, opened 2021-03-29
Box: add `as` prop to use semantic tags

dangerismycat:
For SEO and a11y purposes, we need to improve our support for semantic HTML elements. A simple way to accomplish this is to add a new prop to Box that will allow that component to output more than just `div` elements. This PR adds the prop `as`, which is an enum of block-level elements. This enum should cover our current use cases, and is easily added to in the future.

### Test Plan

The easiest way to verify this is working is to navigate to the Box docs on the preview link, then add something like `as="article"` to a Box example. Inspect the element and note that it now uses the `<article>` tag instead of `<div>`.

### DIFF
--- a/packages/gestalt/src/Box.js
+++ b/packages/gestalt/src/Box.js
@@ -203,10 +203,10 @@ const disallowedProps = [
 ];
 
 type OutputType =
-  | Element<'div'>
   | Element<'article'>
   | Element<'aside'>
   | Element<'details'>
+  | Element<'div'>
   | Element<'figcaption'>
   | Element<'figure'>
   | Element<'footer'>
--- a/packages/gestalt/src/Box.js
+++ b/packages/gestalt/src/Box.js
@@ -25,6 +25,7 @@ import {
   type AlignContent,
   type AlignItems,
   type AlignSelf,
+  type As,
   type BorderStyle,
   type Bottom,
   type Column,
@@ -50,6 +51,7 @@ import {
   AlignContentPropType,
   AlignItemsPropType,
   AlignSelfPropType,
+  AsPropType,
   BorderStylePropType,
   ColorPropType,
   ColumnPropType,
@@ -98,6 +100,7 @@ type Props = {
   alignContent?: AlignContent,
   alignItems?: AlignItems,
   alignSelf?: AlignSelf,
+  as?: As,
   bottom?: Bottom,
   borderStyle?: BorderStyle,
   color?: Color,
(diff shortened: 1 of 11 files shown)


packages/gestalt/src/contexts/ScrollBoundaryContainer.js
```diff
@@ -34,6 +34,8 @@ function ScrollBoundaryContainerProvider({ children }: Props): Element<typeof Pr
   const scrollBoundaryContainerContext = {
     scrollBoundaryContainerRef,
     addRef: useCallback((ref) => {
+      // Cannot call `setScrollBoundaryContainerRef` with `ref` bound to the first parameter because a call signature declaring the expected parameter / return type is missing in  `HTMLElement` [1] but exists in  function type [2].
```
dangerismycat, line 37: I'm a bit confused by this one. Not sure where I should be adding a return type

packages/gestalt/src/Box.js
```diff
@@ -199,19 +202,36 @@ const disallowedProps = [
...
+  | Element<'figcaption'>
+  | Element<'figure'>
+  | Element<'footer'>
+  | Element<'header'>
+  | Element<'main'>
+  | Element<'mark'>
+  | Element<'nav'>
+  | Element<'section'>
+  | Element<'summary'>
+  | Element<'time'>;
```
christianvuerings, line 219: @dangerismycat I would suggest to only allow `block` elements for now and add `inline` elements as needed.

E.g. `<Box as="time" padding={4}>` might not do what folks expect. Unless we make `display: block` the default for now.

packages/gestalt/src/contexts/ScrollBoundaryContainer.js
christianvuerings: Would be good to figure this one out before merging.

packages/gestalt/src/Box.js
dangerismycat: Works for me. Is there a list somewhere of which of these elements are inline vs block?

packages/gestalt/src/contexts/ScrollBoundaryContainer.js
dangerismycat: Does that error message suggest anything to you? I can't tell what it's trying to communicate.

AlbertCarreras: try 

` scrollBoundaryContainerRef: ?HTMLElement,`

dangerismycat: genius! 😅

AlbertCarreras (approved):

PR #1448, opened 2021-03-31
Docs: fix Flow type in Docs for Tabs

AlbertCarreras:


dangerismycat (approved):
Good catch! Thanks for this!

PR #1450, opened 2021-03-31
Link/Button/IconButton/TapArea +composed components: always pass disableOnNavigation to onClick

AlbertCarreras:
`disableOnNavigation`  is not always passed to onClick. Within Link or Internal Link, it does disable the logic from Provider. 

Outside of it, it's just a noop function.

This change changes the flow type so `disableOnNavigation` isn't optional. This prevents having to call `disableOnNavigation` with optional chaining `disableOnNavigation?.()` in the best of the cases or conditional statements within `onClick`:

`disableOnNavigation && disableOnNavigation()
`
`if (disableOnNavigation)  disableOnNavigation()
`

This change won't break code but might require cleaning optional chaining to prevent eslint complains.

### DIFF
--- a/packages/gestalt/src/TapArea.js
+++ b/packages/gestalt/src/TapArea.js
@@ -36,7 +36,7 @@ type BaseTapArea = {|
     | SyntheticKeyboardEvent<HTMLDivElement>
     | SyntheticMouseEvent<HTMLAnchorElement>
     | SyntheticKeyboardEvent<HTMLAnchorElement>,
-    {| disableOnNavigation?: () => void |},
+    {| disableOnNavigation: () => void |},
   >,
   tabIndex?: -1 | 0,
   rounding?: Rounding,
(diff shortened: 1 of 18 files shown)


packages/gestalt/src/TapArea.js
```diff
@@ -119,9 +119,18 @@ const TapAreaWithForwardRef: React$AbstractComponent<unionProps, unionRefs> = fo
     },
   );
 
+  const handleOnKeyPress = (event) => {
+    // Check to see if space or enter were pressed
+    if (!disabled && onTap && keyPressShouldTriggerTap(event)) {
+      // Prevent the default action to stop scrolling when space is pressed
+      event.preventDefault();
+      onTap({ event, disableOnNavigation: () => {} });
```
dangerismycat, line 127: Is this casting the type of disableOnNavigation? If so, I think we want `void` instead of `{}`

dangerismycat (approved):

packages/gestalt/src/TapArea.js
AlbertCarreras: no, this is setting `disableOnNavigation` to  `() => {}`, not a type, but the value

dangerismycat: Oh I see, sometimes hard to read between object literals and renamed destructuring and Flow declarations and… 😅

PR #1451, opened 2021-04-01
Docs: Fix outdated What's New Page

christianvuerings:
https://cdn.jsdelivr.net heavily caches their files so our What's New Page stays outdated for a while.

## Before

image

## After

image

### DIFF
--- a/docs/src/What's New.doc.js
+++ b/docs/src/What's New.doc.js
@@ -13,7 +13,7 @@ function Changelog() {
   useEffect(() => {
     const fetchChangelog = async () => {
       const result = await fetch(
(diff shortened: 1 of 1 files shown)


jimaek:
If you know the latest release you could instead do this and cache would not be a problem https://cdn.jsdelivr.net/gh/pinterest/gestalt@21.4.1/CHANGELOG.md
But not sure how accessible that information is

christianvuerings:
> If you know the latest release you could instead do this and cache would not be a problem https://cdn.jsdelivr.net/gh/pinterest/gestalt@21.4.1/CHANGELOG.md
> But not sure how accessible that information is

That's correct, however it would require us to update the link for every release.

AlbertCarreras (approved):

AlbertCarreras (approved):

PR #1453, opened 2021-04-02
Dropdown: Fix link access through keyboard

ayeshakmaz:
Fix for [BUG-123544](https://jira.pinadmin.com/browse/BUG-123544)

Ensures link access is possible through keyboard navigation

Also ensures only external links open in new tab

(also lists Dropdown under Navigation and Forms)

## Test Plan

current tests

### DIFF
--- a/packages/gestalt/src/Dropdown.js
+++ b/packages/gestalt/src/Dropdown.js
@@ -67,7 +67,11 @@ export default function Dropdown({
   let selectedElement;
   const setOptionRef = (optionRef) => {
     selectedElement = optionRef;
-    if (selectedElement) selectedElement.focus();
+    if (selectedElement && selectedElement.getElementsByTagName('a')[0]) {
+      selectedElement.getElementsByTagName('a')[0].focus();
+    } else if (selectedElement) {
+      selectedElement.focus();
+    }
   };
 
   const containerRef = useRef();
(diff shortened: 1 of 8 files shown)


packages/gestalt/src/Dropdown.js
```diff
@@ -67,7 +67,11 @@ export default function Dropdown({
   let selectedElement;
   const setOptionRef = (optionRef) => {
     selectedElement = optionRef;
-    if (selectedElement) selectedElement.focus();
+    if (selectedElement && selectedElement.getElementsByTagName('a')[0]) {
```
dangerismycat, line 70: nit: a bit more readable if you pull this out above, like
```js
const linkElement = selectedElement?.getElementsByTagName('a')[0];

if (linkElement) {
  linkElement.focus();
} ...
```

dangerismycat:
Fine if it's out of scope for this PR, but we have another bug related to links here: they're always opened in a new tab, even if internal. It looks like the bug is in MenuOption, line 167: we always pass `target="blank"` to Link, but we should only pass that if `isExternal` is true.

packages/gestalt/src/Dropdown.jsdom.test.js
```diff
@@ -407,4 +407,67 @@ describe('Dropdown', () => {
 
     expect(handleSelectMock).toHaveBeenCalledTimes(1);
   });
+
+  it('should call link handleSelect when enter key is pressed', () => {
```
dangerismycat, line 411: 🙏

dangerismycat (approved):
Awesome, thanks for this fix!

PR #1457, opened 2021-04-05
Docs: updated Docs for ZIndex Classes

AlbertCarreras:
Docs: updated Docs for ZIndex Classes

### DIFF
--- a/docs/src/ZIndex Classes.doc.js
+++ b/docs/src/ZIndex Classes.doc.js
@@ -1,181 +1,660 @@
 // @flow strict
 import React, { type Node } from 'react';
-import { Box, Flex, Icon, Text } from 'gestalt';
-import Example from './components/Example.js';
-import Card from './components/Card.js';
 import PageHeader from './components/PageHeader.js';
-import Markdown from './components/Markdown.js';
+import FeedbackCallout from './components/FeedbackCallout.js';
+import MainSection from './components/MainSection.js';
 
 const cards: Array<Node> = [];
 const card = (c) => cards.push(c);
+
+card(<FeedbackCallout componentName="zIndex Classes" />);
 
 card(
   <PageHeader
     name="zIndex Classes"
     fileName="zIndex"
-    description="FixedZIndex and CompositeZIndex are 2 classes that generate z-index for the Box and Sticky components."
+    description="FixedZIndex and CompositeZIndex are utility classes that generate z-indexes for Gestalt components."
   />,
 );
 
 card(
-  <Card name="FixedZIndex">
-    <Flex alignItems="start" direction="column" gap={4}>
-      <Text>
-        FixedZIndex is used for setting fixed z-index values. FixedZIndex must be instantiated with
-        a number.
-      </Text>
-      <Box padding={2} color="lightGray" rounding={2}>
-        <Markdown
-          text="
+  <MainSection name="FixedZIndex">
+    <MainSection.Subsection columns={2}>
+      <MainSection.Card
+        cardSize="lg"
+        description={`
+FixedZIndex is used for setting fixed z-index values. Use this class when you want to create an initial z-index to stack others on top of. FixedZIndex must be instantiated with a number.
 ~~~jsx
 import { FixedZIndex } from 'gestalt';
+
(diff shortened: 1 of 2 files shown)


docs/src/ZIndex Classes.doc.js
```diff
@@ -1,181 +1,660 @@
...
-    <Flex alignItems="start" direction="column" gap={4}>
-      <Text>
-        FixedZIndex is used for setting fixed z-index values. FixedZIndex must be instantiated with
-        a number.
-      </Text>
-      <Box padding={2} color="lightGray" rounding={2}>
-        <Markdown
-          text="
+  <MainSection name="FixedZIndex">
+    <MainSection.Subsection columns={2}>
```
dangerismycat, line 22: Can/should these be full-width?

docs/src/ZIndex Classes.doc.js
```diff
@@ -1,181 +1,660 @@
...
+        description={`
+Extract z-index values from both zIndex classes to use in non-Gestaltifiable HTML Tags. See [z-index in non-Gestalt components](#z-index-in-non-Gestalt-components) to learn more.
 ~~~jsx
-const customButton = <CustomButton zIndex={fixedZindex.index()}/>;
+import { FixedZIndex } from 'gestalt';
+
+const fixedZindex = new FixedZIndex(1);
+
+const canvas = (
+  <canvas
```
dangerismycat, line 217: This is a great example, btw! 👍

docs/src/ZIndex Classes.doc.js
```diff
@@ -1,181 +1,660 @@
...
-            text="
+`}
+      />
+    </MainSection.Subsection>
+    <MainSection.Subsection columns={2}>
+      <MainSection.Card
+        cardSize="lg"
+        type="do"
+        description={`
+Use Layer instead of zIndex when a component needs to visually break out of its parent container, for example, in the case of [modals](/Modal), [sheets](/Sheet) and [tooltips](/Tooltip). See [z-Index in foundational components](#zIndex-in-foundational-components) and [ZIndex in Layer](#zIndex-in-Layer) to learn more.
```
dangerismycat, line 233: I think the Don't example here would be something like
```js
<Box zIndex={1000000}>
  <Modal />
<Box>
```

docs/src/ZIndex Classes.doc.js
```diff
@@ -1,181 +1,660 @@
...
+card(
+  <MainSection name="Variants">
+    <MainSection.Subsection
+      title="zIndex in foundational components"
+      description={`
+[Box](/Box), [Sticky](/Sticky), and [Layer](/Layer) are foundational components that have zIndex props. If any other components need to be positioned within their stacking context, wrap with one of these foundational components to set the z-index.
+
+In the case of Layer, this foundational components creates a new stacking context. Unless there's a conflict with another z-index, don't pass unnecessary zIndex to Layer.
+
+The following example sets a z-index in the Layer wrapping [Sheet](/Sheet) to position Sheet over the page header in the Docs. Set \`HeaderPageZIndex\` below 10 to see the importance of z-index in this example.`}
```
dangerismycat, line 258: `PAGE_HEADER_ZINDEX`

dangerismycat (approved):
This looks great! Such a huge improvement over the previous z-index docs. Thanks for this work!

Go ahead and add that Don't example if you agree, then merge this. There are some grammar nits but it'll be faster if I follow-up with a separate PR for those.

docs/src/ZIndex Classes.doc.js
AlbertCarreras: YEP! good catch

DONE

docs/src/ZIndex Classes.doc.js
AlbertCarreras: DONE

docs/src/ZIndex Classes.doc.js
AlbertCarreras: we have a max-width on markdown, so this is expected

docs/src/ZIndex Classes.doc.js
AlbertCarreras: DONE

PR #1458, opened 2021-04-05
Typeahead: Fix static `value` bug

dangerismycat:
Typeahead initializes its displayed value with the passed-in `value` prop (if present), but ignores any changes to that prop. So if, say an action elsewhere on the page causes a new `value` to be passed in, Typeahead completely ignores it.

This PR adds a useEffect to listen for any changes to `value` and update the internal displayed value variable accordingly.

### Test Plan

I recommend copying the code from [this sandbox](https://codesandbox.io/s/default-item-example-forked-5z7sb?file=/example.js:179-183) into one of the docs examples, then play around and see if you can break it. You should be able to select a new value via the dropdown or by typing, and the programmatic selection via the buttons should work fine as well.

### DIFF
--- a/packages/gestalt/src/Typeahead.js
+++ b/packages/gestalt/src/Typeahead.js
@@ -88,7 +88,12 @@ const TypeaheadWithForwardRef: React$AbstractComponent<Props, HTMLInputElement> 
 
   // Track input value
   const defaultOption: OptionObject | null = findDefaultOption(value);
-  const [search, setSearch] = useState<string>(defaultOption?.label || '');
+  const displayValue = defaultOption?.label ?? '';
+  const [search, setSearch] = useState<string>('');
+  // Make sure we respect any external changes to `value`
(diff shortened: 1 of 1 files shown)


AlbertCarreras:
It's almost there. 

Select a value (7) from the dropdown. Open dropdown again, 7 should be selected.
Select a value (6) with the button. Open dropdown again, 6 should be selected but it's not, 7 remains selected.

dangerismycat:
Ah, good catch. The selected option is being tracked separately from the displayed option for some reason. Updating…

dangerismycat:
Updated to set the selected item as well. This will show a stale hover state, but that's a separate bug that may not be worth fixing for now.

AlbertCarreras:
@dangerismycat can you rerun?

AlbertCarreras (approved):
Thanks for fixing this.

PR #1462, opened 2021-04-07
Callout, Upsell: Convert title to text instead of heading

ayeshakmaz:
Convert the H3 in Callout and Upsell to plain text

## Test Plan

update snap shots, visually ensure the same styling


packages/gestalt/src/headingStyles.js
```diff
@@ -0,0 +1,13 @@
+// @flow strict
```
ayeshakmaz, line 1: Created this to make future updates easier, but open to other format suggestions if this isn't ideal

dangerismycat: Can we compose this in a CSS file instead? Possibly in Heading.css?

ayeshakmaz: Always forget about compose, great call! Updated!

dangerismycat (approved):

PR #1464, opened 2021-04-08
Docs: dogfood Gestalt ESLint rules on our docs

christianvuerings:
Let's [dogfood](https://en.wikipedia.org/wiki/Eating_your_own_dog_food) our ESLint rules on our own docs:

* Helps with spotting issues with our rules
* Ensures our docs are up to the highest coding standards (perhaps this is an exaggeration) 

## Test Plan

* ESLint passes in CI
* Modified docs pages still look good


docs/src/components/FeedbackCallout.js
```diff
@@ -14,26 +14,24 @@ export default function FeedbackCallout({ componentName }: Props): Node {
   const link = `${BASE_LINK}${componentName}`;
 
   return (
-    <Box marginLeft={-1} marginRight={-1}>
```
christianvuerings, line 17: `marginLeft` and `marginRight` are no longer supported on `Box` so this didn't do anything

packages/eslint-plugin-gestalt/src/button-icon-restrictions.js
```diff
@@ -1,6 +1,5 @@
 /**
  * @fileoverview Require specific props when using an icon with Button
- * @author Ryan James <rjames@pinterest.com>
```
christianvuerings, line 3: Removed all authors, no reason to make these public

packages/eslint-plugin-gestalt/src/button-icon-restrictions.js
```diff
@@ -50,7 +49,7 @@ const rule = {
           return;
         }
         importedComponent = decl.specifiers.some((node) => {
-          const isValidComponent = node.imported.name === componentName;
+          const isValidComponent = node.imported?.name === componentName;
```
christianvuerings, line 52: Discovered an issue with a prop called `name` on the `Example.js` page:

https://github.com/pinterest/gestalt/blob/fe1aa734f729e9ce5b249b0b820c7d466a180e18/docs/src/components/Example.js#L40

docs/src/components/FeedbackCallout.js
dangerismycat: weird that Flow wasn't complaining about this given that we removed those props a while ago…

dangerismycat (approved):
Awesome, thanks Christian!

dangerismycat:
Closed without merging?

PR #1465, opened 2021-04-08
Docs: convert ESLint docs to new format

christianvuerings:
Makes it possible to link to a specific rule: `/Eslint%20Plugin#gestaltno-box-marginleft-marginright`

## Test Plan

1. Go to https://gestalt.netlify.app/Eslint%20Plugin
2. Ensure everything looks good

### DIFF
--- a/docs/src/components/MainSectionSubsection.js
+++ b/docs/src/components/MainSectionSubsection.js
@@ -5,7 +5,6 @@ import slugify from 'slugify';
 import CopyLinkButton from './buttons/CopyLinkButton.js';
 import Markdown from './Markdown.js';
 import { copyToClipboard } from './Card.js';
-import { capitalizeFirstLetter } from './utils.js';
 
 type Props = {|
   children?: Node,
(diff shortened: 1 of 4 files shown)


docs/src/components/Markdown.css
```diff
@@ -11,11 +11,16 @@
 .Markdown code {
   background-color: var(--g-colorGray100);
   color: var(--g-colorGray300);
+  display: inline-block;
```
christianvuerings, line 14: @ayeshakmaz Fixes an issue where `padding` was only applied to the first line of code. The root cause was that `code` is an `inline` element by default.

## Before
image

## After
image

docs/src/components/MainSectionSubsection.js
```diff
@@ -32,7 +31,7 @@ const MainSectionSubsection = ({ children, columns = 1, description, title }: Pr
             data-anchor
           >
             <Flex alignItems="center" gap={2}>
-              <Heading size="sm">{capitalizeFirstLetter(title)}</Heading>
+              <Heading size="sm">{title}</Heading>
```
christianvuerings, line 34: @ayeshakmaz Instead of doing this `capitalizeFirstLetter`, I would recommend to fix the issues instead. 

In our case we want to show
```
gestalt/button-icon-restrictions
```

and not

```
Gestalt/button-icon-restrictions
```

as the heading

docs/src/components/Markdown.css
AlbertCarreras: Awesome

AlbertCarreras (approved):
Thanks! Looking good.

dangerismycat (approved):
This is great, thanks Christian!

PR #1471, opened 2021-04-14
Internal: use new JSX transform & remove unnecessary react imports

christianvuerings:
Context: https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html#removing-unused-react-imports

Fixes #1422 from @dangerismycat 

## Test Plan

* CI passes
* Docs still work: https://deploy-preview-1471--gestalt.netlify.app/


packages/gestalt-core/build.js
```diff
@@ -131,19 +132,35 @@ const plugins = (name) => [
...
+          : '@babel/plugin-transform-react-jsx',
+        {
+          runtime: 'automatic',
+          useBuiltIns: true,
+        },
+      ],
+    ],
     exclude: 'node_modules/**',
   }),
+  commonjs(),
```
christianvuerings, line 163: Needed for `react/jsx-runtime` which exports `jsx` on the module export.
Without it, the transformed import `import {jsx as _jsx} from 'react/jsx-runtime'` will fail.

babel.config.js
```diff
@@ -1,10 +1,15 @@
 // @flow strict
+
+const tranformPlugin /*: string */ =
```
christianvuerings, line 3: Uses flow comment types: https://flow.org/en/docs/types/comments/

We use babel to strip the flow types but don't use babel on the babel config file itself.

AlbertCarreras (approved):

PR #1474, opened 2021-04-20
Card: Fix RTL hover bug

AlbertCarreras:
Fixes https://jira.pinadmin.com/browse/BUG-124352

Before: 
image

After:
image


ayeshakmaz (approved):
Looks like you need to update flow types, but otherwise looks great!

packages/gestalt/src/Card.js
```diff
@@ -33,8 +33,7 @@ export default function Card(props: Props): Node {
   };
 
   const classes = classnames(styles.card, {
-    // If, like @chrislloyd, you can't remember Javascript equality rules,
```
dangerismycat, line 36: aw,  I think we should leave the snark 😂

PR #1476, opened 2021-04-20
Docs: Fix responsiveness on PageHeader

christianvuerings:
## Before

Screen Shot 2021-04-20 at 4 51 15 PM

## After

Screen Shot 2021-04-20 at 4 51 52 PM

## Test Plan

1. Go to `/PageHeader`
2. Resize the browser window
3. Ensure the contents of the docs don't overlap the navigation

### DIFF
--- a/docs/src/components/App.js
+++ b/docs/src/components/App.js
@@ -30,7 +30,7 @@ export default function App({ children }: Props): Node {
 
                   <Divider />
 
-                  <Box width="100%">
+                  <Box width="100%" minWidth={0}>
                     <Box padding={4} mdPadding={6} lgPadding={8} width="100%" role="main">
                       {children}
                     </Box>


docs/src/components/App.js
```diff
@@ -30,7 +30,7 @@ export default function App({ children }: Props): Node {
 
                   <Divider />
 
-                  <Box width="100%">
+                  <Box width="100%" minWidth={0}>
```
christianvuerings, line 33: image

More information at https://css-tricks.com/flexbox-truncated-text/#the-solution-is-min-width-0-on-the-flex-child

ayeshakmaz: Gahhh I swear this one thing has been the source of every other bug when it comes to Flex issues - great find!

ayeshakmaz (approved):
🙌

docs/src/components/App.js
dangerismycat: Hm…is this a universal enough fix that we should add it as a hardcoded attribute on FlexItem?

christianvuerings: @dangerismycat I like that idea, to add `minWidth` as a prop to `Flex.Item` and perhaps add a link to that article. I wouldn't make it the default though since it goes against the Flex spec.

PR #1478, opened 2021-04-21
Datapoint: new component

ponori:
[Datapoint TDD](https://paper.dropbox.com/doc/Gestalt-TDD-DataPoint--BGoaOJalmcr0wxQZ7M2C9Oa1Ag-TbnunbB7P3RVjOAnrax4X)
[Datapoint Figma file](https://www.figma.com/file/Kf6sAtvqQLy4utSP9iM5GJ/Datapoint?node-id=1%3A7)


packages/gestalt/src/index.js
```diff
@@ -13,6 +13,7 @@ import Collage from './Collage.js';
 import Collection from './Collection.js';
 import Column from './Column.js';
 import Container from './Container.js';
+import Datapoint from './Datapoint.js';
```
christianvuerings, line 16: @ponori just fyi, this is failing on Netlify / CI since the Linux file system is case sensitive. To fix the builds, we should change this (and other references) to `./DataPoint.js`

Mac OS file system is not case sensitive so that's why it works for you locally.

ponori:
Thanks Christian! Yeah, I initially set up the component to be camel-case, but we agreed to change it to Datapoint after the work began and that’s where this issue originates from. I’ll likely close this PR/branch and just run the `yarn component` script with the new name to make this whole process cleaner. 

Thanks!

PR #1480, opened 2021-04-22
Heading: add accessibilityLevel="none" option

christianvuerings:
Adds `accessibilityLevel="none"` option to `Heading`

## Use Cases

* #1462 where we wanted a `Heading` without any accessibility level. In that case we added a new class `.TextLikeHeadingSm` but that approach is limited to just 1 font size & isn't exposed externally. /cc @ayeshakmaz 
* Pinterest's codebase has a few places where folks create their own components to get around this limitation.

## Test Plan

See flow tests and unit tests.

Screen Shot 2021-04-22 at 7 33 45 AM

### DIFF
--- a/docs/src/Heading.doc.js
+++ b/docs/src/Heading.doc.js
@@ -21,7 +21,7 @@ card(
     props={[
       {
         name: 'accessibilityLevel',
-        type: '1 | 2 | 3 | 4 | 5 | 6',
+        type: '1 | 2 | 3 | 4 | 5 | 6 | "none"',
         description: 'Allows you to override the default heading level',
         href: 'levels',
       },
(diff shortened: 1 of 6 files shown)


AlbertCarreras:
LGTM

docs/src/Heading.doc.js
```diff
@@ -175,7 +175,7 @@ card(
     description="
     For accessibility purposes, we allow you to override the heading level.
 
-    We should have one level 1 per page & levels should be appropriately nested. E.g. level 1 followed by level 2 & level 2 followed by level 2 or level 3.
+    We should have one level 1 per page &amp; levels should be appropriately nested. E.g. level 1 followed by level 2 &amp; level 2 followed by level 2 or level 3. We also allow headings without an accessibility level.
```
AlbertCarreras, line 178: As a follow  up I'd add more info re when to use "none"

themindfuldev:
Thank you very much for doing this, @christianvuerings !

AlbertCarreras (approved):

christianvuerings:
@ayeshakmaz raised a good point in that it might be better to follow the `Box` way of doing it with `as`:
* `<Heading as="h1" />`
* `<Heading as="div" />`

However that's out of scope for this particular diff since it requires removing `accessibilityLabel` and we can do that as a follow up.

ayeshakmaz (approved):

PR #1481, opened 2021-04-23
Modal: added comment on temporal prop to prevent confusion

AlbertCarreras:

### DIFF
--- a/packages/gestalt/src/Modal.js
+++ b/packages/gestalt/src/Modal.js
@@ -150,6 +150,7 @@ const ModalWithForwardRef: React$AbstractComponent<Props, HTMLDivElement> = forw
                     )}
                   </div>
                 )}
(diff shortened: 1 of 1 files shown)


packages/gestalt/src/Modal.js
```diff
@@ -150,6 +150,7 @@ const ModalWithForwardRef: React$AbstractComponent<Props, HTMLDivElement> = forw
                     )}
                   </div>
                 )}
+                {/* _dangerouslyDisableScrollBoundaryContainer must be kept temporally until specific surfaces migrate from Modal to Sheet */}
```
dangerismycat, line 153: "temporarily"?

dangerismycat (approved):

packages/gestalt/src/Modal.js
AlbertCarreras: DONE

PR #1484, opened 2021-04-27
Docs: Make unexpanded code example more obviously interactive

dangerismycat:
- Bump height to hit mid-line
- Change cursor to `zoom-in`

### DIFF
--- a/docs/src/components/ExampleCode.css
+++ b/docs/src/components/ExampleCode.css
@@ -1,0 +1,3 @@
+.LiveEditor__textarea__notExpanded textarea {
+  cursor: zoom-in !important;
+}
--- a/docs/src/components/ExampleCode.css
+++ b/docs/src/components/ExampleCode.css
@@ -1,3 +1,3 @@
 .LiveEditor__textarea__notExpanded textarea {
-  cursor: zoom-in !important;
+  cursor: zoom-in !important; /* stylelint-disable-line declaration-no-important */
 }
--- a/docs/src/components/ExampleCode.js
+++ b/docs/src/components/ExampleCode.js
@@ -1,7 +1,5 @@
 // @flow strict
-import type { Node } from 'react';
-
-import { useEffect, useRef, useState } from 'react';
+import { type Node, useEffect, useRef, useState } from 'react';
 import { Box, Flex } from 'gestalt';
 import { LiveEditor } from 'react-live';
 import clipboardCopy from './clipboardCopy.js';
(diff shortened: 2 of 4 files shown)


docs/src/components/ExampleCode.css
```diff
@@ -0,0 +1,3 @@
+.LiveEditor__textarea__notExpanded textarea {
+  cursor: zoom-in !important;
```
dangerismycat, line 2: The library is also trying to set cursor, so need to override that.

The library doesn't provide a way to add styles for the underlying textarea. This is potentially brittle, but I doubt future versions will move away from using a textarea, so this feels relatively safe.

docs/src/components/ExampleCode.js
```diff
@@ -61,28 +68,29 @@ export default function ExampleCode({ code, name }: {| code: string, name: strin
             borderStyle="sm"
             display="flex"
             overflow="hidden"
+            maxHeight={!expanded ? CODE_EXAMPLE_HEIGHT : undefined}
             position="relative"
             ref={codeExampleRef}
             rounding={2}
-            {...(expanded ? {} : { maxHeight: CODE_EXAMPLE_HEIGHT, overflow: 'hidden' })}
```
dangerismycat, line 67: This was setting these as attributes, not styles, so the overflow was being ignored

docs/src/components/ExampleCode.js
```diff
@@ -61,28 +68,29 @@ export default function ExampleCode({ code, name }: {| code: string, name: strin
...
               flex="grow"
               onFocus={() => {
                 setExpanded(true);
               }}
             >
-              {/* We can not pass in an id for LiveEditor which links to the underlying text area */}
-              {/* eslint-disable-next-line jsx-a11y/label-has-associated-control */}
-              <label>
-                <LiveEditor
-                  className="live-editor-pane"
```
dangerismycat, line 79: This isn't an available prop, and we don't set this class

ayeshakmaz: This was being used in commands.js to avoid Cypress testing the accessibility of the code editor (there was an issue with it detecting the right background color IIRC), so I'd just keep an eye out if those tests still pass. This class name was showing up on the outer div

ayeshakmaz: Ah yup, looks like they're all failing again

ayeshakmaz (approved):
:shipit:

docs/src/components/ExampleCode.js
```diff
@@ -61,28 +68,31 @@ export default function ExampleCode({ code, name }: {| code: string, name: strin
...
-                  }}
-                  padding={16}
-                />
-              </label>
+              <div className={!expanded ? 'LiveEditor__textarea__notExpanded' : undefined}>
+                {/* We can not pass in an id for LiveEditor which links to the underlying text area */}
+                {/* eslint-disable-next-line jsx-a11y/label-has-associated-control */}
+                <label>
+                  <LiveEditor
+                    // Don't remove this className or all the a11y integration tests will fail
```
AlbertCarreras, line 87: what about adding the context Ayesha just gave? 

`// Don't remove this className or all the a11y integration tests will fail. It's used in commands.js to avoid Cypress testing the accessibility of the code editor due to an issue with it detecting the right background color IIRC.`

AlbertCarreras:
There's a funny blue focus
image

AlbertCarreras (approved):

PR #1485, opened 2021-04-28
Button: fix inline for role="link"

AlbertCarreras:
Before
image

After
image


packages/gestalt/src/Link.css
```diff
@@ -9,15 +9,17 @@
   text-decoration: underline;
 }
 
-.inlineBlock {
-  composes: inlineBlock from "./Layout.css";
+.inlineFlex {
+  composes: itemsCenter from "./Layout.css";
+  composes: justifyCenter from "./Layout.css";
+  display: inline-flex;
```
dangerismycat, line 15: Rather than declare this here, we should follow the existing pattern of establishing this style in Layout.css and composing here

packages/gestalt/src/Link.css
```diff
@@ -9,15 +9,17 @@
   text-decoration: underline;
 }
 
-.inlineBlock {
-  composes: inlineBlock from "./Layout.css";
+.inlineFlex {
```
dangerismycat, line 12: It's unexpected to get styles for alignItems and justifyContent from a class called "inlineFlex" — by our typical usage, this would only set the `display` style. Also, from the logic in InternalLink, it looks like we'll get itemsCenter and justifyCenter regardless. So it would be more obvious to add both those explicitly in InternalLink and keep this class focused…or actually not even defining this class (or `flex` below), and just use the existing classes directly from Layout.css.

packages/gestalt/src/InternalLink.js
```diff
@@ -104,15 +104,15 @@ const InternalLinkWithForwardRef: AbstractComponent<Props, HTMLAnchorElement> =
...
-    inline ? linkStyles.inlineBlock : linkStyles.block,
     getRoundingClassName(isTapArea ? rounding || 0 : 'pill'),
     {
       [touchableStyles.tapCompress]: !disabled && tapStyle === 'compress' && isTapping,
       [focusStyles.accessibilityOutline]: !disabled && isFocusVisible,
     },
     isButton
       ? {
-          [linkStyles.buttonLink]: true,
+          [linkStyles.inlineFlex]: inline,
```
dangerismycat, line 114: These will now be flex? Will that affect current usages?

packages/gestalt/src/Link.css
AlbertCarreras: DONE

packages/gestalt/src/Link.css
AlbertCarreras: DONE

packages/gestalt/src/InternalLink.js
AlbertCarreras: image


It was already flex 
image

AlbertCarreras: We are just adding the inline to fix that issue.

packages/gestalt/src/Link.css
AlbertCarreras: DONE

packages/gestalt/src/InternalLink.js
```diff
@@ -127,6 +130,7 @@ const InternalLinkWithForwardRef: AbstractComponent<Props, HTMLAnchorElement> =
       : {},
     isTapArea
       ? {
+          [layoutStyles.block]: true,
```
AlbertCarreras, line 133: TapArea has no inline. Moving here from line 107
image

packages/gestalt/src/InternalLink.js
```diff
@@ -104,15 +105,17 @@ const InternalLinkWithForwardRef: AbstractComponent<Props, HTMLAnchorElement> =
     linkStyles.link,
     focusStyles.hideOutline,
     touchableStyles.tapTransition,
-    inline ? linkStyles.inlineBlock : linkStyles.block,
```
AlbertCarreras, line 107: moving this logic on a per component basis

packages/gestalt/src/InternalLink.js
AlbertCarreras: image
 >>

packages/gestalt/src/InternalLink.js
dangerismycat: Totally fine since we're matching current behavior. Should we include an inline prop in the future? Is it needed?

dangerismycat (approved):
Thanks for adding more tests!

packages/gestalt/src/InternalLink.js
AlbertCarreras: I don't know, we should evaluate that.

PR #1487, opened 2021-04-28
Flex: add `minWidth` prop to Flex.Item

dangerismycat:
Extra-wide children can overflow their parent Flex container, breaking the layout. Adding a `minWidth` prop to Flex.Item fixes this by allowing us to wrap the wide child in `<Flex.Item minWidth={0}>`. See [this blog post](https://css-tricks.com/flexbox-truncated-text/#the-solution-is-min-width-0-on-the-flex-child) for more details.

Screen Shot 2021-04-28 at 1 59 07 PM

## Test Plan
Check out the new "FlexItem-minWidth" example on the Flex docs.


ayeshakmaz (approved):
LGTM 👍

PR #1493, opened 2021-05-04
Fieldset: New component for form accessibility

ayeshakmaz:
## Description

Add a Fieldset component that can be used to wrap form items and specify a legend

### Links

- JIRA: https://jira.pinadmin.com/browse/PDS-2317

### Checklist

- [X] Added Unit and Flow Tests
- [X] Verified Accessibility: keyboard & screen reader interaction
- [X] Checked Dark Mode, Responsiveness, and Right-to-Left support
- [X] Checked Stakeholder feedback (ex. Gestalt designers)


dangerismycat (commented):
I know I'm not the reviewer on this one, I just really wanted to see the implementation 😅

My two comments can def be addressed in future work if you prefer

ponori:
Is the intended usage for a FieldSet for a group of RadioButton components or any form component? I ask b/c grouping textfields in this FieldSet component may result in some type hierarchy challenges.

docs/src/Fieldset.doc.js
```diff
@@ -0,0 +1,222 @@
...
+import PageHeader from './components/PageHeader.js';
+import MainSection from './components/MainSection.js';
+
+const cards: Array<Node> = [];
+const card = (c) => cards.push(c);
+
+card(
+  <PageHeader
+    name="Fieldset"
+    description="Creates an accessible fieldset and legend for a group of related form items"
```
AlbertCarreras, line 14: Nit... not necessary (most new docs follow this pattern)

Fieldset creates an accessible fieldset and legend for a group of related form items, such as RadioButtons(/) and Checkboxes(/)."

docs/src/Fieldset.doc.js
```diff
@@ -0,0 +1,222 @@
...
+card(
+  <PropTable
+    Component={Fieldset}
+    props={[
+      {
+        name: 'legend',
+        type: 'string',
+        required: true,
+        description:
+          "Content for this fieldset's legend that clearly and concisely describes the question being asked.",
```
AlbertCarreras, line 64: Not strong about it, googled Fieldset here and there and they use the word caption. \

Maybe something like 
- "Caption for the form elements grouped in the fieldset" 

I guess I'm iffy with this part `describes the question being asked`. It can be a simple color or gender.

docs/src/Fieldset.doc.js
```diff
@@ -0,0 +1,222 @@
...
+      },
+      {
+        name: 'legendDisplay',
+        type: "'visible' | 'hidden'",
+        defaultValue: 'visible',
+        description:
+          'Whether the legend should be visible or not. If `hidden`, the legend is still available for screen reader users, but does not appear visually.',
+      },
```
AlbertCarreras, line 72: I'd add .. See (legent visibility)[] to learn more.

docs/src/Fieldset.doc.js
```diff
@@ -0,0 +1,222 @@
...
+    ]}
+  />,
+);
+
+card(
+  <MainSection name="Accessibility">
+    <MainSection.Subsection
+      title="When to use Fieldset"
+      description={`Wrapping form fields in a fieldset creates an accessible grouping that signals to users when certain form items are related. The \`legend\` should clearly describe what information is needed from the group of items, whether they're [RadioButtons](/RadioButton), [Checkboxes](/Checkbox) or [TextFields](/TextField). Learn more about the [use of fieldset and legend](https://www.w3.org/WAI/tutorials/forms/grouping/#associating-related-controls-with-fieldset).`}
```
AlbertCarreras, line 87: in a Fieldset

AlbertCarreras: nice resource

docs/src/Fieldset.doc.js
```diff
@@ -0,0 +1,222 @@
...
+        description: `The content of Fieldset, typically [RadioButtons](/RadioButton), [Checkboxes](/Checkbox) or [TextFields](/TextField).`,
+      },
+    ]}
+  />,
+);
+
+card(
+  <MainSection name="Accessibility">
+    <MainSection.Subsection
+      title="When to use Fieldset"
```
AlbertCarreras, line 86: Could we add a good example for Textfield?  

I thought Profile Details was  legend! So i'd be great to show both cases, when Textfield are actually grouped.

image

docs/src/Fieldset.doc.js
AlbertCarreras: Maybe OOS, but we could repro our settings page 
image
very simplified, but to show a more realistic usage and different fieldsets working together withng a form

docs/src/Fieldset.doc.js
```diff
@@ -0,0 +1,222 @@
...
+    </MainSection.Subsection>
+  </MainSection>,
+);
+
+card(<MainSection name="Localization" description={`Be sure to localize the \`legend\` text.`} />);
+
+card(
+  <MainSection name="Variants">
+    <MainSection.Subsection
+      description={`Be default, the \`legend\` is visible above the items in the Fieldset. However, if the form items are labelled by content elsewhere on the page, or a more complex legend is needed, the \`legendDisplay\` prop can be used to visually hide the legend. In this case, it is still available to screen reader users, but will not appear visually on the screen.`}
```
AlbertCarreras, line 160: By default typo

AlbertCarreras: not strong about this, but maybe we can specify what the screen readers will read something like.

In the example below,  screenreaders will read in the following order:

1. Design options
2. Choose which backgrounds to use as placeholders when creating content
3. Which backgrounds would you like to use?
4.

docs/src/Fieldset.doc.js
```diff
@@ -0,0 +1,222 @@
...
+        <Text size="md">Choose which backgrounds to use as placeholders when creating content</Text>
+      </Box>
+      <Fieldset legend="Which backgrounds would you like to use?" legendDisplay="hidden">
+        <Flex direction="column" gap={4}>
+          <Checkbox
+            checked={checkedBotanical}
+            id="coral"
+            label="Coral"
+            subtext="Botanical art in coral and green"
+            image={<Box height={100} width={80}><Image alt="Botanical art in coral and green" src="https://i.ibb.co/7bQQYkX/stock2.jpg" fit="contain" naturalWidth={1} naturalHeight={1}/></Box>}
```
AlbertCarreras, line 183: can we get the forat from the VSCode? Image might get reformated.

docs/src/Fieldset.doc.js
```diff
@@ -0,0 +1,222 @@
...
+card(
+  <PageHeader
+    name="Fieldset"
+    description="Creates an accessible fieldset and legend for a group of related form items"
+    defaultCode={`
+      function RadioButtonExample() {
+        const [favorite, setFavorite] = React.useState(undefined);
+
+        return (
+          <Fieldset legend="What is your favorite pet?">
```
AlbertCarreras, line 20: image

Maybe something more businessy?

docs/src/Fieldset.doc.js
```diff
@@ -0,0 +1,222 @@
...
+      />
+    </MainSection.Subsection>
+  </MainSection>,
+);
+
+card(
+  <MainSection name="Related">
+    <MainSection.Subsection
+      description={`
+      **[Label](/Label)**
```
AlbertCarreras, line 215: Should we add a best practice do/don't explicitly? "Don't use Fieldset for single form elements. Use Label or the build-in label prop on them."

docs/src/Fieldset.doc.js
AlbertCarreras, line 86: "When to use Fieldset", don't feel super strong about it. I see the intention of informing when to use it, but technically the WHEN should be addressed on best practices or the description itself. I see this more as focusing on the a11y part of it. Maybe specifying the benefits of accesible forms.

AlbertCarreras: I'd point out that the example below has a fieldset used for grouping x.

packages/gestalt/src/Fieldset.test.js
```diff
@@ -0,0 +1,26 @@
...
+  });
+
+  it('renders a visually hidden legend', () => {
+    const tree = create(
+      <Fieldset legend="What is your favorite dog?" legendDisplay="hidden">
+        <Checkbox id="Schnauzer" label="Schnauzer" onChange={() => {}} />
+        <Checkbox id="Aussie" label="Aussie" onChange={() => {}} />
+      </Fieldset>,
+    ).toJSON();
+    expect(tree).toMatchSnapshot();
```
AlbertCarreras, line 24: could we do a test that checks that finds (What is your favorite dog?) and returns 1 when visible and 0 when hidden?

docs/src/Checkbox.doc.js
```diff
@@ -274,8 +266,8 @@ function CheckboxExample() {
             setCheckedSp(checked);
           }}
         />
-      </Box>
-    </Box>
+      </Flex>
+    </Fieldset>
   );
 }
 `}
```
AlbertCarreras, line 273: Should we add same comment for Textfield?

packages/gestalt/src/Fieldset.js
```diff
@@ -0,0 +1,34 @@
...
+
+type Props = {|
+  children: Node,
+  legend: string,
+  legendDisplay?: 'visible' | 'hidden',
+|};
+
+export default function Fieldset({ legend, legendDisplay = 'visible', children }: Props): Node {
+  return (
+    <fieldset className={cx(formStyles.unstyled, whitespaceStyles.p0, whitespaceStyles.m0)}>
```
AlbertCarreras, line 20: whitespaceStyles.p0 could we use here `boxWhitespaceStyles.paddingX0` instead of `whitespaceStyles.p0`, just wondering, haven't played with the code, so it might change the style, but it should be the same right?

packages/gestalt/src/Fieldset.js
```diff
@@ -0,0 +1,34 @@
+// @flow strict
+import type { Node } from 'react';
+import cx from 'classnames';
+import Text from './Text.js';
+import labelStyles from './Label.css';
+import formStyles from './FormElement.css';
+import formLabelStyles from './FormLabel.css';
+import boxWhitespaceStyles from './boxWhitespace.css';
```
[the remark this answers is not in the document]
AlbertCarreras: Agree

packages/gestalt/src/Fieldset.js
```diff
@@ -0,0 +1,34 @@
...
+  children: Node,
+  legend: string,
+  legendDisplay?: 'visible' | 'hidden',
+|};
+
+export default function Fieldset({ legend, legendDisplay = 'visible', children }: Props): Node {
+  return (
+    <fieldset className={cx(formStyles.unstyled, whitespaceStyles.p0, whitespaceStyles.m0)}>
+      <legend
+        className={cx(
```
AlbertCarreras, line 22: can we do here `classnames` to be more consistent? 
image

packages/gestalt/src/Fieldset.js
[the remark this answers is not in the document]
AlbertCarreras: What about `Label.css` and `FormLabel.css` i feel we could also combine those 2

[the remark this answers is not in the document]
AlbertCarreras: We should also combine `Label.css` and `FormLabel.css`

packages/gestalt/src/Fieldset.js
```diff
@@ -0,0 +1,34 @@
...
+          labelStyles.label,
+          formLabelStyles.formLabel,
+          boxWhitespaceStyles.paddingX0,
+          legendDisplay === 'hidden' && boxStyles.xsDisplayVisuallyHidden,
```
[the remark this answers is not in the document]
AlbertCarreras: me just wondering..., should we standardized how we do conditionals in classnames? 
like getting an object `{[style]: true/false }` vs true/false & [style]

I think it doesn't matter from a code perspective, just wondering.
image

packages/gestalt/src/Fieldset.js
```diff
@@ -0,0 +1,34 @@
...
+|};
+
+export default function Fieldset({ legend, legendDisplay = 'visible', children }: Props): Node {
+  return (
+    <fieldset className={cx(formStyles.unstyled, whitespaceStyles.p0, whitespaceStyles.m0)}>
+      <legend
+        className={cx(
+          labelStyles.label,
+          formLabelStyles.formLabel,
+          boxWhitespaceStyles.paddingX0,
```
AlbertCarreras, line 25: I was like why do we need this  `boxWhitespaceStyles.paddingX0,`

It seems leyend come with padding. 
image

AlbertCarreras:
Looks great! I added many comments but they shouldn't be blocking the PR. Feel free to take my feedback and do some changes, or not if feels OOS or too much.

AlbertCarreras (approved):
Looks great to me. Thanks for this @ayeshakmaz . I added many comments, but they shouldn't block the PR cos the component works and it's awesome. Feel free to take the feedback I gave and incorporate it or not if they are OOS or unnecessary.

packages/gestalt/src/Fieldset.test.js
ayeshakmaz: Since the legend is only hidden with css, the test would just be looking for the visuallyHidden class name, which is essentially the snapshot test anyway so I think it's probably fine. This is where Visual Regression Testing would be ideal instead

ayeshakmaz:
To answer some of the questions @ponori and @AlbertCarreras , this pilot version is more so for radios and checkboxes, but can easily be used for textfields in situations where the legend is hidden. The next iteration will include more styling options for the legend that would allow for better hierarchy when using this to group TextFields with a visible legend. I didn't add any specific text field examples for now, but we can add those in once we add the hierarchy styling.

docs/src/Fieldset.doc.js
ayeshakmaz: We can add this in the next iteration when we add some styling options too

docs/src/Fieldset.doc.js
ayeshakmaz: Kept the top one fun, but made the others more realistic

PR #1495, opened 2021-05-06
Sheet: New docs and best practices

ayeshakmaz:
## Description

New Sheet docs

### Links

- JIRA: https://jira.pinadmin.com/browse/PDS-2537
- [Paper doc](https://paper.dropbox.com/doc/Sheet--BKIvuwA~9LvsbY43OcUdE_RfAg-pp6qxEdyanIdQbBuE2wo8)

### Checklist

### DIFF
--- a/docs/src/Sheet.doc.js
+++ b/docs/src/Sheet.doc.js
@@ -1,9 +1,8 @@
 // @flow strict
 import type { Node } from 'react';
 import PropTable from './components/PropTable.js';
-import Example from './components/Example.js';
 import PageHeader from './components/PageHeader.js';
-import Card from './components/Card.js';
+import MainSection from './components/MainSection.js';
 
 const cards: Array<Node> = [];
 const card = (c) => cards.push(c);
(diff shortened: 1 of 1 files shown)


docs/src/Sheet.doc.js
```diff
@@ -1,17 +1,24 @@
...
+import MainSection from './components/MainSection.js';
 
 const cards: Array<Node> = [];
 const card = (c) => cards.push(c);
 
 card(
   <PageHeader
     name="Sheet"
-    description="Sheets are side surfaces that allow users to view information or complete sub-tasks in a workflow while keeping the context of the current page."
+    description="Sheets are surfaces that allow users to view optional information or complete sub-tasks in a workflow while keeping the context of the current page. The most common example of Sheet displays content in a panel that opens from the side of the screen for the user to read or input information. Sheets have a default internal padding for content."
```
dangerismycat, line 13: nittiest of nits: "Sheets have default internal padding…" The "a" in there sounds kinda weird?

docs/src/Sheet.doc.js
```diff
@@ -23,472 +30,445 @@ card(
...
         required: false,
         defaultValue: null,
-        description: [
-          'Supply the text that is going to be placed as the Sheet text heading.',
-          'Please do not repeat the same text being passed in the "accessibilitySheetLabel" prop, but instead provide something that identifies the Sheet. For instance, if the heading is "Pin Builder", the accessibilitySheetLabel can be "Create a new Pin".',
-          'Obs: This element will be padded by 32px, similarly to `<Modal>`.',
-        ],
-        href: 'defaultPaddingAndStylingExample',
+        description:
+          "Supply the text that is going to be placed as the Sheet's text heading. Be sure to localize this text. See the [heading variant](#Heading) for more info.",
```
dangerismycat, line 74: "is going to be placed as" is a bit wordy…perhaps something like "Text used for Sheet's heading."?

Also some articles that could be removed here and throughout, but that can always be follow-up work and I won't call them out here 🙂

docs/src/Sheet.doc.js
```diff
@@ -1,17 +1,24 @@
...
 import PropTable from './components/PropTable.js';
-import Example from './components/Example.js';
 import PageHeader from './components/PageHeader.js';
-import Card from './components/Card.js';
+import MainSection from './components/MainSection.js';
 
 const cards: Array<Node> = [];
 const card = (c) => cards.push(c);
 
 card(
```
dangerismycat, line 10: Don't forget the feedback callout! We might actually get some real feedback on this new sandbox approach 🤞

docs/src/Sheet.doc.js
```diff
@@ -23,472 +30,445 @@ card(
...
+        cardSize="lg"
+        iframeContent={`
+        <iframe src="https://codesandbox.io/embed/magical-christian-2kk1c?fontsize=14&hidenavigation=1&theme=light&view=preview"
+        style={{width: '100%', height:'600px', border:'0', borderRadius: '4px', overflow:'hidden'}}
+        title="Accessibility Labels"
+        allow="accelerometer; ambient-light-sensor; camera; encrypted-media; geolocation; gyroscope; hid; microphone; midi; payment; usb; vr; xr-spatial-tracking"
+        sandbox="allow-forms allow-modals allow-popups allow-presentation allow-same-origin allow-scripts"
+      ></iframe>
+        `}
+        defaultCode={`
```
dangerismycat, line 191: Ah, the one downfall of the sandbox approach: it's no longer live-editable within the docs page. Bummer. I don't think that's a blocker by any means, but it is something we should probably communicate to the user. Otherwise I expect confusion and frustration.

docs/src/Sheet.doc.js
```diff
@@ -23,472 +30,445 @@ card(
...
         </Layer>
       )}
     </React.Fragment>
   );
 }
 `}
+      />
+    </MainSection.Subsection>
+    <MainSection.Subsection
+      title="Focus management"
```
dangerismycat, line 312: Should we add an example for this one? Probably not sandboxed, so we can demonstrate how focus is placed when the Sheet opens

docs/src/Sheet.doc.js
```diff
@@ -615,113 +599,501 @@ function SubheadingExample(props) {
...
+Sheet comes in 3 sizes: small (\`sm\`), medium (\`md\`), and large (\`lg\`).
+- Small Sheets (540px) are primarily used for displaying information or acting as a point to link to other content. They are the least commonly used.
+- Medium Sheets (720px) are the standard size offered for content.
+- Large Sheets (900px) should be used in cases where there may be columns of content or navigation where the additional space is required to keep the content at a comfortable reading width.
+`}
+    >
+      <MainSection.Card
+        cardSize="lg"
+        defaultCode={`
+function SizesExample(props) {
```
dangerismycat, line 757: Possible future improvement: use different content for each size to better exemplify when each size should be used

dangerismycat (approved):
This is so great — such a good improvement. Thanks for this!!

docs/src/Sheet.doc.js
ayeshakmaz: oh good point!

docs/src/Sheet.doc.js
ayeshakmaz: Once we fix the focus bug, yes :P

dangerismycat: http://gph.is/1E2ssrH

docs/src/Sheet.doc.js
ayeshakmaz: Yeah I was definitely planning to update the Large size once we have SideNav :D

dangerismycat: Product tagging on Pinbuilder is a pretty neat `md` example that's not so form-heavy (makes good use of the sub-heading and footer as well)

PR #1498, opened 2021-05-09
Internal: combine babel configs

christianvuerings:
## Description

1 source of truth for our babel configuration. Previously we had to copy/paste between these 2 config locations which is error prone.


dangerismycat (approved):
Thanks Christian!!

PR #1501, opened 2021-05-10
ESLint: add rule to disallow non-standard Box props

christianvuerings:
## Description

Lint against non-standard `Box` props since they result in warnings.

Example

```js
<Box backgroundColor="#fff" />
```

Results in

> Warning: React does not recognize the `backgroundColor` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `backgroundcolor` instead. If you accidentally passed it from a parent component, remove it from the DOM element.

### Solution

Add an ESLint rule to disallow non-standard Box props. That way we can detect these issues as we code and will no longer have these warnings show up.

### What do you dislike about this solution?

That we duplicate the list of `Box` props in the ESLint rule. However, that seems better than trying to import `Box` in the ESLint plug-in, running an AST to get the list of available props. That would make the rule slower and harder to maintain.

### Checklist

- [x] Added Unit and Flow Tests

### DIFF
--- a/packages/eslint-plugin-gestalt/src/no-box-disallowed-props.js
+++ b/packages/eslint-plugin-gestalt/src/no-box-disallowed-props.js
@@ -1,0 +1,140 @@
+/**
+ * Error on disallowed props on `Box`
+ */
+
+// @flow strict
+const allowedBaseProps = [
+  'dangerouslySetInlineStyle',
+  'display',
+  'column',
+  'direction',
+  'smDisplay',
+  'smColumn',
+  'smDirection',
+  'mdDisplay',
+  'mdColumn',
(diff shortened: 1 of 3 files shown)


dangerismycat (approved):
Oh, this is nice. :r-plus: from me though I'll hold off on merging it in case @AlbertCarreras or @ayeshakmaz have any concerns

christianvuerings:
@AlbertCarreras or @ayeshakmaz could either of you also take a look?

ayeshakmaz (approved):
LGTM!

AlbertCarreras (approved):
It's gonna be fun to merge

PR #1506, opened 2021-05-10
Bugfix: Add units to `border-radius` values

dangerismycat:
For some reason `border-radius` values without units stopped being respected, seemingly tied to the recent [stylelint](https://github.com/pinterest/gestalt/pull/1504) upgrade. This PR adds units where needed.


christianvuerings (approved):

christianvuerings:
@dangerismycat totally my mistake in https://github.com/pinterest/gestalt/pull/1499

Seems like we should add `px` in 1 more place:

image

PR #1507, opened 2021-05-13
Button: add new transparentWhite color

AlbertCarreras:
Added color to match color='white' in `IconButton`.

Kapture 2021-05-13 at 19 23 42

### Links

- [TDD](paper doc link)
- [Figma](link to figma file)

### Checklist

### DIFF
--- a/docs/src/Button.doc.js
+++ b/docs/src/Button.doc.js
@@ -71,7 +71,7 @@ card(
       },
       {
         name: 'color',
-        type: `'gray' | 'red' | 'blue' | 'transparent' | 'transparentWhiteText' | 'white'`,
+        type: `'gray' | 'red' | 'blue' | 'transparent' |  'transparentWhite' | 'transparentWhiteText' | 'white'`,
         required: false,
         defaultValue: 'gray',
         description: [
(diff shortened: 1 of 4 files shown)


ashleyseto:
I'm having trouble finding the example in the docs. Only a coupe notes for the example.

- It'd be good if we could put these over an image since our guidance will be to use over images
- We generally recommend the icon in the icon button be the dark gray rather than gray so we should show that here.

Thanks for getting this done so quickly!

docs/src/Button.doc.js
```diff
@@ -71,7 +71,7 @@ card(
       },
       {
         name: 'color',
-        type: `'gray' | 'red' | 'blue' | 'transparent' | 'transparentWhiteText' | 'white'`,
+        type: `'gray' | 'red' | 'blue' | 'transparent' |  'transparentWhite' | 'transparentWhiteText' | 'white'`,
```
ayeshakmaz, line 74: This needs to be updated

docs/src/Button.doc.js
```diff
@@ -327,48 +327,86 @@ card(
 );
 
 card(
-  <Combination id="color" name="Color" color={['gray', 'red', 'blue', 'white', 'transparent']}>
+  <Combination
+    id="color"
+    name="Color"
+    color={['gray', 'red', 'blue', 'white', 'transparent', 'transparentWhite']}
```
ayeshakmaz, line 333: Same here

docs/src/Button.doc.js
```diff
@@ -327,48 +327,86 @@ card(
...
+  >
     {(props, i) => <Button id={`example-${i}`} onChange={() => {}} {...props} text="Button" />}
   </Combination>,
 );
 
 card(
   <Example
-    name="Color: transparent"
+    name="Color: transparent and semiTransparentWhite"
```
ayeshakmaz, line 341: Love the dual example!

packages/gestalt/src/Button.js
```diff
@@ -205,7 +213,7 @@ const ButtonWithForwardRef: React$AbstractComponent<unionProps, unionRefs> = for
         wrappedComponent="button"
       >
         {iconEnd ? (
-          <IconEnd text={buttonText} textColor={textColor} icon={iconEnd} size={size} />
+          <IconEnd text={buttonText} textColor="darkGray" icon={iconEnd} size={size} />
```
ayeshakmaz, line 216: I don't think these two should be hard coded, it creates issues when you disabled or select them
Screen Shot 2021-05-14 at 9 48 35 AM

Was there a need to change them?

docs/src/Button.doc.js
AlbertCarreras: DONE

docs/src/Button.doc.js
AlbertCarreras: DONE

packages/gestalt/src/Button.js
AlbertCarreras: LEGACY DEBUGGING... MY BAD

DONE

ayeshakmaz (approved):
LGTM 👍

PR #1508, opened 2021-05-13
Datapoint: remove compression on click

AlbertCarreras:
## Main Change
Replace `Iconbutton` with `TapArea` + `Icon` to remove default click compression in  IconButtons (info Icon). 
Tooltips are rendered on hover and the click event has no effect on hiding/showing the tooltip.
The compression on click is confusing as the user could expect an event after clicking the button.
 
Context: An interactive element is needed inside the Tooltip as Tooltip should be anchored with interactive elements. Interactive elements provide keyboard navigation. On keyboard navigation, the tooltip is rendered automatically. TapArea with tapStyle="none" provides the same navigation without the compression.

## Manual testing

Kapture 2021-05-13 at 19 35 15

Kapture 2021-05-13 at 19 37 43

### BEFORE

Kapture 2021-05-13 at 19 51 15

### AFTER

Kapture 2021-05-13 at 21 07 19

### Links

- [TDD](paper doc link)
- [Figma](link to figma file)

### Checklist

### DIFF
--- a/packages/gestalt/src/Datapoint.js
+++ b/packages/gestalt/src/Datapoint.js
@@ -4,8 +4,9 @@ import PropTypes from 'prop-types';
 import Text from './Text.js';
 import Heading from './Heading.js';
 import Flex from './Flex.js';
-import IconButton from './IconButton.js';
+import Icon from './Icon.js';
 import Tooltip from './Tooltip.js';
+import TapArea from './TapArea.js';
 import DatapointTrend from './DatapointTrend.js';
 
 type TrendObject = {|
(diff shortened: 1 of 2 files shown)


packages/gestalt/src/Datapoint.js
```diff
@@ -41,13 +42,9 @@ export default function Datapoint({
...
         {tooltipText && (
           <Tooltip text={tooltipText} idealDirection="up">
-            <IconButton
-              accessibilityLabel=""
-              size="sm"
-              icon="info-circle"
-              iconColor="gray"
-              padding={1}
-            />
+            <TapArea accessibilityLabel="" rounding="circle" tapStyle="none">
```
dangerismycat, line 45: Can you add a comment here that the accessibilityLabels are taken care of by the Tooltip?

dangerismycat (approved):

packages/gestalt/src/Datapoint.js
AlbertCarreras: DONE

PR #1511, opened 2021-05-17
ESLint: add native React props to no-box-disallowed-props

christianvuerings:
## Description

Certain props like `key` / `id` should be allowed on `Box`, so let's add them to our rule.


packages/eslint-plugin-gestalt/src/no-box-disallowed-props.js
```diff
@@ -4,6 +4,49 @@
 
 // @flow strict
 const allowedBaseProps = [
+  // React / DOM
+  'id',
```
christianvuerings, line 8: This list contains props we found in the Pinterest codebase & ones referenced in https://github.com/jsx-eslint/jsx-ast-utils/blob/c554f28b6b1deaf57d6e244aeac3bd7fc6c31039/src/eventHandlers.js#L16 (do know that some of these don't make sense on `div` / `main` / ..., so I didn't include all of them)

AlbertCarreras (approved):

PR #1513, opened 2021-05-17
TextField, TextArea: Open errorMessage type to accept Node

dangerismycat:
We have need in the product for clickable text in the error messages for TextField, specifically as part of the login flow:
Screen Shot 2021-05-17 at 12 00 54 PM

In order to allow this, we need to update the `errorMessage` type from `string` to `React.Node`. Unfortunately this opens the flood gates to pass anything as the error message, so we'll need to monitor usage to avoid abuse. I added a note about usage in the prop description.

### Reviewers
Paste this code into an example to test the new functionality (replace TextField with TextArea accordingly for that doc).

I specifically didn't update the docs with examples of this behavior because it's something we want to _support_ but not _encourage_.

```js
function Example(props) {
  const [value, setValue] = React.useState('')

  const errorMessage = (

    <React.Fragment>
      This field can't be
      <button
    onClick={() => {
      alert('CLICK')
    }}
    style={{ background: 'transparent', border: 'none', verticalAlign: 'inherit', padding: 0 }}
    type="button"
  >
    <Text key="err_text" color="red" inline size="lg" weight="bold">
      blank!
    </Text>
  </button>
    </React.Fragment>
  );

  return (
    <TextField
      id="aboutme"
      errorMessage={errorMessage}
      onChange={({ value }) => setValue(value)}
      label="With an error message"
      value={value}
    />
  );
}
```

### DIFF
--- a/docs/src/TextArea.doc.js
+++ b/docs/src/TextArea.doc.js
@@ -168,6 +168,31 @@ function Example(props) {
 
 card(
   <Example
+    id="errorMessageExample"
+    name="Example: Error message"
+    description={`
+    A TextArea can display its own error message.
+    To use our errors, simply pass in an \`errorMessage\` when there is an error present and we will     handle the rest.`}
+    defaultCode={`
+function Example(props) {
+  const [value, setValue] = React.useState('')
+  return (
+    <TextArea
+      id="witherror"
+      onChange={({value}) => setValue(value)}
+      errorMessage={!value ? "This field can't be blank!" : null}
+      placeholder="Write something about yourself..."
+      label="With an error message"
+      value={value}
+    />
+  );
+}
+`}
+  />,
+);
+
+card(
+  <Example
     id="refExample"
     name="Example: ref"
     description={`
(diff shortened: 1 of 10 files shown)


docs/src/TextArea.doc.js
```diff
@@ -166,6 +168,31 @@ function Example(props) {
   />,
 );
 
+card(
+  <Example
+    id="errorMessageExample"
+    name="Example: Error message"
+    description={`
+    A TextArea can display its own error message.
+    To use our errors, simply pass in an \`errorMessage\` when there is an error present and we will     handle the rest.`}
```
AlbertCarreras, line 177: can we replace "our" with "built in" or "default"?


also big space there: `we will handle the rest`

docs/src/TextArea.doc.js
```diff
@@ -166,6 +168,31 @@ function Example(props) {
...
+    defaultCode={`
+function Example(props) {
+  const [value, setValue] = React.useState('')
+  return (
+    <TextArea
+      id="witherror"
```
AlbertCarreras, line 183: id="with-error"

docs/src/TextArea.doc.js
AlbertCarreras: we will handle the rest. could also be said as `TextArea handles the error display`

docs/src/TextField.doc.js
```diff
@@ -25,8 +25,10 @@ card(
       },
       {
         name: 'errorMessage',
-        type: 'string',
+        type: 'React.Node',
         href: 'errorMessageExample',
+        description:
+          'For most use cases, pass a string with a helpful error message (be sure to localize!). In certain instances it can be useful to make some text clickable; to support this, you may instead pass a React.Node to wrap text in Link or TapArea.',
```
AlbertCarreras, line 31: `Be sure to localize the error message. ` to keep consistent tone.

AlbertCarreras (approved):
LGTM, left some nit comments

docs/src/TextArea.doc.js
dangerismycat: I wasn't updating any of the existing content — we'll take a pass at this in future work when we overall the whole page

PR #1515, opened 2021-05-17
Video: Add missing video event handlers

valiantvu:
Adds event handling for the following video events: error, loadStart, playing, seeking, and waiting. These event handlers are used for context logging, specifically for the VIDEO_PLAYBACK_PERFORMANCE event. 

More background:
Since these props are missing from the Video component, we currently grab video element ref after the video element has been created and then add the event handlers. This is problematic specifically for error handling since errors can occur before the error handler is attached to the video element.


packages/gestalt/src/Video.js
```diff
@@ -400,6 +412,15 @@ export default class Video extends PureComponent<Props, State> {
     }
   };
 
+  // Sent when an error occurs.
+  handleError: (event: SyntheticEvent<HTMLDivElement>) => void = (event) => {
+    const { onError } = this.props;
+
+    if (onError) {
```
dangerismycat, line 419: nit: you can conditionally call functions to avoid this kind of if statement:
```js
onError?.({ event });
```

dangerismycat (approved):
LGTM! Thanks Michelle!

PR #1516, opened 2021-05-19
Tabs: Add ref to tab items

dangerismycat:
In at least one place in Pinboard, we're passing text wrapped in a Box setting a ref and a height to attach an Experience popover to a tab. That height breaks the Tabs V2 UI, but is necessary to properly space the popover from the entire tab element (since the ref is on the text, not the element as a whole). By allowing the user to set a ref on a tab element, it will no longer be necessary to use the wrapping Box with a height set, this unbreaking the Tabs V2 UI.

### DIFF
--- a/packages/gestalt/src/Tabs.js
+++ b/packages/gestalt/src/Tabs.js
@@ -1,7 +1,5 @@
 // @flow strict
-import type { Node } from 'react';
-
-import { useState } from 'react';
+import { type Node, useState } from 'react';
 import PropTypes from 'prop-types';
 import Box from './Box.js';
 import Flex from './Flex.js';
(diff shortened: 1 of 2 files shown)


packages/gestalt/src/Tabs.js
```diff
@@ -91,6 +90,7 @@ function Tab({
   indicator,
   isActive,
   onChange,
+  ref,
```
AlbertCarreras, line 93: can we use forwardRef here?

packages/gestalt/src/Tabs.js
```diff
@@ -153,6 +153,7 @@ function TabV2({
   index,
   isActive,
   onChange,
+  ref,
```
AlbertCarreras, line 156: same here. I'm  not sure if ref is passed... 
image

dangerismycat: yep, you're right 😬😅

AlbertCarreras (approved):

PR #1521, opened 2021-05-24
Modal: New docs and Best Practices

ayeshakmaz:
### Links

- JIRA: https://jira.pinadmin.com/browse/PDS-2536
- [TDD](paper doc link)
- [Figma](link to figma file)

### Checklist

### DIFF
--- a/docs/src/Modal.doc.js
+++ b/docs/src/Modal.doc.js
@@ -21,7 +21,7 @@ card(
       style={{width: '100%', height:'500px', border:'0', borderRadius: '4px', overflow:'hidden'}}
       title="Modal Main Example"
       sandbox="allow-forms allow-modals allow-popups allow-presentation allow-same-origin allow-scripts"
-    ></iframe>
+      ></iframe>
     `}
   />,
 );
@@ -116,7 +116,7 @@ card(
           style={{width: '100%', height:'500px', border:'0', borderRadius: '4px', overflow:'hidden'}}
           title="Modal required response"
           sandbox="allow-forms allow-modals allow-popups allow-presentation allow-same-origin allow-scripts"
-        ></iframe>
+          ></iframe>
         `}
       />
 
@@ -129,7 +129,7 @@ card(
           style={{width: '100%', height:'500px', border:'0', borderRadius: '4px', overflow:'hidden'}}
           title="Modal with actions"
           sandbox="allow-forms allow-modals allow-popups allow-presentation allow-same-origin allow-scripts"
-        ></iframe>
+          ></iframe>
         `}
       />
       <MainSection.Card
@@ -141,7 +141,7 @@ card(
           style={{width: '100%', height:'500px', border:'0', borderRadius: '4px', overflow:'hidden'}}
           title="Modal without scrolling"
           sandbox="allow-forms allow-modals allow-popups allow-presentation allow-same-origin allow-scripts"
-        ></iframe>
+          ></iframe>
         `}
       />
     </MainSection.Subsection>
@@ -184,12 +184,11 @@ card(
       <MainSection.Card
         cardSize="lg"
         iframeContent={`
-        <iframe src="https://codesandbox.io/embed/trusting-wu-c8514?fontsize=14&hidenavigation=1&theme=light&view=preview"
-        style={{width: '100%', height:'500px', border:'0', borderRadius: '4px', overflow:'hidden'}}
-        title="Accessibility Labels"
-        allow="accelerometer; ambient-light-sensor; camera; encrypted-media; geolocation; gyroscope; hid; microphone; midi; payment; usb; vr; xr-spatial-tracking"
-        sandbox="allow-forms allow-modals allow-popups allow-presentation allow-same-origin allow-scripts"
-      ></iframe>
+          <iframe src="https://codesandbox.io/embed/trusting-wu-c8514?fontsize=14&hidenavigation=1&theme=light&view=preview"
+          style={{width: '100%', height:'500px', border:'0', borderRadius: '4px', overflow:'hidden'}}
+          title="Accessibility Labels"
+          sandbox="allow-forms allow-modals allow-popups allow-presentation allow-same-origin allow-scripts"
+          ></iframe>
         `}
         defaultCode={`
 function AccessibilityExample(props) {
@@ -232,13 +231,12 @@ function AccessibilityExample(props) {
         title='role="alertdialog"'
         cardSize="lg"
         iframeContent={`
-    <iframe src="https://codesandbox.io/embed/trusting-wu-c8514?fontsize=14&hidenavigation=1&theme=light&view=preview"
-    style={{width: '100%', height:'400px', border:'0', borderRadius: '4px', overflow:'hidden'}}
-    title="Accessibility Role ex1"
-    allow="accelerometer; ambient-light-sensor; camera; encrypted-media; geolocation; gyroscope; hid; microphone; midi; payment; usb; vr; xr-spatial-tracking"
-    sandbox="allow-forms allow-modals allow-popups allow-presentation allow-same-origin allow-scripts"
-  ></iframe>
-    `}
+          <iframe src="https://codesandbox.io/embed/trusting-wu-c8514?fontsize=14&hidenavigation=1&theme=light&view=preview"
+            style={{width: '100%', height:'400px', border:'0', borderRadius: '4px', overflow:'hidden'}}
+            title="Accessibility Role ex1"
+            sandbox="allow-forms allow-modals allow-popups allow-presentation allow-same-origin allow-scripts"
+          ></iframe>
+        `}
         defaultCode={`
 function AlertDialogAccessibilityExample(props) {
 const [showModal, setShowModal] = React.useState(false);
@@ -274,13 +272,12 @@ return (
         title='role="dialog" (default)'
         cardSize="lg"
         iframeContent={`
-    <iframe src="https://codesandbox.io/embed/fervent-rosalind-xrzm5?fontsize=14&hidenavigation=1&theme=light&view=preview"
-    style={{width: '100%', height:'450px', border:'0', borderRadius: '4px', overflow:'hidden'}}
-    title="Accessibility Role ex2"
-    allow="accelerometer; ambient-light-sensor; camera; encrypted-media; geolocation; gyroscope; hid; microphone; midi; payment; usb; vr; xr-spatial-tracking"
-    sandbox="allow-forms allow-modals allow-popups allow-presentation allow-same-origin allow-scripts"
-  ></iframe>
-    `}
+          <iframe src="https://codesandbox.io/embed/fervent-rosalind-xrzm5?fontsize=14&hidenavigation=1&theme=light&view=preview"
+          style={{width: '100%', height:'450px', border:'0', borderRadius: '4px', overflow:'hidden'}}
+          title="Accessibility Role ex2"
+          sandbox="allow-forms allow-modals allow-popups allow-presentation allow-same-origin allow-scripts"
+          ></iframe>
+        `}
         defaultCode={`
     function DialogExample(props) {
       const ModalWithHeading = ({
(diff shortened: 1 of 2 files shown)


docs/src/Modal.doc.js
```diff
@@ -23,85 +32,497 @@ card(
...
+        </Layer>
+      )}
+    </React.Fragment>
+  );
+}
+`}
+      />
+    </MainSection.Subsection>
+    <MainSection.Subsection
+      title="Sizes"
```
ayeshakmaz, line 518: Didn't update these yet with real content, but can do that in a second pass once we get examples from product

docs/src/Modal.doc.js
```diff
@@ -1,17 +1,26 @@
...
+import MainSection from './components/MainSection.js';
 
 const cards: Array<Node> = [];
 const card = (c) => cards.push(c);
 
 card(
   <PageHeader
     name="Modal"
-    description="Modals are light temporary views; general purpose containers allowing for a multitude of design choices."
+    description="Modals display content that requires user interaction on a layer above the page content. Modals block the content underneath, preventing users from interacting with anything else besides the Modal.
```
dangerismycat, line 16: nit: the first sentence is a little hard to parse at first glance. Does the content require user interaction, or the Modal? Is the user interaction on a layer of the page content, or is the Modal?

docs/src/Modal.doc.js
```diff
@@ -1,17 +1,26 @@
...
+
+    The most common example of Modal is confirming an action the user has taken. Modal also provides some of the padding required for styling out of the box."
```
dangerismycat, line 18: Not sure if the second sentence (about padding) makes sense here. This paragraph is an overview and that's more of a detail.

docs/src/Modal.doc.js
```diff
@@ -1,17 +1,26 @@
...
+    defaultCode={`
+      <iframe src="https://codesandbox.io/embed/trusting-wu-c8514?fontsize=14&hidenavigation=1&theme=light&view=preview"
+      style={{width: '100%', height:'500px', border:'0', borderRadius: '4px', overflow:'hidden'}}
+      title="Modal Main Example"
+      allow="accelerometer; ambient-light-sensor; camera; encrypted-media; geolocation; gyroscope; hid; microphone; midi; payment; usb; vr; xr-spatial-tracking"
```
dangerismycat, line 23: Do we need to allow all this? Seems like a possible privacy concern. I don't think any of those should be necessary for our Modal in a sandbox?

docs/src/Modal.doc.js
```diff
@@ -23,85 +32,497 @@ card(
...
+          'String that clients such as VoiceOver will read to describe the modal. Always localize the label. See the [Accessibility section](#Accessibility) for more info.',
       },
       {
         name: 'align',
         type: `"center" | "left"`,
         defaultValue: 'center',
-        description: 'Use to specify the alignment of `heading` & `subHeading` strings',
-        href: 'headingOptions',
+        description:
+          'Use to specify the alignment of `heading` & `subHeading` strings. See the [Heading variant](#Heading) for more info.',
```
dangerismycat, line 45: We should've called this prop `headingAlign` or something similar. 😕

docs/src/Modal.doc.js
```diff
@@ -23,85 +32,497 @@ card(
...
+          'Close the modal when you click outside of it. See the [outside click variant](#Preventing-close-on-outside-click) for more info.',
         defaultValue: true,
-        href: 'closeOnOutsideClickExample',
       },
       {
         name: 'footer',
         type: 'React.Node',
-        href: 'sizesExample',
+        description:
+          "Supply the container element(s) or render prop that will be used as Modal's custom footer. See the [Best Practices](#Best-practices) for more info.",
```
dangerismycat, line 64: I don't believe `footer` supports function children. It only accepts a Node, not `() => Node`

docs/src/Modal.doc.js
```diff
@@ -23,85 +32,497 @@ card(
...
+        description:
+          "The text used for Modal's heading. Be sure to localize this text. See the [Heading variant](#Heading) for more info.",
       },
       {
         name: 'onDismiss',
         type: '() => void',
         required: true,
-        href: 'sizesExample',
+        description:
+          'Callback fired when Modal is dismissed by clicking on the a dismiss-type Footer button or clicking on the backdrop outside of the Modal (if `closeOnOutsideClick` is true).',
```
dangerismycat, line 77: Unless I'm missing something, there's no such thing as a "dismiss-type" button, and it's nothing Modal pays attention to anyway. `onDismiss` is only called when the modal is closed by outside click. We perhaps _should_ support this being called by a footer button, but we don't currently.

docs/src/Modal.doc.js
```diff
@@ -23,85 +32,497 @@ card(
...
       },
       {
         name: 'ref',
         type: "React.Ref<'div'>",
         description: 'Forward the ref to the underlying div element',
```
dangerismycat, line 82: This description makes it sound like `ref` is a boolean indicating if the user wants a ref forwarded. Consider something like "Ref that will set on the underlying `<div>` element."

docs/src/Modal.doc.js
```diff
@@ -23,85 +32,497 @@ card(
...
-      All Modals have a max width of 100%.
+  <MainSection name="Best practices">
+    <MainSection.Subsection>
+      <MainSection.Card
+        cardSize="lg"
+        showCode={false}
+        type="do"
+        description="Use Modal when a response is required from the user. Clearly communicate what response is expected and make the action simple and straight forward, such as clicking a button to confirm. The most common responses will be related to confirming or canceling."
+        defaultCode={`
+          <iframe src="https://codesandbox.io/embed/trusting-wu-c8514?fontsize=14&hidenavigation=1&theme=light&view=preview"
```
dangerismycat, line 116: It looks like this is all copy/paste with a different src? We should DRY this out if possible — perhaps a function accepting a `src` and returning the template literal?

docs/src/Modal.doc.js
```diff
@@ -23,85 +32,497 @@ card(
...
+          sandbox="allow-forms allow-modals allow-popups allow-presentation allow-same-origin allow-scripts"
+        ></iframe>
+        `}
+      />
+    </MainSection.Subsection>
+    <MainSection.Subsection>
+      <MainSection.Card
+        cardSize="lg"
+        type="don't"
+        description="Use Modal for content that should have a dedicated surface, like log in flows. Think about the core areas of your product that could appear in navigation. If a dedicated URL would be beneficial, use a full page instead. If the user interaction is an optional sub-task, consider using a [Sheet](/Sheet)."
```
dangerismycat, line 156: nit: `s/"log in"/"login"`

docs/src/Modal.doc.js
```diff
@@ -23,85 +32,497 @@ card(
...
+        type="don't"
+        description="Use Modal for long and complex tasks. Don’t keep the user in a Modal that takes multiple steps to exit. If multiple tasks are required, take the user to a separate page instead."
+        defaultCode={`
+
+`}
+      />
+      <MainSection.Card
+        cardSize="lg"
+        type="don't"
+        description="Avoid adding additional Modals to the product. While these are currently used in some Pinner surfaces for editing, consider using a full page, Sheet, Flyout or inline editing for a better user experience."
```
dangerismycat, line 172: Given our usual format, this reads as "Don't avoid adding additional modals…", which I think is the opposite of our intention. Also: is this specifically talking about Pinner product? Otherwise our style guide recommends against being that specific.

docs/src/Modal.doc.js
```diff
@@ -23,85 +32,497 @@ card(
...
+      )}
+    </React.Fragment>
+  );
+}
+`}
+      />
+    </MainSection.Subsection>
+    <MainSection.Subsection
+      title="Sizes"
+      description={`Modal has 4 size options: small (\`sm\`), medium (\`md\`), large (\`lg\`) or a number representing a custom width.
```
dangerismycat, line 522: We really have three size options plus an escape hatch, right? Assuming we'd prefer people to stick with our t-shirt sizes, we should probably present those as "The Sizes", then mention that custom sizes are possible but not recommended.

docs/src/Modal.doc.js
ayeshakmaz: Thought about that, but we may want different sizing each time. We could abstract out into something with 3 sizes eventually

docs/src/Modal.doc.js
ayeshakmaz: I think for Sheet and Modal we've had to get a bit more specific since we have specific guidelines depending on the surface. We mention it again for sizing and alignment on both as well

dangerismycat: Specificity is fine, just making sure it's intended 🙂

docs/src/Modal.doc.js
```diff
@@ -23,85 +34,492 @@ card(
...
+      description={`
+      We want to make sure Modals have a clear purpose when being read by a screen reader. \`accessibilityModalLabel\` allows us to update the spoken text for the heading prop and give it more context.`}
+    >
+      <MainSection.Card
+        cardSize="lg"
+        iframeContent={`
+        <iframe src="https://codesandbox.io/embed/trusting-wu-c8514?fontsize=14&hidenavigation=1&theme=light&view=preview"
+        style={{width: '100%', height:'500px', border:'0', borderRadius: '4px', overflow:'hidden'}}
+        title="Accessibility Labels"
+        allow="accelerometer; ambient-light-sensor; camera; encrypted-media; geolocation; gyroscope; hid; microphone; midi; payment; usb; vr; xr-spatial-tracking"
```
dangerismycat, line 190: looks like there's still an `allow` to be removed

docs/src/Modal.doc.js
```diff
@@ -23,85 +34,492 @@ card(
...
+`}
+    >
+      <MainSection.Card
+        title='role="alertdialog"'
+        cardSize="lg"
+        iframeContent={`
+    <iframe src="https://codesandbox.io/embed/trusting-wu-c8514?fontsize=14&hidenavigation=1&theme=light&view=preview"
+    style={{width: '100%', height:'400px', border:'0', borderRadius: '4px', overflow:'hidden'}}
+    title="Accessibility Role ex1"
+    allow="accelerometer; ambient-light-sensor; camera; encrypted-media; geolocation; gyroscope; hid; microphone; midi; payment; usb; vr; xr-spatial-tracking"
```
dangerismycat, line 238: another one here too

docs/src/Modal.doc.js
```diff
@@ -23,85 +34,492 @@ card(
...
+`}
+      />
+      <MainSection.Card
+        title='role="dialog" (default)'
+        cardSize="lg"
+        iframeContent={`
+    <iframe src="https://codesandbox.io/embed/fervent-rosalind-xrzm5?fontsize=14&hidenavigation=1&theme=light&view=preview"
+    style={{width: '100%', height:'450px', border:'0', borderRadius: '4px', overflow:'hidden'}}
+    title="Accessibility Role ex2"
+    allow="accelerometer; ambient-light-sensor; camera; encrypted-media; geolocation; gyroscope; hid; microphone; midi; payment; usb; vr; xr-spatial-tracking"
```
dangerismycat, line 280: and here

docs/src/Modal.doc.js
```diff
@@ -23,85 +34,492 @@ card(
         type: 'string',
         required: true,
         description:
-          'String that clients such as VoiceOver will read to describe the modal. Always localize the label.',
-        href: 'accessibility',
+          'String that clients such as VoiceOver will read to describe the modal. Always localize the label. See the [Accessibility section](#Accessibility) for more info.',
       },
       {
         name: 'align',
         type: `"center" | "left"`,
```
dangerismycat, line 41: This is now out of date with @ohariz 's latest PR, so don't forget to rebase. You may want to update the copy.

docs/src/Modal.doc.js
```diff
@@ -23,85 +34,492 @@ card(
...
+);
+
+card(
+  <MainSection name="Variants">
+    <MainSection.Subsection
+      title="Heading"
+      description={`
+      The \`heading\` will render an H1 when a string is passed in and supports multiple alignment options with the \`align\` prop.
+
+      - **Left**
```
dangerismycat, line 366: This will need to be updated after rebasing as well

dangerismycat (approved):

PR #1522, opened 2021-05-24
Text/Heading: update align values

ohariz:
Updating the values of `align` to work better with RTL. `"start"` and `"end"` replace `"left"` and `"right"` as locale-dependent values, and `"forceLeft"`/`"forceRight"` have been introduced for cases where the alignment should not depend on locale

### Links

- JIRA: https://jira.pinadmin.com/browse/BIPLTM-1103

### Checklist


packages/gestalt/src/FormErrorMessage.css.flow
```diff
@@ -0,0 +1,5 @@
+// @flow strict
```
dangerismycat, line 1: Whoops, pretty sure that was my mistake this hadn't been included previously 😅

dangerismycat (approved):
Excellent work — this is great! Really appreciate the rigor to find all uses, including Modal. Thanks for this work! 😊

PR #1525, opened 2021-05-26
Internal: src/ directories: remove test files &  only publish .js.flow

christianvuerings:
## Problem(s)

* Folks are importing files within `src/` directories which results in us not being able to change internal file names
* We publish unnecessary files (e.g. test files) which makes the npm/yarn package bigger than it needs to be

### Links

Approach was inspired by https://github.com/facebook/relay/blob/625552c31ca4f5f35df1698c39f7f1eac176d1e7/gulpfile.js#L278

### DIFF
--- a/scripts/releaseSteps.js
+++ b/scripts/releaseSteps.js
@@ -11,6 +11,9 @@ const { getOctokit, context } = require('@actions/github');
 function packageJSONPath(item) {
   return path.join(__dirname, '..', 'packages', item, 'package.json');
 }
+function srcDirectory(item) {
+  return path.join(__dirname, '..', 'packages', item, 'src');
+}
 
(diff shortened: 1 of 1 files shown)


scripts/releaseSteps.js
```diff
@@ -115,6 +121,20 @@ async function createGitHubRelease({ newVersion, releaseNotes }) {
   return { releaseId, htmlUrl, uploadUrl };
 }
 
+async function cleanSource() {
+  packages.map(async (packageName) => {
```
dangerismycat, line 125: Does this need a Promise.all to run these in parallel?

dangerismycat (approved):
I'm excited to see the reduction in output!

scripts/releaseSteps.js
christianvuerings: @dangerismycat Good spotting - this didn't need to be `async` - I made it sync instead

Issue #448, opened 2019-01-05
Button is not expanding to full width after update from 0.85 to 0.86


polluxzero:
Button wont expand to full width unless external css wrapper is applied.
The codesandbox link shows how the button is currently acting after the update. 
https://codesandbox.io/s/l2ovwwm81l
This Submit button should be full width similar size to the password field above it.

pfarejowicz:
Thanks for the report, looking at the change list, the only thing that stands out to me is #432 which would have touched the CSS for Button. No other changes between those releases touched Button

https://github.com/pinterest/gestalt/compare/d6a62c4a8d6095946f2b64bea00b1be3b4b61889...master

<img width="525" alt="screen shot 2019-01-05 at 8 52 32 pm" src="https://user-images.githubusercontent.com/7877010/50732284-b38b6100-112c-11e9-8bc7-215131a082eb.png">

In your sandbox example, it looks like the class for `width: 100%` is showing up as `undefined` in the element

@btraut that change was related to postcss-next, can you take a look?

PR #1538, opened 2021-06-07
Internal: replace deprecated `postcss-cssnext` with `postcss-preset-env`

dangerismycat:
As noted in #1524, we need to update our current dependency `postcss-cssnext`. This module has been deprecated for a few years now, and we're likely to get more Dependabot alerts in the future that will require more resolutions, since this module won't be updated.

The [module author recommends](https://moox.io/blog/deprecating-cssnext) replacing with `postcss-preset-env`. It's not _quite_ a drop-in replacement, but almost. I used a [previous attempt at this upgrade](https://github.com/pinterest/gestalt/pull/432) (and the [bug it caused](https://github.com/pinterest/gestalt/issues/448#issuecomment-452475281)) and [recommendations from the previous module's author](https://github.com/csstools/postcss-preset-env/issues/32#issuecomment-393754060) to adjust our build configuration.

CSS file sizes before/after:
Screen Shot 2021-06-07 at 2 14 33 PM
Screen Shot 2021-06-07 at 2 16 16 PM
We see no change in `gestalt-future.css`. However, we do see a 309 byte drop in the size of `gestalt.css`.

[gist with before/after of the CSS files](https://gist.github.com/dangerismycat/6890c877f01da48d167976893cc7eada)

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,7 +60,7 @@
     "netlify-cli": "^2.65.1",
     "papaparse": "^5.3.0",
     "postcss": "^8.2.10",
-    "postcss-cssnext": "^3.1.0",
+    "postcss-preset-env": "^6.5.0",
     "postcss-modules": "^3.2.2",
     "prettier": "2.1.2",
     "prop-types": "^15.7.2",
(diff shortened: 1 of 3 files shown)


christianvuerings:
@dangerismycat lgtm - could you add the following to the description of the diff:
* File sizes before and after of the outputted build CSS?
* File outputs (perhaps gists) of the contents of the CSS files?

christianvuerings (approved):
Thanks for doing this!

PR #1544, opened 2021-06-09
Internal: Add `meow` resolution to fix `trim-newlines` vulnerability

dangerismycat:
[Security alert](https://github.com/pinterest/gestalt/security/dependabot/yarn.lock/trim-newlines/open)

`trim-newlines` is used by a couple of versions of `meow` (3.7.0 + 9.0.0). Both of those versions are used by our direct dependencies (`stylelint` + `css-modules-flow-types-cli`), but sadly neither of those modules have been updated.

This PR adds a resolution for the latest version of `meow`. 🤞 this doesn't break anything, since it's a major version jump for `stylelint` and a multiple major version jump for `css-modules-flow-types-cli`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -83,6 +83,7 @@
     "cssnano": "4.1.11",
     "immer": "8.0.1",
     "lodash": "4.17.21",
+    "meow": "10.0.1",
     "multicast-dns": "7.2.3",
     "serialize-javascript": "3.1.0",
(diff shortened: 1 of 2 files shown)


dangerismycat:
Abandoning in favor of a direct resolution of `trim-newlines` itself: https://github.com/pinterest/gestalt/pull/1546

PR #1550, opened 2021-06-09
Internal: add resolution for `postcss` to fix vulnerability

dangerismycat:
[Security alert](https://github.com/pinterest/gestalt/security/dependabot/yarn.lock/postcss/open)

Screen Shot 2021-06-09 at 2 23 28 PM

`postcss` is used by 95 different modules, so a resolution is the only real option.

This PR adds a resolution for `postcss` to the latest fixed version. Unfortunately this is a major version change for several modules, so 🤞 nothing breaks.

### DIFF
--- a/package.json
+++ b/package.json
@@ -86,6 +86,7 @@
     "lodash": "4.17.21",
     "multicast-dns": "7.2.3",
     "normalize-url": "4.5.1",
+    "postcss": "8.2.10",
     "serialize-javascript": "3.1.0",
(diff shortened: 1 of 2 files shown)


dangerismycat:
Unfortunately the major version change included a significant API change, leading to the failed CI runs on this PR. Given that the module is "no longer vulnerable" (🤷) and only of `moderate severity`, I'm abandoning this fix.

PR #1558, opened 2021-06-16
Internal: Add links to docs from code

christianvuerings:
## Description

Add links from our code to our documentation. In VSCode, it means that we now have clickable links to our documentation when we hover on a component.

Screen Shot 2021-06-16 at 10 52 23 AM

### Notes

* We need to strip the URLs from the original source files, otherwise we would see the URLs twice:
Screen Shot 2021-06-16 at 11 33 53 AM

### DIFF
--- a/packages/gestalt/src/Sheet.js
+++ b/packages/gestalt/src/Sheet.js
@@ -273,11 +273,13 @@ SheetWithForwardRef.propTypes = {
   subHeading: PropTypes.node,
 };
 
-/*
+/**
+ * <AnimatedSheet> component: adds animation capabilities
+ * <AnimatedSheetWithForwardRef> component: wrapper for forwarding ref to <AnimatedSheet>, the one which gets exported
+ */
 
-<AnimatedSheet> component: adds animation capabilities
-<AnimatedSheetWithForwardRef> component: wrapper for forwarding ref to <AnimatedSheet>, the one which gets exported
-
+/**
+ * https://gestalt.pinterest.systems/Sheet
  */
 const AnimatedSheetWithForwardRef: React$AbstractComponent<
   AnimatedSheetProps,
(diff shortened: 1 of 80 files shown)


packages/gestalt/src/Sheet.js
```diff
@@ -273,11 +273,13 @@ SheetWithForwardRef.propTypes = {
...
-
-<AnimatedSheet> component: adds animation capabilities
-<AnimatedSheetWithForwardRef> component: wrapper for forwarding ref to <AnimatedSheet>, the one which gets exported
+/**
+ * <AnimatedSheet> component: adds animation capabilities
+ * <AnimatedSheetWithForwardRef> component: wrapper for forwarding ref to <AnimatedSheet>, the one which gets exported
+ */
 
+/**
+ * https://gestalt.pinterest.systems/Spinner
```
dangerismycat, line 282: This doesn't seem right

packages/gestalt/src/InternalLink.js
```diff
@@ -46,6 +46,9 @@ type Props = {|
   wrappedComponent: 'button' | 'iconButton' | 'tapArea',
 |};
 
+/**
+ * https://gestalt.pinterest.systems/Link
```
dangerismycat, line 50: This isn't quite right, as this component is used in number of components aside from Link

packages/gestalt/src/Sheet.js
christianvuerings: @dangerismycat thanks for spotting - updated

packages/gestalt/src/InternalLink.js
christianvuerings: @dangerismycat thanks for spotting - updated

dangerismycat (approved):
Thanks for this, Christian! 🙂

AlbertCarreras:
This is awesome! Thanks

@christianvuerings Can we add it to `scripts/templates/ComponentName.js` as well?

christianvuerings:
> This is awesome! Thanks
> 
> @christianvuerings Can we add it to `scripts/templates/ComponentName.js` as well?

@AlbertCarreras Added, thanks for spotting

PR #1559, opened 2021-06-16
Image: fit children

dangerismycat:
## Description

When we scale an image (`fit` of "cover" or "contain") , we use `<div role="img">` instead of ``. This causes screen readers to ignore reading any child elements — not great for accessibility.

This PR moves the child content to be a sibling of the `<div role="img">` rather than a child, un-hiding the child content from screen readers.

Screen Shot 2021-06-16 at 2 17 17 PM

Screen Shot 2021-06-16 at 2 18 51 PM

TO TEST: Navigate to https://gestalt.pinterest.systems/Image#fit. Add a Button as a child to the examples. Turn your screen reader on, then tab to the button. Note that while you can tab to the button and interact with it, the screen reader doesn't read the button text. Now repeat these steps on the preview docs for this PR. Note that now the screen reader reads the button text.

### Links

- JIRA: https://jira.pinadmin.com/browse/BUG-126875

### Checklist

- [x] Verified Accessibility: keyboard & screen reader interaction

### DIFF
--- a/docs/src/Image.doc.js
+++ b/docs/src/Image.doc.js
@@ -156,7 +156,7 @@ card(
   "
     name="Overlay"
     defaultCode={`
-<Box column={6} paddingX={2}>
+<Box height={500} paddingX={2} width={250}>
   <Image
     alt="Tropic greens: The taste of Petrol and Porcelain | Interior design, Vintage Sets and Unique Pieces agave"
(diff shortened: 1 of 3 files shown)


ayeshakmaz (approved):
LGTM

PR #1560, opened 2021-06-17
Internal: add resolution for `postcss` to fix vulnerability

dangerismycat:
[Security alert](https://github.com/pinterest/gestalt/security/dependabot/yarn.lock/postcss/open)

Screen Shot 2021-06-16 at 5 07 47 PM

Attempting to add a resolution for `postcss` to fix this security vulnerability. Last time this broke things — 🤞 this time is better.

### DIFF
--- a/package.json
+++ b/package.json
@@ -87,6 +87,7 @@
     "lodash": "4.17.21",
     "multicast-dns": "7.2.3",
     "normalize-url": "4.5.1",
+    "postcss": "8.2.10",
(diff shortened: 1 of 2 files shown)


dangerismycat:
Welp, that didn't work (again). I'm going to ignore the security alert for now since it's only moderate severity. Hopefully our dependencies that rely on vulnerable versions of `postcss` update soon 🤞

PR #1564, opened 2021-06-21
Button: rename inline prop to fullWidth + codemod

AlbertCarreras:
Currently Button (and other components) defaults to full width, using an optional boolean `inline` prop to size according to the content. This runs counter to the design guidance that inline buttons should be the default. Let's remove the existing `inline` prop and replace it with an optional boolean `fullWidth` prop.

A codemod will be necessary for the Pinboard upgrade, adding `fullWidth` where `inline` is false or undefined, and removing `inline` where it is true.

Codemod 
image
image
image
image

### Links

- [TDD](paper doc link)
- [Figma](link to figma file)

### Checklist

### DIFF
--- a/docs/src/Button.doc.js
+++ b/docs/src/Button.doc.js
@@ -105,9 +105,9 @@ card(
         name: 'fullWidth',
         type: 'boolean',
         required: false,
-        defaultValue: true,
+        defaultValue: false,
         description:
-          "Default full-width buttons expand to the full width of their container whereas buttons with 'fullWidth' = 'false'  are sized by the text within the button.",
+          'Default buttons are sized by the text within the button whereas full-width buttons expand to the full width of their container.',
         href: 'width',
       },
       {
@@ -236,7 +236,7 @@ card(
   <Example
     name="Basic Button"
     id="basic-button"
-    defaultCode={`<Button text="Medium-sized button" fullWidth={false} />`}
+    defaultCode={`<Button text="Medium-sized button" />`}
   />,
 );
 
@@ -417,10 +417,10 @@ card(
     defaultCode={`
 <React.Fragment>
   <Box padding={2}>
-    <Button text="Inline button" fullWidth={false} />
+    <Button text="Inline button" />
   </Box>
   <Box padding={2}>
-    <Button text="Default full-width button" />
+    <Button fullWidth text="Default full-width button" />
   </Box>
 </React.Fragment>`}
   />,
@@ -431,7 +431,7 @@ card(
     name="Icons"
     id="iconEnd"
     defaultCode={`
-    <Button iconEnd="download" text="Download CVS file" fullWidth={false} />
+    <Button iconEnd="download" text="Download CVS file" />
 `}
   />,
 );
@@ -445,7 +445,6 @@ function Example() {
   const [selected, setSelected] = React.useState(false);
   return (
     <Button
-      fullWidth={false}
       selected={selected}
       onClick={() => {setSelected(!selected)}}
       text={selected ? "Selected" : "Deselected"}
@@ -469,7 +468,6 @@ function ButtonPopoverExample() {
   return (
     <React.Fragment>
       <Button
-        fullWidth={false}
         onClick={() => setSelected(!selected)}
         ref={anchorRef}
         selected={selected}
@@ -505,7 +503,6 @@ function AccessibilityExample() {
     <Button
       accessibilityLabel={selected ? "Unfollow Alberto on Pinterest" : "Follow Alberto on Pinterest"}
       onClick={() => setSelected(!selected)}
-      fullWidth={false}
       selected={selected}
       text={selected ? "Unfollow" : "Follow"}        />
   );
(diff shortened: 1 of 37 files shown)


dangerismycat:
Can you fill out the PR summary so Future Us know the context around these changes?

docs/src/Button.doc.js
```diff
@@ -102,12 +102,12 @@ card(
         href: 'iconEnd',
       },
       {
-        name: 'inline',
+        name: 'fullWidth',
         type: 'boolean',
         required: false,
-        defaultValue: false,
+        defaultValue: true,
```
dangerismycat, line 108: In the long-term, I don't think we want this to default to "true". Is the plan to invert this default value and codemod existing usages at some point?

AlbertCarreras: 'm changing from false to true because `inline` is the opposite of `fullWidth`. 

If Button had inline='false' <Button/> means than now it must be fullWidth='true' <Button />
inline='false' becomes fullWidth='false'

I'm keeping the same logic to prevent major changes and points of failure.

dangerismycat: Right, but I believe the whole point of this change was to update the default behavior. Changing the prop name is more of a side effect to keep the API consistent.

In general, we try to avoid having default boolean props of "true"; it's a cleaner API to have the default behavior be "false" and for the user to pass a boolean prop to indicate the non-default behavior (e.g. how `inline` currently works).

The motivation behind this change was that the vast majority of the time we don't want full-width buttons, so having that as the default state is annoying. The default state should represent the most common usage.

This change should be safe with the proper codemod. In the end, most buttons should be something like `<Button {...otherProps} />` and only full-width buttons should be `<Button fullWidth {...otherProps} />`.

AlbertCarreras: Got it... didn;t read the ticket content. You mentioned that in there. Cool. I'll change it, it does makes sense.

AlbertCarreras: @dangerismycat Fixed. Thanks for the feedback.

AlbertCarreras:
@dangerismycat whenever you get a chance to take a look! thanks

docs/src/Toast.doc.js
```diff
@@ -332,7 +327,7 @@ card(
         src="https://i.ibb.co/Lx54BCT/stock1.jpg"
       />,
     ]}
-    button={[null, <Button key="button-key" inline text="Undo" size="lg" />]}
+    button={[null, <Button key="button-key" fullWidth={false} text="Undo" size="lg" />]}
```
dangerismycat, line 330: these can be removed

packages/gestalt-codemods/24.0.0/button-replace-inline-fullWidth.js
```diff
@@ -0,0 +1,98 @@
...
+      const newAttrs = attrs
+        .map((attr) => {
+          const attrName = attr?.name?.name;
+          const disallowedProps = Object.keys(transformMap);
+
+          if (attrName && disallowedProps.includes(attrName)) {
+            addFullWidth = false;
+
+            if (attr.value === null) {
+              return false;
```
dangerismycat, line 65: nit: the point of these `return false` statements is to indicate that the attribute should be removed as part of the `.filter(Boolean)` sweep, right? I think returning null would better communicate that, as it typically is used to indicate "this is not a value". Booleans are typically valid values. This doesn't change the behavior at all since they're both falsy, but I think it makes the code a bit more readable.

packages/gestalt-codemods/24.0.0/button-replace-inline-fullWidth.js
```diff
@@ -0,0 +1,98 @@
...
+
+      const newAppendAttr = addFullWidth ? [j.jsxAttribute(j.jsxIdentifier('fullWidth'))] : [];
+
+      node.openingElement.attributes = [...newAppendAttr, ...newAttrs];
+
+      return null;
+    })
+    .toSource();
+
+  return fileHasModifications ? transform : null;
```
dangerismycat, line 97: It's wild how many valid, completely different approaches there are to writing codemods. And I have no idea what the pros/cons are for each aside from opinions about readability. At some point this half we should really create documentation on this and work with Core Platform to decide on best practices so there isn't so much context switching from codemod to codemod.

dangerismycat (approved):
This is great! Thank you so much for taking on such a tedious change. I think it'll improve our DX considerably.

docs/src/Button.doc.js
```diff
@@ -420,10 +417,10 @@ card(
     defaultCode={`
 <React.Fragment>
   <Box padding={2}>
-    <Button text="Inline button" inline />
+    <Button text="Inline button" />
   </Box>
   <Box padding={2}>
-    <Button text="Default full-width button" />
+    <Button fullWidth text="Default full-width button" />
```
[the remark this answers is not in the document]
AlbertCarreras: DONE

docs/src/Toast.doc.js
AlbertCarreras: DONE

packages/gestalt-codemods/24.0.0/button-replace-inline-fullWidth.js
AlbertCarreras: DONE

PR #1565, opened 2021-06-21
Text: Add `lineClamp` prop

dangerismycat:
## Description

_Note: prop is undocumented for now. If we ship this prop, `truncate` will be removed and this prop will be documented._

Since we dropped support for IE11, we can now add support for [line-clamp](https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-line-clamp)! 🎉

This CSS property has [wide support](https://caniuse.com/?search=line-clamp) using the Webkit prefix, which requires a few other properties to be set as well. This PR adds support for this via a new optional `lineClamp` prop, which accepts a number (realistically this is a positive integer, but we're only typing as `number`).

For now, this also adds `word-break: break-all` when using `lineClamp`, as this means that `lineClamp={1}` will faithfully reproduce the current behavior while using `truncate`. In the future, we can revisit the necessity/desirability of that property.

_truncate_
truncate

_lineClamp_
lineClamp

_Accessibility_
`lineClamp` retains the same accessibility as `truncate`: the full text is available to the screen reader while being visually truncated. We continue to add the full text to the `title` of the div/span so it is available on hover for sighted users.

### Links

- JIRA: https://jira.pinadmin.com/browse/PDS-2816

### Checklist

- [ ] Added Unit and Flow Tests _to be added when shipped_
- [ ] Added documentation + accessibility tests _to be added when shipped_
- [x] Verified Accessibility: keyboard & screen reader interaction


docs/src/Text.doc.js
```diff
@@ -129,7 +129,7 @@ card(
     name="Colors"
     defaultCode={`
 <Flex alignItems="start" direction="column" gap={4}>
-  <Box color="darkGray">
+  <Box color="darkGray" padding={1}>
```
dangerismycat, line 132: Unrelated thing that's bugged me for a while 😅

docs/src/Text.doc.js
```diff
@@ -147,35 +147,30 @@ card(
...
-    </Text>
-  </Box>
-  <Box marginBottom={2}>
-    <Text weight="bold">truncate:</Text>
-    <Text truncate>
-      This is a long and Supercalifragilisticexpialidocious sentence.
-    </Text>
-  </Box>
-</Box>
+<Flex direction="column" gap={2} maxWidth={180}>
```
dangerismycat, line 150: No material changes, just refactoring to use Flex

docs/src/Text.doc.js
```diff
@@ -225,16 +223,10 @@ card(
...
-  <Box marginBottom={2}>
-    <Text weight="bold">Bold</Text>
-  </Box>
-  <Box marginBottom={2}>
-    <Text italic>Italic</Text>
-  </Box>
-  <Box marginBottom={2}>
-    <Text underline>Underline</Text>
-  </Box>
+<Flex direction="column" gap={2}>
```
dangerismycat, line 226: No material changes, just refactoring to use Flex

packages/gestalt/src/Text.js
```diff
@@ -13,33 +13,40 @@ const SIZE_SCALE = {
   lg: 3,
 };
 
+type Align = 'start' | 'end' | 'center' | 'justify' | 'forceLeft' | 'forceRight';
```
dangerismycat, line 16: These types are all unchanged. Extracted for use in add Flow types for the PropTypes below, removing suppressions

packages/gestalt/src/Text.js
```diff
@@ -62,23 +70,7 @@ export default function Text({
...
-    color === 'navy' && colors.navy,
-    color === 'olive' && colors.olive,
-    color === 'orange' && colors.orange,
-    color === 'orchid' && colors.orchid,
-    color === 'pine' && colors.pine,
-    color === 'purple' && colors.purple,
-    color === 'red' && colors.red,
-    color === 'watermelon' && colors.watermelon,
-    color === 'white' && colors.white,
+    color && colors[color],
```
dangerismycat, line 73: This has no change in the output, just a refactor to simplify

packages/gestalt/src/Text.js
```diff
@@ -91,26 +83,37 @@ export default function Text({
     underline && typography.underline,
     weight === 'bold' && typography.fontWeightBold,
     weight === 'normal' && typography.fontWeightNormal,
-    truncate && typography.truncate,
+    // `lineClamp` overrides `truncate`
+    truncate && !isNotNullish(lineClamp) && typography.truncate,
```
dangerismycat, line 87: It's currently possible to pass both `truncate` and `lineClamp`, though unlikely since `lineClamp` is undocumented. Regardless, `lineClamp` explicitly overrides `truncate`.

packages/gestalt/src/Text.js
```diff
@@ -91,26 +83,37 @@ export default function Text({
...
+    isNotNullish(lineClamp) && typography.lineClamp,
   );
+
   const Tag = inline ? 'span' : 'div';
 
   return (
     <Tag
       className={cs}
-      {...(truncate && typeof children === 'string' ? { title: children } : null)}
+      title={
```
dangerismycat, line 96: If `title` is undefined it drops off an isn't added to the DOM element

packages/gestalt/src/Typography.css
```diff
@@ -75,8 +75,7 @@
 
 html:not([dir="rtl"]) .alignStart,
 html[dir="rtl"] .alignEnd,
-.alignForceLeft,
-{
+.alignForceLeft {
```
dangerismycat, line 78: Weird code style issue our linter didn't like, but also wouldn't autofix

packages/gestalt/src/utils/isNotNullish.js
```diff
@@ -0,0 +1,6 @@
+// @flow strict
+
+// $FlowFixMe[unclear-type] The whole point is that we don't know the type of val
+export default function isNotNullish(val: any): boolean {
```
dangerismycat, line 4: Added a "global" util since this is useful anywhere we have props that could be defined with falsy values

packages/gestalt/src/Text.js
christianvuerings: Nit: Curious to know what the difference in output looks like when using https://github.com/reactjs/react-docgen. Ideally we would use react-docgen to automatically generate our prop table in our docs.

packages/gestalt/src/Text.js
christianvuerings: @dangerismycat it looks like this does include a change in behavior. E.g. we can now use `transparentDarkGray` as a text color which is probably not desired.

Might still be good to have an includes list: `color && ['white', ...].includes(color) && colors[color]`

packages/gestalt/src/Text.js
```diff
@@ -91,26 +83,37 @@ export default function Text({
...
+        (truncate || isNotNullish(lineClamp)) && typeof children === 'string' ? children : undefined
+      }
+      {...(lineClamp ? { style: { WebkitLineClamp: lineClamp } } : {})}
```
christianvuerings, line 99: Might be worth checking in with design. Would we want any constraints on how many lines we allow?

packages/gestalt/src/utils/isNotNullish.js
christianvuerings: I try to avoid util functions like these since it's one more level of abstraction. In this case, can we use `!= null` instead of `val !== null && val !== undefined`?

packages/gestalt/src/Text.js
dangerismycat: Good question. I don't think that's a blocker for this PR, but should probably be resolved before we ship.

packages/gestalt/src/Text.js
dangerismycat: It doesn't look like we currently use that tool…is that something you've had in mind for a future improvement?

packages/gestalt/src/Text.js
dangerismycat: Good catch! Updating

packages/gestalt/src/utils/isNotNullish.js
dangerismycat: I'm a fan of favoring explicit over implicit behavior, hence the two checks instead of using `!=`. But I'll move this function into Text until we prove a more global need

AlbertCarreras:
LGTM. @dangerismycat  I've noticed that `truncate`/`lineClamp` doesn't work with `Flex` but it does with `Box display='flex'` (same behavior on master so nothing to do with this change)

AlbertCarreras (approved):

PR #1566, opened 2021-06-21
Tabs: Ship V2 tab treatment

dangerismycat:
## Description

Back in [this PR](https://github.com/pinterest/gestalt/pull/1440) we added an experimental new treatment for the Tabs component. This PR ships that treatment, removing the old version.

As part of that removal, the `size` prop is deprecated and removed. A codemod is added to assist in this deprecation.

This updates the docs slightly, but the needed full overhaul is out of scope for this work.

### Links

- [Jira](https://jira.pinadmin.com/browse/PDS-2865)
- [Figma](https://www.figma.com/file/vlihOKdfamsqUQhd9ybkQv/Tab-component-redesign?node-id=30%3A2387)


packages/gestalt/src/Tabs.test.js
```diff
@@ -1,41 +1,45 @@
...
-        activeTabIndex={0}
-        onChange={() => {}}
-      />,
-    ).root;
-    const links = instance.findAll((element) => element.type === 'a');
-    expect(links && links[0].props['aria-selected']).toEqual(true);
-    expect(links && links[1].props['aria-selected']).toEqual(false);
-    expect(links && links[2].props['aria-selected']).toEqual(false);
-  });
+  // TODO: we no longer support this, but we should
```
AlbertCarreras, line 6: @dangerismycat we should create a JIRA task

AlbertCarreras (approved):

AlbertCarreras:
LGTM! 
image

packages/gestalt/src/Tabs.test.js
dangerismycat: Good call! https://jira.pinadmin.com/browse/PDS-2876

Issue #1541, opened 2021-06-08
<Table> SortableTableHeader not compatible with sticky column


CalvinChhour:
Platform: (desktop web)

Steps to reproduce:

See [code sandbox](https://codesandbox.io/s/example-multiple-sticky-columns-forked-3i4th?file=/example.js) 
1. Setup table with sticky columns
2. Use SortableTableHeader 

Expected Result: Should be sticky without issues
![intended](https://user-images.githubusercontent.com/6954364/121236099-d3219900-c852-11eb-92e5-1d097157c4fd.gif)

Actual Result: The header doesn't follow the sticky behavior
![unintended](https://user-images.githubusercontent.com/6954364/121236102-d452c600-c852-11eb-929c-7d4fe65b023f.gif)

Please add either a screenshot or a video to help us quickly identify the issue.

ayeshakmaz:
Thanks for reporting this @CalvinChhour. We have a ticket for this functionality in our backlog, I'll keep you posted here when it's been added :)

ayeshakmaz:
Hi @CalvinChhour, this should be working now with v27.3.0. Let us know if you have any issues, and thanks again for reporting!

CalvinChhour:
Looks like things are working on our end, thanks for the update and quick turnaround!

PR #1567, opened 2021-06-22
Table: Add sticky column support to sortable header

ayeshakmaz:
### Summary

Add support for sticky columns on sortable header cells

### Links

- [Jira](https://jira.pinadmin.com/browse/PDS-2857)
- GitHub Issue: https://github.com/pinterest/gestalt/issues/1541

### Checklist

### DIFF
--- a/docs/src/Table.doc.js
+++ b/docs/src/Table.doc.js
@@ -1269,4 +1269,79 @@ card(
   />,
 );
 
+card(
+  <Example
+    id="sortableExample"
+    name="Example: Sortable header cells with sticky column"
+    defaultCode={`
+    function SortableHeaderExample() {
+      const [sortOrder, setSortOrder] = React.useState('desc');
+      const [sortCol, setSortCol] = React.useState('name');
+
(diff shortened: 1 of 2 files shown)


docs/src/Table.doc.js
```diff
@@ -1269,4 +1269,79 @@ card(
...
+card(
+  <Example
+    id="sortableExample"
+    name="Example: Sortable header cells with sticky column"
+    defaultCode={`
+    function SortableHeaderExample() {
+      const [sortOrder, setSortOrder] = React.useState('desc');
+      const [sortCol, setSortCol] = React.useState('name');
+
+      const onSortChange = (col) => {
```
AlbertCarreras, line 1281: The example sorting doesn't work. Could we add a few more columns to trigger the scrollbar automatically?

ayeshakmaz: Made the update to add more columns and leave off the actual sorting

AlbertCarreras (approved):

PR #1571, opened 2021-06-23
Tabs, TapArea, InternalLink: Add support for aria-current

dangerismycat:
### Summary

The previous version of Tabs supported `aria-selected` to indicate the selected tab. However, [that is only correct if used with `role="tab"`](https://www.stefanjudis.com/blog/aria-selected-and-when-to-use-it/), which we removed because that does not refer to a collection of links (we really need to rename Tabs and SegmentedControl). [The proper way to indicate the current location in a collection of links is `aria-current="page"`](https://www.w3.org/WAI/tutorials/menus/structure/), so this PR adds support for that.

Note: this is a purposefully simplified implementation of `aria-current`. `aria-current` is actually an enum, not a boolean. However, for our _current_ (:pun_dog:) needs, this will suffice. If we decide to implement support for other enum values in the future, that can be a purely internal refactor with no changes needed to the external API. (This prop is purposefully undocumented on TapArea for that reason — once we figure out the proper API we can document it.)

### Checklist

### DIFF
--- a/packages/gestalt/src/InternalLink.js
+++ b/packages/gestalt/src/InternalLink.js
@@ -1,7 +1,12 @@
 // @flow strict
-import type { AbstractComponent, Node, Element } from 'react';
-
-import { forwardRef, useImperativeHandle, useRef } from 'react';
+import {
+  forwardRef,
+  type AbstractComponent,
+  type Element,
+  type Node,
+  useImperativeHandle,
+  useRef,
+} from 'react';
 import PropTypes from 'prop-types';
 import classnames from 'classnames';
 import buttonStyles from './Button.css';
@@ -17,6 +22,7 @@ import getRoundingClassName, { RoundingPropType, type Rounding } from './getRoun
 import { useOnLinkNavigation } from './contexts/OnLinkNavigation.js';
 
 type Props = {|
+  accessibilityCurrent?: boolean,
   accessibilityLabel?: string,
   children?: Node,
   colorClass?: string,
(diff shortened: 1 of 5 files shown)


AlbertCarreras:
@dangerismycat

Can you add screenshot of voice over? I tried myself in test URL but can see `current page` as in the [example below](https://www.aditus.io/aria/aria-current/)
image

Re boolean, the simplified implementation is as simple as passing `page` vs undefined and it stays closer to the actual `aria-current` attribute, I don't see much benefit from simplifying it and it masks the behavior of the prop/attribute.

dangerismycat:
Fair point on the implementation.

I'm also not seeing the attribute read by the screen reader, but I do see it in the DOM. @ayeshakmaz , any tips?

ayeshakmaz:
Looks good to me, I see the attribute read: 
CurrentPage

ayeshakmaz (approved):
Looks good to me!

AlbertCarreras:
@ayeshakmaz Are you using Chrome Version 91.0.4472.114 in MacOS Big Sur? I couldn't read it either on polaris.shopify.com so I guess some environments don't support this... or I might be doing something else wrong.

ayeshakmaz:
I'm using Chrome Version 91.0.4472.77 on Catalina. Some aria support can tend to be finicky, this may be one that has less than ideal support at times. All the more reason to update the naming of Tabs and SegmentedControl to helps with better a11y

PR #1576, opened 2021-06-24
Image: Refactor to function component

dangerismycat:
Does what it says on the tin. No material changes expected.

Also updated several internal usages that were missing `color`, a required prop. Something about the way Flow handles defaultProps vs default values meant that the missing prop wasn't detected.

Flow complains if this is removed. This is what I was referring to in the summary; Flow handles default arg values and defaultProps differently, and doesn't understand that this has a default value.

### DIFF
--- a/docs/src/Toast.doc.js
+++ b/docs/src/Toast.doc.js
@@ -327,7 +327,6 @@ card(
       <Image
         key="image-key"
         alt="Saved to home decor board"
-        color="transparent"
         naturalHeight={564}
         naturalWidth={564}
         src="https://i.ibb.co/Lx54BCT/stock1.jpg"
@@ -354,7 +353,6 @@ card(
           <Image
             key="image-key"
             alt="Saved to home decor board"
-            color="transparent"
             naturalHeight={751}
             naturalWidth={564}
             src="https://i.ibb.co/7bQQYkX/stock2.jpg"
--- a/docs/src/Toast.doc.js
+++ b/docs/src/Toast.doc.js
@@ -327,6 +327,7 @@ card(
       <Image
         key="image-key"
         alt="Saved to home decor board"
+        color="transparent"
         naturalHeight={564}
         naturalWidth={564}
         src="https://i.ibb.co/Lx54BCT/stock1.jpg"
@@ -353,6 +354,7 @@ card(
           <Image
             key="image-key"
             alt="Saved to home decor board"
+            color="transparent"
             naturalHeight={751}
             naturalWidth={564}
             src="https://i.ibb.co/7bQQYkX/stock2.jpg"
--- a/packages/gestalt/src/Checkbox.test.js
+++ b/packages/gestalt/src/Checkbox.test.js
@@ -64,7 +64,6 @@ test('Checkbox with an image', () => {
         <Box>
           <Image
             alt=""
-            color="transparent"
             naturalHeight={1}
             naturalWidth={1}
             src="https://i.ibb.co/FY2MKr5/stock6.jpg"
--- a/packages/gestalt/src/Checkbox.test.js
+++ b/packages/gestalt/src/Checkbox.test.js
@@ -64,9 +64,10 @@ test('Checkbox with an image', () => {
         <Box>
           <Image
             alt=""
-            src="https://i.ibb.co/FY2MKr5/stock6.jpg"
+            color="transparent"
             naturalHeight={1}
             naturalWidth={1}
+            src="https://i.ibb.co/FY2MKr5/stock6.jpg"
           />
         </Box>
       }
--- a/packages/gestalt/src/Image.flowtest.js
+++ b/packages/gestalt/src/Image.flowtest.js
@@ -1,9 +1,7 @@
 // @flow strict
 import Image from './Image.js';
 
-const Valid = (
-  <Image alt="foo" color="transparent" naturalHeight={50} naturalWidth={50} src="foo.png" />
-);
+const Valid = <Image alt="foo" naturalHeight={50} naturalWidth={50} src="foo.png" />;
 
 // $FlowExpectedError[prop-missing]
 const MissingProp = <Image />;
--- a/packages/gestalt/src/Image.flowtest.js
+++ b/packages/gestalt/src/Image.flowtest.js
@@ -1,7 +1,9 @@
 // @flow strict
 import Image from './Image.js';
 
-const Valid = <Image alt="foo" naturalHeight={50} naturalWidth={50} src="foo.png" />;
+const Valid = (
+  <Image alt="foo" color="transparent" naturalHeight={50} naturalWidth={50} src="foo.png" />
+);
 
 // $FlowExpectedError[prop-missing]
 const MissingProp = <Image />;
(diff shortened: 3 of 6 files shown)


AlbertCarreras:
> Does what it says on the tin. No material changes expected.
> 
> Also updated several internal usages that were missing `color`, a required prop. Something about the way Flow handles defaultProps vs default values meant that the missing prop wasn't detected.

It wasn't be missing/detected cos' Flow detects defaultProps. 
image

docs/src/Toast.doc.js
```diff
@@ -327,6 +327,7 @@ card(
       <Image
         key="image-key"
         alt="Saved to home decor board"
+        color="transparent"
```
AlbertCarreras, line 330: If it's a default defining `color="transparent"` could be redundant

docs/src/Toast.doc.js
```diff
@@ -353,6 +354,7 @@ card(
           <Image
             key="image-key"
             alt="Saved to home decor board"
+            color="transparent"
```
AlbertCarreras, line 357: same

packages/gestalt/src/Checkbox.test.js
```diff
@@ -64,9 +64,10 @@ test('Checkbox with an image', () => {
         <Box>
           <Image
             alt=""
-            src="https://i.ibb.co/FY2MKr5/stock6.jpg"
+            color="transparent"
```
AlbertCarreras, line 67: not necessary

packages/gestalt/src/Image.flowtest.js
```diff
@@ -1,7 +1,9 @@
 // @flow strict
 import Image from './Image.js';
 
-const Valid = <Image alt="foo" naturalHeight={50} naturalWidth={50} src="foo.png" />;
+const Valid = (
+  <Image alt="foo" color="transparent" naturalHeight={50} naturalWidth={50} src="foo.png" />
```
AlbertCarreras, line 5: same, not necessary

packages/gestalt/src/Image.js
```diff
@@ -29,133 +32,109 @@ type Props = {|
...
-  |} = {
-    color: 'transparent',
-    fit: 'none',
-    importance: 'auto',
-    loading: 'auto',
-  };
+export default function Image({
+  alt,
+  children,
+  color = 'transparent',
```
AlbertCarreras, line 38: here,   `color = 'transparent'`, no need to define it unless passing other values

packages/gestalt/src/Image.js
```diff
@@ -29,133 +32,109 @@ type Props = {|
...
+  srcSet,
+}: Props): Node {
+  const isScaledImage = shouldScaleImage(fit);
 
-  componentDidMount() {
-    if (shouldScaleImage(this.props.fit)) {
-      this.loadImage();
-    }
-  }
+  const handleLoad = useCallback(() => {
```
AlbertCarreras, line 54: nit

`  const handleLoad = useCallback(() => onLoad?.(), [onLoad]);
`

packages/gestalt/src/Image.js
```diff
@@ -29,133 +32,109 @@ type Props = {|
...
+    onLoad?.();
+  }, [onLoad]);
 
-  componentDidUpdate(prevProps: Props) {
-    const { fit, src } = this.props;
-    if (shouldScaleImage(fit) && prevProps.src !== src) {
-      this.loadImage();
-    }
-  }
+  const handleError = useCallback(() => {
```
AlbertCarreras, line 58: same

packages/gestalt/src/Image.js
```diff
@@ -29,133 +32,109 @@ type Props = {|
...
-      role = 'img',
-      sizes,
-      src,
-      srcSet,
-    } = this.props;
+  useEffect(() => {
+    if (isScaledImage) {
+      loadImage();
+    }
+  }, [isScaledImage, loadImage, src]);
```
AlbertCarreras, line 75: can we add a comment for why `src` is added here?

packages/gestalt/src/Image.js
```diff
@@ -29,133 +32,109 @@ type Props = {|
...
-  loadImage() {
+  const loadImage = useCallback(() => {
     if (typeof window !== 'undefined') {
       const image = new window.Image();
-      image.onload = this.handleLoad;
-      image.onerror = this.handleError;
-      image.src = this.props.src;
+      image.onload = handleLoad;
+      image.onerror = handleError;
+      image.src = src;
```
AlbertCarreras, line 67: I think    

` if (shouldScaleImage(fit) && prevProps.src !== src) {
`
is not equivalent to passing `src` to the dep array.

Maybe we could save src in a `useRef` and check it against new values (then update useRef)

packages/gestalt/src/Image.js
AlbertCarreras, line 75: I _think_    

` if (shouldScaleImage(fit) && prevProps.src !== src) {
` line 71
is not equivalent to passing `src` to the dep array cos this `useEffect` gets triggered any time `isScaledImage` and `loadImage` change as well, independently of `src` changing.

Maybe we could save src in a `useRef` and check it against new values (then update useRef)

packages/gestalt/src/Image.test.js
```diff
@@ -4,15 +4,15 @@ import Image from './Image.js';
 
 test('Image matches snapshot', () => {
   const component = renderer.create(
-    <Image alt="foo" naturalHeight={50} naturalWidth={50} src="foo.png" />,
+    <Image alt="foo" color="transparent" naturalHeight={50} naturalWidth={50} src="foo.png" />,
```
AlbertCarreras, line 7: same as previous

packages/gestalt/src/RadioButton.test.js
```diff
@@ -57,9 +57,10 @@ test('RadioButton with image', () => {
         <Box>
           <Image
             alt=""
-            src="https://i.ibb.co/FY2MKr5/stock6.jpg"
+            color="transparent"
```
AlbertCarreras, line 60: same as previous

docs/src/Toast.doc.js
dangerismycat: Flow complains if this is removed. This is what I was referring to in the summary; Flow handles default arg values and defaultProps differently, and doesn't understand that this has a default value.

packages/gestalt/src/Image.js
dangerismycat: That actually changes the functionality, since that would now be returning the onLoad call instead of just calling it. That could change behavior if a handler function was returning something that we were previously dropping on the floor, but now we start to return.

packages/gestalt/src/Image.js
dangerismycat: I get what you're saying but I think that's overthinking it. The basic point is "run this if `fit` or `src` change"; I don't think there will be any practical difference with the added complexity you're suggesting.

AlbertCarreras:
If we are setting a default... wouldn't itbe better to set color prop from required to optional? I feel here the issue is that if required... a default does nothing... I think this could be a better fix.

packages/gestalt/src/Image.js
AlbertCarreras: yeah, reread logic and yes. `isScaledImage` is gating it.

AlbertCarreras (approved):

PR #1578, opened 2021-07-01
Modal, Sheet: deprecate ref (forwardRef removal) + codemod

AlbertCarreras:
Modal, Sheet: deprecate ref (forwardRef removal)

Deprecating refs as there are no cases for accessing the inner tags on any case and could lead to bad practices
Modal
image + gestaltExtension 
image
Sheet
image

Codemod to console.log existing refs in Moalds and Sheets. Run
```
yarn codemod --parser=flow -t=packages/gestalt-codemods/28.0.0/sheet_modal_deprecate_ref_props.js relative/path/to/your/code

```
### Links

- [TDD](link to Paper doc)
- [Figma](link to Figma file)

### Checklist

### DIFF
--- a/docs/src/Modal.doc.js
+++ b/docs/src/Modal.doc.js
@@ -73,11 +73,6 @@ card(
           'Callback fired when Modal is dismissed by clicking on the backdrop outside of the Modal (if `closeOnOutsideClick` is true).',
       },
       {
-        name: 'ref',
-        type: "React.Ref<'div'>",
-        description: 'Ref that will be set on the underlying div element',
-      },
-      {
         name: 'role',
         type: `"alertdialog" | "dialog"`,
         defaultValue: 'dialog',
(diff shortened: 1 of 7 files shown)


dangerismycat (commented):
Awesome, this is great. It looks like there was one use for Modal. What’s your plan for handling that one?

Also: since this is a breaking change, we should add a codemod for consistency even though we don't need it in Pinboard. I think it could probably just find Modal/Sheet usages with `ref` and add a message for them to be handled manually, same as we do for spread props.

AlbertCarreras:
> Awesome, this is great. It looks like there was one use for Modal. What’s your plan for handling that one?
> 
> Also: since this is a breaking change, we should add a codemod for consistency even though we don't need it in Pinboard. I think it could probably just find Modal/Sheet usages with `ref` and add a message for them to be handled manually, same as we do for spread props.

Yeah, should be straighforward

dangerismycat (approved):

PR #1582, opened 2021-07-08
Tabs: Add `bgColor` prop, tweak padding/states

dangerismycat:
### Summary

The current Tabs UI doesn't look great on a colored background:
124195629-ac383a80-da7f-11eb-9724-420651948e48

This PR attempts to address that by making a few changes:
- Adds a new optional `bgColor` prop of `'default' | 'transparent'`. The `'transparent'` variant removes the background and uses RGBA-based colors for the hover/pressed/focused states, so the effect is additive on top of the surrounding background color
- Increases padding around the tab text
- Adjusts the "selected tab" logic to disable the underlying TapArea and the hover/pressed states (disabling prevents keyboard and mouse focus, so no need to address that state)

### Links

- [Jira](https://jira.pinadmin.com/browse/PDS-2922)

### Checklist

### DIFF
--- a/docs/src/Tabs.doc.js
+++ b/docs/src/Tabs.doc.js
@@ -12,6 +12,18 @@ card(
   <PageHeader
     name="Tabs"
     description={`Tabs may be used navigate between multiple URLs. Tabs are intended as page-level navigation - if you're looking at just switching panels please use [SegmentedControl](/SegmentedControl).`}
+    defaultCode={`
+      <Tabs
+        activeTabIndex={0}
+        onChange={() => {}}
+        tabs={[
+          { href: "#", text: "Explore", indicator: "dot" },
+          { href: "#", text: "Shop" },
+          { href: "#", text: "Profiles" },
+        ]}
+        wrap
+      />
+    `}
   />,
 );
 
@@ -25,9 +37,10 @@ card(
         required: true,
       },
       {
-        name: 'tabs',
-        type: `Array<{| href: string, text: React.Node, id?: string, indicator?: 'dot' | number, ref?: {| current: ?HTMLElement |} |}>`,
-        required: true,
+        name: 'bgColor',
+        type: `"default" | "transparent"`,
+        defaultValue: 'default',
+        description: `If Tabs is displayed in a container with a colored background, use this prop to remove the white tab background. See the [background color](#Background-color) example to learn more.`,
       },
       {
         name: 'onChange',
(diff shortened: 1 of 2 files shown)


docs/src/Tabs.doc.js
```diff
@@ -12,6 +12,24 @@ card(
   <PageHeader
     name="Tabs"
     description={`Tabs may be used navigate between multiple URLs. Tabs are intended as page-level navigation - if you're looking at just switching panels please use [SegmentedControl](/SegmentedControl).`}
+    defaultCode={`
```
ayeshakmaz, line 15: 🙌

docs/src/Tabs.doc.js
```diff
@@ -61,8 +85,6 @@ function TabExample() {
   const tabs = [
     { href: "https://pinterest.com", text: "Boards for You", indicator: "dot" },
     { href: "https://pinterest.com", text: "Pins for You" },
-    { href: "https://pinterest.com", text: "1" },
```
ayeshakmaz, line 64: 🙌

packages/gestalt/src/Tabs.js
```diff
@@ -117,16 +136,24 @@ export const TabWithForwardRef: AbstractComponent<TabProps, HTMLElement> = forwa
...
         role="link"
         rounding={TAB_ROUNDING}
-        tapStyle="compress"
+        tapStyle={isActive ? 'none' : 'compress'}
       >
         <Flex alignItems="center" direction="column">
           <Box
-            color={color}
+            // $FlowExpectedError[incompatible-type] Flow doesn't understand the non-RGBA colors are valid for Box
+            color={color.startsWith('rgba') ? undefined : color}
```
ayeshakmaz, line 149: Are the colors allowed in Box exported? Just thinking it might be easier to make the logic read more like "If it's a color Box accepts, use it, otherwise set it as a backgroundColor" instead of replying on it starting with "rgba"

ayeshakmaz (approved):
One question about the logic for color, but more of a nit than anything

packages/gestalt/src/Tabs.js
dangerismycat: They aren't currently, but that's a good idea!

dangerismycat: Dang, no dice. ☹️ Flow still complains.

Issue #1556, opened 2021-06-15
Button: Disabled buttons aren't styled as disabled on Safari / iOS


jackhsu978:
Platform: (desktop web | mobile web)

Steps to reproduce:

1. Open https://gestalt.netlify.app/Button#type-roles in Safari
2. Click on "Disable buttons"

Expected Result: Text colors of all buttons change.
![Screen Shot 2021-06-15 at 11 13 09 AM](https://user-images.githubusercontent.com/5341184/122102900-cb21a600-cdca-11eb-87b7-6fa352d2c6c3.png)

Actual Result: Only the last one (the link) has text color changed. Text colors for the buttons don't change.
![Screen Shot 2021-06-15 at 10 57 27 AM](https://user-images.githubusercontent.com/5341184/122102911-d07ef080-cdca-11eb-87ab-634c17ce5f30.png)

I found that versions prior to v5.10.0 work. The PR in v5.10.0 https://github.com/pinterest/gestalt/pull/951 introduced this bug. It looks like iOS/Safari has a UI bug where a button with `pointer-events: none` doesn't get repainted after its color CSS attribute has been updated.

NOTE: this is also [logged in JIRA](https://jira.pinadmin.com/browse/BUG-127975).

dangerismycat:
This was fixed by #1584.

PR #1584, opened 2021-07-13
Button: Add key to fix Safari bug

dangerismycat:
### Summary

As noted in #1556, there's weird behavior on Safari browsers (mobile and desktop) relating to text color and changing disabled state. This seems to be an issue specific to Safari itself and how it handles `pointer-events: none` vis-a-vis repaints.

This PR implements the simplest fix for this issue: adding a key to the underlying DOM element based on the `disabled` prop, forcing a re-render whenever that prop changes.

To review:
- Use Safari 😛 
- Paste this code (or something similar) in one of the Button examples:
```js
function Demo() {
  const [disabled, setDisabled] = React.useState(true);

  React.useEffect(() => setTimeout(() => setDisabled(false), 1000), []);

  return <Button color="red" text="Button" disabled={disabled} />;
}
```
- The Button should flip from disabled -> enabled after a second, and the text color should be correct. (Try this on the prod docs to see the bug in action)

### Links

- [Jira](https://jira.pinadmin.com/browse/BUG-127975)

### DIFF
--- a/packages/gestalt/src/Button.js
+++ b/packages/gestalt/src/Button.js
@@ -124,6 +124,10 @@ const ButtonWithForwardRef: React$AbstractComponent<unionProps, unionRefs> = for
     text,
   } = props;
 
+  // Key necessary for text color to update on Safari when disabled changes
+  // See https://github.com/pinterest/gestalt/issues/1556 for more context
+  const key = String(disabled);
+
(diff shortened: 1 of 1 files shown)


ayeshakmaz (approved):
Works for me!

PR #1585, opened 2021-07-14
Lint rules: Add `no-box-useless-props` rule

dangerismycat:
### Summary

We see a lot of unfortunate Box props usage, where props are applied that are either useless without other props, or conflict with other props. This is provides a frustrating DX.

This PR adds a lint rule to check for three categories of useless props:
- `alignContent`, `alignItems`, `direction`, `justifyContent`, or `wrap` without `display="flex"`
- `bottom`, `left`, `right`, or `top` without `position="absolute"`
- `fit` and `maxWidth` used together, since `fit` sets `maxWidth` under the hood

The rule and tests are written to be easily extensible, so we can provide further guidance with future updates.

### Links

- [Jira](https://jira.pinadmin.com/browse/PDS-2910)

### DIFF
--- a/packages/eslint-plugin-gestalt/src/__fixtures__/no-box-useless-props/invalid/absolute-bottom.js
+++ b/packages/eslint-plugin-gestalt/src/__fixtures__/no-box-useless-props/invalid/absolute-bottom.js
@@ -1,0 +1,5 @@
+import { Box } from 'gestalt';
+
+export default function TestComponent() {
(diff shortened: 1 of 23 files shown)


ayeshakmaz (approved):
Looks great! So easy to understand, only thing that might be worth adding is an invalid test where multiple flex props are specified without display set, since that's a very common real-world use, but approving either way

PR #1596, opened 2021-07-22
Docs: Fix useFocusVisible example on Accesibility page

jportella93:
I was browsing the docs and realized they were switched.

### DIFF
--- a/docs/src/useFocusVisible.doc.js
+++ b/docs/src/useFocusVisible.doc.js
@@ -33,7 +33,7 @@ function Example() {
   return (
     <Flex alignItems="start" direction="column" gap={4}>
       <Flex alignItems="center" gap={4}>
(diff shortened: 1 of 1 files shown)


ayeshakmaz:
Hi Jon, thanks so much for checking out our new Accessibility page! Admittedly, this example is a bit confusing - I'll make a new PR to make it more clear, but the current labels are actually correct.

Essentially, when you use :focus-visible, or useFocusVisible, you are saying: if a user uses a mouse or finger to touch/click the button, don't show the focus ring, but if they Tab to it with a keyboard, DO show a focus ring. So the top button is using useFocusVisible, while the bottom example is not. It's a bit counterintuitive, so I'll work on making that more clear. Thanks again for calling this out though!

PR #1599, opened 2021-07-23
Docs: Add usage guidelines for multiple components

ayeshakmaz:
### Summary

Add Usage guidelines for Button, ButtonGroup, Tooltip, Link, Callout, Upsell, and Toast

### Links

- [Jira](https://jira.pinadmin.com/browse/PDS-2927)

### DIFF
--- a/docs/src/Button.doc.js
+++ b/docs/src/Button.doc.js
@@ -233,6 +233,32 @@ card(
 );
 
 card(
+  <MainSection name="Usage guidelines">
+    <MainSection.Subsection columns={2}>
+      <MainSection.Card
+        cardSize="md"
+        type="do"
+        title="When to Use"
+        description={`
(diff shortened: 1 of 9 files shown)


docs/src/Button.doc.js
```diff
@@ -232,6 +232,32 @@ card(
   />,
 );
 
+card(
+  <MainSection name="Usage guidelines">
```
ponori, line 236: @ayeshakmaz The one nit I have that should in no way block this shipping is to add the name of the component in the title so we can rely on that as a title in our Gestalt docs Figma plugin.

ponori (approved):

docs/src/Button.doc.js
ayeshakmaz: Chatted on Slack, but we're going to wait for this change for now

PR #1606, opened 2021-08-03
Upsell: Fix Upsell.Form onSubmit API to match convention

dangerismycat:
Typically our handlers are `({ event }) => {}`. Upsell.Form's onSubmit is `(event) => {}`. This PR fixes that handler's arguments to match our convention.

### DIFF
--- a/docs/src/Upsell.doc.js
+++ b/docs/src/Upsell.doc.js
@@ -120,7 +120,7 @@ card(
       {
         name: 'onSubmit',
         type:
-          '(SyntheticMouseEvent<HTMLButtonElement> | SyntheticKeyboardEvent<HTMLButtonElement> | SyntheticMouseEvent<HTMLAnchorElement> | SyntheticKeyboardEvent<HTMLAnchorElement>) => void',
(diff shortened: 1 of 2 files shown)


docs/src/Upsell.doc.js
```diff
@@ -120,7 +120,7 @@ card(
       {
         name: 'onSubmit',
         type:
-          '(SyntheticMouseEvent<HTMLButtonElement> | SyntheticKeyboardEvent<HTMLButtonElement> | SyntheticMouseEvent<HTMLAnchorElement> | SyntheticKeyboardEvent<HTMLAnchorElement>) => void',
+          '({ event: SyntheticMouseEvent<HTMLButtonElement> | SyntheticKeyboardEvent<HTMLButtonElement> | SyntheticMouseEvent<HTMLAnchorElement> | SyntheticKeyboardEvent<HTMLAnchorElement> }) => void',
```
AlbertCarreras, line 123: exact typep?

AlbertCarreras (approved):

docs/src/Upsell.doc.js
dangerismycat: duh, good catch 😅

PR #1610, opened 2021-08-04
Video: add backgroundColor prop

AlbertCarreras:
Add backgroundColor prop to fill the video's placeholder
Previously black color by default, now we can pass an alternative color: transparent.

This provides a fix for Box/Mask: Box/Mask containing Video shows unexpected corner border when hiding overflowing content
image

Fixes bugs:
- https://jira.pinadmin.com/browse/BUG-123359

BEFORE 
image
AFTER
image

BEFORE
image
AFTER
image

### Links

- [TDD](link to Paper doc)
- [Figma](link to Figma file)

### Checklist

### DIFF
--- a/packages/gestalt/src/Video.js
+++ b/packages/gestalt/src/Video.js
@@ -19,6 +19,7 @@ type Source =
 
 type ObjectFit = 'fill' | 'contain' | 'cover' | 'none' | 'scale-down';
 type CrossOrigin = 'anonymous' | 'use-credentials';
+type BackgroundColor = 'black' | 'transparent';
 
 type Props = {|
   accessibilityHideCaptionsLabel?: string,
@@ -30,7 +31,7 @@ type Props = {|
   accessibilityPlayLabel: string,
   accessibilityUnmuteLabel: string,
   aspectRatio: number,
-  backgroundColor: 'black' | 'transparent',
+  backgroundColor: BackgroundColor,
   captions: string,
   crossOrigin?: CrossOrigin,
   children?: Node,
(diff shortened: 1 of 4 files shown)


AlbertCarreras:
From @dangerismycat comments

How does this handle instances where we would show the bar? Does the bar no longer show? If so, is there a strategy to detect when the black background should be used vs transparent?
- The bar is shown but it’s transparent. I don’t think there’s a was to control if the video size matches the aspectRatio passed to the Video cmp. If they match, and no bg fill is shown, then transparent can be used without issues. I think we could document a little bit more when the filler shows, we have to rewrite the docs so I didn’t put much thoughts on adding new examples. This bug was previously closed as “Blending corners in videos is something that browsers don’t handle well and fixing it inside the library would require hacks. If this critical for your feature you might have to build a layer on top of video or some kind of custom component. ” as per Brent’s feedback. So this is a pretty straightforward solution that comes with caveats / tradeoffs (edited) 

It’s the “after” of the bunny video that worries me. The fix in Gestalt seems fine — it’s about how that is used in Pinboard. Moving from showing black bars to not showing them (transparent bars) is a big visual change that dramatically changes how we display videos, and could have big changes in user behavior (not to mention the design change). I think there are two options:
- Get signoff from Design (not sure who, but possibly designers on several teams since this would have wide-ranging impacts) to change to transparent bars, and possibly run an experiment to quantify user changes
- Find some way to reliably determine when the bars would be shown and when they would not, and use that to determine the backgroundColor for that Video

packages/gestalt/src/Video.js
```diff
@@ -224,11 +227,13 @@ export default class Video extends PureComponent<Props, State> {
   };
 
   static defaultProps: {|
+    backgroundColor: 'black' | 'transparent',
```
dangerismycat, line 230: nit: repeating this props isn't great, would be better to define once as its own type

dangerismycat (approved):

christianvuerings:
@AlbertCarreras checking in - are we good to merge?

packages/gestalt/src/Video.js
AlbertCarreras: DONE

PR #1611, opened 2021-08-04
Text: Remove `truncate`, document `lineClamp`

dangerismycat:
We previously added an undocumented `lineClamp` prop to Text, allowing us to use `-webkit-line-clamp`. After running [an experiment in Pinboard](https://helium.pinadmin.com/web_pin_rep_line_clamp), we've determined that replacing `truncate` with `lineClamp={1}` will not have detrimental metrics effects.

This PR removes `truncate` and adds documentation for `lineClamp`. It also includes a codemod to transform `<Text truncate />` to `<Text lineClamp={1} />`.

### DIFF
--- a/docs/src/Text.doc.js
+++ b/docs/src/Text.doc.js
@@ -48,6 +48,13 @@ card(
         href: 'styles',
       },
       {
+        name: 'lineClamp',
+        type: 'number',
+        description:
+          'Visually truncate the text to the specified number of lines. This also adds the `title` attribute if `children` is a string.',
+        href: 'overflow',
+      },
+      {
         name: 'overflow',
         type: `"normal" | "breakWord" | "noWrap"`,
         defaultValue: 'breakWord',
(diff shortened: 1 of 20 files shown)


packages/gestalt-codemods/next/__testfixtures__/text-replace-truncate-lineClamp.input.js
```diff
@@ -0,0 +1,14 @@
+// @flow strict
+import { Box, Text } from 'gestalt';
```
AlbertCarreras, line 2: can we add Text ad RenamedText to check for aliases?

docs/src/Text.doc.js
```diff
@@ -47,6 +47,13 @@ card(
         defaultValue: false,
         href: 'styles',
       },
+      {
+        name: 'lineClamp',
+        type: 'number',
+        description:
+          'Visually truncate the text to the specified number of lines. This also adds the `title` attribute if `children` is a string.',
```
AlbertCarreras, line 54: This also adds the `title` attribute if `children` is a string.', This could be a little more educational, it might not be obvious what it means for some.

AlbertCarreras (approved):
LGTM. Thanks for this change.

AlbertCarreras:
Also, the codemod /next/ I think would be /30/

PR #1613, opened 2021-08-07
Eslint plugin: prefer-box-as-tag rule w/ autofix

AlbertCarreras:
### Summary
new Eslint rule:

prefer-box-as-tag: Prefer Box: prevent HTML tags supported in Box through the `as` prop

Kapture 2021-08-11 at 18 46 55

With Eslint rules, this PR initiates an AST helper library.

Context: Gestalt Tooling Ecosystem Exploration Doc

OTHER REFACTORS TO CLEAN UP / STANDARDIZE TESTS

### Links

- [TDD](link to Paper doc)
- [Figma](link to Figma file)

### Checklist

### DIFF
--- a/packages/eslint-plugin-gestalt/src/__fixtures__/prefer-box-as-tag/valid.js
+++ b/packages/eslint-plugin-gestalt/src/__fixtures__/prefer-box-as-tag/valid.js
@@ -1,0 +1,9 @@
+import { Box } from 'gestalt';
+export default function TestElement() {
+  return (
+    <Box>
+      <ul />
+      <frame />
+    </Box>
+  );
+}
--- a/packages/eslint-plugin-gestalt/src/eslintASTFixers.js
+++ b/packages/eslint-plugin-gestalt/src/eslintASTFixers.js
@@ -1,20 +1,20 @@
 // @flow strict
-import { getNamedImportsComponents } from './eslintASTHelpers.js';
+import { getNamedImportsComponents, getTextNodeFromSourceCode } from './eslintASTHelpers.js';
 
 // $FlowFixMe[unclear-type]
-type GenericType = any;
+type GenericNode = {| [string]: any |};
 
 /** =================  FIXERS =================
 Fixers are the functions executed in the fix method inside context.report
 */
 
 type InsertGestaltImportTopFileFixerType = ({|
-  context: GenericType,
-  fixer: GenericType,
-  gestaltImportNode: GenericType,
+  context: GenericNode,
+  fixer: GenericNode,
+  gestaltImportNode: GenericNode,
   newComponentName: string,
-  programNode: GenericType,
-|}) => GenericType;
+  programNode: GenericNode,
+|}) => GenericNode;
 
 /** This function updates the imports to include the new Gestalt component if needed. If there's no previous Gestalt import, it's preprended at the top of the file. It mantains aliased imports.
  */
@@ -53,12 +53,12 @@ export const updateGestaltImportFixer: InsertGestaltImportTopFileFixerType = ({
 };
 
 type RenameTagFixerType = ({|
-  context: GenericType,
-  elementNode: GenericType,
-  fixer: GenericType,
+  context: GenericNode,
+  elementNode: GenericNode,
+  fixer: GenericNode,
   newComponentName: string,
   tagName: string,
-|}) => GenericType;
+|}) => $ReadOnlyArray<GenericNode>;
 
 /** This function renames a given tag name inside a node: `tagName` is replaced with `newComponentName`
 Examples 1:
@@ -74,13 +74,59 @@ export const renameTagFixer: RenameTagFixerType = ({
   tagName,
 }) => {
   return [elementNode.openingElement, elementNode.closingElement]
-    .map((node) =>
-      node
-        ? fixer.replaceText(
+    .map((node) => {
+      // $FlowFixMe[incompatible-type] Flow is not detecting the method filter(Boolean)
+      if (!node) return false;
+
+      return fixer.replaceText(
+        node,
+        getTextNodeFromSourceCode({ context, elementNode: node }).replace(
+          tagName,
+          newComponentName,
+        ),
+      );
+    })
+    .filter(Boolean);
+};
+
+type RenameTagWithPropsFixerType = ({|
+  additionalPropsString: string,
+  context: GenericNode,
+  elementNode: GenericNode,
+  fixer: GenericNode,
+  newComponentName: string,
+  tagName: string,
+|}) => $ReadOnlyArray<GenericNode>;
+
+/** This function is a more complex version of renameTagFixer. It has the same tag replacement functionality, but it also rebuild the props in the opening tag to include a new prop. The new prop must be formatted as a string, p.e. `as="article"`
+Examples 1:
+"\<div\>\<\/div\>" if tagName="div", newComponentName="Box", and completePropsString=`as="article"` returns "\<Box as="article"\>\<\/Box\>"
+*/
+export const renameTagWithPropsFixer: RenameTagWithPropsFixerType = ({
+  additionalPropsString,
+  context,
+  elementNode,
+  fixer,
+  newComponentName,
+  tagName,
+}) => {
+  return [elementNode.openingElement, elementNode.closingElement]
+    .map((node, index) => {
+      // $FlowFixMe[incompatible-type] Flow is not detecting the method filter(Boolean)
+      if (!node) return false;
+
+      const completeOpeningNode = `<${newComponentName} ${additionalPropsString}${
+        elementNode.closingElement ? '' : ' /'
+      }>`;
+      return index === 0
+        ? fixer.replaceText(node, completeOpeningNode)
+        : fixer.replaceText(
             node,
-            context.getSourceCode().getText(node).replace(tagName, newComponentName),
-          )
-        : false,
-    )
+            getTextNodeFromSourceCode({ context, elementNode: node }).replace(
+              tagName,
+              newComponentName,
+            ),
+          );
+    })
     .filter(Boolean);
 };
--- a/packages/eslint-plugin-gestalt/src/eslintASTHelpers.js
+++ b/packages/eslint-plugin-gestalt/src/eslintASTHelpers.js
@@ -1,19 +1,31 @@
 // @flow strict
 
 // $FlowFixMe[unclear-type]
-type GenericType = any;
+type GenericNode = {| [string]: any |};
 
 /** =================  HELPERS =================
  */
 
-type GetPropertiesFromVariable = ({|
-  variableNode: GenericType,
-|}) => GenericType;
+type GetTextNodeFromSourceCodeType = ({|
+  context: GenericNode,
+  elementNode: GenericNode,
+|}) => string;
+
+/** This function returns the text from a node as it's shown in the code source.
+ */
+export const getTextNodeFromSourceCode: GetTextNodeFromSourceCodeType = ({
+  context,
+  elementNode,
+}) => context.getSourceCode().getText(elementNode);
+
+type GetPropertiesFromVariableType = ({|
+  variableNode: GenericNode,
+|}) => ?GenericNode;
 
 /** This function returns the properties  of a variable (variableNode).
 Examples: const a = { key: "value"} >> returns nodes containing information (key: "value")
 */
-export const getPropertiesFromVariable: GetPropertiesFromVariable = ({ variableNode }) =>
+export const getPropertiesFromVariable: GetPropertiesFromVariableType = ({ variableNode }) =>
   variableNode?.resolved?.defs[0]?.node?.init?.properties;
 
 type KeyValuesType = {|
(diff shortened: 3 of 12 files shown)


packages/eslint-plugin-gestalt/src/prefer-box-as-tag.js
```diff
@@ -0,0 +1,99 @@
...
+      url: 'https://gestalt.pinterest.systems/Eslint%20Plugin#gestaltprefer-box-lonely-ref',
+    },
+    fixable: 'code',
+    schema: ([]: $ReadOnlyArray<empty>),
+    messages: {
+      disallowed: `Use <Box as="{{ tagName }}"></Box>.`,
+    },
+  },
+
+  // $FlowFixMe[unclear-type]
```
christianvuerings, line 40: @AlbertCarreras it would be great to have a `ESLintRule` flow type so we
* Ensure the type is consistent between rules
* We enforce certain properties to be set (e.g. `meta.docs.description`

```
/**
 * Flowtype definitions for eslint.ts
 * Generated by Flowgen from a Typescript Definition
 * Flowgen v1.14.1
 */

// @flow

declare type ESLintRuleMetaData = {|
  docs: {|
    /**
     * provides the short description of the rule in the [rules index](https://eslint.org/docs/rules/)
     */
    description: string,

    /**
     * specifies the heading under which the rule is listed in the [rules index](https://eslint.org/docs/rules/)
     */
    category?: string,

    /**
     * is whether the `"extends": "eslint:recommended"` property in a [configuration file](https://eslint.org/docs/user-guide/configuring#extending-configuration-files) enables the rule
     */
    recommended?: boolean,

    /**
     * specifies the URL at which the full documentation can be accessed
     */
    url?: string,

    /**
     * specifies whether rules can return suggestions (defaults to false if omitted)
     */
    suggestion?: boolean,
  |},
  messages?: {|
    [messageId: string]: string,
  |},
  fixable?: 'code' | 'whitespace',
  schema?: any,
  deprecated?: boolean,
  type?: 'problem' | 'suggestion' | 'layout',
|};

declare interface ESLintRule {
  create(context: any): any;
  meta?: ESLintRuleMetaData;
}
```

packages/eslint-plugin-gestalt/src/eslintASTHelpers.js
```diff
@@ -0,0 +1,233 @@
...
+type GetNamedImportsComponentsType = ({| importNode: GenericType |}) => $ReadOnlyArray<
+  $ReadOnlyArray<string>,
+>;
+
+export const getNamedImportsComponents: GetNamedImportsComponentsType = ({ importNode }) => {
+  const namedImports = importNode.specifiers.map((node) => [node.imported.name, node?.local?.name]);
+  return namedImports;
+};
+
+/* ↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓  getHtmlTag ↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓
```
christianvuerings, line 194: @AlbertCarreras Could we use regular JSCode format instead?

```js
type IsTagType = ({| elementNode: GenericType, tagName: string |}) => boolean;

/**
 * This function checks is a given node (elementNode) contains a given tag (tagName), and returns true if so.
 * Examples:
 *   `<div />` if tagName="div" returns true
 *   `<div />` if tagName="button" returns false
 */
export const isTag: IsTagType = ({ elementNode, tagName }) => elementNode?.name?.name === tagName;
```

It actually works better with VSCode's hover capabilities:

image
image

packages/eslint-plugin-gestalt/src/eslintASTHelpers.js
```diff
@@ -0,0 +1,233 @@
...
+This function returns the tag of a given node (elementNode).
+Examples:
+<div /> >> returns "div"
+<div /> >> returns "button"
+*/
+type GetHtmlTagType = ({| elementNode: GenericType |}) => string;
```
christianvuerings, line 200: @AlbertCarreras I'm curious, what's the reason for not inlining this type if it's not used anywhere else?

packages/eslint-plugin-gestalt/src/prefer-box-as-tag.js
```diff
@@ -0,0 +1,99 @@
+// @flow strict
+import { getHtmlTag, hasImport } from './eslintASTHelpers.js';
+import { renameTagFixer, updateGestaltImportFixer } from './eslintASTFixers.js';
+/**
+ * @fileoverview Prefer Box: prevent HTML tags supported in Box through the `as` prop
+ * @author Alberto Carreras <acarreras@pinterest.com>
```
christianvuerings, line 6: Up for discussion but I think we should get rid of `author` from our ESLint rules:

* We don't do it for our other files (e.g. components)
* We can get that information by looking at the git history of a file.

packages/eslint-plugin-gestalt/src/prefer-box-as-tag.test.js
```diff
@@ -0,0 +1,55 @@
+// @flow strict
+import { RuleTester } from 'eslint';
+import { readFileSync } from 'fs';
+import path from 'path';
+import rule, { SUPPORTED_HTML_TAGS } from './prefer-box-as-tag.js';
+
+const ruleTester = new RuleTester();
```
christianvuerings, line 7: We can pass in the parserOptions to `RuleTester` directly so we don't have to repeat them separately for the invalid or valid test cases:

```
const ruleTester = new RuleTester({ parserOptions: { ecmaVersion: 2015 } });
```

See https://eslint.org/docs/developer-guide/nodejs-api#ruletester

packages/eslint-plugin-gestalt/src/prefer-box-as-tag.test.js
```diff
@@ -0,0 +1,55 @@
...
+
+ruleTester.run('prefer-box-as-tag', rule, {
+  valid: [
+    {
+      code: validCode,
+      parserOptions,
+    },
+  ],
+  invalid: [[invalidHTMLTagWithPropsInput, invalidHTMLTagWithPropsOutput]].map((code) => ({
+    code: code[0],
```
christianvuerings, line 50: @AlbertCarreras might be easier to read as

```js
invalid: [[invalidHTMLTagWithPropsInput, invalidHTMLTagWithPropsOutput]].map(([input, output]) => ({
    code: input,
    parserOptions,
    output: output,
    errors: ERROR_MESSAGES,
  })),
```

packages/eslint-plugin-gestalt/src/eslintASTHelpers.js
```diff
@@ -0,0 +1,233 @@
+// @flow strict
+
+// $FlowFixMe[unclear-type]
+type GenericType = Object;
```
christianvuerings, line 4: @AlbertCarreras `GenericType` is `Object` which is `any`. Can we avoid extracting it and use `any` everywhere? I'm also wondering if we can have types for certain items like `node`.

packages/eslint-plugin-gestalt/src/__fixtures__/prefer-box-as-tag/invalid/invalid-HTML-tag-with-props-input.js
```diff
@@ -0,0 +1,18 @@
+import { Box } from 'gestalt';
+export default function TestElement() {
+  return (
+    <Box>
+      <article></article>
```
christianvuerings, line 5: Could you add a test case where an element has attributes?

```jsx
<article data-test-id="test">
```

packages/eslint-plugin-gestalt/src/__fixtures__/prefer-box-as-tag/valid.js
```diff
@@ -0,0 +1,9 @@
+import { Box } from 'gestalt';
+export default function TestElement() {
+  return (
+    <Box>
+      <ul />
+      <frame />
```
christianvuerings, line 6: Good old times, so happy we no longer have to use `<frame />` in today's web development: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/frame

AlbertCarreras:
@christianvuerings  I didn't finish it! But taking notes from your comments! I'll amend the previous one with this comments

This was under development!

packages/eslint-plugin-gestalt/src/prefer-box-as-tag.js
AlbertCarreras: Thanks! DONE

packages/eslint-plugin-gestalt/src/eslintASTHelpers.js
AlbertCarreras: Yeah... good point. DONE

packages/eslint-plugin-gestalt/src/prefer-box-as-tag.js
AlbertCarreras: DONE

packages/eslint-plugin-gestalt/src/prefer-box-as-tag.test.js
AlbertCarreras: DONE!

packages/eslint-plugin-gestalt/src/prefer-box-as-tag.test.js
AlbertCarreras: DAH, haha DONE

packages/eslint-plugin-gestalt/src/eslintASTHelpers.js
AlbertCarreras: trying to build a library so I'm breaking down function in small functions. This one might be very irrelevant but it requires knowing that nodes have opening and closing nodes and then access .name.name from them.

packages/eslint-plugin-gestalt/src/eslintASTHelpers.js
AlbertCarreras: DONE

packages/eslint-plugin-gestalt/src/eslintASTFixers.js
```diff
@@ -74,13 +74,59 @@ export const renameTagFixer: RenameTagFixerType = ({
...
+type RenameTagWithPropsFixerType = ({|
+  additionalPropsString: string,
+  context: GenericNode,
+  elementNode: GenericNode,
+  fixer: GenericNode,
+  newComponentName: string,
+  tagName: string,
+|}) => $ReadOnlyArray<GenericNode>;
+
+/** This function is a more complex version of renameTagFixer. It has the same tag replacement functionality, but it also rebuild the props in the opening tag to include a new prop. The new prop must be formatted as a string, p.e. `as="article"`
```
dangerismycat, line 101: Does this handle a component that already contains props, and we're adding another prop to it? If so, does it make more sense to keep the props as objects and only stringify as part of the return of this function? Or is it easier to just deal with strings the whole time?

packages/eslint-plugin-gestalt/src/eslintASTHelpers.js
```diff
@@ -180,7 +218,17 @@ export const getNamedImportsComponents: GetNamedImportsComponentsType = ({ impor
   return namedImports;
 };
 
-type IsTagType = ({| elementNode: GenericType, tagName: string |}) => boolean;
+type GetHtmlTagType = ({| elementNode: GenericNode |}) => string;
+
+/** This function returns the tag of a given node (elementNode).
+Examples:
+\<div \/\> returns "div"
+<div \/\> returns "button"
```
dangerismycat, line 226: ?

packages/eslint-plugin-gestalt/src/prefer-box-as-tag.js
```diff
@@ -0,0 +1,104 @@
...
+  'section',
+  'summary',
+];
+
+const rule: ESLintRule = {
+  meta: {
+    type: 'suggestion',
+    docs: {
+      description:
+        'Prefer Box: prevent HTML tags supported in Box through the `as` prop: "article" , "aside" | "details" | "div" | "figcaption" | "figure" | "footer" | "header" | "main" | "nav" | "section" | "summary", instead',
```
dangerismycat, line 29: nit: rather than repeat the tags here, use a template literal and map the array above to the format needed here

dangerismycat (approved):

packages/eslint-plugin-gestalt/src/eslintASTFixers.js
AlbertCarreras: The API is basically working with Text
image

AlbertCarreras: There might be better solutions, I just found this one pretty straightforward

AlbertCarreras: What happened here is that I had to change both tags AND props. And it cannot be done in different fixers because fixers cannot overlap "fixing objects must not be overlapped". So I found it easier to build the opening/closing tags.

packages/eslint-plugin-gestalt/src/eslintASTHelpers.js
AlbertCarreras: DONE

packages/eslint-plugin-gestalt/src/prefer-box-as-tag.js
AlbertCarreras: good catch!

PR #1616, opened 2021-08-11
Color: Add design tokens package and color page

ayeshakmaz:
### Summary

Introducing a new package "gestalt-design-tokens" to house design tokens across multiple formats.

### Links

- [Jira](https://jira.pinadmin.com/browse/PDS-2446)
- [Figma](https://www.figma.com/file/sx8tp1BSjmS98cStHtacKo/Gestalt-x-Brand-Colors?node-id=1376%3A0)


packages/gestalt-design-tokens/package.json
```diff
@@ -0,0 +1,40 @@
+{
+  "name": "gestalt-design-tokens",
+  "version": "0.1.0",
```
christianvuerings, line 3: @ayeshakmaz this version should be the same as the `gestalt` / `gestalt-datepicker` / ... versions & auto bump when we're doing a release:

https://github.com/pinterest/gestalt/blob/c0c3fd53884c2b4e690f540b04870a4a219c73dc/scripts/releaseSteps.js#L22
https://github.com/pinterest/gestalt/blob/c0c3fd53884c2b4e690f540b04870a4a219c73dc/.github/workflows/release.yml#L86-L91

packages/gestalt-design-tokens/dist/ios/StyleDictionaryColor.m
```diff
@@ -0,0 +1,68 @@
+
+//
+// StyleDictionaryColor.m
+//
+
+// Do not edit directly
+// Generated on Tue, 10 Aug 2021 23:17:55 GMT
```
christianvuerings, line 7: @ayeshakmaz could we add a CI check to make sure these generated files are up to date on master?

Example for how we do it for CSS flow types: https://github.com/pinterest/gestalt/blob/c0c3fd53884c2b4e690f540b04870a4a219c73dc/scripts/test.sh#L29-L41

packages/gestalt-design-tokens/README.md
```diff
@@ -0,0 +1,181 @@
+# Basic Style Dictionary
```
christianvuerings, line 1: @ayeshakmaz could we update this file and make it specific to Gestalt? Perhaps we can use a link to the `style-dictionary` docs?

package.json
```diff
@@ -97,11 +97,11 @@
     "a11y:generate": "./scripts/generateAccessibilitySpecs.js",
     "a11y:validate": "./scripts/validateAccessibilitySpecs.js",
     "svgIcons:validate": "./scripts/validateIconSvgSpecs.js",
-    "build": "cd packages/gestalt && yarn build && cd ../gestalt-datepicker && yarn build && cd ../eslint-plugin-gestalt && yarn build",
+    "build": "cd packages/gestalt && yarn build && cd ../gestalt-datepicker && yarn build && cd ../eslint-plugin-gestalt && yarn build && cd ../gestalt-design-tokens && npm run build",
```
christianvuerings, line 100: @ayeshakmaz could we make sure to also use `yarn` (not `npm`) for the `gestalt-design-tokens` package?

docs/src/components/ColorPalette.js
```diff
@@ -0,0 +1,45 @@
...
+import type { Node } from 'react';
+import { Box, Text } from 'gestalt';
+
+type Props = {|
+  name: string,
+  tokenId: string,
+|};
+
+const ColorPalette = ({ name, tokenId }: Props): Node => {
+  const tokenNumbers = [0, 50, 100, 200, 300, 400, 450, 500, 600, 700, 800, 900];
```
christianvuerings, line 11: @ayeshakmaz is it possible to import the `gestalt-design-tokens` package instead and get the values that way?

packages/gestalt-design-tokens/README.md
ayeshakmaz: Totally, will do 👍

packages/gestalt-design-tokens/README.md
```diff
@@ -0,0 +1,181 @@
...
+
+This example code is bare-bones to show you what this framework can do. If you have the style-dictionary module installed globally, you can `cd` into this directory and run:
+```bash
+style-dictionary build
```
AlbertCarreras, line 5: I'd add a little more guidance on how to get it running
command to install globally >> cd <path> to get there >> style-dictionary build

docs/src/Color.doc.js
```diff
@@ -0,0 +1,61 @@
...
+    description={`
+    Our color palettes are shared between Brand and Gestalt, and represent our full range of options. For further Brand guidance, please reference the [Brand guidelines website](https://brand.pinterest.com/).
+    `}
+  />,
+);
+
+card(
+  <MainSection
+    name="Full Palettes"
+    description="In order to ensure accessible contrast for color pairings, we require a dark font color to be used for any colors 400 and below. For 500 and above, we recommend using white."
```
dangerismycat, line 23: How do we define "dark font color"?

docs/src/Color.doc.js
```diff
@@ -0,0 +1,61 @@
...
+  >
+    <MainSection.Subsection title="Colors">
+      <Flex direction="column" gap={12}>
```
dangerismycat, line 26: nit: rather than hardcode all this markup manually, a couple of helper components that you can use to map an array of `{ name, tokenId }` would be nice. This would be annoying to update if we want to, say, adjust the `gap`.

docs/src/Color.doc.js
```diff
@@ -0,0 +1,61 @@
...
+    `}
+  />,
+);
+
+card(
+  <MainSection
+    name="Full Palettes"
+    description="In order to ensure accessible contrast for color pairings, we require a dark font color to be used for any colors 400 and below. For 500 and above, we recommend using white."
+  >
+    <MainSection.Subsection title="Colors">
```
dangerismycat, line 25: I'm only seeing the first three? This doesn't seem right:
Screen Shot 2021-08-13 at 2 04 33 PM

docs/src/Color.doc.js
```diff
@@ -0,0 +1,61 @@
...
+import ColorPalette from './components/ColorPalette.js';
+
+const cards: Array<Node> = [];
+const card = (c) => cards.push(c);
+
+card(
+  <PageHeader
+    name="Color"
+    description={`
+    Our color palettes are shared between Brand and Gestalt, and represent our full range of options. For further Brand guidance, please reference the [Brand guidelines website](https://brand.pinterest.com/).
```
dangerismycat, line 15: Do we have any other design guidance we should provide here, or are we deferring to the Brand guidelines for that? Should we establish what our primary/secondary/tertiary colors are, call out Shopping blue, anything like that? Should there be any mention of dark mode here?

docs/src/components/ColorPalette.js
```diff
@@ -0,0 +1,43 @@
+// @flow strict
+import type { Node } from 'react';
+import { Box, Text } from 'gestalt';
+// $FlowFixMe[untyped-import]
```
dangerismycat, line 4: nit: this can be `$FlowExpectedError`, since I don't think we ever expect to fix this

docs/src/components/ColorPalette.js
```diff
@@ -0,0 +1,43 @@
...
+  tokenId: string,
+|};
+
+const ColorPalette = ({ name, tokenId }: Props): Node => {
+  const tokenNumbers = [0, 50, 100, 200, 300, 400, 500, 600, 700, 800, 900];
+  const colorBoxes = tokenNumbers.map((number, idx) => {
+    const textColor = number > 400 ? 'white' : 'darkGray';
+    return tokens[`color-${tokenId}-${number}`] ? (
+      <Box
+        key={`${idx}-${tokenId}`}
```
dangerismycat, line 18: nit: this should probably use `${number}-${tokenId}`, since it's typically frowned upon to use indices as keys

docs/src/components/ColorPalette.js
```diff
@@ -0,0 +1,43 @@
...
+import tokens from 'gestalt-design-tokens/dist/json/variables.json';
+
+type Props = {|
+  name: string,
+  tokenId: string,
+|};
+
+const ColorPalette = ({ name, tokenId }: Props): Node => {
+  const tokenNumbers = [0, 50, 100, 200, 300, 400, 500, 600, 700, 800, 900];
+  const colorBoxes = tokenNumbers.map((number, idx) => {
```
dangerismycat, line 14: nit: this isn't reused, so it should either be inlined into the main markup below, or pulled out to a separate child component

docs/src/components/sidebarIndex.js
```diff
@@ -31,7 +31,7 @@ const sidebarIndex: Array<sidebarIndexType> = [
   },
   {
     sectionName: 'Foundation',
-    pages: ['Heading', 'Icon', 'Text'],
+    pages: ['Color', 'Heading', 'Icon', 'Text'],
```
dangerismycat, line 34: This seems like an odd section choice to me, since Color isn't a component. Would "Guidelines" be a better fit?

packages/gestalt-design-tokens/README.md
```diff
@@ -0,0 +1,181 @@
...
+✔︎  build/ios-swift/StyleDictionary.swift
+
+ios-swift-separate-enums
+✔︎  build/ios-swift/StyleDictionaryColor.swift
+✔︎  build/ios-swift/StyleDictionarySize.swift
+```
+
+Good for you! You have now built your first style dictionary! Moving on, take a look at what we have built. This should have created a build directory and it should look like this:
+```
+├── README.md
```
dangerismycat, line 44: COOL!!!! Holy cow this is so cool. Really, really nice work figuring all this out! This is so awesome!! 🎉

packages/gestalt-design-tokens/build.js
```diff
@@ -0,0 +1,5 @@
+// @flow strict
+// $FlowFixMe[untyped-import]
```
dangerismycat, line 2: same nit here about the Flow suppression

dangerismycat (commented):
I'll defer to @christianvuerings for the final approval, but this is looking great! Excellent work!

docs/src/Color.doc.js
ayeshakmaz: This is just the initial representation of the color palette, so next step is to create a Design Tokens page that showcases our Alias tokens, which is when dark mode and primary/secondary/etc. would come in

docs/src/Color.doc.js
ayeshakmaz: I left this for now, but I can revisit when I do the design tokens page

docs/src/Color.doc.js
ayeshakmaz: This should be updated now that the package is published

christianvuerings (approved):

PR #1623, opened 2021-08-12
Lint rules: Add support for dynamic/dangerous styles in `no-box-useless-props`

dangerismycat:
This PR adds support for a few more scenarios, which should eliminate most/all suppressions in Pinboard:
- `display={condition ? 'flex' : 'block'}`
- `dangerouslySetInlineStyle={{ __style: { display: 'grid' }
}}`
- `dangerouslySetInlineStyle={{ __style: { display: condition ? 'grid' : 'block' }
}}`

### DIFF
--- a/packages/eslint-plugin-gestalt/src/__fixtures__/no-box-useless-props/valid/flex-dynamic-dangerous.js
+++ b/packages/eslint-plugin-gestalt/src/__fixtures__/no-box-useless-props/valid/flex-dynamic-dangerous.js
@@ -1,0 +1,10 @@
+import { Box } from 'gestalt';
+
+export default function TestComponent() {
+  return (
+    <Box
+      alignContent="start"
+      dangerouslySetInlineStyle={{ __style: { display: true ? 'inline-flex' : 'flow-root' } }}
+    />
+  );
+}
--- a/packages/eslint-plugin-gestalt/src/no-box-useless-props.js
+++ b/packages/eslint-plugin-gestalt/src/no-box-useless-props.js
@@ -41,8 +41,20 @@ function getAttributeValue(attributeValue): ?(string | $ReadOnlyArray<string>) {
   }
 
   const valueExpression = attributeValue.expression;
+  // ternary
   if (valueExpression.type === 'ConditionalExpression') {
     return getExpressionValues(valueExpression);
+  }
+  // variable
+  if (valueExpression.type === 'Identifier') {
+    // This could be a variable defined within the component/file,
+    // or it could be imported from elsewhere: we don't know. If needed in the future,
+    // we could check to see if the variable is defined in this file, and therefore
+    // check the value. However, as of Aug '21 there are only two instances of `display`
+    // using a variable in Pinboard, and both are passed-in props, so we can't know
+    // the values. For now we'll give variables the benefit of the doubt and treat them
+    // as 'flex' so we don't throw unnecessary errors.
+    return 'flex';
   }
   return undefined;
 }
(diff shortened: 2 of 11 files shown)


packages/eslint-plugin-gestalt/src/no-box-useless-props.js
```diff
@@ -8,16 +8,73 @@ import { type ESLintRule } from './eslintFlowTypes.js';
...
+
+  const valueExpression = attributeValue.expression;
+  if (valueExpression.type === 'ConditionalExpression') {
+    return getExpressionValues(valueExpression);
+  }
+  return undefined;
+}
+
+// $FlowExpectedError[unclear-type]
+function getDangerouslySetStyles(attributeValue): null | { [string]: Object } {
```
christianvuerings, line 51: @dangerismycat (nit) could we replace `Object` with a more specific type?

packages/eslint-plugin-gestalt/src/__fixtures__/no-box-useless-props/valid/flex-dynamic-dangerous.js
```diff
@@ -0,0 +1,10 @@
+import { Box } from 'gestalt';
+
+export default function TestComponent() {
+  return (
+    <Box
+      alignContent="start"
+      dangerouslySetInlineStyle={{ __style: { display: true ? 'inline-flex' : 'flow-root' } }}
```
christianvuerings, line 7: @dangerismycat could we also add 2 tests for both `display` and  `dangerouslySetInlineStyle .style.display` where the display style is a reference?

E.g. these should not throw an ESLint exception:

```js
const display = condition ? 'flex' : 'block';
return (
    <Box
      alignContent="start"
      display={display}
```

packages/eslint-plugin-gestalt/src/no-box-useless-props.js
dangerismycat: We could, but that goes down a bit of a rabbit hole. It'd be great if we could find exported Flow types for eslint (I can find Typescript), but I'm not sure if it's worth our time to manually convert all the TS types to Flow.

packages/eslint-plugin-gestalt/src/__fixtures__/no-box-useless-props/valid/flex-dynamic-dangerous.js
dangerismycat: Can do. I hadn't gone too far down _that_ rabbit hole, as it can be deep (a prop value that's a reference to a variable that is defined by a ternary that references two variables that…). But I can include this one!

dangerismycat: Actually, on further consideration I don't think this is worth pursuing for now. Using a regex (`/[\s\n]display=\{[a-z]+\}/`) I only turned up two examples of this being set on Box (Testable and NavigationItem/Sidebar), both of which take `display` as a prop. I don't know if determining the possible values there is even possible with an AST (I don't think it is), but I don't think it's worth spending the time on for now.

packages/eslint-plugin-gestalt/src/no-box-useless-props.js
christianvuerings: @dangerismycat understood - not a big priority, I was more wondering if we looked into it.

packages/eslint-plugin-gestalt/src/__fixtures__/no-box-useless-props/valid/flex-dynamic-dangerous.js
christianvuerings: Spoke offline with @dangerismycat - there was a misunderstanding. I just wanted to make sure that we only throw ESLint exceptions when we're absolutely certain there is an issue.

christianvuerings (approved):

PR #1624, opened 2021-08-12
 Avatar, AvatarGroup, AvatarPair: Add usage guidelines

ponori:


docs/src/AvatarGroup.doc.js
```diff
@@ -111,6 +111,29 @@ See the [Accessibility guidelines](#Accessibility) for details on proper usage.`
...
 
+card(
+  <MainSection name="Usage guidelines">
+    <MainSection.Subsection columns={2}>
+      <MainSection.Card
+        cardSize="md"
+        type="do"
+        title="When to Use"
+        description={`
+          - For the general display of groups of people, companies and/or brands.
```
ayeshakmaz, line 122: probably worth mentioning that this can also be used as a button to add more collaborators, which is unique to AvatarGroup

docs/src/AvatarPair.doc.js
```diff
@@ -28,6 +29,30 @@ card(
...
+card(
+  <MainSection name="Usage guidelines">
+    <MainSection.Subsection columns={2}>
+      <MainSection.Card
+        cardSize="md"
+        type="do"
+        title="When to Use"
+        description={`
+          - To represent a group of people, companies and/or brands in a square format.
+          - In cases where the parent element can represent either a single person/company or a group of people/companies.
```
ayeshakmaz, line 41: I'm not sure I totally understand this one.... What does parent element refer to?

docs/src/AvatarGroup.doc.js
ponori: Good call - I'll add!

docs/src/AvatarPair.doc.js
ponori: Yeah... I'm having a hard time communicating this one. Basically, there are cases where UI will represent a individual account or a group of accounts in the same location (essentially, Inbox). I'm fumbling over how to best communicate this.

ayeshakmaz: This sounds much better, love it!

ayeshakmaz (approved):
:shipit:

PR #1625, opened 2021-08-12
Sheet: new onAnimationEnd prop + fix for bug

AlbertCarreras:
### Summary

This PR adds onAnimationEnd to Sheet so events can get triggered once the component is fully rendered or removed. Examples can be placing educational focus on elements inside Sheet.

To detect which animation is (in vs out) the callback `({ animationState }) => void` has access to `animationState`

Kapture 2021-08-12 at 19 57 13

ALSO, fix for [BUG-131248 Sheet dismiss button unreachable when heading is missing
](https://jira.pinadmin.com/browse/BUG-131248)

### Links

- [TDD](link to Paper doc)
- [Figma](link to Figma file)

### Checklist

### DIFF
--- a/docs/src/Sheet.doc.js
+++ b/docs/src/Sheet.doc.js
@@ -74,6 +74,12 @@ card(
           "The text used for Sheet's heading. Be sure to localize this text. See the [heading variant](#Heading) for more info.",
       },
       {
+        name: 'onAnimationEnd',
+        type: "({ animationState: 'in' | 'out' }) => void",
+        description:
+          'Callback fired when the Sheet in/out animations end. See the [animation](#Animation) variant to learn more. ',
+      },
+      {
         name: 'onDismiss',
         type: '() => void',
         required: true,
@@ -1033,6 +1039,8 @@ function ClosingExample(props) {
       - Done red button (children)
       - Left arrow red icon button (children)
       - Close button (footer)
+
+      Sheet also provides \`onAnimationEnd\`, a callback that gets triggered at the end of each animation. The callback has access to \`animationState\` to identify the end of each 'in' and 'out' animation for cases where each event don't trigger the same response.
       `}
     >
       <MainSection.Card
--- a/docs/src/Sheet.doc.js
+++ b/docs/src/Sheet.doc.js
@@ -1040,7 +1040,7 @@ function ClosingExample(props) {
       - Left arrow red icon button (children)
       - Close button (footer)
 
-      Sheet also provides \`onAnimationEnd\`, a callback that gets triggered at the end of each animation. The callback has access to \`animationState\` to identify the end of each 'in' and 'out' animation for cases where each event don't trigger the same response.
+      Sheet also provides \`onAnimationEnd\`, a callback that gets triggered at the end of each animation. The callback has access to \`animationState\` to identify the end of each 'in' and 'out' animation for cases where the two events trigger different responses..
       `}
     >
       <MainSection.Card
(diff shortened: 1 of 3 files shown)


docs/src/Sheet.doc.js
```diff
@@ -1033,6 +1039,8 @@ function ClosingExample(props) {
       - Done red button (children)
       - Left arrow red icon button (children)
       - Close button (footer)
+
+      Sheet also provides \`onAnimationEnd\`, a callback that gets triggered at the end of each animation. The callback has access to \`animationState\` to identify the end of each 'in' and 'out' animation for cases where each event don't trigger the same response.
```
dangerismycat, line 1043: grammar nit: "where each event doesn't trigger the same response"

ayeshakmaz, line 1043: Maybe: "Sheet also provides \`onAnimationEnd\`, a callback that gets triggered at the end of each animation. The callback has access to \`animationState\` to identify the end of each 'in' and 'out' animation for cases where the two events trigger different responses."

packages/gestalt/src/Sheet.js
```diff
@@ -117,14 +121,18 @@ function Sheet(props: SheetProps): Node {
...
     onDismiss,
     size = 'sm',
     subHeading,
   } = props;
 
   const [showTopShadow, setShowTopShadow] = useState<boolean>(false);
   const [showBottomShadow, setShowBottomShadow] = useState<boolean>(false);
-  const { animationState, onAnimationEnd } = useAnimation();
+  const {
+    animationState: animationStateHook,
```
dangerismycat, line 133: nit: this name is a bit odd, since it makes it sound like this variable is a hook itself. Perhaps stick with the convention you used for the next one, `animationStateFromHook`?

dangerismycat (approved):

packages/gestalt/src/Sheet.js
```diff
@@ -142,6 +150,11 @@ function Sheet(props: SheetProps): Node {
     };
   }, [onDismiss]);
 
+  const handleOnAnimationEnd = useCallback(() => {
+    onAnimationEndFromHook?.();
+    onAnimationEnd?.({ animationState: animationStateHook === 'in' ? 'in' : 'out' });
```
ayeshakmaz, line 155: Can this not be simplified to "animationState: animationStateHook"?

packages/gestalt/src/Sheet.js
```diff
@@ -213,7 +227,13 @@ function Sheet(props: SheetProps): Node {
                 )}
                 {!heading && (
                   <Box display="flex" flex="grow" justifyContent="end" marginBottom={8}>
-                    <Box flex="none" paddingX={6} paddingY={7} position="absolute">
+                    <Box
+                      flex="none"
+                      paddingX={6}
+                      paddingY={7}
+                      position="absolute"
+                      zIndex={new FixedZIndex(1)}
```
AlbertCarreras, line 235: FIX for bug https://jira.pinadmin.com/browse/BUG-131248

docs/src/Sheet.doc.js
AlbertCarreras: DONE

packages/gestalt/src/Sheet.js
AlbertCarreras: OOH I didn't notice that subtlety, yeah 
DONE

packages/gestalt/src/Sheet.js
AlbertCarreras: @ayeshakmaz I didn't understand the question, can you rephrase it?

AlbertCarreras: OOH! `animationStateHook` returns only 'in' & 'out' but the type is `export type AnimationStateType = 'in' | 'postIn' | 'out' | 'postOut' | null;` We wanna simplify the type to just   `'in' | 'out'` and we just pass `animationStateHook` we'd have to extend its type. We can only access 'in' & 'out' at the end of the animation, but the type doesn't match with the behavior in this particular moment.

PR #1628, opened 2021-08-16
Lint rules: Add `prefer-flex` rule w/autofix

dangerismycat:
### Summary

Discoverability of Flex has been an issue since its initial release, and we have a ton of Box usages that could very easily use Flex instead. We'd like to encourage the use of Flex to increase awareness of the component, reduce Box usage, better separate concerns (layout vs other UI concerns), and allow simpler markup with `gap`.

This rule detects usages where Box has `display="flex"` and only has props that are common to both Box and Flex.

The autofix replaces the Box with Flex, adds a Flex import if needed (and removes the Box import if there are no other usages in the file), and adds `alignItems` and `justifyContent` with Box defaults if not already set (since Box and Flex use different defaults for those props). 

### Links

- [Jira](https://jira.pinadmin.com/browse/PDS-2808)


packages/eslint-plugin-gestalt/src/prefer-flex.js
```diff
@@ -0,0 +1,168 @@
...
+  'minWidth',
+  'overflow',
+  'width',
+  'wrap',
+];
+
+export const errorMessage =
+  "Please use Flex for flexbox layouts. If you are wrapping children in Boxes to set margin/padding, try using Flex's `gap` prop instead!";
+
+const rule = {
```
AlbertCarreras, line 34: can we match the patterns re Flow and metadata from the other files? Gestalt restrictions


```
import { type ESLintRule } from './eslintFlowTypes.js';

const rule: ESLintRule = {
  meta: {
    type: 'suggestion',
    docs: {
      description:
      category: 'Gestalt restrictions',
      recommended: true,
      url: 'https://gestalt.pinterest.systems/Eslint%20Plugin#gestaltprefer-flex',
    },
    fixable: 'code',
    schema: ([]: $ReadOnlyArray<empty>),
    messages: {
      preferFlex: "Please use Flex for flexbox layouts. If you are wrapping children in Boxes to set margin/padding, try using Flex's `gap` prop instead!",
    },
  },

```

packages/eslint-plugin-gestalt/src/prefer-flex.js
```diff
@@ -0,0 +1,168 @@
...
+    schema: [
+      {
+        type: 'object',
+        additionalProperties: false,
+      },
+    ],
+  },
+
+  // $FlowFixMe[unclear-type]
+  create(context: Object): Object {
```
AlbertCarreras, line 50: create(context) {

dangerismycat: Not sure why Flow doesn't care in the rules you've written recently, but I get `Cannot build a typed interface for this module.` 😕

dangerismycat: Ah nm, it was because you're declaring the ESLintRule type

packages/eslint-plugin-gestalt/src/__fixtures__/prefer-flex/invalid/multiple-box-renamed.output.js
```diff
@@ -0,0 +1,11 @@
+import { Fragment } from 'react';
+import { Box as GestaltBox, Flex } from 'gestalt';
+
+export default function TestElement() {
+  return (
+    <Fragment>
+      <GestaltBox marginBottom={3} rounding={2} />
+      <Flex alignItems="center" justifyContent="start" />
```
AlbertCarreras, line 8: You could try here renaming Flex to RenamedFlex, it should work right away

packages/eslint-plugin-gestalt/src/testHelpers.js
```diff
@@ -1,4 +1,8 @@
 // @flow strict
```
AlbertCarreras, line 1: I loved this file

packages/eslint-plugin-gestalt/src/eslintASTFixers.js
```diff
@@ -125,19 +130,19 @@ export const renameTagWithPropsFixer: RenameTagWithPropsFixerType = ({
   newComponentName,
   tagName,
 }) => {
-  return [elementNode.openingElement, elementNode.closingElement]
+  const openingElement =
+    elementNode.type === 'JSXOpeningElement' ? elementNode : elementNode.openingElement;
+  const completeOpeningNode = `<${newComponentName} ${modifiedPropsString}${
```
AlbertCarreras, line 135: const completeOpeningNode = `<${finalNewComponentName} ${modifiedPropsString}${

packages/eslint-plugin-gestalt/src/eslintASTFixers.js
```diff
@@ -125,19 +130,19 @@ export const renameTagWithPropsFixer: RenameTagWithPropsFixerType = ({
   newComponentName,
   tagName,
 }) => {
-  return [elementNode.openingElement, elementNode.closingElement]
+  const openingElement =
```
AlbertCarreras, line 133: nit: this is repeated in many places, I created `getOpeningElement`

packages/eslint-plugin-gestalt/src/prefer-flex.test.js
```diff
@@ -0,0 +1,62 @@
...
+  getPathFormatterByRuleName,
+  getRuleTester,
+  getTestTypePrepender,
+  readTestByPath,
+} from './testHelpers.js';
+
+const ruleName = 'prefer-flex';
+const ruleTester = getRuleTester();
+const pathFormatter = getPathFormatterByRuleName(ruleName);
+
```
AlbertCarreras, line 13: for my last one, I added 

`const buildInvalidTest = (name) => readTestByPath(pathFormatter(invalidPrepender(name)));`

AlbertCarreras (approved):

PR #1629, opened 2021-08-17
Eslint plugin: `prefer-box-no-classname` rule w/ autofix + merged into `prefer-box-lonely-ref`

AlbertCarreras:
### Summary
prefer-box-no-disallowed: "Prefer Box: prevent `<div>` tags that don't contain disallowed attributes: `className`, `onClick`. Use Gestalt Box, instead"

This is a list of the [eslint-plugin-jsx-a11y](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y) eslints that cover div but won't cover Box. Other props [ `role`, `onMouseOver`, `onMouseOut`, `accessKey`, `autoFocus`, `tabIndex`] are also considered ignored or disallowed to prevent losing coverage:
- aria-activedescendant-has-tabindex: AFFECTS div
- aria-props: AFFECTS div
- aria-proptypes: AFFECTS div
- aria-role: AFFECTS div
- interactive-supports-focus: AFFECTS div somewhat
- mouse-events-have-key-events: AFFECTS div
- no-access-key: AFFECTS div
- no-autofocus: AFFECTS div
- no-noninteractive-tabindex: AFFECTS div
- no-static-element-interactions: AFFECTS div but not that much disallowing onClick
- role-has-required-aria-props: AFFECTS div
- role-supports-aria-props: AFFECTS div
- tabindex-no-positive: AFFECTS div 

Good to go on Pinboard:

`TIMING=1 ./node_modules/.bin/eslint ~/code/pinboard/webapp/app 
No errors thrown. 

image

Example:
Kapture 2021-08-19 at 01 55 04

### Links

- [TDD](link to Paper doc)
- [Figma](link to Figma file)

### Checklist


packages/eslint-plugin-gestalt/src/eslintASTFixers.js
```diff
@@ -70,21 +72,31 @@ export const renameTagFixer: RenameTagFixerType = ({
   context,
   elementNode,
   fixer,
+  gestaltImportNode,
   newComponentName,
+  replaceRegexCallback = ({ input }) => input,
```
dangerismycat, line 77: This seems a bit specific to `dangerouslySetInlineStyle`. I wonder if it would make more sense to build that prop (and any others) as a string outside of this function, then use RenameTagWithPropsFixerType's `additionalPropsString` arg to build it into the fixed component?

AlbertCarreras: This is the less "invasive" approach. Adding one additional props would require removing the previous style prop.  I could also reconstruct all props and alphabetize, but that would make the PR review even more opaque to the important changes. I might change the approach when I work on other Eslints, there must be a more universal way. For now, I think I wanna keep this one.

dangerismycat: What do you think of [the approach I took with the Box->Flex rule](https://github.com/pinterest/gestalt/pull/1628/files#diff-ec763b6499c1e98321bf995aaa9c01807756a1e9301928bdc3834af524a9b4f2)?

Avoiding regex keeps the code more readable, and it seems a lot simpler to me to keep the rule-specific logic outside of these helpers. The helpers can remove a current prop or add a new one, and the logic within the rule can determine what needs to be removed and what needs to be added (and how the new prop(s) should be formatted).

AlbertCarreras: DONE

packages/eslint-plugin-gestalt/src/eslintASTHelpers.js
```diff
@@ -160,7 +181,7 @@ type GetComponentFromAttributeType = ({| nodeAttribute: GenericNode |}) => Gener
 Example:
 \<div {...props} \/\> returns div node for the spread props attribute
 */
-export const getComponenFromAttribute: GetComponentFromAttributeType = ({ nodeAttribute }) =>
+const getComponenFromAttribute: GetComponentFromAttributeType = ({ nodeAttribute }) =>
```
dangerismycat, line 184: nit: typo, `getComponentFromAttribute`

packages/eslint-plugin-gestalt/src/eslintASTHelpers.js
```diff
@@ -235,7 +254,15 @@ Example 1:
 Example 2:
 \<div \/\> >> tagName="button" returns false
 */
-export const isTag: IsTagType = ({ elementNode, tagName }) => elementNode?.name?.name === tagName;
+const isTag: IsTagType = ({ elementNode, tagName }) => elementNode?.name?.name === tagName;
+
+type HasSpreadAttributesType = ({| elementNode: GenericNode |}) => boolean;
+/** This function checks is a given node (elementNode) contains spread attributs
```
dangerismycat, line 260: nit: typo `attributes`

packages/eslint-plugin-gestalt/src/eslintASTHelpers.js
```diff
@@ -249,7 +276,78 @@ Example 1:
...
   elementNode?.attributes?.length === 1 &&
   elementNode.attributes[0]?.name?.name === attribute;
+
+type HasAttributesType = ({|
+  elementNode: GenericNode,
+  tagName: string,
+  attributes: $ReadOnlyArray<string>,
+|}) => boolean;
+
+/** This function checks is a given tag (tagName) in a node (elementNode) contains a given attribute (attribute), and returns true if so.
```
dangerismycat, line 290: nit: typo `…checks if a given…`

packages/eslint-plugin-gestalt/src/prefer-box-no-disallowed.js
```diff
@@ -0,0 +1,178 @@
...
+        'tabIndex',
+      ];
+
+      const ignoreAttributes = [
+        ...boxDisallowedAttributes,
+        ...ignoreEslintPluginJsxA11yConflictingAttributes,
+      ];
+
+      // First, return if div should stay unmodified
+      if (
```
dangerismycat, line 75: This block reads so nicely 😍

packages/eslint-plugin-gestalt/src/eslintASTFixers.js
dangerismycat: Nice! Looks great. This arg can now be removed, right?

packages/eslint-plugin-gestalt/src/eslintASTHelpers.js
AlbertCarreras: DONE

packages/eslint-plugin-gestalt/src/eslintASTHelpers.js
AlbertCarreras: DONE

packages/eslint-plugin-gestalt/src/eslintASTHelpers.js
AlbertCarreras: DONE

AlbertCarreras:
https://github.com/pinterest/gestalt/pull/1629#discussion_r692527014

Yeah, I left because I thought we could use it in the future, preoptimization... so not needed it.

dangerismycat (approved):

PR #1633, opened 2021-08-18
Lint rules: Fix name bug in no-box-useless-props

dangerismycat:
Screen Shot 2021-08-17 at 5 28 15 PM


packages/eslint-plugin-gestalt/src/no-box-useless-props.js
```diff
@@ -69,7 +69,7 @@ function getDangerouslySetStyles(attributeValue): null | { [string]: Object } {
   return styleObject.value?.properties?.reduce(
     (acc, { key, value }) => ({
       ...acc,
-      [key.name]: value,
+      [key?.name]: value,
```
christianvuerings, line 72: @dangerismycat could we add a test case which failed on the previous line?

christianvuerings (approved):

PR #1635, opened 2021-08-18
Modal: Add flow docs for props

christianvuerings:
## Summary

Add documentation for every prop in `Modal`.

Screen Shot 2021-08-18 at 6 33 55 AM

### Pros
* See documentation in your IDE
* Quicker navigation to a specific section of the docs from within IDE

### Cons
* Duplicate types: we now inline these types so we have better documentation while coding. See Next steps to show how we can mitigate this downside.

## Next steps after this PR

* Use [React Docgen](https://github.com/reactjs/react-docgen): parse the flow docs out & use those in our documentation
* Remove PropTypes: we should be able to do this as soon as we have SSR integration tests & monitor the console

modal-flow-docs

### DIFF
--- a/packages/gestalt/src/Modal.js
+++ b/packages/gestalt/src/Modal.js
@@ -15,23 +15,61 @@ import { ScrollBoundaryContainerWithForwardRef as InternalScrollBoundaryContaine
 import { ScrollBoundaryContainerProvider } from './contexts/ScrollBoundaryContainer.js';
 import modalStyles from './Modal.css';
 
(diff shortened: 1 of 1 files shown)


AlbertCarreras (approved):

christianvuerings:
Going to land as is and start working on these follow ups

dangerismycat:
Hm…ok. I'd really like to sort out the imported types issue before we implement this more widely. I don't like the idea of duplicating types  — that sounds like a recipe for bugs.

PR #1640, opened 2021-08-19
Image: Add crossorigin property

ayeshakmaz:
### Summary

Add ability to specify cross origin for Images

### Checklist

### DIFF
--- a/packages/gestalt/src/Image.js
+++ b/packages/gestalt/src/Image.js
@@ -12,6 +12,7 @@ type Props = {|
   alt: string,
   children?: Node,
   color: string,
+  crossOrigin?: 'anonymous' | 'use-credentials',
   elementTiming?: string,
   fit?: 'contain' | 'cover' | 'none',
   importance?: 'high' | 'low' | 'auto',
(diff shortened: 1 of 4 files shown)


packages/gestalt/src/Image.js
```diff
@@ -34,6 +35,7 @@ export default class Image extends PureComponent<Props> {
     alt: PropTypes.string.isRequired,
     children: PropTypes.node,
     color: PropTypes.string,
+    crossOrigin: PropTypes.string,
```
dangerismycat, line 38: This should also be an enum

packages/gestalt/src/Image.js
```diff
@@ -144,6 +147,7 @@ export default class Image extends PureComponent<Props> {
         <img
           alt={alt}
           className={styles.img}
+          crossOrigin={crossOrigin}
```
dangerismycat, line 150: From [the MDN page](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#attr-crossorigin), it looks like providing the attribute at all changes the behavior. So we should only apply the attribute if the prop is defined

ayeshakmaz: Yeah I tests this, and the property does not appear on the img when it's not defined, you can inspect the docs and see that crossorigin only gets added when the prop is defined

dangerismycat: Nice!

dangerismycat (approved):

PR #1641, opened 2021-08-19
Eslint plugin: updated `gestalt/no-dangerous-style-duplicates` w/ autofix

AlbertCarreras:
### Summary
- This PR cleans up the `no-dangerous-style-duplicates` using best practices set on previous Eslint rules (reusable helpers, etc)
- This PR adds autofix for all previous supported cases
- Some other basic cleanups

This one complements the previous one `gestalt/prefer-box-no-disallowed` .
Now we have the full funnel: `div` with `style` attribute > Box with `dangerouslySetInlineStyle` > Box with props

### Links

- [TDD](link to Paper doc)
- [Figma](link to Figma file)

### Checklist


packages/eslint-plugin-gestalt/src/helpers/eslintASTHelpers.js
```diff
@@ -290,7 +318,7 @@ type HasAttributesType = ({|
   attributes: $ReadOnlyArray<string>,
 |}) => boolean;
 
-/** This function checks if a given tag (tagName) in a node (elementNode) contains a given attribute (attribute), and returns true if so.
+/** This function checks is a given tag (tagName) in a node (elementNode) contains a given attribute (attribute), and returns true if so.
```
dangerismycat, line 321: looks like this is a regression

packages/eslint-plugin-gestalt/src/helpers/eslintASTHelpers.js
```diff
@@ -331,6 +359,25 @@ const getLocalComponentImportName: GetLocalComponentImportNameType = ({
   return (componentNameMatch && componentNameMatch[1]) ?? componentName;
 };
 
+type IsGestaltComponentType = ({|
+  elementNode: GenericNode,
+  gestaltImportNode: GenericNode,
+  componentName: string,
+|}) => boolean;
+
+/** This function ... */
```
dangerismycat, line 368: I'm not sure if I follow these "This function" comments?

packages/eslint-plugin-gestalt/src/helpers/styleValidators.js
```diff
@@ -0,0 +1,104 @@
...
+  '#e2e2e2': `color="lightWash"`,
+  '#dadada': `color="darkWash"`,
+};
+
+const borderRadiusMap = {
+  '50%': `rounding="circle"`,
+  '999px': `rounding="circle"`,
+};
+
+const borderMap = {
```
dangerismycat, line 34: I think a regex might be better for this? e.g. it looks like this would miss `lightgray 1px solid`…the order doesn't matter for the three values so I don't think we want to enumerate every possible combination

Or perhaps look for the presence of certain substrings using String.includes

packages/eslint-plugin-gestalt/src/helpers/styleValidators.js
```diff
@@ -0,0 +1,104 @@
...
+}
+
+export const validateBackgroundColor = (value: string): ?string => colorMap[value];
+
+export const validateBorder = (value: string): ?string => {
+  // If the value is a string:
+  // 1) convert everything to lowerCase (css is case-insensitive)
+  // 2) sort the values since some found uses have the wrong order
+  const cleanValue =
+    value && value.toLowerCase ? value.toLowerCase().split(' ').sort().join(' ') : value;
```
dangerismycat, line 72: nit: simplify this to
```js
const cleanValue = value?.toLowerCase?().split(' ').sort().join(' ') ?? value;
```

packages/eslint-plugin-gestalt/src/helpers/styleValidators.js
dangerismycat: Wait…it looks like you are using a regex below for this. I'm confused why this map is needed?

dangerismycat (approved):

packages/eslint-plugin-gestalt/src/helpers/eslintASTHelpers.js
AlbertCarreras: AGGG sorry, I forgot to write them!

packages/eslint-plugin-gestalt/src/helpers/styleValidators.js
AlbertCarreras: DONE

packages/eslint-plugin-gestalt/src/helpers/styleValidators.js
AlbertCarreras: This is from Jenny's. I only reorganized the file. The logic hasn't changed.

packages/eslint-plugin-gestalt/src/helpers/eslintASTHelpers.js
AlbertCarreras: DONE

packages/eslint-plugin-gestalt/src/helpers/eslintASTHelpers.js
AlbertCarreras: DONE

packages/eslint-plugin-gestalt/src/helpers/styleValidators.js
AlbertCarreras: LEAVING AS IS

PR #1643, opened 2021-08-20
Internal: fix gestalt-design-tokens variables import

christianvuerings:
### Before
`@import` in the middle of the file
image

### After
CSS gets inlined
image

### DIFF
--- a/packages/gestalt/src/Colors.css
+++ b/packages/gestalt/src/Colors.css
@@ -1,5 +1,3 @@
-@import "gestalt-design-tokens/dist/css/variables.css";
-
 :root {
   /* primary colors */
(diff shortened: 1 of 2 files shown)


ayeshakmaz (approved):
Thank you!!

christianvuerings:
@ayeshakmaz @AlbertCarreras @dangerismycat just as a fyi this increased our CSS size from 48kb to 52kb, an 8.3% increase.

We heavily optimize our class names but can't do that for for CSS variables, since the naming is shared.

dangerismycat:
Dang. Do you expect a noticeable perf hit with that increase?

PR #1644, opened 2021-08-21
Internal: Upgrade Flow to 0.158.0

dangerismycat:
Upgrading Flow to 0.158.0


AlbertCarreras (approved):

christianvuerings:
@dangerismycat we should be good to merge (after rebasing). The slowness we experience in Pinterest's codebase with flow v0.158.0 was fixed by setting `experimental.refactor=false`: https://github.com/facebook/flow/issues/8736#issuecomment-904876353

PR #1648, opened 2021-08-23
Flex: Add `flexBasis` prop to Flex.Item

dangerismycat:
### Summary

Extending the usefulness of Flex.Item by adding [`flex-basis`](https://developer.mozilla.org/en-US/docs/Web/CSS/flex-basis)

### Links

- [Jira](https://jira.pinadmin.com/browse/PDS-2458)

### Checklist

### DIFF
--- a/docs/pages/flex.js
+++ b/docs/pages/flex.js
@@ -137,6 +137,43 @@ card(
 );
 
 card(
+  <PropTable
+    name="Flex.Item"
+    id="Flex.Item"
+    Component={Flex?.Item}
+    props={[
+      {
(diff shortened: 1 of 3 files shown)


docs/src/Flex.doc.js
```diff
@@ -135,26 +135,6 @@ card(
   />,
 );
 
-card(
```
christianvuerings, line 138: @dangerismycat could you rebase? With Next.js, the docs pages live under `pages/`

ayeshakmaz (approved):
LGTM!

PR #1669, opened 2021-08-30
Docs: auto generate prop table from component

christianvuerings:
## Summary

Before: duplicated props in the component file & docs
After: generate prop table docs from component props

## Changes

- Automatically generate the prop table in our docs with `react-docgen`
- Introduce `<Page />` which now takes regular React components as `children`. Will allow us to get rid of the `cards` array in `CardPage`

## Differences found

- `heading` prop takes a `Node` and a `string` in the previous docs. However, only `Node` is specified in the `heading` component prop.

## Links

- https://github.com/reactjs/react-docgen


docs/pages/modal.js
```diff
@@ -1,217 +1,141 @@
...
-const card = (c) => cards.push(c);
-
-card(
-  <PageHeader
-    name="Modal"
-    description="Modals display content that requires user interaction. They appear on a layer above the page and therefore block the content underneath, preventing users from interacting with anything else besides the Modal.
+export default function ModalPage({ generatedDocGen }: {| generatedDocGen: DocGen |}): Node {
+  return (
+    <Page title="Modal">
+      <PageHeader
```
christianvuerings, line 13: `<Page />` now takes regular React children, so we no longer need the `cards` prop

docs/pages/modal.js
```diff
@@ -1,217 +1,141 @@
...
+style={{width: '100%', height:'500px', border:'0', borderRadius: '4px', overflow:'hidden'}}
+title="Modal Main Example"
+sandbox="allow-forms allow-modals allow-popups allow-presentation allow-same-origin allow-scripts"
+></iframe>
     `}
-  />,
-);
-
-card(
-  <PropTable
```
christianvuerings, line 28: See `<GeneratedPropTable />` for the replacement

dangerismycat (approved):

PR #1671, opened 2021-08-31
Text: use generated docs

christianvuerings:
### Summary

Docs for `<Text />` now use the generated version. See #1669 for the initial PR

### Changes

* Use `react-docgen` docs for `<Text />`
* Add `href` support (some of our docs still had this)
* Fix issue when a flow union starts with `|`
* Support booleans & numbers for default values

### Preview

https://deploy-preview-1671--gestalt.netlify.app/text

### DIFF
--- a/packages/gestalt/src/Text.js
+++ b/packages/gestalt/src/Text.js
@@ -21,16 +21,74 @@ type Overflow = 'normal' | 'breakWord' | 'noWrap';
 type Size = 'sm' | 'md' | 'lg';
 
 type Props = {|
-  align?: Align,
+  /**
+   * `"start"` and `"end"` should be used for regular alignment since they flip with locale direction. `"forceLeft"` and `"forceRight"` should only be used in special cases where locale direction should be ignored, such as tabular or numeric text.
+   *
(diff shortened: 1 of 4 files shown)


packages/gestalt/src/Text.js
```diff
@@ -21,16 +21,74 @@ type Overflow = 'normal' | 'breakWord' | 'noWrap';
...
 
 type Props = {|
-  align?: Align,
+  /**
+   * `"start"` and `"end"` should be used for regular alignment since they flip with locale direction. `"forceLeft"` and `"forceRight"` should only be used in special cases where locale direction should be ignored, such as tabular or numeric text.
+   *
+   * Default: 'start'
+   * Link: https://gestalt.pinterest.systems/text#align
+   */
+  align?: 'start' | 'end' | 'center' | 'justify' | 'forceLeft' | 'forceRight',
```
christianvuerings, line 30: @dangerismycat as soon as you've created the babel plugin to be able to use imported types, we can change this back.

AlbertCarreras (approved):
This is awesome.

PR #1678, opened 2021-09-07
Docs: Update link to accessible design deck

ayeshakmaz:
### Summary

A quick pr to fix some doc typos and add a note about Links in Tooltips.

We also got approval to make our accessible design deck public, so updating the link there


AlbertCarreras (approved):
LGTM

PR #1679, opened 2021-09-08
Design Tokens: Update format for Android

ayeshakmaz:
### Summary

Update space tokens to use `dimen` format with `dp` for the Android output, as this is what Android [expects for dimensions](https://developer.android.com/guide/topics/resources/more-resources#Dimension) instead of `px`.


AlbertCarreras:
@ayeshakmaz Could you add a little bit more context to this change? So we can follow along?

AlbertCarreras (approved):
LGTM, just add a little bit of context re why this change is needed.

PR #1680, opened 2021-09-08
Docs: use generated flow types for Popover

AlbertCarreras:
### Links

- [TDD](link to Paper doc)
- [Figma](link to Figma file)

### Checklist

### DIFF
--- a/packages/gestalt/src/Popover.js
+++ b/packages/gestalt/src/Popover.js
@@ -2,24 +2,60 @@
 import type { Node } from 'react';
 import PropTypes from 'prop-types';
 import Controller from './Controller.js';
-import { type Role } from './Contents.js';
 
 type Color = 'blue' | 'orange' | 'red' | 'white' | 'darkGray';
 type Size = 'xs' | 'sm' | 'md' | 'lg' | 'xl' | 'flexible' | number;
 type IdealDirection = 'up' | 'right' | 'down' | 'left';
+type Role = 'dialog' | 'listbox' | 'menu';
 
 type Props = {|
+  /**
+   * The reference element, typically [Button](/Button) or [IconButton](/IconButton), that Popover uses to set its position.
+   */
   anchor: ?HTMLElement,
+  /**
+   * The content shown in Popover.
+   */
   children?: Node,
+  /**
+   * The background color of Popover. See the [color and caret](#Color-and-caret) variant to learn more.
+   */
   color?: Color,
+  /**
+   * Callback fired on keyboard navigation (keydown event).
+   */
   handleKeyDown?: (event: SyntheticKeyboardEvent<HTMLElement>) => void,
+  /**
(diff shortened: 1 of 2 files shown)


packages/gestalt/src/Popover.js
```diff
@@ -9,18 +9,69 @@ type Size = 'xs' | 'sm' | 'md' | 'lg' | 'xl' | 'flexible' | number;
 type IdealDirection = 'up' | 'right' | 'down' | 'left';
 
 type Props = {|
+  /**
+   * The reference element, typically [Button](/Button) or [IconButton](/IconButton), that Popover uses to set its position.
```
christianvuerings, line 13: @AlbertCarreras could you make sure to add full URLs? These docs are also used by Flow and otherwise won't work inside of our IDE's.

https://github.com/pinterest/gestalt/blob/4698948d06bc6313084cd0a8ba6ecb1c3eff9df6/packages/gestalt/src/Modal.js#L28

So instead:

```
The reference element, typically [Button](https://gestalt.pinterest.systems/Button) or [IconButton](https://gestalt.pinterest.systems/IconButton), that Popover uses to set its position.
```

packages/gestalt/src/Popover.js
```diff
@@ -9,18 +9,69 @@ type Size = 'xs' | 'sm' | 'md' | 'lg' | 'xl' | 'flexible' | number;
...
+  /**
+   * Specifies the preferred position of Popover relative to its anchor element. See the [ideal direction](#Ideal-direction) variant to learn more.
+   */
+  idealDirection?: 'up' | 'right' | 'down' | 'left',
+  /**
+   * Callback fired when Popover requests to be closed. Must be used to control Popover’s on/off display state.
+   */
   onDismiss: () => void,
+  /**
+   * Properly positions Popover relative to its anchor element. Set to false when used within [Layer](/Layer. See the [with Layer](#With-layer) variant to learn more.
```
christianvuerings, line 44: @AlbertCarreras looks a bit broken, I think there is a square bracket missing at the end of `[Layer](/Layer`
image

packages/gestalt/src/Popover.js
dangerismycat: Is that something we can configure at a higher level? Add an assumption that all relative links are relative from a certain path?

christianvuerings: @dangerismycat not at the moment.

A potential way to fix this would be to add the base URL `https://gestalt.pinterest.systems/` for relative links when we publish. However, it breaks if we want to test out these links within VSCode while developing gestalt. So for now, I would suggest to keep adding `https://gestalt.pinterest.systems/` to all urls in these docs.

christianvuerings:
@AlbertCarreras Could you address the comments so we can merge this one?

* https://github.com/pinterest/gestalt/pull/1680#pullrequestreview-749653557
* https://github.com/pinterest/gestalt/pull/1680#pullrequestreview-749654335

PR #1682, opened 2021-09-08
Internal: add visual regression tests

christianvuerings:
### Summary

Adds visual regression tests to Gestalt. We added tests for 2 components: `Button` and `Text`, both light and dark mode tests.

#### Notes

* We always give the screenshot a specific name. That way, we can use it in our docs
* We use `display="inlineBlock"` to only capture the component (not the whole page)
* Uses a Cypress plug-in called [cypress-plugin-snapshots](https://github.com/meinaart/cypress-plugin-snapshots)

### Example to show how to update tests

https://user-images.githubusercontent.com/127199/132586626-c5bf973b-070d-47a8-b329-930124e34e9d.mp4


package.json
```diff
@@ -99,21 +100,21 @@
   "scripts": {
     "a11y:generate": "./scripts/generateAccessibilitySpecs.js",
     "a11y:validate": "./scripts/validateAccessibilitySpecs.js",
-    "svgIcons:validate": "./scripts/validateIconSvgSpecs.js",
```
christianvuerings, line 102: alphabetized these

cypress/integration/visual-test/Button-dark_spec.js
```diff
@@ -0,0 +1,11 @@
+describe('Button dark mode visual regression check', () => {
```
dangerismycat, line 1: Can we combine the dark and light mode tests into one file? I'm thinking that could reduce boilerplate slightly and keep the directory a bit cleaner. Are there compelling reasons to keep them separate?

cypress/integration/visual-test/Button_spec.js
```diff
@@ -0,0 +1,11 @@
+describe('Button visual regression check', () => {
```
dangerismycat, line 1: Given how similar these tests are, this seems like something we could script (possibly two scripts: one to generate test files for all our existing components, and another that's added to the new component scaffolding script)

docs/pages/visual-test/Button.js
```diff
@@ -0,0 +1,7 @@
+// @flow strict
+import type { Node } from 'react';
+import { Button } from 'gestalt';
+
+export default function ButtonSpec(): Node {
+  return <Button color="red" text="Save a Pin" />;
```
dangerismycat, line 6: How deep should we go with making sure we're testing every possible visual variation of a component? I would assume we'd want to test everything, but curious about your opinion

package.json
dangerismycat: thank you! 😊

dangerismycat:
This is super cool!

cypress/integration/visual-test/Button_spec.js
christianvuerings: @dangerismycat agreed - that seems like a good follow up.

docs/pages/visual-test/Button.js
christianvuerings: @dangerismycat we can go as deep as we want but I would start with adding a visual diff test for every component, only after we could add more tests.

dangerismycat (approved):
Approving as I don't think any of my questions should be blockers, and I'm interested to see this in action!

cypress/integration/visual-test/Button-dark_spec.js
christianvuerings: @dangerismycat I tried that but it doesn't work as expected - I always get back `cy.screenshot()` only works with a screenshot area with a height greater than zero.` on the second test, even when I change the assertions.

e.g.

```js
describe('Text visual regression check', () => {
  it('Compares screenshots: default', () => {
    cy.visit('/visual-test/Text');
    cy.get('[data-test-id="visual-test"]').should('contain', 'Text size small');
    cy.get('[data-test-id="visual-test"]').toMatchImageSnapshot({
      name: 'Text',
    });
  });

  it('Compares screenshots: dark mode', () => {
    cy.visit('/visual-test/Text-dark');
    cy.get('[data-test-id="visual-test"]').should('contain', 'Text size small (dark)');
    cy.get('[data-test-id="visual-test"]').toMatchImageSnapshot({
      name: 'Text-dark',
    });
  });
});
```

So going to leave this as is for now

dangerismycat: Huh, weird. Well, not a big deal 🙂

PR #1684, opened 2021-09-09
Internal: Keep URLs in flow docs

christianvuerings:
### Summary

When using the flow comments in Pinterest's codebase, I noticed all the URLs to the docs are missing. That's caused by #1561 which strips all the URLs from the `src` file, which is not what we want.

Instead, in this diff, we remove all comments (apart from [terser pure annotation comments](https://terser.org/docs/api-reference.html#annotations)) from the `dist/` directory (build files)

Screen Shot 2021-09-09 at 5 58 44 AM

### DIFF
--- a/packages/gestalt-core/build.js
+++ b/packages/gestalt-core/build.js
@@ -142,10 +142,11 @@ const plugins = (name) => [
     preferConst: true,
   }),
   babel({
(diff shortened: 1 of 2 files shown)


AlbertCarreras (approved):
There are a few warnings there

PR #1677, opened 2021-09-02
Toast: Replace `color` with `variant`

dangerismycat:
### Summary

We'd like to remove direct color specification for Toast, re-aligning to "default" and "error" variants. This PR removes the `color` prop and replaces it with `variant`. The included codemod converts `color="red"` to `variant="error"`, `color="white"` to `variant="default"`.

### Links

- [Jira](https://jira.pinadmin.com/browse/PDS-2836)

### Checklist

- [x] Added documentation + accessibility tests

PR #1686, opened 2021-09-09
Codemods: Suppress constant condition warnings on Toast color/variant codemod

dangerismycat:


christianvuerings (approved):

packages/gestalt-codemods/34.0.0/__testfixtures__/ternary-color.input.js
```diff
@@ -3,6 +3,7 @@ import { Toast } from 'gestalt';
 
 export default function TestComponent() {
   return (
+    // eslint-disable-next-line no-constant-condition
```
christianvuerings, line 6: @dangerismycat should we make these `error` so it fails the build instead of silently going in?

dangerismycat: I'm not sure I follow?

christianvuerings: @dangerismycat `no-constant-condition` is set as a warning now, so it didn't fail the CI build (which is what we want).

image

2 potential fixes:

* Set `--max-warnings` to `0` https://eslint.org/docs/user-guide/command-line-interface
* Convert the warning to an `error` in our `.eslintrc.json` file:
   ```json
   "rules": {
       "no-constant-condition": "error",
   ```

I recommend the second option since it works way better with IDE extensions

AlbertCarreras: https://github.com/pinterest/gestalt/pull/1677#discussion_r704700926

AlbertCarreras: Can we just set the true in a const in scope? (That was my initial suggestion)

dangerismycat: > Can we just set the true in a const in scope? (That was my initial suggestion)

That wouldn't fix this — we're still hardcoding something to a boolean, which is what this rule is designed to catch

PR #1687, opened 2021-09-10
Layouts: Refactor form layout to use Flex

dangerismycat:
### Summary

Does what it says on the tin

### Links

- [Jira](https://jira.pinadmin.com/browse/PDS-2513)

### DIFF
--- a/docs/pages/layouts.js
+++ b/docs/pages/layouts.js
@@ -17,105 +17,80 @@ card(
 
 card(
   <Example
-    description="Responsive &amp; RTL friendly form layout."
+    description="Responsive &amp; RTL-friendly form layout."
     name="Form Layout"
     defaultCode={`
-<Box
-  display="flex"
-  marginStart={-3}
-  marginEnd={-3}
-  marginBottom={-3}
-  marginTop={-3}
-  wrap
-  width="100%"
+<Flex
   direction="column"
+  gap={6}
   maxWidth={800}
+  width="100%"
+  wrap
 >
-  <Box flex="grow" paddingX={3} paddingY={3}>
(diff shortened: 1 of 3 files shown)


docs/pages/layouts.js
```diff
@@ -17,105 +17,80 @@ card(
...
-      </Box>
-      <Box flex="grow" paddingX={3} paddingY={3}>
-        <TextField
-          label="TextField 3"
-          id="textfield3"
-          onChange={() => {}}
-          placeholder="Placeholder"
-        />
-      </Box>
+  <Box
```
ayeshakmaz, line 41: I think the elimination of this outer box has changed the responsive behavior:

With Box:

Screen Shot 2021-09-09 at 8 50 12 PM

New Version:

Screen Shot 2021-09-09 at 8 50 19 PM

dangerismycat: That was intentional, as I was seeing the stacking (wrapping) behavior in the prod example. Do you think the two adjacent fields shouldn't stack at a certain width? I added the minWidth to prevent the fields from getting too small. Totally open to changing this though

ayeshakmaz (approved):
LGTM 👍

packages/gestalt-codemods/34.0.0/__testfixtures__/ternary-color.input.js
```diff
@@ -3,6 +3,7 @@ import { Toast } from 'gestalt';
 
 export default function TestComponent() {
   return (
+    // eslint-disable-next-line no-constant-condition
```
dangerismycat, line 6: these were unintentional — didn't branch off master, but landed that other PR first

PR #1688, opened 2021-09-10
Docs: add autogenerated docs for AvatarGroup (forwardRef), automatically get default value & add ability to link to specific prop in docs

christianvuerings:
### Summary

@AlbertCarreras noticed that `AvatarGroup` didn't export any types when using the generated docs. That was due to the fact that the component had a `forwardRef`.

# Changes
* **Use autogenerated docs for AvatarGroup** (forwardRef)
* **Automatically get default value**: `react-docgen` already knows about the value so no need to duplicate it.
* **Add ability to link to specific prop in proptable**: makes it easy to reference other props from other components

# Found issues

## `Modal` had 2 more types with default values

Since we previously were setting these manually, we didn't discover these

### Before
image

### After
image

### DIFF
--- a/docs/components/GeneratedPropTable.js
+++ b/docs/components/GeneratedPropTable.js
@@ -4,22 +4,6 @@ import type { DocGen } from './docgen.js';
 
 import PropTable from './PropTable.js';
 
-function getDefaultValue(
-  description?: string,
-): {|
-  description?: string,
-  defaultValue?: string,
-|} {
-  const input = description ?? '';
-  const match = input.match(/(?<main>Default: (?<defaultValue>.*))/);
-  const groups = match?.groups ?? {};
-
-  return {
-    description: groups.main ? input.replace(groups.main, '') : description,
-    defaultValue: groups.defaultValue?.replace(/'/g, ''),
-  };
-}
-
 function getHref(
   description?: string,
 ): {|
@@ -46,17 +30,15 @@ export default function GeneratedPropTable({
   // Using Object.keys because of https://github.com/facebook/flow/issues/2174
   const props = Object.keys(generatedDocGen.props)
     .map((key: string) => {
-      const { flowType, description, required } = generatedDocGen.props[key];
+      const { flowType, description, required, defaultValue } = generatedDocGen.props[key];
 
       // Filter out PropType only types & excluded props
       if (!flowType || excludeProps.includes(key)) {
         return null;
       }
-
-      const { description: descriptionWithoutDefaultValue = '', defaultValue } = getDefaultValue(
+      const { description: descriptionWihoutLink, href } = getHref(
         description?.replace(/https:\/\/gestalt\.pinterest\.systems/, ''),
       );
-      const { description: descriptionWihoutLink, href } = getHref(descriptionWithoutDefaultValue);
 
       return {
         name: key,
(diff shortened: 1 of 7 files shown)


docs/components/GeneratedPropTable.js
```diff
@@ -4,22 +4,6 @@ import type { DocGen } from './docgen.js';
 
 import PropTable from './PropTable.js';
 
-function getDefaultValue(
```
christianvuerings, line 7: No longer required, since we get the default value from `react-docgen`

docs/components/PropTable.js
```diff
@@ -172,18 +173,27 @@ export default function PropTable({
                     acc.push(
                       <tr key={i}>
                         <Td shrink border={!propNameHasSecondRow}>
-                          <Flex gap={2}>
-                            <Text overflow="normal" underline={!!href}>
-                              {href ? (
-                                <Link href={`#${href}`}>
+                          <Box
+                            id={`${propsId}-${name}`}
```
christianvuerings, line 177: Adds ability to link to a specific prop in each prop table

packages/gestalt/src/AvatarGroup.js
```diff
@@ -47,7 +90,7 @@ const AvatarGroupWithForwardRef: React$AbstractComponent<Props, UnionRefs> = for
     onClick,
     role,
     size = 'fit',
-  },
+  }: Props,
```
christianvuerings, line 93: @AlbertCarreras this was the "trick" to get the auto generated docs to work. Manually specifying the `Props` on the component.

packages/gestalt/src/AvatarGroup.js
```diff
@@ -7,28 +7,71 @@ import AddCollaboratorsButton from './AvatarGroupAddCollaboratorsButton.js';
...
+   */
   href?: string,
+  /**
+   * Callback fired when the component is clicked (pressed and released) with a mouse or keyboard. See the [role](https://gestalt.pinterest.systems/AvatarGroup#Role) variant to learn more and see [TapArea's `onTap`](https://gestalt.pinterest.systems/taparea#Props-onTap) for more info about `OnTapType`.
+   */
   onClick?: OnTapType,
+  /**
+   * Forward the ref to the underlying div or anchor element. See the [role](https://gestalt.pinterest.systems/AvatarGroup#Role) variant to learn more.
+   */
+  ref?: UnionRefs, // eslint-disable-line react/no-unused-prop-types
```
christianvuerings, line 64: Bit controversial, since we do have to disable a linter but...

1) We now can see `ref` in the generated props table
2) We get flow docs for the `ref` on `AvatarGroup`

packages/gestalt/src/Text.js
```diff
@@ -24,13 +24,11 @@ type Props = {|
   /**
    * `"start"` and `"end"` should be used for regular alignment since they flip with locale direction. `"forceLeft"` and `"forceRight"` should only be used in special cases where locale direction should be ignored, such as tabular or numeric text.
    *
-   * Default: 'start'
```
christianvuerings, line 27: These are now all fetched automatically with `react-docgen`

AlbertCarreras (approved):

packages/gestalt/src/AvatarGroup.js
dangerismycat: weird…this seems like something react-docgen would have figured out by now, given that forwardRefs are popular and not new

packages/gestalt/src/AvatarGroup.js
dangerismycat: I wonder if we should keep a list somewhere of these unintuitive "gotchas" we're figuring out about getting auto-generated props to work?

PR #1689, opened 2021-09-10
ScrollBoundaryContainer: use generated docs (multiple exports)

christianvuerings:
### Summary

@AlbertCarreras noticed that generating the docs for `ScrollBoundaryContainer` didn't work because of multiple exports:

Screen Shot 2021-09-10 at 6 55 03 AM

Fix is to split the components up into multiple files (also makes it more straightforward to see the dependencies of each one)

### Issues found

We previously didn't specify that `children` is a required prop.

Screen Shot 2021-09-10 at 7 08 46 AM


AlbertCarreras (approved):
Solution: splitting. ha!

PR #1690, opened 2021-09-13
Toast: Add experimental prop for dark gray color

dangerismycat:
### Summary

We're looking to move the default Toast color back to dark gray, but this time we'd like to make the change behind an experiment to gauge user behavior changes. This PR adds an undocumented prop to use dark gray instead of white for the toast background. It also adds an override (specific to the experimental color treatment) to ensure that all contained text is the correct color, regardless of any colors set on more complex `text` Node props.

### Links

- [Jira](https://jira.pinadmin.com/browse/PDS-2837)

### Checklist

### DIFF
--- a/packages/gestalt/src/Toast.css
+++ b/packages/gestalt/src/Toast.css
@@ -1,0 +1,7 @@
+.textColorOverrideWhite * {
+  color: var(--g-colorGray0) !important; /* stylelint-disable-line declaration-no-important */
+}
+
+.textColorOverrideDarkGray * {
+  color: var(--g-colorGray300) !important; /* stylelint-disable-line declaration-no-important */
+}
--- a/packages/gestalt/src/Toast.js
+++ b/packages/gestalt/src/Toast.js
@@ -33,31 +33,32 @@ export default function Toast({
   const isDarkMode = colorSchemeName === 'darkMode';
   const isErrorVariant = variant === 'error';
 
-  let containerColor = _dangerouslyUseDarkGray ? 'darkGray' : 'white';
-  let textColor = _dangerouslyUseDarkGray ? 'white' : 'darkGray';
+  let containerColor = 'white';
+  let textColor = 'darkGray';
+  let textElement = text;
 
+  if (_dangerouslyUseDarkGray) {
+    containerColor = isDarkMode ? 'white' : 'darkGray';
+    textColor = isDarkMode ? 'darkGray' : 'white';
+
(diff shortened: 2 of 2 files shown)


packages/gestalt/src/Toast.js
```diff
@@ -23,10 +27,37 @@ export default function Toast({
...
-  const textColor = isErrorVariant ? 'white' : undefined;
+
+  let containerColor = _dangerouslyUseDarkGray ? 'darkGray' : 'white';
+  let textColor = _dangerouslyUseDarkGray ? 'white' : 'darkGray';
+
+  if (isErrorVariant) {
+    // Error variant does not currently support dark mode
+    containerColor = 'red';
+    textColor = 'white';
+  } else if (isDarkMode) {
```
AlbertCarreras, line 43: is DarkMode the same than default variant when `_dangerouslyUseDarkGray=false`?

```
containerColor = 'white';
    textColor = 'darkGray';
```

dangerismycat: Yep: colors are inverted so the toasts pop against dark backgrounds.

packages/gestalt/src/Toast.js
```diff
@@ -23,10 +27,38 @@ export default function Toast({
...
+
+  let containerColor = 'white';
+  let textColor = 'darkGray';
+  let textElement = text;
+
+  if (_dangerouslyUseDarkGray) {
+    containerColor = isDarkMode ? 'white' : 'darkGray';
+    textColor = isDarkMode ? 'darkGray' : 'white';
+
+    // If `text` is a Node, we need to override any text colors within to ensure they all match
```
AlbertCarreras, line 44: matcha betta TankU

AlbertCarreras (approved):
LGTM!

packages/gestalt/src/Toast.css
```diff
@@ -0,0 +1,7 @@
+.textColorOverrideWhite * {
+  color: var(--g-colorGray0) !important; /* stylelint-disable-line declaration-no-important */
```
christianvuerings, line 2: @dangerismycat when we roll it out, will we be able to avoid the `!important` and `*`? Both of these are anti patterns

dangerismycat: I don't think so, and that's the part I was unsure about. We need to be able to control the color of the `text` when it's a Node, and I couldn't think of a better solution.

PR #1702, opened 2021-09-21
Box: use generated props table

dangerismycat:
### Summary

This PR updates the Box docs page to use a generated props table. Note the filtering to remove the responsive props, which didn't look great with the naive props alphabetization.
Screen Shot 2021-09-20 at 5 08 25 PM
Screen Shot 2021-09-20 at 5 37 02 PM


AlbertCarreras:
BEFORE
image

AFTER
image

You cannot export the types, they have to be on the same file.

image

AlbertCarreras:
Also, the default only get picked up if the props are destructured

AlbertCarreras (changes requested):
Sorry, left comments but didn't click the button!

docs/pages/box.js
```diff
@@ -1,258 +1,70 @@
...
+
+export default function SearchFieldPage({ generatedDocGen }: {| generatedDocGen: DocGen |}): Node {
+  return (
+    <Page title="Box">
+      <PageHeader
+        name="Box"
+        description="Box is a component primitive that can be used to build the foundation of pretty much any other component. It keeps details like spacing, borders and colors consistent with the rest of Gestalt, while allowing the developer to focus on the content."
+      />
+
+      <GeneratedPropTable generatedDocGen={generatedDocGen} excludeProps={ignoredProps} />
```
dangerismycat, line 55: We already had a means to ignore props in the props table, so using that now

docs/pages/box.js
```diff
@@ -1,258 +1,70 @@
...
+  'lgPadding',
+  'smPaddingX',
+  'mdPaddingX',
+  'lgPaddingX',
+  'smPaddingY',
+  'mdPaddingY',
+  'lgPaddingY',
+];
+
+export default function BoxPage({ generatedDocGen }: {| generatedDocGen: DocGen |}): Node {
```
dangerismycat, line 47: Most of this content is the same, but I copyedited here and there. Not really any major changes.

packages/gestalt/src/Box.js
```diff
@@ -80,96 +54,818 @@ extra runtime typechecks in the transformers for performance.
 
 */
 
+type UnionRefs = HTMLDivElement | HTMLAnchorElement;
+
+// These types are mostly defined in boxTypes.js. Unfortunately to get generated docs
```
dangerismycat, line 59: This makes me sad. ☹️ I already have a ticket to write a Babel plugin that will inline imported types, which I'll get to in the near-ish future. Once that's merged I'll undo these changes, but I didn't want this PR held up until I have the time for that work.

packages/gestalt/src/Box.js
```diff
@@ -80,96 +54,818 @@ extra runtime typechecks in the transformers for performance.
...
+  /**
+   * An "escape hatch" used to apply styles not otherwise available on Box.
+   */
+  dangerouslySetInlineStyle?: {|
+    __style: { [key: string]: string | number | void },
+  |},
+
+  /**
+   * Aligns a flex container's lines within when there is extra space in the cross-axis, similar to how justify-content aligns individual items within the main-axis.
+   * Learn more about Flexbox layouts on [MDN](https://developer.mozilla.org/en-US/docs/Learn/CSS/CSS_layout/Flexbox).
```
dangerismycat, line 78: Added links to MDN Flexbox docs after reading some survey feedback that these props had confusing names (i.e. the person wasn't familiar with Flexbox already).

packages/gestalt/src/Box.js
```diff
@@ -80,96 +54,818 @@ extra runtime typechecks in the transformers for performance.
...
+   * Default: 'stretch'
```
dangerismycat, line 79: Since these default values are largely browser defaults or set in CSS, we need to use this feature to define them. This mostly applies to Box and Flex.

packages/gestalt/src/Box.js
```diff
@@ -80,96 +54,818 @@ extra runtime typechecks in the transformers for performance.
...
+  lgPaddingY?: 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10 | 11 | 12,
+  /**
+   * See the [absolute positioning](https://gestalt.pinterest.systems/box#Absolute-positioning) variant for more info.
+   * Default: 'static'
+   */
+  position?: 'static' | 'absolute' | 'relative' | 'fixed',
+  /**
+   * Ref that is forwarded to the underlying input element. See the [using as a ref](https://gestalt.pinterest.systems/box#Using-as-a-ref) variant for more info.
+   */
+  ref?: UnionRefs,
```
dangerismycat, line 832: Interestingly, this type _is_ resolved correctly. It appears the line is at imported types.

packages/gestalt/src/Box.js
```diff
@@ -241,57 +450,58 @@ BoxWithForwardRef.propTypes = {
...
-  lgColumn: ColumnPropType,
-
   alignContent: AlignContentPropType,
   alignItems: AlignItemsPropType,
   alignSelf: AlignSelfPropType,
   as: AsPropType,
   bottom: PropTypes.bool,
   borderStyle: BorderStylePropType,
   color: ColorPropType,
+
```
dangerismycat, line 460: alphabetizin' and matching Props order

packages/gestalt/src/Box.js
```diff
@@ -80,96 +54,331 @@ extra runtime typechecks in the transformers for performance.
 
 */
 
+type UnionRefs = HTMLDivElement | HTMLAnchorElement;
```
dangerismycat, line 57: Interestingly, these types _are_ resolved correctly. It appears the dividing line is imported types.

AlbertCarreras:
Nit, we could manually override

image
 to
image

AlbertCarreras:
The spacing could be a little bit more consistent 

image

image

image

AlbertCarreras:
Discrepancy here to explore. 
image

This is not expected : `HTMLAnchorElement`

AlbertCarreras:
This is also a discrepancy 
image

dangerismycat:
> Nit, we could manually override
> 
> image
> to
> image

I'm not sure what to do about this one yet. Technically it's more correct now, just more verbose. I think this one should slide for now and we can work with Design on a possible better solution.

dangerismycat:
> This is also a discrepancy
> image

This one I'm also not too worried about right now, as I don't think it actually means much to most users — they're still going to head over to the ZIndexClasses docs to figure out what to use. This will be addressed by the future Babel plugin.

AlbertCarreras (approved):

PR #1707, opened 2021-09-23
Module: iconButton prop

themindfuldev:
### Summary

Add an `IconButton` element to be placed after the title for a supplemental help CTA. Will not be displayed if `title` is not provided. Not to be used with `badgeText` or `icon`.

Also, refactor types so that these props are mutually exclusive: `badgeText`, `icon` and `iconButton`.

### Links

- [Jira](https://jira.pinadmin.com/browse/PDS-3079)
- [TDD](https://paper.dropbox.com/doc/Gestalt-TDD-Modules-help-IconButton--BS4TjZGhn~rEVeVBa3YZgGeqAg-SFDuQbfYhXfiOdorq2vrn)
- [Figma](https://www.figma.com/file/vjhfBsOtHw0wVg67vqwz1v/01-.-Web-Sticker-Sheet?node-id=4068%3A7200)

### Checklist

### DIFF
--- a/packages/gestalt/src/ModuleExpandable.js
+++ b/packages/gestalt/src/ModuleExpandable.js
@@ -29,7 +29,7 @@ export default function ModuleExpandable({
     setExpandedId(getExpandedId(expandedIndex));
   }, [expandedIndex, setExpandedId]);
 
-  const onModuleClickedHandler = useCallback(
+  const buildOnModuleClickHandler = useCallback(
     (index: number) => (isExpanded: boolean): void => {
       if (onExpandedChange) {
         onExpandedChange(isExpanded ? null : index);
@@ -56,7 +56,7 @@ export default function ModuleExpandable({
               iconButton={props.iconButton ? props.iconButton : undefined}
               id={`${id}-${index}`}
               isCollapsed={expandedId !== index}
-              onModuleClicked={onModuleClickedHandler(index)}
+              onModuleClicked={buildOnModuleClickHandler(index)}
               summary={summary}
               title={title}
               type={type}
--- a/packages/gestalt/src/ModuleExpandable.js
+++ b/packages/gestalt/src/ModuleExpandable.js
@@ -42,7 +42,7 @@ export default function ModuleExpandable({
   return (
     <Box borderStyle="shadow" rounding={4}>
       {items.map((props: PublicModuleExpandableItemProps, index) => {
-        const { children, summary, title, type } = props;
+        const { children, iconAccessibilityLabel, summary, title, type } = props;
 
         return (
           <Fragment key={index}>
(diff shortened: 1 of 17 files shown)


packages/gestalt/src/ModuleExpandable.js
```diff
@@ -28,43 +22,50 @@ export default function ModuleExpandable({
...
   onExpandedChange,
-}: Props): Node {
+}: PublicModuleExpandableProps): Node {
   const [expandedId, setExpandedId] = useState<?number>(getExpandedId(expandedIndex));
 
   useEffect(() => {
     setExpandedId(getExpandedId(expandedIndex));
   }, [expandedIndex, setExpandedId]);
 
+  const onModuleClickedHandler = useCallback(
```
dangerismycat, line 32: nit: this is a function to build the handler rather than the handler itself, so something like `buildOnModuleClickHandler` would be more accurate

packages/gestalt/src/ModuleTitle.flowtest.js
```diff
@@ -0,0 +1,32 @@
+// @flow strict
```
dangerismycat, line 1: Thanks for adding this file too!

packages/gestalt/src/ModuleTitle.js
```diff
@@ -1,62 +1,64 @@
...
+          <Icon accessibilityLabel={accessibilityLabel} color={color} icon={iconName} />
+        </Flex.Item>
       )}
 
-      <Flex.Item minWidth={0}>
-        <Text color={color} lineClamp={1} weight="bold">
-          {title}
-        </Text>
-      </Flex.Item>
+      {title && (
```
dangerismycat, line 41: I think we'll need stricter enforcement than this, as the user can add a `$FlowFixMe` and add all three possibilities. We also want to make sure that even if the user passes, say, both iconButton and badgeText, we only actually show one of them.

dangerismycat (commented):
So great! This is almost there — just need a bit better enforcement in ModuleTitle.

packages/gestalt/src/ModuleTitle.js
themindfuldev: Thanks @dangerismycat , I just added something to prevent they all rendering together

packages/gestalt/src/ModuleTitle.js
```diff
@@ -1,62 +1,56 @@
...
-    error: {
-      icon: 'workflow-status-problem',
-      color: 'red',
-    },
-  };
+export default function ModuleTitle(props: ModuleTitleProps): Node {
+  const { iconAccessibilityLabel = '', title, type = 'info' } = props;
 
-  const { color, icon: iconName } = TYPE_ICON_ATTRIBUTES[type];
+  const decoration = ['icon', 'badgeText', 'iconButton'].find((prop) => !!props[prop]);
```
dangerismycat, line 16: Oh, that's really clever — I don't think I ever would have thought of this approach. But I still find it very readable. Nice!

dangerismycat (approved):
Thanks again for this work!!

PR #1724, opened 2021-10-01
Status: make `title` optional, add `accessibilityLabel`

dangerismycat:
### Summary

There is a need for the Status icons without the text (or with different text), so this PR makes that prop optional. To keep things accessible, this also adds an optional accessibility label prop.

Ideally we would use union types to make one of {accessbilityLabel, title} required, but that doesn't currently work with our generated props tables. That will be a good upgrade once we figure out the union type story.

### Links

- [Jira](https://jira.pinadmin.com/browse/PDS-3190)

### Checklist

### DIFF
--- a/docs/pages/status.js
+++ b/docs/pages/status.js
@@ -82,7 +82,7 @@ If using \`title\` to describe what the icon represents, \`accessibilityLabel\` 
 
       <MainSection
         name="Localization"
-        description={`Be sure to localize the \`title\` and \`subtext\` props. Note that localization can lengthen text by 20 to 30 percent.`}
+        description={`Be sure to localize the \`title\`, \`subtext\` and \`accessibilityLabel\` props. Note that localization can lengthen text by 20 to 30 percent.`}
       />
 
       <MainSection name="Variants">
--- a/docs/pages/status.js
+++ b/docs/pages/status.js
@@ -45,6 +45,42 @@ export default function SearchFieldPage({ generatedDocGen }: {| generatedDocGen:
       </MainSection>
 
       <MainSection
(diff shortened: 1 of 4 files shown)


docs/pages/status.js
```diff
@@ -44,6 +44,42 @@ export default function SearchFieldPage({ generatedDocGen }: {| generatedDocGen:
...
+  <Text align="center" weight="bold">
+    Dynamic retargeting
+  </Text>
+</Flex>`}
+          />
+        </MainSection.Subsection>
+      </MainSection>
+
       <MainSection
         name="Localization"
```
AlbertCarreras, line 84: Can we add `accessibilityLabel` to 

```
MainSection
        name="Localization
```?

**Review by AlbertCarreras — APPROVED** [review]

File: docs/pages/status.js [hunk 2]
```diff
@@ -44,6 +44,42 @@ export default function SearchFieldPage({ generatedDocGen }: {| generatedDocGen:
...
+  <Status accessibilityLabel="This item has a problem" type="problem" />
+  <Text weight="bold" size="lg">Dynamic retargeting</Text>
+</Flex>`}
+          />
+          <MainSection.Card
+            cardSize="md"
+            defaultCode={`
+<Flex alignItems="end" direction="column" gap={1}>
+  <Status title="This item has a problem" type="problem" />
+  <Text align="center" weight="bold">
```
AlbertCarreras, line 75: I think align="center" is unnecessary

docs/pages/status.js
dangerismycat: Good catch!

PR #1726, opened 2021-10-03
Internal: remove prop-types

christianvuerings:
### Impact

* **Performance**: 12.4% Decrease in Gestalt's prod bundle size: 1,837 to 1,609 kB
* **Maintainability**: Avoid duplication between PropTypes and Flow types. Also results in having to extra less types which were shared between PropTypes and Flow.
* **Improved autogenerated docs support**: Enforce `Props` type on components with forwardRef
* **Linting**: Lint on required props which always have a default value


dangerismycat:
I thought we needed to support PropTypes for some consumers of Gestalt? Is that no longer the case?

christianvuerings:
@dangerismycat no longer the case. Main thing we wanted to wait for is Next.js and the generated prop table. Right now the duplication of types is causing more trouble than it's worth.

dangerismycat (approved):
Oh, awesome. I wish I could r-plus this a thousand times then 😄

PR #1746, opened 2021-10-20
Datapoint: text sizing adjustments

ponori:
Adjusted the size of the title to 14px  (from 12px) and the size of the larger datapoint value to 28px (from 36px)

Before 
image

After
image


dangerismycat:
This looks fine to me as well, but please don't merge until we've verified that this won't break existing layouts in Pinboard.

PR #1760, opened 2021-10-29
NumberField: add new component

dangerismycat:
### Summary

We've had requests to add a version of TextField that is specific to numerical input. This makes numbers easier to work with in a couple of ways:
- Inputs (`value`) and outputs (in handlers) are all numbers, not strings, so the dev doesn't need to parse and stringify
- Add min/max/step props, mirroring those `<input>` attributes

This PR also converts TextField to use generated docs, and cleans up some copy slightly. Converting the docs page to our newer style is out of scope, as is converting NumberField's docs to the newer style (since it's so similar to TextField, so copy/paste/modify).

### Links

- [Jira](https://jira.pinadmin.com/browse/PDS-2860)

### Checklist

### DIFF
--- a/docs/pages/numberfield.js
+++ b/docs/pages/numberfield.js
@@ -178,7 +178,9 @@ function Example(props) {
         id="ref example"
         name="Example: ref"
         description={`
-    Set a ref on NumberField to use the [Constraint validation API](https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Constraint_validation) or to anchor a Popover-based element.
+    Set a ref on NumberField to use the [Constraint Validation API](https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Constraint_validation) or to anchor a Popover-based element.
+
+    Note that while the arrow buttons will not exceed the min/max (if set), the user is free to enter any number using the keyboard. Validation should be performed explicitly using the Constraint Validation API to ensure the value is within the specified range.
   `}
         defaultCode={`
 function Example(props) {
@@ -206,6 +208,7 @@ function Example(props) {
       }}
       placeholder="Enter a number from 1–9"
       ref={ref}
+      step={2}
       value={value}
     />
   );
--- a/docs/pages/numberfield.js
+++ b/docs/pages/numberfield.js
@@ -1,0 +1,246 @@
+// @flow strict
+import { type Node } from 'react';
+import Example from '../components/Example.js';
+import PageHeader from '../components/PageHeader.js';
+import Card from '../components/Card.js';
+import MainSection from '../components/MainSection.js';
+import Page from '../components/Page.js';
+import GeneratedPropTable from '../components/GeneratedPropTable.js';
+import docgen, { type DocGen } from '../components/docgen.js';
+
+export default function NumberFieldPage({ generatedDocGen }: {| generatedDocGen: DocGen |}): Node {
+  return (
+    <Page title="NumberField">
+      <PageHeader name="NumberField" description={generatedDocGen?.description} />
+
+      <GeneratedPropTable generatedDocGen={generatedDocGen} />
+
+      <MainSection name="Usage guidelines">
+        <MainSection.Subsection columns={2}>
+          <MainSection.Card
+            cardSize="md"
+            type="do"
+            title="When to Use"
+            description={`
+          - Any time succinct numerical data needs to be entered by a user.
+        `}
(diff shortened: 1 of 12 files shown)


packages/gestalt/src/NumberField.js
```diff
@@ -0,0 +1,152 @@
...
+   */
+  size?: 'md' | 'lg',
+  /**
+   * Indicates the amount the value will increase or decrease when using the input's arrows.
+   */
+  step?: number,
+  /**
+   * The current value of the input.
+   */
+  value?: number | typeof undefined,
```
AlbertCarreras, line 93: typeof undefined, is "undefined": a string, shouldn't this be `undefined`

packages/gestalt/src/NumberField.js
```diff
@@ -0,0 +1,152 @@
+// @flow strict
+import { forwardRef, type Node } from 'react';
+import InternalTextField from './InternalTextField.js';
+
+// <input> deals with strings, but we only want numbers for this component.
+// So we parse what we get from InternalTextField and we stringify what we give it.
```
AlbertCarreras, line 6: nit 
So we stringify what we give to InternalTexField and then parse what we get from InternalTextField

packages/gestalt/src/NumberField.js
dangerismycat:
```
Cannot use `undefined` as a type. A name can be used as a type only if it refers to a type, interface, class, or enum definition. To get the type of a non-class value, use `typeof`.Flow(value-as-type)
```
Though `undefined` as a type is really just an alias for `void`, so I updated to use the latter.

…unless you think that makes the docs less readable?

packages/gestalt/src/NumberField.js
```diff
@@ -0,0 +1,152 @@
...
+      onChange={parseHandlerValue(onChange)}
+      onFocus={parseHandlerValue(onFocus)}
+      onKeyDown={parseHandlerValue(onKeyDown)}
+      placeholder={placeholder}
+      size={size}
+      step={step}
+      ref={ref}
+      type="number"
+      // See comment above — we need to stringify what we give InternalTextField
+      value={value === undefined ? value : String(value)}
```
AlbertCarreras, line 145: nit

value === undefined ? value : String(value)
vs
value === undefined ? undefined : String(value)

?

docs/pages/numberfield.js
```diff
@@ -0,0 +1,246 @@
...
+  );
+}
+`}
+      />
+
+      <Example
+        id="ref example"
+        name="Example: ref"
+        description={`
+    Set a ref on NumberField to use the [Constraint validation API](https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Constraint_validation) or to anchor a Popover-based element.
```
AlbertCarreras, line 181: Can we explain a little bit more in here. From my observations, using the native mouse and buttons to increase/decrease the number will follow the max/min/steps. 

However, if entered manually outside the min/max range, it's still a valud entry and the only way to check for validity is ref.current.checkValidity()

AlbertCarreras (approved):
LGTM, with a couple comments

docs/pages/numberfield.js
```diff
@@ -0,0 +1,246 @@
...
+import Card from '../components/Card.js';
+import MainSection from '../components/MainSection.js';
+import Page from '../components/Page.js';
+import GeneratedPropTable from '../components/GeneratedPropTable.js';
+import docgen, { type DocGen } from '../components/docgen.js';
+
+export default function NumberFieldPage({ generatedDocGen }: {| generatedDocGen: DocGen |}): Node {
+  return (
+    <Page title="NumberField">
+      <PageHeader name="NumberField" description={generatedDocGen?.description} />
```
ayeshakmaz, line 14: can you add a defaultCode example here?

dangerismycat: I wasn't going to update this to the new docs style, as we can do that in the future and this is basically a copy/paste/tweak from TextField's docs

packages/gestalt/src/NumberField.js
dangerismycat: It doesn't really matter, but I like the philosophy of "we're always returning the value, just sometimes we stringify it" as opposed to "sometimes we return the value, sometimes we return something else" — even though in practice it doesn't change anything here. It just seems more "correct" to me philosophically.

PR #1762, opened 2021-11-03
Internal: upgrade Eslint to 8.1.0, + configs+plugins

dangerismycat:
see title


.eslintrc.json
```diff
@@ -8,15 +8,13 @@
     "plugin:jsx-a11y/recommended",
     "plugin:@next/next/recommended",
     "plugin:react/recommended",
-    "prettier",
-    "prettier/flowtype",
```
dangerismycat, line 12: No need for these anymore, they're now included in `prettier`

.eslintrc.json
```diff
@@ -8,15 +8,13 @@
...
-    "prettier/react"
+    "prettier"
   ],
   "env": {
     "browser": true,
     "es6": true
   },
-  "parser": "babel-eslint",
+  "parser": "@babel/eslint-parser",
```
dangerismycat, line 17: `babel-eslint` is deprecated, replaced by this parser. Upgrading is necessary to work with eslint 8.x.x

.eslintrc.json
```diff
@@ -65,6 +63,7 @@
     "no-await-in-loop": "off",
     "no-constant-condition": "error",
     "no-return-await": "off",
+    "prefer-arrow-callback": "off",
```
dangerismycat, line 66: This wanted to change all our forwardRef functions to arrow functions, losing the function name. While I typically prefer arrow function callbacks, I like the communication of naming the function where it provides clarification.

.eslintrc.json
```diff
@@ -75,6 +74,7 @@
     "react/no-array-index-key": "off",
     "react/react-in-jsx-scope": "off",
     "react/require-default-props": "off",
+    "react/sort-comp": "off",
```
dangerismycat, line 77: This is all about sorting class members, which is largely irrelevant to our codebase. The errors it raised weren't worth worrying about.

package.json
```diff
@@ -45,17 +45,17 @@
...
-    "eslint-plugin-jest": "^24.4.0",
-    "eslint-plugin-jsx-a11y": "^6.3.1",
-    "eslint-plugin-react": "^7.21.3",
-    "eslint-plugin-react-hooks": "^4.1.2",
+    "eslint-plugin-flowtype": "^8.0.3",
+    "eslint-plugin-import": "^2.25.2",
+    "eslint-plugin-jest": "^25.2.2",
+    "eslint-plugin-jsx-a11y": "^6.4.1",
+    "eslint-plugin-react": "^7.26.1",
+    "eslint-plugin-react-hooks": "^4.2.1-alpha-255221c86-20211103",
```
dangerismycat, line 58: We'll have to use this alpha release for now as the latest stable release isn't compatible with eslint 8.x.x. Hopefully the next stable release happens sometime soon

packages/gestalt/src/VideoPlayhead.js
```diff
@@ -38,6 +38,7 @@ export default class VideoPlayhead extends PureComponent<Props, State> {
     }
   };
 
+  // eslint-disable-next-line class-methods-use-this
```
dangerismycat, line 41: I agree with this rule in general, but don't think it's worth worrying about here. As with `react/sort-comp`, this will fall away as we refactor our few remaining class components.

ayeshakmaz (approved):
LGTM

package.json
ayeshakmaz: Should we add a slackbot reminder or something to check on this in a month or so?

PR #1766, opened 2021-11-04
ComboBox: added lineClamp to ComboBoxOptionItem

AlbertCarreras:
### Summary

Before 
image
After 
image

### Links

- [TDD](link to Paper doc)
- [Figma](link to Figma file)

### Checklist

### DIFF
--- a/packages/gestalt/src/ComboBox.js
+++ b/packages/gestalt/src/ComboBox.js
@@ -333,6 +333,7 @@ const ComboBoxWithForwardRef: React$AbstractComponent<Props, HTMLInputElement> =
                     id={id}
                     index={index}
                     key={`${option.label}${index}`}
+                    lineClamp={1}
                     option={option}
                     onSelect={({ event, item }) => handleSelectOptionItem({ event, item })}
                     selected={selectedOption ?? selectedItem}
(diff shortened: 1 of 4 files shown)


packages/gestalt/src/ComboBox.js
```diff
@@ -337,6 +337,7 @@ const ComboBoxWithForwardRef: React$AbstractComponent<Props, HTMLInputElement> =
                     onSelect={({ event, item }) => handleSelectOptionItem({ event, item })}
                     selected={selectedOption ?? selectedItem}
                     setHoveredItemIndex={setHoveredItemIndex}
+                    truncate={2}
```
ayeshakmaz, line 340: I think I'm missing something, but why 2 here and 1 elsewhere?

AlbertCarreras: @ayeshakmaz So, I could do 1... The issue here is that they want this combobox in very narrow width field... so one line would show next to no text.  Adding 2 line clamp could give a little bit more context for each line. But I can keep 1 for consistency

image

dangerismycat (approved):

ayeshakmaz (approved):

PR #1767, opened 2021-11-04
Internal: Changes test id prop to 'data-test-id'

manvoid:
### Summary

The data test id prop in the Pinterest codebase is usually 'data-test-id'. This pr changes the OptionItem props to be the same and avoid confusion.

This also includes a change in the setup of react testing library so it uses 'data-test-id' for the getByTestId query (the default is 'data-testid'). This is done via the configure function inside rtl. For more info look [https://testing-library.com/docs/dom-testing-library/api-configuration](https://testing-library.com/docs/dom-testing-library/api-configuration). Pinboard uses `data-test-id`, so this change matches the configuration on Pinboard.

### Checklist

### DIFF
--- a/packages/gestalt/src/ScrollBoundaryContainer.jsdom.test.js
+++ b/packages/gestalt/src/ScrollBoundaryContainer.jsdom.test.js
@@ -11,7 +11,7 @@ describe('ScrollBoundaryContainer', () => {
 
     const { getByTestId } = render(
       <ScrollBoundaryContainer overflow="scroll">
-        <Box data-testid="childrenId" ref={ref} />
+        <Box data-test-id="childrenId" ref={ref} />
       </ScrollBoundaryContainer>,
     );
 


scripts/setupJest.js
```diff
@@ -1 +1,6 @@
 import '@testing-library/jest-dom/extend-expect.js';
+
+// Set up react testing library.
+import { configure } from '@testing-library/react';
```
AlbertCarreras, line 4: can you add a little bit more context on this change?

AlbertCarreras (approved):
LGTM

scripts/setupJest.js
AlbertCarreras: This is needed so packages/gestalt/src/ScrollBoundaryContainer.jsdom.test.js  can pass. Also, Pinboard uses `data-test-id` configuration so this change mirrors the configuration on Pinabord.

PR #1774, opened 2021-11-09
Masonry: delay updating React scroll state until scroll events stop coming in

valbooth28:
### Summary
**TLDR:** This changes the `updateScrollPosition` strategy in Masonry from: making sure that updating isn't called multiple times w/in 100ms, to: not calling `updateScrollPosition` at all until scroll callbacks stop occurring.

The browser API for scroll callbacks is called multiple times per scroll events. Therefore, any code called within the scroll handler callback can very easily slow down scroll. This seems to be the idea behind the `throttle`ing here, to ensure multiple callbacks aren't called within 100ms of each other: 

https://github.com/pinterest/gestalt/blob/6021664f7341a7f3f7a52887e388c1203da77e41/packages/gestalt/src/Masonry.js#L135

https://github.com/valbooth28/gestalt/blob/aefb6127eacbe19a23bdfdc0ecf8d326c4eb6dac/packages/gestalt/src/throttle.js#L10

At the start of a scroll, however, no scroll callback has been called, so the React state update is called _immediately_, slowing down the 1,2,3... invocations of the `addEventListener('scroll'` events. When scrolling a page during Performance Profiling in Chrome DevTools, it is explicitly highlighted as slowing down the handler:

Scroll Before
Scroll Before 2

### Links

- [Jira](https://jira.pinadmin.com/browse/PERF-10133)
- Further documentation about Scroll RUM data logging seen in the gifs available upon request

#### Before
_See above screenshots as well. PWT values are milliseconds per frame (or 1/FPS)_

scroll PWT before

#### After
_PWT values are milliseconds per frame (or 1/FPS)_
Scroll After Code

scroll PWT after

### Checklist

### DIFF
--- a/packages/gestalt/src/Masonry.js
+++ b/packages/gestalt/src/Masonry.js
@@ -1,17 +1,16 @@
 // @flow strict
 import type { ComponentType, Node } from 'react';
-
 import { Component as ReactComponent } from 'react';
 import debounce from './debounce.js';
 import FetchItems from './FetchItems.js';
 import styles from './Masonry.css';
 import ScrollContainer from './ScrollContainer.js';
-import throttle from './throttle.js';
 import type { Cache } from './Cache.js';
 import MeasurementStore from './MeasurementStore.js';
 import { getElementHeight, getRelativeScrollTop, getScrollPos } from './scrollUtils.js';
 import { DefaultLayoutSymbol, UniformRowLayoutSymbol } from './legacyLayoutSymbols.js';
 import defaultLayout from './defaultLayout.js';
+import throttle from './throttle.js';
 import uniformRowLayout from './uniformRowLayout.js';
 import fullWidthLayout from './fullWidthLayout.js';
 import LegacyMasonryLayout from './layouts/MasonryLayout.js';
@@ -87,6 +86,9 @@ type Props<T> = {|
    * This is required if the grid is expected to be scrollable.
    */
   scrollContainer?: () => HTMLElement,
+  deferScrollPositionUpdate?: boolean,
+  virtualBoundsTop?: number,
+  virtualBoundsBottom?: number,
   virtualBoundsTop?: number,
   virtualBoundsBottom?: number,
   /**
@@ -109,6 +111,16 @@ const RESIZE_DEBOUNCE = 300;
 // The amount of extra buffer space for populating visible items.
 const VIRTUAL_BUFFER_FACTOR = 0.7;
 
+// Duration for the function that we'll use to see if scroll is still happening.
+// One mouse scroll creates many addEventListener('scroll', ()=>{}) type events,
+// so if this timeout is reached (not reset by another scroll event), we know scroll
+// is over and perform post scroll tasks.
+<<<<<<< HEAD
+const SCROLL_TIMEOUT_WAIT = 100;
+=======
+const SCROLL_TIMEOUT_WAIT=100;
+>>>>>>> e766dbd8... chore: minor code review changes. Smoke tested w/Pinterest
+
 const layoutNumberToCssDimension = (n) => (n !== Infinity ? n : undefined);
 
 /**
@@ -129,10 +141,8 @@ export default class Masonry<T: { ... }> extends ReactComponent<Props<T>, State<
     }
   }, RESIZE_DEBOUNCE);
 
-  // Using throttle here to schedule the handler async, outside of the event
-  // loop that produced the event.
   // $FlowFixMe[signature-verification-failure]
-  updateScrollPosition = throttle(() => {
+  updateScrollPosition = () => {
     if (!this.scrollContainer) {
       return;
     }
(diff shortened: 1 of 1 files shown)


packages/gestalt/src/Masonry.js
```diff
@@ -145,7 +142,24 @@ export default class Masonry<T: { ... }> extends ReactComponent<Props<T>, State<
...
     });
-  });
+  };
+
+  scrollTimeoutFn: ? TimeoutID = null;
+
+  // The browser scroll event API will fire LOTS of events. This waits until the events
+  // stop coming in to set the final scroll position
+  // $FlowFixMe[signature-verification-failure]
+  handleScroll = () => {
```
valbooth28, line 152: I tried to have a void function signature here:

packages/gestalt/src/Masonry.js
```diff
@@ -145,7 +142,24 @@ export default class Masonry<T: { ... }> extends ReactComponent<Props<T>, State<
...
+    if (this.scrollTimeoutFn) {
+      clearTimeout(this.scrollTimeoutFn);
+    }
+    this.scrollTimeoutFn = setTimeout(this.updateScrollPosition, 100);
+  };
+
+  // $FlowFixMe[signature-verification-failure]
+  clearScrollTimeout = (): void => {
```
valbooth28, line 160: Likewise: 
Screen Shot 2021-11-09 at 2 34 05 PM

packages/gestalt/src/Masonry.js
```diff
@@ -145,7 +142,24 @@ export default class Masonry<T: { ... }> extends ReactComponent<Props<T>, State<
...
+  scrollTimeoutFn: ? TimeoutID = null;
+
+  // The browser scroll event API will fire LOTS of events. This waits until the events
+  // stop coming in to set the final scroll position
+  // $FlowFixMe[signature-verification-failure]
+  handleScroll = () => {
+    if (this.scrollTimeoutFn) {
+      clearTimeout(this.scrollTimeoutFn);
+    }
+    this.scrollTimeoutFn = setTimeout(this.updateScrollPosition, 100);
```
valbooth28, line 156: 100ms based on previous throttle value, happy to change.

packages/gestalt/src/Masonry.js
dangerismycat: Hm…that's odd. I'm not super familiar with typing class methods, but that looks like it should be the correct syntax to me. You could try more explicit syntax, which may work:
```js
handleScroll: () => void = () => {
  ...
}
```
TBH I don't like this syntax as I find it difficult to read, but if it works… ¯\_(ツ)_/¯

packages/gestalt/src/Masonry.js
dangerismycat: I'm not sure if it should change, but do you mind making it a top-of-file const with a descriptive name? If that is a knob we'll potentially want to tweak in the future, it'd be nice to make it easier to find.

packages/gestalt/src/Masonry.js
```diff
@@ -145,7 +142,24 @@ export default class Masonry<T: { ... }> extends ReactComponent<Props<T>, State<
     this.setState({
       scrollTop: getScrollPos(scrollContainer),
     });
-  });
+  };
+
+  scrollTimeoutFn: ?TimeoutID = null;
```
dangerismycat, line 147: nit: this isn't a function, it's a timeout id, so the name seems a bit misleading

dangerismycat (commented):
This seems fine to me. @liuyenwei do you have any concerns?

packages/gestalt/src/Masonry.js
valbooth28: The double parens did it - thanks!

chrislloyd:
This is a large enough change that it should be run as an experiment to validate user behavior. Synthetic dev-approximated results are promising, but lets see what happens in the wild. Happy to approve if you put this behind a flag.

chrislloyd (approved):
Thanks for the speedy turnaround! LGTM

packages/gestalt/src/Masonry.js
```diff
@@ -87,6 +87,7 @@ type Props<T> = {|
    * This is required if the grid is expected to be scrollable.
    */
   scrollContainer?: () => HTMLElement,
+  deferScrollPositionUpdate?: boolean,
```
dangerismycat, line 90: A couple of small requests:
- Can you add a `_` prefix to the prop name? We're in the process of converting our docs props tables to auto-generate from the component Props types, and we're likely to implement screening out undocumented props by that prefix.
- Can you add an explanatory comment with context around this prop?

dangerismycat (approved):

packages/gestalt/src/Masonry.js
valbooth28: Friendly callout that you'll need to disable the dangling eslint rule when you do that: 
Screen Shot 2021-11-16 at 4 21 08 PM

PR #1776, opened 2021-11-10
WIP: union types exploration

dangerismycat:
Attempting to use union types with auto-generated props. Thus far unsuccessful.


dangerismycat:
Leaving this here to document explorations thus far.

PR #1777, opened 2021-11-10
Internal: upgrade stylelint to 14.0.1

dangerismycat:
Also upgrades `stylelint-config-prettier`, `stylelint-config-standard`, `stylelint-order`


packages/gestalt/src/Pulsar.css.flow
```diff
@@ -1,8 +1,8 @@
 // @flow strict
 /* This file is automatically generated by css-modules-flow-types */
 declare module.exports: {|
-  +'AppearAnimation': string,
```
christianvuerings, line 4: @dangerismycat since can be imported in JavaScript, wouldn't it be better to avoid kebab-case?

E.g. `styles.AppearAnimation` instead of `styles['appear-animation']`

christianvuerings (approved):

packages/gestalt/src/Pulsar.css.flow
dangerismycat: Hm…good point. I kept kebab-case as most existing usages were in that case, and I was hesitant to change the animations related to the searchbox since I think that relies on matching some code we don't own. I hadn't considered importing in to JS. Do we do that with animations currently? Do you expect we'll need to in the future?

christianvuerings: @dangerismycat we probably don't right now but I do expect it in the future - especially if we start to become more animation heavy in gestallt

dangerismycat: I'm going to go ahead and merge this PR now, and we can revisit the lint rule if necessary in the future. I don't have strong feelings about it at all.

PR #1779, opened 2021-11-10
Table: Update design guidelines in Docs

AlbertCarreras:
### Summary

Main change: Added design guidelines to Table component

Other improvement:
Refactor subcomponent proptable heading to show up on the right navigation bar.
Moved the subcomponent props to a sublevel in the right nav sidebar to give more sense of structure and readability it.

## Before

image

## After

image

The props tables are VERY long, so it might take a lot of scrolling to get to the Guidelines. So I thought that Variants is for engineers as well as the Subcomponent section, so I placed it at the end assuming engineers already go to Variants to figure out doubts etc,

We know engineers wanted the Props at the top… but very long Props sections including subcomponents might be too much to get to the sections designers go more often.

So by calling it Subcomponents we make it more eng focused so it can just go over Variants, which in most cases is also eng focused.

### Links

- [TDD](link to Paper doc)
- [Figma](link to Figma file)

### Checklist

### DIFF
--- a/docs/pages/table.js
+++ b/docs/pages/table.js
@@ -1,88 +1,18 @@
 // @flow strict
 import type { Node } from 'react';
 import { Table } from 'gestalt';
-import Card from '../components/Card.js';
-import Example from '../components/Example.js';
 import PageHeader from '../components/PageHeader.js';
 import PropTable from '../components/PropTable.js';
-import CardPage from '../components/CardPage.js';
 import MainSection from '../components/MainSection.js';
+import Page from '../components/Page.js';
 
-const cards: Array<Node> = [];
-const card = (c) => cards.push(c);
-
-card(
-  <PageHeader
-    name="Table"
-    description="The Table contains the following composable elements: Table, Table.Body, Table.Cell, Table.Footer, Table.Header, Table.HeaderCell, Table.Row, Table.SortableHeaderCell."
-  />,
-);
-
-card(
-  <PropTable
-    Component={Table}
-    props={[
-      {
-        name: 'accessibilityLabel',
-        type: 'string',
-        description: 'Label for screen readers to announce Table.',
-        required: true,
-      },
-      {
-        name: 'children',
-        type: 'React.Node',
-      },
-      {
-        name: 'borderStyle',
-        type: `"sm" | "none"`,
-        description: 'Specify a border width for table: "sm" is 1px',
-        defaultValue: 'none',
-      },
-      {
-        name: 'maxHeight',
-        type: `number | string`,
-        description: `Use numbers for pixels: maxHeight={100} and strings for percentages: maxHeight="100%"`,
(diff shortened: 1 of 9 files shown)


docs/pages/table.js
```diff
@@ -145,90 +75,1666 @@ card(
...
+        <BaseRow
+          id={tableID}
+          disabled
+          text="This ad is great"
+          campaign="Awareness"
+        />
+        <BaseRow
+          id={tableID}
+          checked={true}
+          text="Macy's pincycle"
```
ayeshakmaz, line 210: I think we should avoid real company names (Macys and Best Buy), and only use fake names

docs/pages/table.js
```diff
@@ -145,90 +75,1666 @@ card(
...
+      </Table.Body>
+    </Table>
+  );
+}
+`}
+          />
+          <MainSection.Card
+            cardSize="md"
+            type="don't"
+            description={`Don’t
```
ayeshakmaz, line 448: Need to remove the "Don't" in description

docs/pages/table.js
```diff
@@ -145,90 +75,1666 @@ card(
...
+Note that localization can lengthen text by 20 to 30 percent; follow our guidelines on concise content and headings to account for localization.
+
+Wrap important table content instead of truncating. Use truncation only for secondary content, and include a tooltip to show the full text on hover.
+`}
+      />
+      <MainSection name="Accessibility">
+        <MainSection.Subsection
+          title="Labels"
+          description={`
+Use  \`accessibilityLabel\` to properly announce the content of the table. For example, use “Ad Groups Table” instead of “Table”.
```
ayeshakmaz, line 1306: This seems counterintuitive to the sentence below. Maybe: "For example, use “Campaign Status Information” instead of “Info”.

docs/pages/table.js
```diff
@@ -145,90 +75,1666 @@ card(
...
+            },
+            {
+              name: 'status',
+              required: true,
+              type: '"active" | "inactive"',
+            },
+          ]}
+        />
+      </MainSection>
+      s
```
ayeshakmaz, line 1732: random "s" 😛

ayeshakmaz (changes requested):

docs/pages/table.js
AlbertCarreras: DONE

docs/pages/table.js
AlbertCarreras: DONE

docs/pages/table.js
AlbertCarreras: DONE

docs/pages/table.js
AlbertCarreras: DONE

ayeshakmaz (approved):

PR #1785, opened 2021-11-18
TextField: Update docs with Best Practices, Accessibility, Localization

dangerismycat:
Also converted existing docs content to the new style

### Links

- [Paper doc](https://paper.dropbox.com/doc/TextField--BWXFhiAjCZ6_squa2ml3VpL4Ag-CQUTbHQNer4tlXZ2PhY6H)

### DIFF
--- a/cypress/integration/accessibility_TextField_spec.js
+++ b/cypress/integration/accessibility_TextField_spec.js
@@ -6,7 +6,7 @@ describe('TextField Accessibility check', () => {
 
   it('Tests accessibility on the TextField page', () => {
     cy.checkA11y({
-      exclude: [['#best-practices-dont-label']],
+      exclude: '#best-practices-dont-label',
     });
   });
 });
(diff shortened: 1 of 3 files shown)


cypress/integration/accessibility_TextField_spec.js
```diff
@@ -5,6 +5,11 @@ describe('TextField Accessibility check', () => {
   });
 
   it('Tests accessibility on the TextField page', () => {
-    cy.checkA11y();
+    cy.checkA11y(null, {
+      rules: {
+        // Disabled to allow for label-less "Don't" example
+        label: { enabled: false },
```
christianvuerings, line 11: @dangerismycat instead of disabling the rule everywhere, should we exclude specific areas?

```js
cy.checkA11y({
    exclude: ['[data-a11y-test-disabled]']
})
```
https://github.com/component-driven/cypress-axe/issues/55

dangerismycat: That's a much better approach! Didn't realize that was possible. Fixing that now

christianvuerings (approved):

cypress/integration/accessibility_TextField_spec.js
dangerismycat: I tried every possible syntax combination I could think of and couldn't get anything to work. I left a comment about this better approach for Future Us, but we'll have to disable the rule altogether for now. :(

PR #1800, opened 2021-12-04
Docs: Center main content for large screens

ponori:
Moved the main block so that it's center-aligned on larger monitors

### DIFF
--- a/cypress/integration/accessibility_OnLinkNavigationProvider_spec.js
+++ b/cypress/integration/accessibility_OnLinkNavigationProvider_spec.js
@@ -7,4 +7,15 @@ describe('OnLinkNavigationProvider Accessibility check', () => {
   it('Tests accessibility on the OnLinkNavigationProvider page', () => {
     cy.checkA11y();
   });
+
+  // Keep disabled until converting to new Code editor
+  cy.configureAxe({
+    rules: [
+      {
+        id: 'color-contrast',
+        enabled: false,
+      },
+    ],
+  });
+  cy.checkA11y();
 });
--- a/cypress/integration/accessibility_OnLinkNavigationProvider_spec.js
+++ b/cypress/integration/accessibility_OnLinkNavigationProvider_spec.js
@@ -4,17 +4,16 @@ describe('OnLinkNavigationProvider Accessibility check', () => {
     cy.injectAxe();
   });
 
-  // Keep disabled until converting to new Code editor
-  cy.configureAxe({
-    rules: [
-      {
-        id: 'color-contrast',
-        enabled: false,
-      },
-    ],
-  });
-
   it('Tests accessibility on the OnLinkNavigationProvider page', () => {
+    cy.configureAxe({
+      rules: [
+        {
+          id: 'color-contrast',
+          enabled: false,
+        },
+      ],
+    });
+
     cy.checkA11y();
   });
 });
--- a/cypress/integration/accessibility_OnLinkNavigationProvider_spec.js
+++ b/cypress/integration/accessibility_OnLinkNavigationProvider_spec.js
@@ -4,10 +4,6 @@ describe('OnLinkNavigationProvider Accessibility check', () => {
     cy.injectAxe();
   });
 
-  it('Tests accessibility on the OnLinkNavigationProvider page', () => {
-    cy.checkA11y();
-  });
-
   // Keep disabled until converting to new Code editor
   cy.configureAxe({
     rules: [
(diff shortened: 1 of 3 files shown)


cypress/integration/accessibility_OnLinkNavigationProvider_spec.js
```diff
@@ -7,4 +7,15 @@ describe('OnLinkNavigationProvider Accessibility check', () => {
...
+
+  // Keep disabled until converting to new Code editor
+  cy.configureAxe({
+    rules: [
+      {
+        id: 'color-contrast',
+        enabled: false,
+      },
+    ],
+  });
```
ayeshakmaz, line 19: These lines should go above line 8, and then you don't need to repeat cy.checkA11y();

cypress/integration/accessibility_Upsell_spec.js
```diff
@@ -15,4 +15,15 @@ describe('Upsell Accessibility check', () => {
...
+  // Keep disabled until converting to new Code editor
+  cy.configureAxe({
+    rules: [
+      {
+        id: 'color-contrast',
+        enabled: false,
+      },
+    ],
+  });
+  cy.checkA11y();
```
ayeshakmaz, line 28: instead of adding this here, you'll want to edit lines 7-17 above to look like: 
```

it('Tests accessibility on the Upsell page', () => {
    cy.configureAxe({
      rules: [
        {
          id: 'scrollable-region-focusable',
          enabled: false,
        },
{
          id: 'color-contrast',
          enabled: false,
        },
      ],
    });
    cy.checkA11y();
  });
```

ayeshakmaz: (with better spacing :P )

ayeshakmaz (changes requested):

cypress/integration/accessibility_Upsell_spec.js
ponori: Done!

cypress/integration/accessibility_OnLinkNavigationProvider_spec.js
ponori: Done!

ayeshakmaz: sooo this should look like:
```

it('Tests accessibility on the OnLinkNavigationProvider page', () => {
  cy.configureAxe({
    rules: [
      {
        id: 'color-contrast',
        enabled: false,
      },
    ],
  });

    cy.checkA11y();
  });
  
```
  Just like your Upsell one now

docs/components/AppLayout.js
```diff
@@ -22,8 +22,19 @@ export default function AppLayout({ children }: Props): Node {
...
+            padding={4}
+            mdPadding={6}
+            lgPadding={8}
+            marginBottom={12}
+            width="100%"
+            role="main"
+            lgDisplay="flex"
+            justifyContent="center"
+          >
+            <Box width="100%" maxWidth={1440}>
```
ayeshakmaz, line 35: Can I ask where 1440 comes from?

ponori: You certainly can and should! I think I was attempting to reverse-engineer the max-size of the current content width, but it seems pretty obvious that I got the math wrong. I think we can change the max-width to 1544 - would that make sense to you?

ayeshakmaz: Hmm I'm getting 1552, 1312 for the main column + max 240 for the right-side nav?

ponori:
What’s strange about this is the sidebar doesn’t seem to grow to 240. When I measure how the DOM elements render out, I get 1544.

1312 (main content width) + 32 (sidebar margin) + 200 (sidebar width)

ayeshakmaz:
Ah yeah there's some funky-ness with that sidebar that I don't think is worth worrying about given future plans, so this looks good to me!

ponori:
Awesome!

ayeshakmaz (approved):

docs/components/AppLayout.js
```diff
@@ -22,8 +22,19 @@ export default function AppLayout({ children }: Props): Node {
...
+            padding={4}
+            mdPadding={6}
+            lgPadding={8}
+            marginBottom={12}
+            width="100%"
+            role="main"
+            lgDisplay="flex"
+            justifyContent="center"
+          >
+            <Box width="100%" maxWidth={1544}>
```
dangerismycat, line 35: This maxWidth is kind of a magic number…can this be a top-of-file const with an explanatory name, and perhaps comment?

ponori:
I’m all for whatever you think is best.

PR #1808, opened 2021-12-10
Tooltip: Add example alongside Label and TextField

ayeshakmaz:
#### What changed?

Updated the Best Practice example to show amore contextual use of the Tooltip + IconButton pairing

#### Why?

Hoping to help provide more guidance on where to place info IconButtons in forms

### Checklist

### DIFF
--- a/docs/pages/tooltip.js
+++ b/docs/pages/tooltip.js
@@ -203,15 +203,27 @@ card(
         type="do"
         description="Use Tooltip to add supplementary information about a feature, typically paired with an `info-circle` [IconButton](/iconbutton)."
         defaultCode={`
-<Tooltip text="Total ad spend in the select time period">
-  <IconButton
-    accessibilityLabel="Additional info."
-    bgColor="white"
-    icon="info-circle"
-    iconColor="darkGray"
-    size="lg"
-  />
-  </Tooltip>
+        <Flex direction="column" gap={1}>
+          <Flex alignItems="center" gap={1}>
+            <Label for="end-date-field">
+              <Text size="sm">Business URL</Text>
+            </Label>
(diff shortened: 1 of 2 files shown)


packages/gestalt/src/Icon.js
```diff
@@ -90,6 +90,8 @@ const flipOnRtlIconNames = [
 
 /**
  * [Icons](https://gestalt.pinterest.systems/icon) are the symbolic representation of an action or information, providing visual context and improving usability.
+ *
+ * See the [Iconography and SVG guidelines](/iconography_and_svgs) to explore the full icon library.
```
ayeshakmaz, line 94: Added this so that it's visible without opening the props table

docs/pages/tooltip.js
```diff
@@ -203,15 +203,27 @@ card(
...
-  />
-  </Tooltip>
+        <Flex direction="column" gap={1}>
+          <Flex alignItems="center" gap={1}>
+            <Label htmlFor="business-url-field">
+              <Text size="sm">Business URL</Text>
+            </Label>
+            <Tooltip text="This is the site users will be redirected to when interacting with your ad">
+              <IconButton
+                accessibilityLabel="Additional info."
```
dangerismycat, line 213: the period seems odd?

packages/gestalt/src/Icon.js
dangerismycat: can you use the full URL here so the link works in IDEs as well?

dangerismycat (approved):

docs/pages/tooltip.js
ayeshakmaz: hah it does, removing

PR #1809, opened 2021-12-10
Module: use generated proptable

christianvuerings:
#### What changed?

* Use generated proptable for the `Module` component
* Inline `Module` flow types
* Remove `overrideTypes` (inlined)

#### Why?

Inlining the props has the following benefits:

* No more weird props spreading in order to make flow happy
* You can quickly see what the props are for a component you are working on without having to go through a habit hole of combined flow types.


docs/pages/module.js
```diff
@@ -5,77 +5,18 @@ import PropTable from '../components/PropTable.js';
...
-          {
-            name: 'type',
-            href: 'static-error',
-            type: '"info" | "error"',
-            defaultValue: 'info',
-            description:
-              'If set to `error`, displays error icon and changes title to red text. Be sure to provide an `iconAccessibilityLabel` when set to `error`.',
-          },
-        ]}
+        description="A Module is a container that holds content about one subject. Its contents can be visible at all times, or expand and collapse as individual modules or a group of modules."
```
dangerismycat, line 17: Can this also be moved to the component file along with the props?

docs/pages/module.js
```diff
@@ -582,3 +523,19 @@ function ModuleExample5() {
...
 }
+
+export async function getStaticProps(): Promise<{| props: {| generatedDocGen: DocGen |} |}> {
+  const generatedDocGen = await docgen('Module');
+
+  generatedDocGen.props.icon = {
+    ...generatedDocGen.props.icon,
+    flowType: {
+      name: 'string',
+      raw: 'Icon[icon]',
```
dangerismycat, line 534: Screen Shot 2021-12-13 at 11 14 13 AM

The current `string` type for this prop isn't helpful, but I'm not sure that this is either. Perhaps the description should include a link to `/icon` to guide the user to valid icon names?

docs/components/docgen.js
```diff
@@ -43,16 +43,3 @@ export default async function docgen(component: string): Promise<DocGen> {
 
   return parsed;
 }
-
-export function overrideTypes(docGen: DocGen, typeOverrides: {| [string]: string |}): DocGen {
```
dangerismycat, line 47: Given that this functionality is used in several places, is there a reason to remove this helper function and inline the logic?

docs/pages/module.js
christianvuerings: @dangerismycat Definitely, updated.

docs/pages/module.js
christianvuerings: @dangerismycat This uses Flow Indexed Access types: https://flow.org/en/docs/types/indexed-access/. I also added a link to https://gestalt.netlify.app/iconography_and_svgs

docs/components/docgen.js
christianvuerings: @dangerismycat I added it back in to unblock this PR. Originally I inlined it for the following reasons:

* The function has the wrong abstraction: it always assumes that the type is `union` (for module I needed `string`). We could add another prop to allow for a different type but that only makes the function more complex.
* `typeOverrides` felt like a weird format, I was expecting `{ name: 'Icon', raw: '', name: '' }`

In general I feel like we're making our code base a bit too DRY by over-abstracting code which makes it harder to see what the code is supposed to do.

More resources on this topic:
* https://www.deconstructconf.com/2019/dan-abramov-the-wet-codebase
* https://www.philosophicalhacker.com/post/when-to-dry/

christianvuerings:
@dangerismycat I addressed all the comments - mind taking another look? Feel free to self-merge

docs/pages/module.js
dangerismycat: TIL!

docs/components/docgen.js
dangerismycat: Totally valid reason — I think you could have left these changes in the PR. We can re-add them in a follow-up

dangerismycat: Also: I get the over-abstraction pain as well. I was just glancing through the Button types the other day and was so confused. It felt like that tweet where Abramov was revisiting Redux and thought it was way too complex 😅

dangerismycat (approved):

PR #1810, opened 2021-12-11
ComboBox: performance optimization

AlbertCarreras:
## Why

When using Gestalt ComboBox with a huge list (more than 500 items) we are seeing very low performance rendering ComboBox items:
https://codesandbox.io/s/gestalt-huge-list-performance-09sqg?file=/index.js:0-818 (use the list of names in the codesandbox for component testing )

If you start typing you are going to see that the thread gets blocked for several seconds. 

## What changed? 

This PR makes ComboBox use its own ComboBoxItem which is build with native HTML element rather than Gestalt components. It simplifies the API of  ComboBoxItem. Detailed changes are described in the HOW section.

Gains after changes: 66.7% decrease in the timing for rendering 2000 items (~900 ms ->  ~<300 ms) 

## How
- Decoupling ComboBoxItem cmp from Dropdown
- Remove useVisibleFocus
- Implement useCallback
- cleanup ComboBoxItem API
- remove Flex from ComboBoxItem
- use requestAnimationFrame in componentDidMount within Contents<Popover *
- remove Box
- use useMemo for ComboBoxItem list

* NOTE on replacing `setTimeOut` with `requestAnimationFrame` in `componentDidMount` within `Contents`<`Popover`
Replacing `setTimeOut` with `requestAnimationFrame`  is a pretty safe change, prioritizing layout vs CPU.
`requestAnimationFrame` executes a callback function whenever the browser is ready to start the next repaint task. The callback is basically called when it's good to call it.
`setTimeOut` takes on the event loop and sets a fix amount of milliseconds to run a callback.

[Advantage of requestAnimationFrame
](https://www.codingdefined.com/2015/11/why-requestanimationframe-is-better.html)
1. The scheduling is much more accurate than setInterval() and setTimeout().
2. The requestAnimationFrame uses High Resolution Timer which helps it to do time based animation. Time based animation smoothens the animation even if the frame rates are not constant.
3. Multiple animations are possible using requestAnimationFrame.
4. CPU optimization is done using requestAnimationFrame by not drawing when tab or window is not visible.

##  Performance analysis
Measuring performance with [React.Profiler
](https://github.com/pinterest/gestalt/pull/1810#issue-800364439)

Console.log logs the following params:
image

Check the following document to see performance changes for each refactor step.
https://paper.dropbox.com/doc/ComboBox-Perf-Improvement--BX3765UbtpPwkhWKaXt~ceVzAQ-93MBaiXL9hLJeOQAMSIln

##  Next steps

[Virtualization or windowing](https://praekiko.medium.com/what-is-windowing-also-i-have-heard-about-react-window-and-react-virtualized-c29dc843f4e0) would be the next step to bring ComboBox performance to the next level.

### Checklist

[ X ] Verified accessibility: keyboard & screen reader interaction


packages/gestalt/src/ComboBox.js
```diff
@@ -151,87 +160,164 @@ const ComboBoxWithForwardRef: React$AbstractComponent<Props, HTMLInputElement> =
...
 
-  const textfieldIconButton =
-    (controlledInputValue && controlledInputValue !== '') ||
-    (textfieldInput && textfieldInput !== '') ||
-    (tags && tags.length > 0)
-      ? 'clear'
-      : 'expand';
+  // ==== MAPPING ComboBoxItem ====
+  const comboBoxItemList = useMemo(() => {
+    const buildComboBoxItemCallback = ({ label: comboBoxItemlabel, subtext, value }, index) => {
```
dangerismycat, line 293: I think this might be more readable if you inline this into the `.map` call

packages/gestalt/src/ComboBoxItem.js
```diff
@@ -0,0 +1,93 @@
+// @flow strict
+import { forwardRef, type Node } from 'react';
+import Text from './Text.js';
+import Icon from './Icon.js';
+import focusStyles from './Focus.css';
+
+export type OptionItemType = {|
```
dangerismycat, line 7: Should this type name be updated to match the component/file name? (I don't feel strongly about this, just want to make sure it's intentional)

packages/gestalt/src/ComboBoxItem.js
```diff
@@ -0,0 +1,93 @@
...
+      aria-selected={isSelected}
+      className={focusStyles.hideOutline}
+      id={`${id}-item-${index}`}
+      onClick={handleOnTap}
+      onKeyPress={handleEventPreventDefault}
+      onMouseDown={handleEventPreventDefault}
+      onMouseEnter={handleOnMouseEnter}
+      ref={isHovered ? ref : null}
+      role="option"
+      style={{ cursor: 'pointer', width: '100%' }}
```
dangerismycat, line 52: Rather than inlining all these styles, can we import from the various CSS files and use those classnames? That will prevent style bloat.

packages/gestalt/src/Contents.js
```diff
@@ -91,15 +91,17 @@ class Contents extends Component<Props, State> {
   };
 
   componentDidMount() {
-    const { onResize, onKeyDown } = this.props;
+    const { onResize, onKeyDown, shouldFocus } = this.props;
     const { popoverRef } = this.state;
 
-    setTimeout(() => {
-      if (this.props.shouldFocus && popoverRef) {
+    function draw() {
```
dangerismycat, line 97: I'm not sure I understand the name of this function. What is being drawn? It seems like something along the lines of `focusPopoverRef` would be more accurate?

packages/gestalt/src/DropdownItem.js
```diff
@@ -46,7 +46,6 @@ export default function DropdownItem({
           id={id}
           index={index}
           key={`${option.value + index}`}
-          lineClamp={1}
```
dangerismycat, line 49: Are these removals intentional? What purpose do they serve? If we do need to change Dropdown / OptionItem, can that happen in a separate PR, or is it related to this ComboBox work?

packages/gestalt/src/ComboBox.js
AlbertCarreras: DONE

packages/gestalt/src/ComboBoxItem.js
AlbertCarreras: DONE

packages/gestalt/src/ComboBoxItem.js
AlbertCarreras: Previously, we had all these styles within classNames. 
image

AlbertCarreras: It was one of the steps while simplifying code to try to reduce timing. I'll add back, cos I think it's not doing much benefit in terms of performance.

packages/gestalt/src/Contents.js
dangerismycat: Makes sense. I think this new name communicates better as the use of requestAnimationFrame is an implementation detail — the function name should communicate _what_ the function is doing, not _when_ it gets called.

packages/gestalt/src/DropdownItem.js
dangerismycat: Oh I see, lineClamp was only needed for ComboBox?

dangerismycat (approved):
LGTM! Feel free to merge assuming the answer to "was lineClamp only needed for ComboBox?" is "yes" 🙂

PR #1813, opened 2021-12-13
Docs: add logo!!

dangerismycat:
Also moves Pinterest logo to footer, makes header light gray, makes all header buttons/text dark gray

_Light mode_
Screen Shot 2021-12-13 at 1 00 41 PM
Screen Shot 2021-12-13 at 1 00 50 PM

_Dark mode_
Screen Shot 2021-12-13 at 1 01 04 PM
Screen Shot 2021-12-13 at 1 01 11 PM

_Small screens_
Screen Shot 2021-12-13 at 1 00 15 PM
Screen Shot 2021-12-13 at 1 00 24 PM


ponori:
LGTM! I'll let @AlbertCarreras and/or @ayeshakmaz chime in from a code standpoint.

ayeshakmaz (commented):
Does the 500 image get used?

ayeshakmaz (approved):
Looks great to me! Very exciting!!

docs/components/Header.js
```diff
@@ -1,11 +1,13 @@
...
 import { type Node, useCallback, useEffect, useState } from 'react';
-import { Box, Flex, FixedZIndex, Text, Icon, IconButton, Sticky } from 'gestalt';
+import { Box, Flex, FixedZIndex, Text, IconButton, Sticky } from 'gestalt';
+import Image from 'next/image';
 import DocSearch from './DocSearch.js';
 import HeaderMenu from './HeaderMenu.js';
 import Link from './Link.js';
 import trackButtonClick from './buttons/trackButtonClick.js';
 import { useNavigationContext } from './navigationContext.js';
+import gestaltLogo from '../public/gestalt-logo-250.png';
```
christianvuerings, line 10: @dangerismycat could we use an inline SVG instead (Just like we do with our icons)? It would result in:

* Avoiding blurriness when zooming in:
image

* No longer flash the logo on load:
logo-flash-on-load

dangerismycat: yep, will take care of that in a follow-up PR

PR #1814, opened 2021-12-13
Internal: CardPage.js deprecation in all Docs pages in favor of Page.js

AlbertCarreras:
#### What changed?

Deprecated `CardPage.js` from all left instances in Gestalt docs by migrating to use `Page.js`. Most components now use `docgen` to generate the component description except a few instances where errors were thrown, which requires further investigation.

#### Why?

Removal of already deprecated file.


dangerismycat (approved):

docs/components/docgen.js
```diff
@@ -25,10 +25,16 @@ export type DocGen = {|
...
+  componentName,
+  alternativeSubdirectory = '',
+}: {|
+  componentName: string,
+  alternativeSubdirectory?: string,
+|}): Promise<DocGen> {
   const filePath = path.join(
     nextConfig().serverRuntimeConfig.GESTALT_ROOT,
-    `/packages/gestalt/src/${component}.js`,
+    alternativeSubdirectory || `/packages/gestalt/src/${componentName}.js`,
```
christianvuerings, line 37: @AlbertCarreras this is a code smell - could we either

* Avoid this extra parameter by moving `OnLinkNavigationProvider` or `ColorSchemeProvider`?
* Make the import explicit for every component?

PR #1816, opened 2021-12-15
Internal: upgrade Prettier to 2.5.1

dangerismycat:
Checked, but no associated packages need upgrading as well


dangerismycat:
I'm tired of fighting stupid Prettier, abandoning this for now since it doesn't really matter

PR #1817, opened 2021-12-15
Upsell: convert to use generated props table

dangerismycat:
Not worrying about Upsell.Form for now. We'll probably need to move it out to its own file in the future, but we should figure out the sidebar plan for that.

### DIFF
--- a/packages/gestalt/src/Upsell.js
+++ b/packages/gestalt/src/Upsell.js
@@ -13,36 +13,28 @@ import styles from './Upsell.css';
 import headingStyles from './Heading.css';
 import typography from './Typography.css';
 import useResponsiveMinWidth from './useResponsiveMinWidth.js';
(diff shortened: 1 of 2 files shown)


packages/gestalt/src/Upsell.js
```diff
@@ -13,37 +13,29 @@ import styles from './Upsell.css';
...
   stacked?: boolean,
   type: string,
-|}): Node => {
+|};
+
+const UpsellAction = ({ data, stacked, type }: UpsellActionProps): Node => {
   const color = type === 'primary' ? 'red' : 'gray';
   const { accessibilityLabel, disabled, href, label, onClick, rel, target } = data;
 
+  const commonProps = {
```
ayeshakmaz, line 29: love this

ayeshakmaz (approved):
LGTM 👍

PR #1821, opened 2021-12-16
Card, Checkbox, Column: refactor to use generated props tables

dangerismycat:
Only adding content where needed — these will still need another pass later to modernize the prop descriptions

### DIFF
--- a/packages/gestalt/src/Column.js
+++ b/packages/gestalt/src/Column.js
@@ -6,7 +6,15 @@ import styles from './Column.css';
 type Columns = 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10 | 11 | 12;
 
 type ColumnProps = {|
+  /**
+   *
+   */
   children?: Node,
+  /**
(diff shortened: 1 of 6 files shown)


christianvuerings (approved):

packages/gestalt/src/Column.js
```diff
@@ -6,15 +6,23 @@ import styles from './Column.css';
...
+   *
+   */
   children?: Node,
+  /**
+   * The number of units in a 12-unit width that this element will occupy.
+   *
+   * Also available in responsive sizes: `smSpan`, `mdSpan`, `lgSpan`
+   */
   span: Columns,
   smSpan?: Columns,
```
christianvuerings, line 19: @dangerismycat should we also add comments to the responsive `span` sizes? We aren't using them for the generated prop table but they should still be useful for our flow docs.

dangerismycat: That's a good point — we're filtering these out of the prop table by prop name, so comments won't affect that. I'll add those here and in Box in a follow-up PR

PR #1822, opened 2021-12-16
Docs: Fixed broken Popover link on Accessibility page

hectoid:
#### What changed?

Fixed the Popover link in the ARIA attributes section so that it is not capitalized.

#### Why?
was generating a 404.


dangerismycat (approved):
LGTM!

PR #1826, opened 2021-12-17
Docs: use inline SVG for header logo

dangerismycat:
As noted by @christianvuerings [in another PR](https://github.com/pinterest/gestalt/pull/1813#discussion_r770368520), using an inline SVG for our logo in the header would allow crisper zoomed display and obviate the need to load the image.

### DIFF
--- a/docs/components/GestaltLogo.js
+++ b/docs/components/GestaltLogo.js
@@ -9,24 +9,25 @@ type Props = {|
 export default function GestaltLogo({ height, width }: Props): Node {
   return (
     <svg
-      width={width ?? 500}
+      aria-label="Home"
+      fill="none"
       height={height ?? 500}
       viewBox="0 0 500 500"
-      fill="none"
+      width={width ?? 500}
       xmlns="http://www.w3.org/2000/svg"
     >
       <path d="M410 249.924H250L410 409.848V249.924Z" fill="#26C0B4" />
       <path
-        fillRule="evenodd"
         clipRule="evenodd"
         d="M90 249.924C90 338.248 161.635 409.848 250 409.848V249.924H90Z"
         fill="#75BFFF"
+        fillRule="evenodd"
       />
       <path
-        fillRule="evenodd"
         clipRule="evenodd"
         d="M250.019 90C161.653 90 90.0186 161.601 90.0186 249.924H250.019V90Z"
         fill="#00857C"
+        fillRule="evenodd"
       />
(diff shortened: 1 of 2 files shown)


dangerismycat:
@christianvuerings do you mind reviewing this whenever you get a minute? Just want to make sure I'm not missing something with the SVG implementation. Feel free to merge if you :rplus: assuming you think it's safe. 🙂

christianvuerings (approved):

christianvuerings:
@dangerismycat looks great, thanks for doing this. The logo now looks crisp and immediately shows up on load.

## Before
image

## After
image

docs/components/GestaltLogo.js
```diff
@@ -0,0 +1,33 @@
+// @flow strict
+import { type Node } from 'react';
+
+type Props = {|
+  height?: number,
```
christianvuerings, line 5: @dangerismycat Nit & out of scope for this PR

What do you think about making Gestalt props read only? It would
* encourage best practices
* avoid certain bugs

2 ways we could do this:

* `$ReadOnly` utility type: https://flow.org/en/docs/types/utilities/#toc-readonly
* Add a `+` for every prop: https://flow.org/en/docs/types/interfaces/#toc-interface-property-variance-read-only-and-write-only

docs/components/GestaltLogo.js
```diff
@@ -0,0 +1,33 @@
...
+  width?: number,
+|};
+
+export default function GestaltLogo({ height, width }: Props): Node {
+  return (
+    <svg
+      width={width ?? 500}
```
christianvuerings, line 12: @dangerismycat should we add a `aria-label` to make the logo icon more accessible? Similar to how we do this with `Icon`:
https://github.com/pinterest/gestalt/blob/2482e777d9938fae50f5e01fc36e57a3cf10c416/packages/gestalt/src/Icon.js#L126

If the answer is no, should we explicitly set `aria-hidden` to `true` on the `svg`?

dangerismycat: Yes, good catch! Updating now

docs/components/GestaltLogo.js
dangerismycat: Sounds like a great idea, but I want to talk through any trade-offs / side effects with the rest of the team. I added it to our agenda for next week's Eng meeting.

PR #1838, opened 2022-01-05
Docs: fix changelog render on What's New page

christianvuerings:
#### What changed?

Fixes a few rendering issues on the What's New page

#### Why?

So the layout no longer looks broken to our users

### Before

Screen Shot 2022-01-05 at 9 04 23 AM

### After

Screen Shot 2022-01-05 at 9 06 00 AM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5340,7 +5340,7 @@ Run codemods for breaking changes in order:
 
 - Icon: add calendar icon to gestalt (#512)
 - Icon: add lightning bolt icon (#513)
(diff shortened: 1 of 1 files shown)


dangerismycat (approved):
These are pretty old commits…is it safe to assume that our current process precludes the possibility of adding links in commit messages?

christianvuerings:
> These are pretty old commits…is it safe to assume that our current process precludes the possibility of adding links in commit messages?

Afaik you can still add markdown links in commit messages but it's something I haven't seen a lot lately.

PR #1839, opened 2022-01-05
Docs: fix links to zIndex page

christianvuerings:
#### What changed?

Fixes links to the zIndex page in our docs

#### Why?

To avoid a broken experience for our users

#### How did you find these?

Thanks to Algolia's new crawler, we can now automatically surface invalid links:

Screen Shot 2022-01-05 at 9 55 31 AM

#### Repro

1. Go to https://gestalt.netlify.app/dropdown
2. Click on `zIndex Classes`
3. Results in: 

```
{"errorType":"TypeError","errorMessage":"Request path contains unescaped characters","trace":["TypeError [ERR_UNESCAPED_CHARACTERS]: Request path contains unescaped characters","    at new NodeError (internal/errors.js:322:7)","    at new ClientRequest (_http_client.js:155:13)","    at request (http.js:94:10)","    at /var/task/.netlify/functions-internal/___netlify-handler/bridge.js:169:19","    at new Promise (<anonymous>)","    at Bridge.launcher (/var/task/.netlify/functions-internal/___netlify-handler/bridge.js:167:12)","    at async Runtime.handler (/var/task/.netlify/functions-internal/___netlify-handler/___netlify-handler.js:123:40)"]}

```

### DIFF
--- a/docs/pages/box.js
+++ b/docs/pages/box.js
@@ -741,7 +741,7 @@ function BoxPopoverExample() {
         </MainSection.Subsection>
 
         <MainSection.Subsection
-          description={`It's possible to use Box with external elements using the CSS \`z-index\` property by capturing those values in controlled objects. The example below shows using a \`FixedZIndex\` for a value that comes from somewhere else, and a \`CompositeZIndex\` to layer the Box on top of it. Visit our [Z-Index documentation](/zindex%20classes) for more details on how to use these utility classes.`}
+          description={`It's possible to use Box with external elements using the CSS \`z-index\` property by capturing those values in controlled objects. The example below shows using a \`FixedZIndex\` for a value that comes from somewhere else, and a \`CompositeZIndex\` to layer the Box on top of it. Visit our [Z-Index documentation](/zindex_classes) for more details on how to use these utility classes.`}
           title="Z-Index"
         >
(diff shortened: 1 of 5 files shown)


dangerismycat (approved):
Nice! Two questions:
- If this crawl is happening automatically, can we set up alerts on it to notify us when it finds broken links?
- Can we be smarter about how we handle escaping/unescaping characters in links?

christianvuerings:
@dangerismycat Good questions - see answers below. Also, I plan to give a quick overview of the Algolia changes to the team on Jan 18 (I'm traveling next week)

> If this crawl is happening automatically, can we set up alerts on it to notify us when it finds broken links?

Right now it crawls every week, however it would be better if we change that to every deploy:

image

Algolia has a Netlify plug-in which automatically crawls on demand that but I'm not sure how much work is involved to move us over: https://www.globenewswire.com/news-release/2020/10/06/2104357/0/en/Algolia-Streamlines-Search-for-Jamstack-Powered-Websites-With-New-Plugin-for-Netlify.html

> Can we be smarter about how we handle escaping/unescaping characters in links?

It looks like Next.js already has some good defaults - using lower case and replace spaces with underscores. So for now, I would suggest following their guidance.

PR #1843, opened 2022-01-06
Collage, Container, Datapoint, Fieldset, Heading: use generated props tables

dangerismycat:
Refactor docs to use generated props tables: Collage, Container, Datapoint, Fieldset, Heading

Also, removed unneeded React imports (probably from an older version of the scaffolding script)

### DIFF
--- a/packages/gestalt/src/Heading.js
+++ b/packages/gestalt/src/Heading.js
@@ -32,7 +32,7 @@ type Props = {|
    */
   accessibilityLevel?: AccessibilityLevel,
   /**
-   * `"start"` and `"end"` should be used for regular alignment since they flip with locale direction. `"forceLeft"` and `"forceRight"` should only be used in special cases where locale direction should be ignored, such as tabular or numeric text. See [example below](#align) for more details.
+   * `"start"` and `"end"` should be used for regular alignment since they flip with locale direction. `"forceLeft"` and `"forceRight"` should only be used in special cases where locale direction should be ignored, such as tabular or numeric text. See [Align example](https://gestalt.pinterest.systems#align) for more details.
    */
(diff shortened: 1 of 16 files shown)


packages/gestalt/src/Heading.js
```diff
@@ -4,39 +4,81 @@ import cx from 'classnames';
...
+  /**
+   * Visually truncate the text to the specified number of lines. This also adds the `title` attribute if `children` is a string, which displays the full text on hover in most browsers. See [example below](#overflowTruncation) for more details.
+   */
   lineClamp?: number,
+  /**
+   * How truncation is handled when text overflows the line. See [example below](#overflowTruncation) for more details.
+   */
   overflow?: Overflow,
+  /**
+   * The font size of the text. See [example below](#sizes) for more details.
```
christianvuerings, line 76: @dangerismycat could we make sure to include `https://gestalt.pinterest.systems/` in all the URLs? That way these links will also work in our flow docs:

```
The font size of the text. See [example](https://gestalt.pinterest.systems/heading#sizes) for more details.
```

christianvuerings: Also, should we change `See example below` to `See example` - the "below" portion doesn't make a lot of sense in flow docs.

AlbertCarreras:
Screen Shot 2022-01-06 at 1 49 37 PM

AlbertCarreras (approved):
LGTM! @dangerismycat Thanks for these changes. Agree with Christian re absolute URLs + removing "below". R+ing so you can merge when you fix them.

packages/gestalt/src/Heading.js
dangerismycat: Ah good point, still clearing the cobwebs from my holiday brain 😅

PR #1849, opened 2022-01-07
Image, Label, Layer, Letterbox, Mask: use generated props tables

dangerismycat:
Mostly leaving existing content intact, except where missing. We'll take another pass in the future to update the content and to use the newer format


AlbertCarreras:
labelDisplay prop

AlbertCarreras:
How do we feel about Layer's  Indexable? 
we should inline the Flow type. 


Screen Shot 2022-01-06 at 9 36 19 PM
9

AlbertCarreras (approved):
Thanks for changes, + with a couple notes.

PR #1859, opened 2022-01-12
Docs: Add Home page

ayeshakmaz:
#### What changed?

Add new Home page for landing 

#### Why?

Provide context and guidance around what Gestalt is with a new Home page

### Links

- [Jira](https://jira.pinadmin.com/browse/GESTALT-3891)
- [Figma](https://www.figma.com/file/w8h95UbqjI9VIl80xuxLAi/Gestalt-Home-Page-2022?node-id=263%3A4966)

### Checklist


docs/components/HomePageCard.js
```diff
@@ -0,0 +1,52 @@
...
+type Props = {|
+  image?: string,
+  description: string,
+  title: string,
+  color: string,
+  href: string,
+|};
+
+const HomePageCard = ({ image, description, title, color, href }: Props): Node => (
+  <Mask rounding={4}>
```
christianvuerings, line 14: @ayeshakmaz Looks like the `Mask` wrapper component breaks the focus styles on the `HomePageCard` component:

# Before

image


# After

image

We can probably fix this by setting the border radius inline

docs/graphics/HomePageIcons.js
```diff
@@ -0,0 +1,414 @@
+// @flow strict
+import { type Node } from 'react';
+
+export function HowToWork(): Node {
+  return (
+    <svg
```
christianvuerings, line 6: @ayeshakmaz since these are no longer just icons, could we import svg files instead? E.g. by using `babel-plugin-react-svg`

That way, it'll also be easier to see a before / after when we update the SVG

docs/pages/home.js
```diff
@@ -0,0 +1,96 @@
...
+              <Flex gap={6} direction="column">
+                <Heading>Gestalt is Pinterest&rsquo;s design system</Heading>
+                <Text size="lg">
+                  We’re here to help you build experiences that inspire people to create the life
+                  they love
+                </Text>
+
+                <Button text="See what’s new" role="link" rel="nofollow" href="/whats_new" />
+              </Flex>
+              <Flex.Item minWidth="60%">
```
christianvuerings, line 34: @ayeshakmaz on mobile this results in a horizontal scroll bar - could we fix this? 
image

docs/components/HomePageCard.js
```diff
@@ -0,0 +1,59 @@
...
+          <Flex direction="column" height={320}>
+            <Box
+              display="flex"
+              alignItems="center"
+              justifyContent="center"
+              height="60%"
+              width="100%"
+              dangerouslySetInlineStyle={{
+                __style: {
+                  backgroundColor: `var(--color-${color}-450)`,
```
christianvuerings, line 28: @ayeshakmaz interesting - should we not allow these colors to be set as `color` on `<Box />`?

docs/components/Header.js
```diff
@@ -34,11 +36,16 @@ function Header() {
...
-                <Box dangerouslySetInlineStyle={{ __style: { marginBottom: '1px' } }}>Gestalt</Box>
+                <Box dangerouslySetInlineStyle={{ __style: { marginBottom: '1px' } }}>
+                  Gestalt Design System
+                </Box>
               </Flex>
             </Box>
           </Link>
         </Text>
+        <Text size="sm" weight="bold">
+          v{GestaltPackageJson.version}
```
christianvuerings, line 47: @ayeshakmaz should we add spacing between the version number and the search field?
image

docs/components/Navigation.js
```diff
@@ -60,6 +61,17 @@ export default function Navigation(): Node {
           maxHeight="calc(100% - 60px)"
           minWidth={MIN_NAV_WIDTH_PX}
         >
+          <SidebarCategorizationButton
```
christianvuerings, line 64: @ayeshakmaz how do we now switch the sidebar view on smaller width screeens?

image

docs/components/Page.js
```diff
@@ -50,18 +53,20 @@ export default function Page({ title: page, children }: Props): Node {
...
-        marginStart={4}
-        mdMarginStart={6}
-        lgMarginStart={8}
-        display="none"
-        lgDisplay="block"
-        flex="none"
-      >
-        <Toc cards={sections} />
-      </Box>
+      {!isFullWidth && (
```
christianvuerings, line 56: @ayeshakmaz isn't it odd that when `isFullWidth` is `true`, we don't render the sections?

ayeshakmaz: Maybe isFullWidth isn't the best name, but this is essentially saying if the page content is full width and doesn't have sections, than don't render the right-side nav

docs/components/HomePageCard.js
dangerismycat: I'm curious about this as well

docs/pages/home.js
ayeshakmaz: Fixed, chatted with Hector and we remove it on mobile

docs/components/HomePageCard.js
ayeshakmaz: For now, we're limiting Box colors to semantic tokens only to avoid the situation we were in before with loads of color options being available

docs/components/Navigation.js
ayeshakmaz: Added the toggles to the mobile view, need to double check with Hector about placement

docs/graphics/HomePageIcons.js
ayeshakmaz: This is a new package correct? We don't currently use it anywhere?

docs/components/Page.js
```diff
@@ -7,11 +7,14 @@ import Toc from './Toc.js';
...
   children: Node,
   title: string,
+  isFullWidth?: boolean,
 |};
 
-export default function Page({ title: page, children }: Props): Node {
+export default function Page({ title: page, children, isFullWidth = false }: Props): Node {
   const sections = Children.toArray(children);
 
+  const DETAIL_PAGE_MAX_WIDTH = 1312;
```
dangerismycat, line 16: nittest of nits: these kind of consts are easier to find when at the top of a file, just below the imports

docs/components/buttons/SidebarCategorizationButton.js
```diff
@@ -10,23 +10,35 @@ type Props = {|
 |};
 
 export default function SidebarCategorizationButton({ onClick, sidebarOrganisedBy }: Props): Node {
-  const sidebarOrganisedByCopy =
-    sidebarOrganisedBy === 'categorized' ? 'Alphabetical' : 'Categorize';
+  const [selected, setSelected] = React.useState(sidebarOrganisedBy);
```
dangerismycat, line 13: Why is this local state necessary? It seems like it's duplicating what's being handled in the parent, and could possibly lead to a buggy UI from drift

docs/components/Page.js
dangerismycat: Perhaps the name should be more direct then, like `hideRightNav`?

docs/pages/home.js
```diff
@@ -0,0 +1,101 @@
...
+} from '../graphics/HomePageIcons.js';
+
+export default function HomePage(): Node {
+  return (
+    <Page title="Welcome to Gestalt" isFullWidth>
+      <Box width="100%">
+        <Flex direction="column">
+          {/* Hero */}
+          <HomePageSection>
+            <Box display="flex" alignItems="center" justifyContent="between">
```
dangerismycat, line 24: we clearly don't enable our lint rules for our own repo 😬 😂 

any reason why this isn't Flex?

dangerismycat (commented):
Possibly out of scope for this PR, but I find the small-screen experience a bit disorienting when the dropdown appears:
Screen Shot 2022-03-01 at 4 26 28 PM

Not sure if a background color or drop shadow would help? Def something we can chat about and fix in the future

docs/pages/home.js
ayeshakmaz: Yeah, I need both the minWidth and display none on the second item inside this container, so I needed to use Box to avoid the Flex.Item being added

dangerismycat: Ah, makes sense. Probably worth leaving a comment about so it doesn't get changed if we start running our lint rules on this codebase, or in other future refactors

docs/components/Page.js
```diff
@@ -4,12 +4,21 @@ import { Box, Flex, Link, Text } from 'gestalt';
...
   title: string,
+  showSideNav?: boolean,
+  showEditLink?: boolean,
 |};
 
-export default function Page({ title: page, children }: Props): Node {
+export default function Page({
+  title: page,
+  children,
+  showSideNav = true,
```
dangerismycat, line 19: Boolean props defaulting to `true` are a bit of an anti-pattern to me, as the API is then a bit awkward:
```js
<Component booleanProp={false} />
```
Consider inverting the logic to `hideSideNav` and `hideEditLink`, such that the default usage is just
```js
<Page />
```
and the special cases are simpler, like
```js
<Page hideSideNav hideEditLink />
```

The current boolean props read better in the contained markup, but the component API is more important IMO.

docs/pages/home.js
```diff
@@ -0,0 +1,100 @@
+// @flow strict
+import { Box, Button, Flex, Heading, Text } from 'gestalt';
+import { type Node } from 'react';
+import HomePageCard from '../components/HomePageCard.js';
+import HomePageSection from '../components/HomePageSection.js';
+import Page from '../components/Page.js';
+import Hero from '../graphics/HeroGraphic.js';
```
dangerismycat, line 7: Should we lazy-load this and the SVGs? Doesn't seem like a blocker to me for this PR, but a possible future improvement

dangerismycat (approved):

PR #1867, opened 2022-01-18
Box: Add new background colors

ayeshakmaz:
#### What changed?

Update Box to support new design token based colors. 

#### Why?

We're working on transitioning away from color names for Box backgrounds ("eggplant") and over to purpose-based design tokens (like "infoBase"). This PR adds the new options as possibilities and updates the documentation. A future PR + codemod will properly deprecate all the old values. 

### Links

- [Jira](https://jira.pinadmin.com/browse/GESTALT-2966)

### Checklist


docs/pages/box.js
```diff
@@ -419,26 +419,32 @@ function Example() {
             )}
           </CombinationNew>
         </MainSection.Subsection>
-        <MainSection.Subsection title="Colors">
+        <MainSection.Subsection
+          title="Colors"
+          description={`
+          The following colors can be used to change the background of Box.
+
+          ⚠️ Note that the previous options ('red', 'white', 'lightGray', 'gray', 'darkGray', 'green', 'pine', 'olive', 'blue', 'navy', 'midnight', 'purple', 'orchid', 'eggplant', 'maroon', 'watermelon', 'orange') are still valid but will be deprecated soon.
```
dangerismycat, line 427: What do you think about adding a section helping with this transition? A table or something that shows "prevColor --> newColor"

ayeshakmaz: We don't have this mapping yet (part of why I've split this into 2 PRs), but I'll definitely do that for the next PR to help show which are deprecated vs which have new color mappings

dangerismycat: nit: if it's easy, can we display "inverse" on a dark background?

dangerismycat (approved):
This looks fine — I imagine you're coordinating merging with a wider announcement about the change? How do you plan to address the temporary mismatch between the colors used in code and the ones we show examples of in the docs?

ayeshakmaz:
Definitely going to pair with an announcement. For the docs, I'm going to leave as is, and once we deprecate the other colors, I'll build an internal component for the Box doc examples

AlbertCarreras:
@ayeshakmaz Some of the new colors `infoBase`, `shopping`, `education` are the same color. Should we add any guidance on which color names to use when same color apply. I can see cases where `infoBase`, `shopping`, `education` are used interchangeably and future under the hood changes in the color values producing errors.

ayeshakmaz:
I can add a line about using these colors based on their intended purposes, given by the name. So only use `education` if you are showing an educational UI element for example

PR #1870, opened 2022-01-19
Video: Add required accessibilityProgressbarLabel prop

christianvuerings:
#### What changed?

* Adds a required `accessibilityProgressbarLabel` prop to video
* Fixes accessibility issue on video docs page

#### Why?

Each element with a `role="progressbar"` is required to have an accessible label.

#### Notes

* **This is a breaking change** since we add a required prop to an existing component.
* To fix the issue, you can add `accessibilityProgressbarLabel="Progress bar"` to `<Video />` elements
* Example of other video players setting this attribute: https://vimeo.com/38839299

### Links

- [Axe aria-progressbar-name rule](https://dequeuniversity.com/rules/axe/4.1/aria-progressbar-name?application=axeAPI)


packages/gestalt/src/Video.js
```diff
@@ -28,6 +28,7 @@ type Props = {|
   accessibilityMuteLabel: string,
   accessibilityPauseLabel: string,
   accessibilityPlayLabel: string,
+  accessibilityProgressbarLabel: string,
```
dangerismycat, line 31: nit: should this use "ProgressBar" since those are typically two separate words?

dangerismycat (approved):

packages/gestalt/src/Video.js
christianvuerings: Definitely - updates

PR #1878, opened 2022-01-25
ColorSchemeProvider: use autogenerated props table

dangerismycat:
and minor cleanup

### DIFF
--- a/docs/pages/colorschemeprovider.js
+++ b/docs/pages/colorschemeprovider.js
@@ -60,15 +60,6 @@ function Example(props) {
           />
         </MainSection.Subsection>
       </MainSection>
-
-      <MainSection name="Related">
-        <MainSection.Subsection
-          description={`
-      **[Link](/link)** / **[Button](/button)** / **[IconButton](/iconbutton)** / **[TapArea](/taparea)**  / **[DropDown](/dropdown)** / **[Callout](/callout)** / **[Upsell](/upsell)** / **[ActivationCard](/activationcard)**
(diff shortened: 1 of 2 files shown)


packages/gestalt/src/contexts/ColorSchemeProvider.js
```diff
@@ -129,20 +129,37 @@ const getTheme = (colorScheme: ?ColorScheme) =>
...
+|};
+
 /**
- * [ColorSchemeProvider](https://gestalt.pinterest.systems/colorschemeprovider) is an optional [React context provider](https://reactjs.org/docs/context.html#contextprovider) to enable dark mode
+ * [ColorSchemeProvider](https://gestalt.pinterest.systems/colorschemeprovider) is an optional [React context provider](https://reactjs.org/docs/context.html#contextprovider) to enable dark mode.
  */
 export default function ColorSchemeProvider({
   children,
-  colorScheme,
+  colorScheme = 'light',
```
dangerismycat, line 152: We've been effectively applying light mode as the default in the logic of `getTheme`. This change makes that default explicit.

We'd also previously been stating that light mode was the default in the docs — this change is necessary to continue providing that info with autogenerated docs.

docs/pages/colorschemeprovider.js
```diff
@@ -77,6 +60,7 @@ function Example(props) {
           />
         </MainSection.Subsection>
       </MainSection>
+
       <MainSection name="Related">
```
dangerismycat, line 64: @AlbertCarreras is this section still relevant? Is this holdover content from when the providers were combined into one?

AlbertCarreras: This is not longer relevant. Should be removed.

PR #1880, opened 2022-01-26
Tooltip: Deprecate links in Tooltips

ayeshakmaz:
#### What changed?

Removed the ability to put a link in a Tooltip

#### Why?

Links in Tooltips are inaccessible for keyboard and screen reader users, and therefore create usability issues

### Links

- [Jira](https://jira.pinadmin.com/browse/GESTALT-2506)

### Checklist

- [X] Removed unit and Flow Tests
- [X] Removed documentation + accessibility tests

### DIFF
--- a/docs/pages/tooltip.js
+++ b/docs/pages/tooltip.js
@@ -76,9 +76,9 @@ export default function DocsPage({ generatedDocGen }: {| generatedDocGen: DocGen
             type="do"
             description="Use Tooltip to describe the function of an interactive element, typically [Icon Button](/iconbutton), in as few words as possible."
             defaultCode={`
-<Tooltip text="Send Pin">
+<Tooltip text="Share Pin with others">
   <IconButton
-    accessibilityLabel=""
+    accessibilityLabel="Share"
     bgColor="white"
     icon="share"
     iconColor="darkGray"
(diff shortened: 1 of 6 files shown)


docs/pages/tooltip.js
```diff
@@ -221,9 +211,9 @@ When using Tooltip with [IconButton](/iconbutton), avoid repetitive labeling. Th
           <MainSection.Card
```
AlbertCarreras, line 211: @ayeshakmaz Should we review the guidance on line 193? 

`When using Tooltip with [IconButton](/iconbutton), avoid repetitive labeling. The \`accessibilityLabel\` provided to IconButton should describe the intent of the button, not the icon itself. For instance, use “Settings” instead of “Cog icon”. Tooltip \`text\` can expand upon that intention, as seen with the \`cog\` IconButton. If Tooltip \`text\` and IconButton \`accessibilityLabel\` contain the same content, pass an empty string to \`accessibilityLabel\`, as seen with the \`send\` IconButton.`}`

AlbertCarreras: @ayeshakmaz Is this redundancy an accepted practice now?
Screen Shot 2022-01-26 at 10 35 49 PM

ayeshakmaz: Ahh yes I need to update that. For now, the redundancy is expected

ayeshakmaz: Actually, I'll update the docs with how to avoid the redundancy now

dangerismycat:
@ayeshakmaz This PR is very old now and will require considerable conflict resolutions, so I'm closing for now. When we decide to pick this work back up we can either re-open this PR or create a new one (the latter may be easier).

PR #1882, opened 2022-01-27
IconButton, Tooltip: Update accessibility guidance

ayeshakmaz:
#### What changed?

Avoid duplicate, redundant tooltips on IconButton by allowing accessibilityLabel to be empty on Tooltip

#### Why?

Right now, we require an accessibilityLabel on IconButton, and when that gets wrapped in a tooltip, we create a redundant experience when the accessibilityLabel and Tooltip text match (simple iconButtons like share). This change recommends people override the tooltip accessibilityLabel to be an empty string, thereby allowing screen readers to only read out the content on the button itself. 

### Checklist

### DIFF
--- a/docs/pages/iconbutton.js
+++ b/docs/pages/iconbutton.js
@@ -522,7 +522,7 @@ function OrderDropdownExample() {
         <MainSection.Subsection
           title="ARIA attributes"
           description={`
-IconButton conveys the component behavior using iconography. IconButton requires \`accessibilityLabel\`, a text description for screen readers to announce and communicate the represented [Icon](/icon). In the example below, the screen reader reads: "More Options. The label should describe the intent of the action, not the Icon itself."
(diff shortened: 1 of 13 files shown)


docs/pages/iconbutton.js
```diff
@@ -522,7 +522,7 @@ function OrderDropdownExample() {
         <MainSection.Subsection
           title="ARIA attributes"
           description={`
-IconButton conveys the component behavior using iconography. IconButton requires \`accessibilityLabel\`, a text description for screen readers to announce and communicate the represented [Icon](/icon). In the example below, the screen reader reads: "More Options."
+IconButton conveys the component behavior using iconography. IconButton requires \`accessibilityLabel\`, a text description for screen readers to announce and communicate the represented [Icon](/icon). In the example below, the screen reader reads: "More Options. The label should describe the intent of the action, not the Icon itself."
```
dangerismycat, line 525: We might want to bold this text? I see "lock icon"-esque labels all the time

dangerismycat (approved):

PR #1885, opened 2022-01-27
ComboBox: updated Docs to use autogenerated prop table

AlbertCarreras:
#### What changed?

ComboBox: updated Docs to use autogenerated prop table

#### Why?

### DIFF
--- a/packages/gestalt/src/ComboBox.js
+++ b/packages/gestalt/src/ComboBox.js
@@ -19,7 +19,7 @@ import Popover from './Popover.js';
 import Text from './Text.js';
 import InternalTextField from './InternalTextField.js';
 import Tag from './Tag.js';
-import ComboBoxItem, { type ComboBoxItemType } from './ComboBoxItem.js';
+import ComboBoxItem from './ComboBoxItem.js';
 import { ESCAPE, TAB, ENTER, UP_ARROW, DOWN_ARROW } from './keyCodes.js';
 import handleContainerScrolling, {
   KEYS,
@@ -27,6 +27,11 @@ import handleContainerScrolling, {
 } from './utils/keyboardNavigation.js';
 
 type Size = 'md' | 'lg';
+type OptionType = {|
+  label: string,
+  subtext?: string,
+  value: string,
+|};
 
 type Props = {|
   /**
@@ -110,11 +115,7 @@ type Props = {|
    */
   onSelect?: ({|
     event: SyntheticInputEvent<HTMLElement> | SyntheticKeyboardEvent<HTMLElement>,
-    item: {|
-      label: string,
-      subtext?: string,
-      value: string,
-    |},
+    item: OptionType,
   |}) => void,
   /**
    * Specify a short description that suggests the expected input for the field.
@@ -128,11 +129,7 @@ type Props = {|
   /**
    * The selected option in ComboBox for controlled components. See [controlled ComboBox](https://gestalt.pinterest.systems/combobox#Controlled-vs-Uncontrolled) variant to learn more.
    */
-  selectedOption?: {|
-    label: string,
-    subtext?: string,
-    value: string,
-  |},
+  selectedOption?: OptionType,
   /**
    * Defines the height of ComboBox: md: 40px, lg: 48px. Width is defined by parent component.
    */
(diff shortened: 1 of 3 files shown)


packages/gestalt/src/ComboBox.js
```diff
@@ -11,59 +11,132 @@ import {
...
 } from './utils/keyboardNavigation.js';
 
 type Size = 'md' | 'lg';
+type OptionType = {|
+  label: string,
+  subtext?: string,
+  value: string,
+|};
 
 type Props = {|
```
dangerismycat, line 36: Do you mind alphabetizing these props while you're in here, please?

packages/gestalt/src/ComboBox.js
```diff
@@ -11,59 +11,132 @@ import {
...
+  |}>,
+  /**
+   * The text shown when the input value returns no matches.
+   */
   noResultText: string,
-  // OPTIONAL
+  /**
+   * When disabled, ComboBox looks inactive and cannot be interacted with. If tags are passed, they will appear disabled as well and cannot be removed. See [tags](https://gestalt.pinterest.systems/combobox#Tags) variant to learn more.
+   */
   disabled?: boolean,
```
dangerismycat, line 64: Default values aren't getting picked up. You'll need to move the props destructuring into the function signature for docgen to see the default values.

packages/gestalt/src/ComboBox.js
```diff
@@ -11,59 +11,132 @@ import {
...
   id: string,
+  /**
+   * Provide a label to identify the ComboBox field.
+   */
   label: string,
-  options: $ReadOnlyArray<ComboBoxItemType>,
+  /**
+   * The data for each selection option. See [subtext](https://gestalt.pinterest.systems/combobox#With-subtext) variant to learn more.
+   */
+  options: $ReadOnlyArray<{|
```
dangerismycat, line 52: It looks like this is the OptionType above? You can use that type here and docgen will show the correct type in the docs. That only breaks when importing types from other files or when types are heavily nested.

packages/gestalt/src/ComboBox.js
```diff
@@ -11,59 +11,132 @@ import {
...
   noResultText: string,
-  // OPTIONAL
+  /**
+   * When disabled, ComboBox looks inactive and cannot be interacted with. If tags are passed, they will appear disabled as well and cannot be removed. See [tags](https://gestalt.pinterest.systems/combobox#Tags) variant to learn more.
+   */
   disabled?: boolean,
+  /**
+   * Provide feedback when an error on selection occurs. See [error message](https://gestalt.pinterest.systems/combobox#Error-message) variant.
+   */
   errorMessage?: Node,
```
dangerismycat, line 68: This is different from the currently-displayed type (string). Do we actually want to restrict this to a string? Is there need for the flexibility of a Node?

packages/gestalt/src/ComboBox.js
```diff
@@ -11,59 +11,132 @@ import {
...
+   */
   onBlur?: ({|
     event: SyntheticFocusEvent<HTMLInputElement> | SyntheticEvent<HTMLInputElement>,
     value: string,
   |}) => void,
+  /**
+   * Callback when user types into the control input field.
+   */
   onChange?: ({|
     value: string,
```
dangerismycat, line 92: Can this be shifted to `{ event, value }` to match the order of the other hander types?

packages/gestalt/src/ComboBox.js
AlbertCarreras: DONE

packages/gestalt/src/ComboBox.js
AlbertCarreras: It doesn't! It only picks direct usages. As soon as you do Array etc if only shows the name! It took me a while to figure out.

packages/gestalt/src/ComboBox.js
AlbertCarreras: This follows most errorMessage types. I'll leave as is, we can decide touching the API later.

packages/gestalt/src/ComboBox.js
AlbertCarreras: DONE

packages/gestalt/src/ComboBox.js
AlbertCarreras: image

packages/gestalt/src/ComboBox.js
dangerismycat: Ah, makes sense. Thanks for adding them!

packages/gestalt/src/ComboBox.js
dangerismycat: Dang, I thought it took one more level of depth for docgen to lose the type. Womp womp.

packages/gestalt/src/ComboBox.js
```diff
@@ -11,59 +11,136 @@ import {
...
+  /**
+   * Specify a short description that suggests the expected input for the field.
+   */
   placeholder?: string,
-  selectedOption?: ComboBoxItemType,
+  // The ref prop is unused and listed here just for documentation purposes.
+  /**
+   * Forward the ref to the underlying component container element. See the [Ref](https://gestalt.pinterest.systems/combobox#Ref) variant to learn more about focus management.
+   */
+  ref?: Ref<'input'>, // eslint-disable-line react/no-unused-prop-types
```
dangerismycat, line 132: I hadn't noted this previously as I'd commented about it on another PR, but can you add logic to display this as `React.Ref<'input'>`?

dangerismycat (approved):

packages/gestalt/src/ComboBox.js
AlbertCarreras: DONE

PR #1890, opened 2022-01-31
[NumberField]: Remove type="number" and create codemod

EduhCosta:
### Summary

To finish this task: https://jira.pinadmin.com/browse/GESTALT-3345 is necessary remove the type "number" of `TextField` component and create a codemod to validate this change.

#### What changed?

Remove the 'number' of types of TextField component and create a Codmod file to alert/throw erros to next uses of this prop.

#### Why?

This is necessary because we create the NumberField component and today this component is responsible for handled the number's input fields;

### Links

- [Jira](https://jira.pinadmin.com/browse/GESTALT-3345)


packages/gestalt-codemods/43.0.0/remove-textFields-number-prop.js
```diff
@@ -0,0 +1,69 @@
...
+      // Not type prop, bail
+      if (propName !== 'type') {
+        return;
+      }
+
+      const propValue = attr?.value?.value;
+
+      if (propValue === 'number') {
+        throw new Error(
+          'Please, prefer use the `NumberField` component rather than use the `type="number"` prop. ' +
```
AlbertCarreras, line 59: Please, replace TextField using deprecated  `type="number"` with [NumberField](https://gestalt.pinterest.systems/numberfield). Location: ${file.path} @line: ${node.loc.start.line}

packages/gestalt/src/TextField.js
```diff
@@ -79,7 +79,7 @@ type Props = {|
   /**
    * The type of input. For non-telephone numerical input, please use [NumberField](https://gestalt.pinterest.systems/numberfield).
    */
-  type?: 'date' | 'email' | 'number' | 'password' | 'tel' | 'text' | 'url',
+  type?: 'date' | 'email' | 'password' | 'tel' | 'text' | 'url',
```
AlbertCarreras, line 82: Remove comment above. Doesn't apply anymore.

packages/gestalt/src/TextField.js
```diff
@@ -77,9 +77,9 @@ type Props = {|
    */
   tags?: $ReadOnlyArray<Element<typeof Tag>>,
   /**
-   * The type of input. For non-telephone numerical input, please use [NumberField](https://gestalt.pinterest.systems/numberfield).
```
dangerismycat, line 80: This comment should remain. We need to guide users to the NumberField component.

packages/gestalt/src/TextField.js
dangerismycat: The comment absolutely still applies — we want to guide users to NumberField.

AlbertCarreras: Then we are doing a Related section inside Prop tables. Here I'd update Textfield to link to the Variants section and add a comment there.

EduhCosta: Done

AlbertCarreras (approved):

PR #1891, opened 2022-01-31
DatePicker: updated Docs to use autogenerated prop table

AlbertCarreras:
#### What changed?

DatePicker: updated Docs to use autogenerated prop table

#### Why?


packages/gestalt-datepicker/src/DatePicker.js
```diff
@@ -11,27 +17,84 @@ import dateFormat from './dateFormat.js';
...
+  /**
+   * Disable dates outside a min date. See the [delimited selection period example](https://gestalt.pinterest.systems/datepicker#maxMinDates) to learn more.
+   */
   minDate?: Date,
-  // $FlowFixMe[unclear-type]
-  nextRef?: ElementRef<*>,
+  /**
+   * Required for date range selection. Pass the complimentary range date picker ref object to DatePicker to autofocus on the unselected date range field. See the [date range picker example](https://gestalt.pinterest.systems/datepicker#rangePicker) to learn more.
+   */
+  nextRef?: Ref<'input'>,
```
dangerismycat, line 67: Can you add logic around `docs/components/GeneratedPropTable.js:100` so that this displays as `React.Ref`?

packages/gestalt-datepicker/src/DatePicker.js
```diff
@@ -110,6 +173,8 @@ const DatePickerWithForwardRef: React$AbstractComponent<Props, HTMLDivElement> =
       (rangeSelector === 'end' && !rangeStartDate)
     ) {
       if (nextRef && submitted) {
+        // $FlowFixMe[prop-missing]
```
dangerismycat, line 176: Hm…are you sure the new type for `nextRef` is correct? Can we avoid these suppressions?

packages/gestalt-datepicker/src/DatePicker.js
```diff
@@ -11,27 +17,84 @@ import dateFormat from './dateFormat.js';
...
+  nextRef?: Ref<'input'>,
+  /**
+   *
+   */
   onChange: ({|
     event: SyntheticInputEvent<HTMLInputElement>,
     value: Date,
   |}) => void,
+  /**
+   * Date format for locale.
```
dangerismycat, line 76: This doesn't seem correct? This seems like a common use for the placeholder text, but is the text itself not more generic?

packages/gestalt-datepicker/src/DatePicker.js
```diff
@@ -11,27 +17,84 @@ import dateFormat from './dateFormat.js';
 import { type LocaleData } from './LocaleDataTypes.js';
 
 type Props = {|
+  /**
+   *
```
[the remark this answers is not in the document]
AlbertCarreras: DONE

packages/gestalt-datepicker/src/DatePicker.js
AlbertCarreras: FIXED.

packages/gestalt-datepicker/src/DatePicker.js
AlbertCarreras: FIXED

packages/gestalt-datepicker/src/DatePicker.js
```diff
@@ -1,37 +1,133 @@
...
+  options?: {|
+    weekStartsOn?: 0 | 1 | 2 | 3 | 4 | 5 | 6,
+    firstWeekContainsDate?: 1 | 2 | 3 | 4 | 5 | 6 | 7,
+  |},
+|};
+// flowlint unclear-type:error
 
 type Props = {|
+  /**
+   *  See the [disabled example](https://gestalt.pinterest.systems/datepicker#disabled) to learn more.
```
dangerismycat, line 54: Can you add actual descriptions for the props that are missing them? More than just "see example". I've been using these refactors as an excuse to make sure that every prop has a description.

packages/gestalt-datepicker/src/DatePicker.js
```diff
@@ -11,27 +17,84 @@ import dateFormat from './dateFormat.js';
...
+   */
   includeDates?: $ReadOnlyArray<Date>,
+  /**
+   *
+   */
   label?: string,
+  /**
+   * DatePicker accepts imported locales from the open source date utility library date-fns. See the [locales example](https://gestalt.pinterest.systems/datepicker#localeData) to learn more.
+   */
   localeData?: LocaleData,
```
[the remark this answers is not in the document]
dangerismycat: oof, agreed — but it is at least the actual type, so users won't have to search the source code

packages/gestalt-datepicker/src/DatePicker.js
```diff
@@ -1,37 +1,127 @@
...
 import classnames from 'classnames';
 import { Icon, Box, Label, Text } from 'gestalt';
 import DatePickerTextField from './DatePickerTextField.js';
 import styles from './DatePicker.css';
 import dateFormat from './dateFormat.js';
-import { type LocaleData } from './LocaleDataTypes.js';
+
+// LocaleData type from https://github.com/date-fns/date-fns/blob/81ab18785146405ca2ae28710cdfbb13a294ec50/src/locale/af/index.js.flow
+// flowlint unclear-type:off
+export type LocaleData = {|
```
dangerismycat, line 12: Does this type need to be exported?

packages/gestalt-datepicker/src/DatePicker.js
```diff
@@ -1,37 +1,127 @@
...
+   */
   excludeDates?: $ReadOnlyArray<Date>,
+  /**
+   * More information about how to complete the DatePicker field. See the [helper text example](https://gestalt.pinterest.systems/datepicker#helperText) to learn more.
+   */
   helperText?: string,
+  /**
+   *
+   */
   id: string,
```
dangerismycat, line 65: This prop still needs a description.

dangerismycat (approved):

packages/gestalt-datepicker/src/DatePicker.js
AlbertCarreras: DONE

packages/gestalt-datepicker/src/DatePicker.js
AlbertCarreras: DONE

PR #1892, opened 2022-01-31
Dropdown: updated Docs to use autogenerated prop table

AlbertCarreras:
#### What changed?

Dropdown: updated Docs to use autogenerated prop table

Added a second `docgen` function (`multipledocgen`) that allows retrieving multiple components' proptables in the same docs page to allow more flexibility for designers to redesign and improve subcomponent discoverability
#### Why?

### DIFF
--- a/docs/components/docgen.js
+++ b/docs/components/docgen.js
@@ -51,6 +51,37 @@ export default async function docgen({
   return parsed;
 }
 
+export async function multipledocgen({
+  componentName,
+  alternativeSubdirectory = '',
+}: {|
+  componentName: Array<string> | string,
+  alternativeSubdirectory?: string,
+|}): Promise<Array<DocGen>> {
+  const parsedPropTables = await Promise.all(
+    (Array.isArray(componentName) ? componentName : [componentName]).map(async (cmp) => {
+      const filePath = path.join(
+        nextConfig().serverRuntimeConfig.GESTALT_ROOT,
+        alternativeSubdirectory || `/packages/gestalt/src/${cmp}.js`,
+      );
+      const contents = await fs.promises.readFile(filePath, 'utf-8');
+      const parsed = parse(contents);
+
+      if (parsed.description) {
+        parsed.description = parsed.description
+          // Remove the first markdown link from the description so we don't link to the page itself
+          .replace(/\[(.*?)\][[(].*?[\])]/, '$1')
+          // Remove images from the description
+          .replace(/!\[(.*?)\][[(].*?[\])]/g, '');
+      }
+
+      return parsed;
+    }),
+  );
+
+  return parsedPropTables;
+}
+
 export function overrideTypes(docGen: DocGen, typeOverrides: {| [string]: string |}): DocGen {
   Object.keys(typeOverrides).forEach((key) => {
     if (docGen?.props?.[key]) {
--- a/docs/components/docgen.js
+++ b/docs/components/docgen.js
@@ -26,18 +26,13 @@ export type DocGen = {|
   |},
 |};
 
-export default async function docgen({
-  componentName,
-  alternativeSubdirectory = '',
-}: {|
-  componentName: string,
-  alternativeSubdirectory?: string,
-|}): Promise<DocGen> {
-  const filePath = path.join(
+const getFilePath = (componentName: string, alternativeSubdirectory?: string): string =>
+  path.join(
     nextConfig().serverRuntimeConfig.GESTALT_ROOT,
     alternativeSubdirectory || `/packages/gestalt/src/${componentName}.js`,
   );
-  const contents = await fs.promises.readFile(filePath, 'utf-8');
+
+const getParsedContent = (contents): DocGen => {
   const parsed = parse(contents);
 
   if (parsed.description) {
@@ -47,8 +42,19 @@ export default async function docgen({
       // Remove images from the description
       .replace(/!\[(.*?)\][[(].*?[\])]/g, '');
   }
-
   return parsed;
+};
+
+export default async function docgen({
+  componentName,
+  alternativeSubdirectory = '',
+}: {|
+  componentName: string,
+  alternativeSubdirectory?: string,
+|}): Promise<DocGen> {
+  const filePath = getFilePath(componentName, alternativeSubdirectory);
+  const contents = await fs.promises.readFile(filePath, 'utf-8');
+  return getParsedContent(contents);
 }
 
 export async function multipledocgen({
@@ -60,22 +66,9 @@ export async function multipledocgen({
 |}): Promise<Array<DocGen>> {
   const parsedPropTables = await Promise.all(
     (Array.isArray(componentName) ? componentName : [componentName]).map(async (cmp) => {
-      const filePath = path.join(
-        nextConfig().serverRuntimeConfig.GESTALT_ROOT,
-        alternativeSubdirectory || `/packages/gestalt/src/${cmp}.js`,
-      );
+      const filePath = getFilePath(cmp, alternativeSubdirectory);
       const contents = await fs.promises.readFile(filePath, 'utf-8');
-      const parsed = parse(contents);
-
-      if (parsed.description) {
-        parsed.description = parsed.description
-          // Remove the first markdown link from the description so we don't link to the page itself
-          .replace(/\[(.*?)\][[(].*?[\])]/, '$1')
-          // Remove images from the description
-          .replace(/!\[(.*?)\][[(].*?[\])]/g, '');
-      }
-
-      return parsed;
+      return getParsedContent(contents);
     }),
   );
 
--- a/docs/components/docgen.js
+++ b/docs/components/docgen.js
@@ -63,16 +63,18 @@ export async function multipledocgen({
 }: {|
   componentName: Array<string> | string,
   alternativeSubdirectory?: string,
-|}): Promise<Array<DocGen>> {
-  const parsedPropTables = await Promise.all(
+|}): Promise<{| [string]: DocGen |}> {
+  const docGenMap = {};
+  await Promise.all(
     (Array.isArray(componentName) ? componentName : [componentName]).map(async (cmp) => {
       const filePath = getFilePath(cmp, alternativeSubdirectory);
       const contents = await fs.promises.readFile(filePath, 'utf-8');
-      return getParsedContent(contents);
+      docGenMap[cmp] = getParsedContent(contents);
+      return cmp;
     }),
   );
 
-  return parsedPropTables;
+  return docGenMap;
 }
 
 export function overrideTypes(docGen: DocGen, typeOverrides: {| [string]: string |}): DocGen {
--- a/docs/pages/dropdown.js
+++ b/docs/pages/dropdown.js
@@ -10,13 +10,13 @@ import GeneratedPropTable from '../components/GeneratedPropTable.js';
 export default function DropdownPage({
   generatedDocGen,
 }: {|
-  generatedDocGen: Array<DocGen>,
+  generatedDocGen: {| [string]: DocGen |},
 |}): Node {
   return (
     <Page title="Dropdown">
       <PageHeader
         name="Dropdown"
-        description={generatedDocGen[0]?.description}
+        description={generatedDocGen.Dropdown?.description}
         badge="pilot"
         defaultCode={`
       function IntroMenuButtonDropdownExample() {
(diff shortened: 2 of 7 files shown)


docs/components/docgen.js
```diff
@@ -50,6 +50,37 @@ export default async function docgen({
...
+export async function multipledocgen({
+  componentName,
+  alternativeSubdirectory = '',
+}: {|
+  componentName: Array<string> | string,
+  alternativeSubdirectory?: string,
+|}): Promise<Array<DocGen>> {
+  const parsedPropTables = await Promise.all(
+    (Array.isArray(componentName) ? componentName : [componentName]).map(async (cmp) => {
+      const filePath = path.join(
```
dangerismycat, line 62: Instead of repeating this logic, can any parts of this be extracted into shared helper functions?

docs/pages/dropdown.js
```diff
@@ -1,44 +1,22 @@
...
+  generatedDocGen,
+}: {|
+  generatedDocGen: Array<DocGen>,
+|}): Node {
   return (
     <Page title="Dropdown">
       <PageHeader
         name="Dropdown"
-        description={generatedDocGen?.description}
+        description={generatedDocGen[0]?.description}
```
dangerismycat, line 19: Using array indices is pretty brittle. Instead of an array, can we use an object/map to allow for access by component name?

packages/gestalt/src/Dropdown.js
```diff
@@ -83,16 +83,38 @@ const renderChildrenWithIndex = (childrenArray) => {
...
-
 type Props = {|
+  /**
+   * Ref for the element that the Dropdown will attach to, will most likely be a [Button](/button). See the [Accessibility](https://gestalt.netlify.app/dropdown#Accessibility) guidelines to learn more.
+   */
   anchor?: ?HTMLElement,
+  /**
+   * Must be instances of Dropdown.Item, Dropdown.Link or Dropdown.Section components. See the [Types of items](https://gestalt.netlify.app/dropdown#Types-of-items) variant to learn more.
+   */
   children: Node,
```
dangerismycat, line 94: This type is presented quite differently currently. Though `Node` is more accurate, we may want to keep displaying the previous type for now, until we officially support using container components to organize items.

packages/gestalt/src/Dropdown.js
```diff
@@ -83,16 +83,38 @@ const renderChildrenWithIndex = (childrenArray) => {
   }, []);
 };
 
-type IdealDirection = 'up' | 'right' | 'down' | 'left';
-
 type Props = {|
+  /**
+   * Ref for the element that the Dropdown will attach to, will most likely be a [Button](/button). See the [Accessibility](https://gestalt.netlify.app/dropdown#Accessibility) guidelines to learn more.
+   */
   anchor?: ?HTMLElement,
```
dangerismycat, line 90: This `?` is getting lost for some reason. Do you know why?

packages/gestalt/src/Dropdown.js
```diff
@@ -83,16 +83,38 @@ const renderChildrenWithIndex = (childrenArray) => {
...
+  /**
+   * Preferred direction for the Dropdown to open.
+   */
+  idealDirection?: 'up' | 'right' | 'down' | 'left',
+  /**
+   * Callback fired when the menu is closed.
+   */
   onDismiss: () => void,
+  /**
+   * An object representing the zIndex value of the Dropdown menu. Learn more about [zIndex classes](https://gestalt.netlify.app/zindex_classes)'
```
dangerismycat, line 116: typo: there's a stray apostrophe at the end of the line

packages/gestalt/src/DropdownItem.js
```diff
@@ -1,31 +1,63 @@
...
+      |}
+    | $ReadOnlyArray<{|
+        label: string,
+        subtext?: string,
+        value: string,
+      |}>
+    | null,
+  /**
+   * Private prop used for accessibility purposes
+   */ index?: number,
```
dangerismycat, line 56: We previously did not document this prop. We should either filter it out manually, or (preferred) rename it to `_index` to indicate that it's a private prop, and so that it's filtered out automatically.

packages/gestalt/src/DropdownItem.js
```diff
@@ -1,31 +1,63 @@
...
 |};
 
 /**
- * https://gestalt.pinterest.systems/dropdown
+ * Subcomponent of [Dropdown](https://gestalt.pinterest.systems/dropdown)
```
dangerismycat, line 60: Can we add a more verbose description here, and  can that be displayed just above the props table in the docs?

packages/gestalt/src/DropdownLink.js
```diff
@@ -1,31 +1,55 @@
...
     SyntheticMouseEvent<HTMLAnchorElement> | SyntheticKeyboardEvent<HTMLAnchorElement>,
     {| dangerouslyDisableOnNavigation: () => void |},
   >,
+  /**
+   * Object detailing the label, value, and optional subtext for this item.
+   */
   option: OptionItemType,
+  /**
+   * Private prop used for accessibility purposes
+   */ index?: number,
```
dangerismycat, line 47: same here, this also needs to be filtered out

packages/gestalt/src/DropdownLink.js
```diff
@@ -1,31 +1,55 @@
...
-  index?: number,
-|};
-
-type Props = {|
-  ...PublicProps,
-  ...PrivateProps,
-|};
-
+/**
+ * Subcomponent of [Dropdown](https://gestalt.pinterest.systems/dropdown)
```
dangerismycat, line 51: Ditto here on a more verbose description

packages/gestalt/src/DropdownLink.js
```diff
@@ -1,31 +1,55 @@
...
+   */
   href: string,
+  /**
+   * When true, adds an arrow icon to the end of the item to signal this item takes users to an external source and opens the link in a new tab. Do not add if the item navigates users within the app. See the [Best practices](https://gestalt.pinterest.systems/dropdown#Best-practices) for more info.
+   */
   isExternal?: boolean,
+  /**
+   * Callback fired when clicked (pressed and released) with a mouse or keyboard. See [OnLinkNavigationProvider](https://gestalt.pinterest.systems/onlinknavigationprovider) to learn more about link navigation.
+   */
   onClick?: AbstractEventHandler<
```
dangerismycat, line 37: I've been manually copying over the type of AbstractEventHandler, since that abstraction means nothing to the docs reader. It's a pretty easy change, and IMHO makes the type a bit more readable as well.

packages/gestalt/src/DropdownSection.js
```diff
@@ -5,12 +5,18 @@ import Text from './Text.js';
...
   children: Node,
+  /**
+   * Label for the section. See the [Sections](https://gestalt.pinterest.systems/dropdown#Sections) variant for more info.
+   */
   label: string,
 |};
 
 /**
- * https://gestalt.pinterest.systems/dropdown
+ * Subcomponent of [Dropdown](https://gestalt.pinterest.systems/dropdown)
```
dangerismycat, line 19: ditto here

packages/gestalt/src/DropdownSection.js
```diff
@@ -5,12 +5,18 @@ import Text from './Text.js';
 import styles from './Dropdown.css';
 
 type Props = {|
+  /**
+   * Any [Dropdown.Items](https://gestalt.pinterest.systems/dropdown#Dropdown.ItemProps) and/or [Dropdown.Links](https://gestalt.pinterest.systems/dropdown#Dropdown.LinkProps) to be rendered
+   */
   children: Node,
```
dangerismycat, line 11: ditto here on displaying the more restrictive type until we officially support more flexibility

dangerismycat (commented):
Thanks for addressing sub-component props tables! That gives us more flexibility for our future docs redesign.

packages/gestalt/src/Dropdown.js
AlbertCarreras: I'll get fixed when merging Popover changes.

packages/gestalt/src/Dropdown.js
AlbertCarreras: DONE

packages/gestalt/src/Dropdown.js
AlbertCarreras: DONE

packages/gestalt/src/DropdownSection.js
AlbertCarreras: DONE

packages/gestalt/src/DropdownLink.js
AlbertCarreras: DONE

packages/gestalt/src/DropdownLink.js
AlbertCarreras: DONE

packages/gestalt/src/DropdownSection.js
AlbertCarreras: DONE

packages/gestalt/src/DropdownLink.js
AlbertCarreras: DONE

docs/pages/dropdown.js
dangerismycat: It looks like the component name is available…after awaiting the Promise.all, it looks like you could use reduce to create a map keyed on the component name?

docs/components/docgen.js
```diff
@@ -46,8 +41,37 @@ export default async function docgen({
...
+}: {|
+  componentName: string,
+  alternativeSubdirectory?: string,
+|}): Promise<DocGen> {
+  const filePath = getFilePath(componentName, alternativeSubdirectory);
+  const contents = await fs.promises.readFile(filePath, 'utf-8');
+  return getParsedContent(contents);
+}
+
+export async function multipledocgen({
```
dangerismycat, line 59: It looks like these two functions could be combined into one? The only difference seems to be if `componentName` is a string or array of strings, and if the return value is a DocGen promise or an array of DocGen promises. The logic of `multipleDocGen` looks like it could work for single or multiple components?

docs/pages/dropdown.js
```diff
@@ -99,64 +77,7 @@ export default function DropdownPage({ generatedDocGen }: {| generatedDocGen: Do
...
-          },
-          {
-            name: 'zIndex',
-            type: 'interface Indexable { index(): number; }',
-            description:
-              'An object representing the zIndex value of the Dropdown menu. Learn more about [zIndex classes](/zindex_classes)',
-          },
-        ]}
-      />
+      <GeneratedPropTable generatedDocGen={generatedDocGen[0]} excludeProps={['index']} />
```
dangerismycat, line 80: Since `index` is an internal-only prop (rather than a repeated prop, like margins and paddings on Box), it'd be better to refactor to use `_index` instead, matching our convention. Then the prop would be automatically filtered out without needing to use `excludeProps`.

AlbertCarreras: I'll handle both internal props onKeyDow and index in follow up PR.

docs/components/docgen.js
AlbertCarreras: I tried to have a single function but had to change API for function and had issues with Flowtyping it. So kept it simple. Also, we don't know how long we are going to use these, as it seems they might end up going into different pages,

dangerismycat: That's a good point!

packages/gestalt/src/DropdownItem.js
AlbertCarreras: Same, follow up PR

docs/components/docgen.js
AlbertCarreras: Refactored it to a map. I played  around with reduce but Promise.all  neews iterable  so keep  the map method and  returned a map rather than  the array from the Promise.all.

docs/pages/dropdown.js
AlbertCarreras: Refactored it to a map, however Promise.all must return iterable and reducer method didn't seem to work  well  with Promise.all.

docs/components/docgen.js
dangerismycat: This works too. So much nicer to access the component info by name. Nice work!

docs/pages/dropdown.js
```diff
@@ -625,77 +546,23 @@ function TruncationDropdownExample() {
         </MainSection.Subsection>
       </MainSection>
       <MainSection name="Subcomponents" />
-      <PropTable
+      <GeneratedPropTable
```
dangerismycat, line 549: I'm still seeing `index` in the tables for Item and Link…looks like that needs to be added to `excludeProps` here

AlbertCarreras: DONE

dangerismycat (approved):

PR #1895, opened 2022-01-31
Popover: updated Docs to use autogenerated prop table

AlbertCarreras:
#### What changed?

Popover: updated Docs to use autogenerated prop table

#### Why?

### DIFF
--- a/docs/components/GeneratedPropTable.js
+++ b/docs/components/GeneratedPropTable.js
@@ -110,6 +110,7 @@ export default function GeneratedPropTable({
         description: descriptionWithoutTypeOverride?.trim(),
         required,
         defaultValue: defaultValueOverride ?? defaultValue?.value?.replace(/'/g, ''),
+        nullable: flowType?.nullable,
         href,
       };
     })
--- a/docs/components/GeneratedPropTable.js
+++ b/docs/components/GeneratedPropTable.js
@@ -110,7 +110,7 @@ export default function GeneratedPropTable({
         description: descriptionWithoutTypeOverride?.trim(),
         required,
         defaultValue: defaultValueOverride ?? defaultValue?.value?.replace(/'/g, ''),
-        nullable: flowType?.nullable,
+        nullable: flowType?.nullable || false,
         href,
       };
     })
--- a/packages/gestalt/src/Popover.js
+++ b/packages/gestalt/src/Popover.js
@@ -1,19 +1,19 @@
 // @flow strict
 import type { Node } from 'react';
 import Controller from './Controller.js';
-import { type Role } from './Contents.js';
 
 type Color = 'blue' | 'orange' | 'red' | 'white' | 'darkGray';
 type Size = 'xs' | 'sm' | 'md' | 'lg' | 'xl' | 'flexible' | number;
 type IdealDirection = 'up' | 'right' | 'down' | 'left';
+type Role = 'dialog' | 'listbox' | 'menu';
 
 type Props = {|
   /**
-   * The reference element, typically [Button](https://gestalt.pinterest.systems/button) or [IconButton](https://gestalt.pinterest.systems/iconbutton), that Popover uses to set its position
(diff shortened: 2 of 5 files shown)


packages/gestalt/src/Popover.js
```diff
@@ -1,24 +1,60 @@
...
-import { type Role } from './Contents.js';
 
 type Color = 'blue' | 'orange' | 'red' | 'white' | 'darkGray';
 type Size = 'xs' | 'sm' | 'md' | 'lg' | 'xl' | 'flexible' | number;
 type IdealDirection = 'up' | 'right' | 'down' | 'left';
+type Role = 'dialog' | 'listbox' | 'menu';
 
 type Props = {|
+  /**
+   * The reference element, typically [Button](https://gestalt.pinterest.systems/button) or [IconButton](https://gestalt.pinterest.systems/iconbutton), that Popover uses to set its position
```
dangerismycat, line 12: nit: prop descriptions should all end with periods

packages/gestalt/src/Popover.js
```diff
@@ -1,24 +1,60 @@
...
+   */
   anchor: ?HTMLElement,
```
dangerismycat, line 14: This is appearing in the preview without the leading `?`…any idea why?

packages/gestalt/src/Popover.js
```diff
@@ -1,24 +1,60 @@
...
+   */
   children?: Node,
+  /**
+   * The background color of Popover. See the [color and caret](https://gestalt.pinterest.systems/popover#Color-and-caret) variant to learn more.
+   */
   color?: Color,
+  /**
+   *
+   */
   handleKeyDown?: (event: SyntheticKeyboardEvent<HTMLElement>) => void,
```
dangerismycat, line 26: Huh, I never noticed this doesn't conform to our event standard. We should probably update that at some point

packages/gestalt/src/Popover.js
```diff
@@ -1,24 +1,60 @@
...
+   */
   role?: Role,
+  /**
+   * Puts the focus on Popover when it’s triggered. See [accessibility](https://gestalt.pinterest.systems/popover#Accessibility) to learn more.
+   */
   shouldFocus?: boolean,
+  /**
+   * Shows a caret on Popover. See the [color and caret](https://gestalt.pinterest.systems/popover#Color-and-caret) variant to learn more.
+   */
   showCaret?: boolean,
```
dangerismycat, line 54: Some of the default values aren't being shown in the table, as with this prop. You need to move the `props` destructuring into the function signature for docgen to pick up on it.

packages/gestalt/src/Popover.js
AlbertCarreras: DONE

packages/gestalt/src/Popover.js
AlbertCarreras: Good catch, the property "nullable" in Flowtype was not covered.

packages/gestalt/src/Popover.js
AlbertCarreras: I think it was added as an internal prop, but agree, we should refactor it

packages/gestalt/src/Popover.js
AlbertCarreras: DONE

docs/components/GeneratedPropTable.js
```diff
@@ -110,6 +110,7 @@ export default function GeneratedPropTable({
         description: descriptionWithoutTypeOverride?.trim(),
         required,
         defaultValue: defaultValueOverride ?? defaultValue?.value?.replace(/'/g, ''),
+        nullable: flowType?.nullable,
```
dangerismycat, line 113: nice find!

packages/gestalt/src/Popover.js
dangerismycat: I _think_ we're not using this anywhere in Pinboard, so perhaps we should add a `_` prefix to make it officially internal?

dangerismycat (approved):
Thanks for this work!

PR #1899, opened 2022-02-02
Badge: fix layout issues

dangerismycat:
Badge currently behaves in funky ways in Flex layouts. Unless `gap` is set, Badge will expand to fill vertical space, and the `position` prop becomes useless. This PR fixes that behavior by setting a specific height on Badge.

Test code:
```js
<Flex>
  <Text>Some text that uses Badge component</Text>
  <Badge text="New" />
</Flex>
```
  

_Before, no gap_
Screen Shot 2022-02-02 at 3 40 32 PM

_Before, with gap_
Screen Shot 2022-02-02 at 3 40 46 PM

_After, no gap_
Screen Shot 2022-02-02 at 3 43 20 PM

### DIFF
--- a/packages/gestalt/src/Badge.css
+++ b/packages/gestalt/src/Badge.css
@@ -5,7 +5,7 @@
   composes: fontWeightBold from './Typography.css';
   composes: white from './Colors.css';
   font-size: 10px;
-  height: fit-content;
+  height: min-content;
   margin-top: -4px;
   padding: 2px;
   white-space: nowrap;
--- a/packages/gestalt/src/Badge.css
+++ b/packages/gestalt/src/Badge.css
@@ -5,6 +5,7 @@
   composes: fontWeightBold from './Typography.css';
   composes: white from './Colors.css';
   font-size: 10px;
+  height: fit-content;
   margin-top: -4px;
   padding: 2px;
   white-space: nowrap;
--- a/packages/gestalt/src/Badge.js
+++ b/packages/gestalt/src/Badge.js
@@ -28,5 +28,5 @@ export default function Badge({ position = 'middle', text }: Props): Node {
     [colors.darkGray]: colorSchemeName === 'darkMode',
   });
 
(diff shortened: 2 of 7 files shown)


packages/gestalt/src/Badge.css
```diff
@@ -5,6 +5,7 @@
   composes: fontWeightBold from './Typography.css';
   composes: white from './Colors.css';
   font-size: 10px;
+  height: fit-content;
```
dangerismycat, line 8: This is the fix.

packages/gestalt/src/Badge.js
```diff
@@ -28,5 +28,5 @@ export default function Badge({ position = 'middle', text }: Props): Node {
     [colors.darkGray]: colorSchemeName === 'darkMode',
   });
 
-  return <span className={cs}>{text}</span>;
+  return <div className={cs}>{text}</div>;
```
dangerismycat, line 31: This change doesn't seem to be necessary given the `inline-block` style set on the span, but in general I find reasoning about divs more intuitive than spans. Can revert this part if there's a good reason to keep this as a span.

packages/gestalt/src/Badge.css
christianvuerings: @dangerismycat can we avoid `height: fit-content`? Firefox doesn't support it yet. https://caniuse.com/mdn-css_properties_height_fit-content

Instead would `align-self` be an option? e.g. `align-self: center;`

dangerismycat: Good catch! 😅 

`align-self` seems odd in that this isn't actually about the alignment of the component within the flex layout, but the its height. Ideally we get the element to keep its normal height in a flex layout but still allow the Flex to align it. How about `height= min-content`? [That seems to have much better support](https://caniuse.com/?search=height%3A%20min-content):
Screen Shot 2022-02-03 at 5 23 52 PM

PR #1900, opened 2022-02-03
Dropdown: prefixed private props (index) with underscore

AlbertCarreras:
#### What changed?

index private prop in Dropdown.Item and Dropdown.Link are prefixed with underscore ('_') following new pattern

#### Why?

Following new pattern to name private props

### DIFF
--- a/docs/pages/dropdown.js
+++ b/docs/pages/dropdown.js
@@ -551,14 +551,12 @@ function TruncationDropdownExample() {
         name="Dropdown.Item"
         id="Dropdown.Item"
         generatedDocGen={generatedDocGen.DropdownItem}
-        excludeProps={['_index']}
       />
       <GeneratedPropTable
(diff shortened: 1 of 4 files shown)


docs/pages/dropdown.js
```diff
@@ -551,14 +551,14 @@ function TruncationDropdownExample() {
         name="Dropdown.Item"
         id="Dropdown.Item"
         generatedDocGen={generatedDocGen.DropdownItem}
-        excludeProps={['index']}
+        excludeProps={['_index']}
```
dangerismycat, line 554: You shouldn't need to manually exclude these — we filter out any props with an underscore prefix

dangerismycat (approved):

docs/pages/dropdown.js
AlbertCarreras: DONE

PR #1906, opened 2022-02-03
Internal: upgrade axe-core

dangerismycat:


dangerismycat:
Abandoning this bump for now. We should keep the version in sync with Pinboard's version to prevent mismatches.

PR #1908, opened 2022-02-03
Heading, Text: Make all 6 font sizes available + codemod

ayeshakmaz:
#### What changed?

Update Heading and Text to support font sizes 100 - 600 from our design tokens

| Size | Text | TextArea | px |
|------|------|----------|----|
| 100  | sm   |     --     | 12 |
| 200  | md   |    --      | 14 |
| 300  | lg   |     --     | 16 |
| 400  |   --   | sm       | 20 |
| 500  |   --   | md       | 28 |
| 600  |  --    | lg       | 36 |

#### Why?

In order to improve accessibility and the flexibility of our system, both Heading and Text are now available in all 6 sizes. This prevents people from using Heading when they just need large Text, or using Text when they really need a small Heading, like in mobile scenarios 

### Links

- [Jira](https://jira.pinadmin.com/browse/GESTALT-3512)

### Checklist


packages/gestalt/src/Text.js
```diff
@@ -10,14 +10,8 @@ function isNotNullish(val): boolean {
...
 
-const SIZE_SCALE = {
-  sm: 100,
-  md: 200,
-  lg: 300,
-};
-
 type Overflow = 'normal' | 'breakWord' | 'noWrap';
-type Size = 'sm' | 'md' | 'lg';
+type Size = '100' | '200' | '300' | '400' | '500' | '600';
```
ayeshakmaz, line 14: This is the main change

packages/gestalt/src/Heading.js
```diff
@@ -11,21 +11,17 @@ function isNotNullish(val): boolean {
...
+  '300': 4,
+  '400': 3,
+  '500': 2,
+  '600': 1,
 };
 
 type AccessibilityLevel = 1 | 2 | 3 | 4 | 5 | 6 | 'none';
 type Overflow = 'normal' | 'breakWord';
-type Size = 'sm' | 'md' | 'lg';
+type Size = '100' | '200' | '300' | '400' | '500' | '600';
```
ayeshakmaz, line 24: Along with this, the other main change

AlbertCarreras:
Can we add table of equivalencies?

packages/gestalt-codemods/45.0.0/__testfixtures__/heading.input.js
```diff
@@ -0,0 +1,12 @@
+// @flow strict
+import {Box, Heading} from 'gestalt';
```
AlbertCarreras, line 2: Can you do a test for renaming the cmp

Heading as RenamedHeading?

cypress/integration/accessibility_Heading_spec.js
```diff
@@ -5,6 +5,15 @@ describe('Heading Accessibility check', () => {
   });
 
   it('Tests accessibility on the Heading page', () => {
+    // Heading docs showcase all sizes, breaking the order
```
dangerismycat, line 8: Rather than disabling the rule entirely, can we add test ids to the incorrect sections and exclude those? (What I was trying on that other PR and failed at — maybe you can get it to work here?)

docs/pages/text.js
```diff
@@ -92,34 +92,62 @@ export default function TextPage({ generatedDocGen }: {| generatedDocGen: DocGen
       />
       <Example
         description={`
-    You can apply \`size\` options to define the size of the text.
+    You can apply \`size\` options to define the size of the text. These font sizes follow those available through our [Design Tokens](/design_tokens#Font-size).
```
dangerismycat, line 95: Do we need to provide any additional guidance here about when to use various sizes, and Text vs Heading?

docs/pages/heading.js
```diff
@@ -16,25 +16,41 @@ export default function DocsPage({ generatedDocGen }: {| generatedDocGen: DocGen
       <Example
         id="sizes"
         name="Example: Sizes"
+        description="These font sizes follow those available through our [Design Tokens](/design_tokens#Font-size)."
```
dangerismycat, line 19: Also here on additional guidance on sizes and Text vs Heading

packages/gestalt-codemods/45.0.0/replace-typography-sizes.js
```diff
@@ -0,0 +1,83 @@
...
+
+      const attrs = node.openingElement.attributes;
+
+      if (attrs.some((attr) => attr.type === 'JSXSpreadAttribute')) {
+        throw new Error(
+          `Remove Dynamic Heading and Text properties and rerun codemod. Location: ${file.path} @line: ${node.loc.start.line}`,
+        );
+      }
+
+      node.openingElement.attributes = attrs
```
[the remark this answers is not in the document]
dangerismycat: You can also handle those cases within the codemod, which is relatively straightforward if literal values are used rather than variables. Check out the Box->Flex lint rule for an example, or also play around with astexplorer to see what ternaries are parsed to

packages/gestalt/src/Heading.css
```diff
@@ -4,19 +4,7 @@
...
-
-.fontSize500 {
-  font-size: var(--font-size-500);
-}
-
-.fontSize600 {
-  font-size: var(--font-size-600);
-}
-
 .TextLikeHeadingSm {
```
dangerismycat, line 7: Is this class still needed?

dangerismycat (commented):
For your own sanity, you may want to split these changes out to two PRs. That will make the Pinboard bumps a bit easier.

Also: it feels like we're not providing enough additional guidance around this change. We previously had wide-open Text/Heading sizes like this, which we restricted to the current scheme because people were misusing them. These changes throw those doors open again — how are we guiding designers and engineers to use this new power correctly?

cypress/integration/accessibility_Heading_spec.js
dangerismycat: Now that we've proven disabling sections of the page, perhaps we can use that here rather than disabling the rule?

docs/pages/text.js
```diff
@@ -92,34 +92,62 @@ export default function TextPage({ generatedDocGen }: {| generatedDocGen: DocGen
       />
       <Example
         description={`
-    You can apply \`size\` options to define the size of the text.
+    You can apply \`size\` options to define the size of the text. These font sizes follow those available through our [Design Tokens](/design_tokens#Font-size). If your text needs to be a [semantic heading (H1-H6)](/https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Heading_Elements), use [Heading](/heading) instead.
```
dangerismycat, line 95: Future improvement: expand on what text would need to be a semantic heading

packages/gestalt-codemods/45.0.0/replace-typography-sizes.js
```diff
@@ -0,0 +1,116 @@
...
+        );
+      }
+
+      const isHeadingComponent = node.openingElement.name.name.includes('Heading');
+
+      node.openingElement.attributes = attrs
+        .map((attr) => {
+          if (attr?.name?.name === 'size' && attr.value.type === 'JSXExpressionContainer') {
+            if (
+              attr.value.expression.type === 'ConditionalExpression' &&
```
dangerismycat, line 74: How does this handle something like `size={condition ? 'sm' : foo}`? You don't necessarily need to codemod `foo`, but you do need to be alerted to it

cypress/integration/accessibility_Heading_spec.js
ayeshakmaz: I think in this case it still makes sense to disable the rule for the whole page, the same way we do for PageHeader. Any example on the Heading page will break the heading order, so rather than disable it for every single example, current and future, I think it's a better developer experience to know that's being taken care of at the page level, and disabling it has no adverse affect on the accessibility of our product since heading-order isn't something we can control

ayeshakmaz: Plus when we disable using the class name, we are disabling every accessibility check. By disabling just the heading-order test on the page, the rest of our accessibility tests will run on each example (contrast, link names, etc.)

dangerismycat: Good points! 👍

packages/gestalt-codemods/46.0.0/replace-typography-sizes.js
```diff
@@ -0,0 +1,120 @@
...
+          `Remove Dynamic Heading and Text properties and rerun codemod. Location: ${file.path} @line: ${node.loc.start.line}`,
+        );
+      }
+
+      const isHeadingComponent = node.openingElement.name.name.includes('Heading');
+
+      node.openingElement.attributes = attrs
+        .map((attr) => {
+          if (attr?.name?.name === 'size' && attr.value.type === 'JSXExpressionContainer') {
+            if (
```
dangerismycat, line 73: It's difficult to tell in the GH UI, but I _think_ this will still miss something like `size={condition ? 'sm' : foo}`. The way I'm reading this is that the old value will be transformed, but there will be no alert about the `foo` variable, which will almost certainly point to old values. Can you add test cases for this specific scenario to make sure the codemod works as expected?

packages/gestalt-codemods/45.0.0/replace-typography-sizes.js
```diff
@@ -0,0 +1,116 @@
+/*
```
dangerismycat, line 1: It looks like there are now two codemods here?

dangerismycat (commented):
Since the changes are no longer breaking, this should no longer be a major version change. It won't be a major version change until the old values are removed.

Given that changes are no longer breaking, this PR would be a lot easier to review if the internal value changes were separate from the Heading/Text changes. This would also make it easier to see that there are currently two codemods in the PR.

PR #1909, opened 2022-02-03
Popover: rename handleOnKeyDown to onKeyDown and standardize event type 

AlbertCarreras:
#### What changed?

Popover's `handleOnKeyDown` prop is renamed to `onKeyDown` and the event type is standardized from `(event)` to `({| event |})`

#### Why?

Dropdown implemented `handleOnKeyDown` to support keyboard navigation within Dropdown. ComboBox followed the same implementation. `handleOnKeyDown` has stayed undocumented, and moving forward there were 2 options: keep it private prefixing underscore (`_handleOnKeyDown`) or documenting it.

Popover is a building component and there's a clear use case for building custom components and needing `onKeyDown` to allow keyboard navigation within Popover's children.

NOTE: This is a breaking change as we are changing API, but prop was private so there are no uses in Pinboard and it's safe to merge.
Screen Shot 2022-02-03 at 7 05 19 PM

### Checklist
- [X ] Added documentation + accessibility tests

### DIFF
--- a/packages/gestalt/src/Controller.js
+++ b/packages/gestalt/src/Controller.js
@@ -22,7 +22,7 @@ type OwnProps = {|
   border?: boolean,
   caret?: boolean,
   children?: ReactNode,
-  handleKeyDown?: (event: SyntheticKeyboardEvent<HTMLElement>) => void,
+  onKeyDown?: ({| event: SyntheticKeyboardEvent<HTMLElement> |}) => void,
   id?: ?string,
   idealDirection?: 'up' | 'right' | 'down' | 'left',
   onDismiss: () => void,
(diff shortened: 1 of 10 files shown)


packages/gestalt/src/Controller.js
```diff
@@ -81,11 +81,11 @@ class Controller extends Component<Props, State> {
   }
 
   handleKeyDown: (event: SyntheticKeyboardEvent<HTMLElement>) => void = (event) => {
-    const { handleKeyDown, onDismiss } = this.props;
+    const { onKeyDown, onDismiss } = this.props;
     if (event.keyCode === ESCAPE) {
       onDismiss();
     }
-    if (handleKeyDown) handleKeyDown(event);
+    if (onKeyDown) onKeyDown({ event });
```
dangerismycat, line 88: nit: `onKeyDown?.({ event });`

packages/gestalt/src/Popover.js
```diff
@@ -21,9 +21,9 @@ type Props = {|
    */
   color?: Color,
   /**
-   *
+   * Callback for key stroke events allowing keyboard navigation in Popover's childrens.
```
dangerismycat, line 24: nit: children is already plural, no need for the "s" suffix

dangerismycat (approved):

packages/gestalt/src/Popover.js
AlbertCarreras: DONE

packages/gestalt/src/Controller.js
AlbertCarreras: DONE

PR #1911, opened 2022-02-04
Button: refactor to fix Dropdown positioning shift on rerender

AlbertCarreras:
#### What changed?

Fixed bug due to Dropdown's positioning shift on rerender.

Solution: moving compression and style to a new direct div children of the button parent tag.  This allows the button element to not compress on click and the ref passed to Dropdown's Popover to stay constant.

Discarded solutions: 

- add timeout to render Dropdown with a delay equivalent to the ease-out compression transition in Button. This prevents Dropdown to use the anchor's position based on the initial Button position on compression. Requires leaking info from Button to Dropdown to testing.

- add event listeners to keyup, keydown, transitionstart, and transitionend inside Dropdown that control the component's rendering. All of them had some sort of incompatibility with mouse/keyboard navigation. For example, transitionstart and transitionend would work to detect the compression but keyboard navigation doesn't compress Button. Therefore, keyup, keydown must be also used. However, with keydown the Dropdown doesn't open as the event happens before the Dropdown gets mounted. With keyup, however, the Button's  focus stays in Button if we keep the Enter key  pressed continuously as the focus doesn't move to the Dropdown creating an infinite toggling loop in the Button state.

The most straightforward and clear solution was to implement timeout, solution that works for all cases.

Updated tests with useFakeTimers to advance timers.

#### Why?

Bug:  https://jira.pinadmin.com/browse/BUG-121306

Steps to reproduce:
1. Visit Dropdown in Gestalt: https://gestalt.netlify.app/Dropdown
2.  Open the example and hover over an option
3. View the slight jump

https://www.loom.com/share/ea724aa7420a47a098fc693c36e16ef0

The root cause is within Dropdown, not Popover. ComboBox doesn’t have this issue for instance. What’s happening is that Button is the anchor for Popover and Button compresses, setting the anchor position on compression. On hover on the second option item, Dropdown rerenders and the anchor has a slightly different uncompressed position.


packages/gestalt/src/Button.test.js
```diff
@@ -23,12 +23,14 @@ describe('<Button />', () => {
   });
   test('Custom white text color on transparent background', () => {
     const instance = create(<Button color="transparentWhiteText" text="Hello World" />).root;
-    expect(instance.find((element) => element.type === 'div').props.className).toContain('white');
+    expect(instance.findAll((element) => element.type === 'div')[1].props.className).toContain(
```
dangerismycat, line 26: I'm not sure what a better solution is, but using the array index here seems a bit brittle

dangerismycat (approved):

packages/gestalt/src/Button.test.js
AlbertCarreras: I looked into https://reactjs.org/docs/test-renderer.html#testinstancefindbytype and couldn't find a better alternative to detect the Text cmp

PR #1912, opened 2022-02-04
Video: use autogen props table

dangerismycat:
Note that a number of formerly "required" props are no longer… @christianvuerings does that seem correct to you?

### DIFF
--- a/docs/pages/video.js
+++ b/docs/pages/video.js
@@ -1,282 +1,19 @@
 // @flow strict
 import type { Node } from 'react';
-import PropTable from '../components/PropTable.js';
 import Example from '../components/Example.js';
-import PageHeader from '../components/PageHeader.js';
+import GeneratedPropTable from '../components/GeneratedPropTable.js';
 import Page from '../components/Page.js';
+import PageHeader from '../components/PageHeader.js';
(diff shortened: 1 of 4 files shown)


docs/pages/video.js
```diff
@@ -477,3 +214,11 @@ function Example() {
     </Page>
   );
 }
+
+export async function getStaticProps(): Promise<{| props: {| generatedDocGen: DocGen |} |}> {
+  const generatedDocGen = await docgen({ componentName: 'Video' });
+
+  return {
+    props: { generatedDocGen: deepCloneReplacingUndefined(generatedDocGen) },
```
AlbertCarreras, line 222: Why are we using this one here? `deepCloneReplacingUndefined`

AlbertCarreras (approved):
LGTM

docs/pages/video.js
dangerismycat: It's a problem with (most) class components: the way some methods are parsed yields `undefined` values, which aren't valid JSON. So then when that tree is serialized, it throws. I'm not sure if we'll have this problem with any other components, but it's necessary for Masonry and Video (Image seems fine).

PR #1928, opened 2022-02-08
Link: deprecate private "disabled" prop + codemod

AlbertCarreras:
#### What changed?

Link: deprecate unused disabled prop + codemod

#### Why?

reference PR https://github.com/pinterest/gestalt/pull/1169

This prop was old code pre InternalLink cmp when Button role='link" used Link cmp and disabled state was needed. After moving to InternalLink, this code became unused.

### DIFF
--- a/packages/gestalt-codemods/46.0.0/link_deprecate_disabled_prop.js
+++ b/packages/gestalt-codemods/46.0.0/link_deprecate_disabled_prop.js
@@ -1,0 +1,72 @@
+/*
+ * Converts
+ *   <Link disabled />
+ * To
+ *   <Link />
+ */
+
+// Run
+// yarn codemod --parser=flow -t=packages/gestalt-codemods/46.0.0/link_deprecate_disabled_prop.js relative/path/to/your/code
+
+export default function transformer(file, api) {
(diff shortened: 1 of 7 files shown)


packages/gestalt-codemods/46.0.0/link_deprecate_disabled_prop.js
```diff
@@ -0,0 +1,72 @@
...
+
+      if (!localIdentifierName.includes(node.openingElement.name.name)) {
+        return null;
+      }
+
+      const attrs = node.openingElement.attributes;
+
+      if (attrs.some((attr) => attr.type === 'JSXSpreadAttribute')) {
+        throw new Error(
+          `Remove Dynamic Text properties and rerun codemod. Location: ${file.path} @line: ${node.loc.start.line}`,
```
dangerismycat, line 46: s/Text/Link

dangerismycat (approved):

packages/gestalt-codemods/46.0.0/link_deprecate_disabled_prop.js
AlbertCarreras: DONE

PR #1931, opened 2022-02-09
Internal: update eslint-config-airbnb to 19.0.4, fix/suppress new errors

dangerismycat:
This PR updates [`eslint-config-airbnb`](https://github.com/airbnb/javascript) to 19.0.4. This is a major version change which added/updated a number of lint rules yielding new errors on our existing codebase. This PR also fixes or suppresses all new errors, most of which fall into these buckets (others are commented about inline on the PR or in the code):

[`function-component-definition`](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/function-component-definition.md)
This may be the most controversial change. This enforces that all named functions are function declarations, and that unnamed functions are function expressions (as arrow functions). This isn't actually new guidance, just Airbnb finally getting their lint rule in line with older guidance. I'm personally a big fan of this rule but I know not everyone will be! 😅 (This also autofixes which is a total lifesaver)

[`no-unstable-nested-components`](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-unstable-nested-components.md)
This may be another controversial change. Basically any function defined within a function component must be memoized, typically with useCallback. Alternatively, the internal function can be extracted, which is great where it's valuable to be specific about its dependencies.

[`prefer-regex-literals`](https://eslint.org/docs/rules/prefer-regex-literals)
There are two different ways of defining regexes in JS, and the simpler one is preferred. I updated some of our regexes where they were either dead-simple to read or to test, and suppressed others. We should fix those in the future.


docs/pages/_document.js
```diff
@@ -44,7 +44,7 @@ gtag('config', 'UA-12967896-44');
 `,
             }}
           />
-          <link rel="shortcut icon" href="/pinterest_favicon.png" />
+          <link rel="icon" href="/pinterest_favicon.png" />
```
dangerismycat, line 47: "shortcut icon" isn't a valid value, but "icon" is

docs/pages/iconography_and_svgs.js
```diff
@@ -38,29 +51,16 @@ export default function IconPage(): Node {
...
-        }}
-      >
-        <Box padding={2}>
-          {/* $FlowFixMe[prop-missing] */}
-          <Icon color="darkGray" accessibilityLabel={iconName} icon={iconName} />
-        </Box>
-      </TapArea>
-    </Tooltip>
-  );
+  const buildHandleIconClick = (iconName: string) => () => {
```
dangerismycat, line 54: This was inlined in ClickableIcon's onTap prop

packages/eslint-plugin-gestalt/src/index.js
```diff
@@ -1,4 +1,5 @@
 // @flow strict
+/* eslint-disable import/no-import-module-exports */
```
dangerismycat, line 2: We should probably fix this in the future:
https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-import-module-exports.md

packages/gestalt-datepicker/rollup.config.js
```diff
@@ -1,5 +1,5 @@
-// eslint-disable-next-line flowtype/require-valid-file-annotation
-import plugins from '../gestalt-core/build.js'; // eslint-disable-line import/no-relative-parent-imports
+// eslint-disable-next-line flowtype/require-valid-file-annotation, import/no-relative-parent-imports, import/no-relative-packages
```
dangerismycat, line 1: Not sure if this is something we actually need to fix:
https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-relative-packages.md

packages/gestalt/src/AnimationController.js
```diff
@@ -150,12 +158,14 @@ function AnimationController({
     }
   }, [animationState, onDismissEnd]);
 
+  const contextValue = useMemo(() => ({ animationState, setAnimationState }), [animationState]);
```
dangerismycat, line 161: Basically the same as the rule about unstable functions — this makes sure the context value is stable

packages/gestalt/src/utils/keyboardNavigation.js
```diff
@@ -36,7 +36,9 @@ const handleContainerScrolling = ({
 
   const containerHeight = container.getClientRects()[0].height;
   const overScroll =
-    nextSelectedOption instanceof HTMLElement && nextSelectedOption?.offsetHeight / 3;
+    nextSelectedOption instanceof HTMLElement && nextSelectedOption?.offsetHeight
```
dangerismycat, line 39: Didn't like potentially doing arithmetic on `undefined`, which would yield `NaN`

docs/components/MainSectionCard.js
```diff
@@ -63,20 +63,23 @@ const MainSectionCard = ({
...
-      display="flex"
-      height={CARD_SIZE_NAME_TO_PIXEL[cardSize]}
-      justifyContent="center"
-      padding={8}
-      position="relative"
-      rounding={2}
-    >
-      {cardChildren}
-    </Box>
+  const PreviewCard = useCallback(
```
christianvuerings, line 66: @dangerismycat wouldn't it be even better to move `PreviewCard` out of `MainSectionCard`?

docs/pages/masonry.js
```diff
@@ -74,6 +75,21 @@ const getPins = () => {
   return Promise.resolve(pinList);
 };
 
+function gridComponent({ data }: { data: Pin, ... }) {
```
christianvuerings, line 78: @dangerismycat should we make this an actual React component? `function GridComponent`?

packages/eslint-plugin-gestalt/src/index.js
christianvuerings: @dangerismycat I think it might be ok for this file since it is the source of the `main` module: https://github.com/pinterest/gestalt/blob/master/packages/eslint-plugin-gestalt/package.json#L7
> Reports the use of import declarations with CommonJS exports in any module except for the [main module](https://docs.npmjs.com/files/package.json#main).

christianvuerings (approved):

docs/components/MainSectionCard.js
dangerismycat: I started with that initially, but came back to this to see how it felt. There is value in defining a function within the scope of another function to take advantage of the parent scope without needing to pass in everything, especially to avoid needing to repeat type definitions. Something like this really doesn't feel like another component, but just a quick helper. For anything that felt like a separate component, I extracted the function.

I'd like to enable this rule in Pinboard someday, so I wanted to explore this pattern. It feels ok to me — retaining the parent scope convenience while also making sure the function is stable.

docs/pages/masonry.js
dangerismycat: Yes, good call. Given typical Masonry usage in Pinboard, I was thinking `comp` needed to be a simple function and would be called as a function internally, but that's wrong — it's rendered as a component. Updating now

PR #1936, opened 2022-02-11
Heading, Text: Update internal usages to use new size values

ayeshakmaz:
#### What changed?

For any instances of Heading or Text inside our components, switch out size values as follows:

| Size | Text | TextArea | px |
|------|------|----------|----|
| 100  | sm   |     --     | 12 |
| 200  | md   |    --      | 14 |
| 300  | lg   |     --     | 16 |
| 400  |   --   | sm       | 20 |
| 500  |   --   | md       | 28 |
| 600  |  --    | lg       | 36 |

#### Why?

In order to deprecate our old values, we need to convert any current instances over. Also helps us eliminate the use of accessibilityLevel="none" on Heading (which will be deprecated in a future ticket)

### Links

- [Jira](https://jira.pinadmin.com/browse/GESTALT-3889)

### Checklist

- [X] Updated unit and Flow Tests

### DIFF
--- a/packages/gestalt/src/Heading.flowtest.js
+++ b/packages/gestalt/src/Heading.flowtest.js
@@ -3,7 +3,7 @@ import Heading from './Heading.js';
 
 const Valid = <Heading size="400">Heading</Heading>;
 
-const ValidccessibilityLevel1 = (
+const ValidAccessibilityLevel1 = (
   <Heading size="400" accessibilityLevel={1}>
     Heading
   </Heading>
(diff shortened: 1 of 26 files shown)


packages/gestalt/src/Heading.flowtest.js
```diff
@@ -3,7 +3,7 @@ import Heading from './Heading.js';
 
 const Valid = <Heading size="400">Heading</Heading>;
 
-const ValidccessibilityLevel1 = (
+const ValidAccessibilityLevel1 = (
```
ayeshakmaz, line 6: Typo fix

AlbertCarreras (approved):

PR #1939, opened 2022-02-14
DeviceTypeProvider: add new component

dangerismycat:
#### What changed?

This PR introduces a new component, DeviceTypeProvider. The associated docs page is skeletal for now, as this component is an alpha that may change. The docs page can be navigated to directly but has not yet been added to the sidebar index, so it is somewhat hidden for now. I will flesh out the rest of the docs page before this component officially launches.

#### Why?

We'd like to start offering device-specific UI for certain components, e.g. migrating Pinboard's MobileModal into Gestalt and rendering that UI for mobile devices. Using the same `'desktop' | 'mobile' | 'tablet'` device types as Pinboard, this context provider is the first step in this project, making device type available within components. Future work will add logic within components to use this device type.

I'm imagining this will be used something like
```js
// App.js

return (
  <DeviceTypeProvider value={useDeviceType()}>
    {children}
  </DeviceTypeProvider>
);
```
```js
// Gestalt Modal.js

const deviceType = useDeviceType();

return deviceType === 'mobile' ? <MobileModal /> : <Modal />;
```

### Links

- [Jira](https://jira.pinadmin.com/browse/GESTALT-3477)

### Checklist

- [ ] ~Added unit and Flow Tests~ Don't seem necessary for a context this simple

### DIFF
--- a/packages/gestalt/src/contexts/DeviceTypeProvider.js
+++ b/packages/gestalt/src/contexts/DeviceTypeProvider.js
@@ -1,0 +1,31 @@
+// @flow strict
+import { type Context, type Node, createContext, useContext } from 'react';
+
+const defaultDeviceType = 'desktop';
+
+type DeviceType = 'desktop' | 'phone' | 'tablet';
+
+const DeviceTypeContext: Context<DeviceType> = createContext<DeviceType>(defaultDeviceType);
+
+type Props = {|
+  /**
+   *
+   */
+  children: Node,
+  /**
+   * The device type as determined by logic within your app. This will be used within certain Gestalt components to render device-specific UI.
+   */
+  deviceType: DeviceType,
+|};
+
+/**
+ * *ALPHA - DO NOT USE YET - MAY HAVE BREAKING CHANGES IN THE NEAR FUTURE*
+ * [DeviceTypeProvider](https://gestalt.pinterest.systems/devicetypeprovider) is an optional [React Context provider](https://reactjs.org/docs/context.html#contextprovider) to enable device-specific UI for Gestalt components that support it.
+ */
+export default function DeviceTypeProvider({ children, deviceType }: Props): Node {
+  return <DeviceTypeContext.Provider value={deviceType}>{children}</DeviceTypeContext.Provider>;
+}
+
+export function useDeviceType(): DeviceType {
+  return useContext(DeviceTypeContext) ?? defaultDeviceType;
+}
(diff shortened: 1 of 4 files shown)


packages/gestalt/src/contexts/DeviceTypeProvider.js
```diff
@@ -0,0 +1,31 @@
+// @flow strict
+import { type Context, type Node, createContext, useContext } from 'react';
+
+const defaultDeviceType = 'desktop';
+
+type DeviceType = 'desktop' | 'phone' | 'tablet';
```
christianvuerings, line 6: Currently we show/hide content through breakpoints (e.g. `smDisplay` / `mdDisplay` and `lgDisplay`) instead of device type. It would be good to clarify when we should use what.

dangerismycat:
That's a really good point. This is less about showing/hiding content and
more about switching completely from one UI to another. That's definitely
something we should cover in the guidance when this is released officially.

packages/gestalt/src/contexts/DeviceTypeProvider.js
```diff
@@ -0,0 +1,31 @@
...
+
+/**
+ * *ALPHA - DO NOT USE YET - MAY HAVE BREAKING CHANGES IN THE NEAR FUTURE*
+ * [DeviceTypeProvider](https://gestalt.pinterest.systems/devicetypeprovider) is an optional [React Context provider](https://reactjs.org/docs/context.html#contextprovider) to enable device-specific UI for Gestalt components that support it.
+ */
+export default function DeviceTypeProvider({ children, deviceType }: Props): Node {
+  return <DeviceTypeContext.Provider value={deviceType}>{children}</DeviceTypeContext.Provider>;
+}
+
+export function useDeviceType(): DeviceType {
```
liuyenwei, line 29: is this hook something we intend to expose as a public API or should it only be used internally within gestalt?

liuyenwei: nm, ignore me.  only the Provider is exported below, not the hook 👍

dangerismycat: Yep, I was planning on the hook only being necessary internally. We can export it in the future if for some reason it's needed.

AlbertCarreras (approved):
LGTM. For the future, I'd consider merging with ColorSchemeProvider into a ConfigProvider

PR #1940, opened 2022-02-15
Docs: Updated Status docs to include design guidelines

ponori:
#### What changed?

Added Best Practices and Related section and updated Accessibility section of Status docs.

### DIFF
--- a/docs/pages/status.js
+++ b/docs/pages/status.js
@@ -87,7 +87,7 @@ export default function SearchFieldPage({ generatedDocGen }: {| generatedDocGen:
           <MainSection.Card
             cardSize="md"
             type="do"
-            description="Use a title when the status it represents is unique, specific and critical for the user to know."
+            description="Use `title` when the status it represents is unique, specific and critical for the user to know."
             defaultCode={`
 <Status type="inProgress" title="Pending review" />
 `}
@@ -95,7 +95,7 @@ export default function SearchFieldPage({ generatedDocGen }: {| generatedDocGen:
           <MainSection.Card
             cardSize="md"
             type="don't"
-            description="Use Status' sub text to display extraneous messaging."
+            description="Use Status' `subText` to display extraneous messaging."
             defaultCode={`
 <Status type="problem" title="Error" subtext="Please try again" />
 `}
--- a/docs/pages/status.js
+++ b/docs/pages/status.js
@@ -64,7 +64,7 @@ export default function SearchFieldPage({ generatedDocGen }: {| generatedDocGen:
           <MainSection.Card
             cardSize="md"
             type="do"
-            description="Place Status close to the status subject in question to provide context and reference. It can be placed as an inline element or paired side by side as needed.  "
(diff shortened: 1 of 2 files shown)


docs/pages/status.js
```diff
@@ -43,7 +43,65 @@ export default function SearchFieldPage({ generatedDocGen }: {| generatedDocGen:
...
+            type="don't"
+            description="Use SVGs or images that resemble the Status’ symbols to denote status."
+            defaultCode={`
+<Icon icon="workflow-status-problem" size="24" accessibilityLabel="This item has an error" />
+`}
+          />
+          <MainSection.Card
+            cardSize="md"
+            type="do"
+            description="Place Status close to the status subject in question to provide context and reference. It can be placed as an inline element or paired side by side as needed.  "
```
dangerismycat, line 67: kinda wordy…instead of "to the status subject in question", how about "to its subject"?

docs/pages/status.js
```diff
@@ -43,7 +43,65 @@ export default function SearchFieldPage({ generatedDocGen }: {| generatedDocGen:
...
+  <Status accessibilityLabel="This item is paused" type="halted" />
+  <Text weight="bold" size="300">Campaign paused</Text>
+</Flex>
 
+`}
+          />
+          <MainSection.Card
+            cardSize="md"
+            type="do"
+            description="Use a title when the status it represents is unique, specific and critical for the user to know."
```
dangerismycat, line 90: props should be denoted with backticks and don't need an article, so
> Use `title` when the status it represents…

docs/pages/status.js
```diff
@@ -43,7 +43,65 @@ export default function SearchFieldPage({ generatedDocGen }: {| generatedDocGen:
...
+            defaultCode={`
+<Status type="inProgress" title="Pending review" />
+`}
+          />
+          <MainSection.Card
+            cardSize="md"
+            type="don't"
+            description="Use Status' sub text to display extraneous messaging."
```
dangerismycat, line 98: ditto here: `subtext` refers to the prop, so needs backticks and should match the prop name exactly

dangerismycat (approved):

docs/pages/status.js
ponori: There was some initial feedback that utilizing prop names may not necessarily be as clear to designers, so we decided to pull is back to something less technical.

docs/pages/status.js
ponori: Same comment for this one. I think there's solid arguments on both sides here and I don't know if I have an informed opinion of how much friction this would cause designers.

docs/pages/status.js
ponori: 👍

docs/pages/status.js
ponori: Done!

docs/pages/status.js
ponori: Done!

PR #1956, opened 2022-02-24
Internal: upgrade postcss and cssnano

dangerismycat:
Attempting to address https://github.com/pinterest/gestalt/security/dependabot/29

Though our version of postcss isn't vulnerable, several dependency versions are. This PR upgrade cssnano and all our postcss dependencies to minimize our exposure to the vulnerability.


dangerismycat:
this isn't a drop-in replacement and we're probably fine, so abandoning for now

PR #1958, opened 2022-03-01
Masonry: add a11y roles

dangerismycat:
Masonry is a list of items, but we don't present it as such semantically. That's not great for accessibility.

As best I can tell, there are three rendering scenarios for Masonry. In two of them, we render some form of a container with items. In the remaining one we render an empty container. For the first two, this PR adds `role="list"` to the main container, and `role="listitem"` to the individual Pin items. No changes were made to the empty container div, since (I think) it doesn't need a role.

### DIFF
--- a/packages/gestalt/src/Masonry.js
+++ b/packages/gestalt/src/Masonry.js
@@ -15,6 +15,8 @@ import fullWidthLayout from './fullWidthLayout.js';
 import LegacyMasonryLayout from './layouts/MasonryLayout.js';
 import LegacyUniformRowLayout from './layouts/UniformRowLayout.js';
 
+type Position = {| top: number, left: number, width: number, height: number |};
+
 type Layout =
   | typeof DefaultLayoutSymbol
   | typeof UniformRowLayoutSymbol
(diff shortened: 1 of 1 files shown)


packages/gestalt/src/Masonry.js
```diff
@@ -382,8 +384,7 @@ export default class Masonry<T: { ... }> extends ReactComponent<Props<T>, State<
     this.forceUpdate();
   }
 
-  // $FlowFixMe[unclear-type]
-  renderMasonryComponent: (itemData: T, idx: number, position: *) => Node = (
+  renderMasonryComponent: (itemData: T, idx: number, position: Position) => Node = (
```
dangerismycat, line 387: unrelated Flow cleanup

packages/gestalt/src/Masonry.js
```diff
@@ -541,7 +545,7 @@ export default class Masonry<T: { ... }> extends ReactComponent<Props<T>, State<
         ? Math.max(...positions.map((pos) => pos.top + pos.height))
         : 0;
       gridBody = (
-        <div style={{ width: '100%' }} ref={this.setGridWrapperRef}>
+        <div style={{ width: '100%' }} ref={this.setGridWrapperRef} role="list">
           <div className={styles.Masonry} style={{ height, width }}>
```
ayeshakmaz, line 549: I believe the role="list" needs to be at this level. With the screenreader, I'm hearing "List 1 item" because it's only seeing this div, whereas when I move it, I get "list 15 items"

ayeshakmaz (changes requested):

packages/gestalt/src/Masonry.js
dangerismycat: you are absolutely correct! 🤦

liuyenwei (approved):

ayeshakmaz (approved):

PR #1965, opened 2022-03-03
Docs: Sentence case titles/navigation

ponori:
### Summary
Applied a sentence-case format for all titles and navigation elements

#### What changed?

Converted some capital letters to lowercase.

#### Why?

Sentence case is the standard format of Pinterest's Writing Standards, so we're just working to align to the standard.


docs/pages/faq.js
```diff
@@ -8,9 +8,9 @@ import Page from '../components/Page.js';
 
 export default function ContainerPage(): Node {
   return (
-    <Page title="Frequently Asked Questions">
-      <PageHeader name="Frequently Asked Questions" showSourceLink={false} />
-      <Card name="Gestalt Usage">
+    <Page title="Frequently asked questions">
```
dangerismycat, line 11: Should these remain capitalized, given the "FAQ" acronym?

docs/pages/faq.js
```diff
@@ -86,7 +86,7 @@ padding 0 .. 12
             />
           </Box>
 
-          <Heading size="400">How can add a new Icon?</Heading>
+          <Heading size="400">How can add a new icon?</Heading>
           <Text>
             If you need an Icon that is not listed on our{' '}
```
dangerismycat, line 91: "icon" shouldn't be capitalized here either, since it's referring to an icon, not our Icon component

docs/pages/faq.js
```diff
@@ -183,13 +183,13 @@ yarn cypress run
             />
           </Text>
 
-          <Heading size="400">What is Gestalt Teachings and how do I add a video?</Heading>
+          <Heading size="400">What is gestalt teachings and how do I add a video?</Heading>
```
dangerismycat, line 186: Gestalt should be capitalized

AlbertCarreras (approved):
LGTM despite 2 tiny notes

docs/pages/faq.js
ponori: I think sentence case makes sense here - but totally open to change if we strongly believe it should be capitalize.

docs/pages/faq.js
ponori: Done!

docs/pages/faq.js
ponori: Ugh, my bad. Thanks and fixed.

PR #1943, opened 2022-02-17
Codemod

AlbertCarreras:
Thanks for creating a PR! Please follow this template and delete items/sections that are not relevant to your changes, including these instructions.

Please also make sure your [PR title](https://github.com/pinterest/gestalt/#releasing) matches our format: `ComponentName: Description`

#### What changed?

From a high level, what are the changes this PR introduces? (No need to recount line-by-line, we can see that.)

#### Why?

What is the purpose of this PR? Please include the context around these changes for Future Us. In addition to _what_ is changing, we need to know _why_ these changes are needed. Imagine someone is looking at these changes a year from now and needs to know why they were made.

### Links

- [Jira](https://jira.pinadmin.com/browse/GESTALT-XXXX)
- [TDD](link to Paper doc)
- [Figma](link to Figma file)

### Checklist

- [ ] Added unit and Flow Tests
- [ ] Added documentation + accessibility tests
- [ ] Verified accessibility: keyboard & screen reader interaction
- [ ] Checked dark mode, responsiveness, and right-to-left support
- [ ] Checked stakeholder feedback (e.g. Gestalt designers)

PR #1969, opened 2022-03-07
Tooling:  new `modifyPropValue` & `detectManualReplacement` codemods

AlbertCarreras:
#### What changed?

This PR is a follow up to: [Tooling](https://gestalt.pinterest.systems/tooling): generic codemods to rename components & modify props ([#1943](https://github.com/pinterest/gestalt/pull/1943))

This PR:
- adds a new ReadMe.md file to provide information for those developing new generic codemods
-  adds a new set of helpers using the jscodeshift API based in Collections, NodePaths and Nodes.
- adds 2 new codemods `modifyPropValue` & `detectManualReplacement`. 
- refactors `modifyProp` codemod to use new helper library
- adds some more context to the Tooling page around available generic codemods

#### Why?

As part of the Tooling and Infra OKR, we are building new codemods that would prevent the Gestalt team from writing custom codemods on each major upgrade.  See list of [previous custom codemods](https://docs.google.com/spreadsheets/d/1dXhVWhmOCiDVYAzivT1MmoT25F17rvTeXAiIjri6dck/edit#gid=0) or [codemod folder](https://github.com/pinterest/gestalt/tree/master/packages/gestalt-codemods).

These set of 4 codemods, replaces aprox 88.50% of manual codemods created previously.
image

### Checklist


packages/gestalt-codemods/generic-codemods/utils.js
```diff
@@ -5,261 +5,278 @@ import {
...
- * renameJSXElement: Renames the JSX element with the name value provided
- * E.g. previousComponentName = Box & nextComponentName =  RenamedBox
- * input: <Box /> >> output: <RenamedBox />
+ * deepCloneNode: Returns a collection containing the Gestalt import declaration node-path
  */
-const renameJSXElement = ({ JSXNode, nextComponentName }: RenameJSXElementType): void => {
-  // TODO: implement deep cloning or prevent reassignment
-  const newJSXNode = { ...JSXNode };
-  newJSXNode.openingElement.name.name = nextComponentName;
+const deepCloneNode = <T>({ node }: { node: T }): T => JSON.parse(JSON.stringify(node));
```
dangerismycat, line 160: Nodes are by definition serializable, right? Will this ever hit something like a function or `undefined` that would cause it to bonk?

packages/gestalt-codemods/generic-codemods/utils.js
```diff
@@ -5,261 +5,278 @@ import {
...
+const buildAttributeFromValue = ({
+  j,
+  prop,
+  value,
+}: {
+  j: JSCodeShift,
+  prop: string,
+  value?: string | boolean | number,
+}): ?JSXNodeType => {
+  switch (typeof value) {
```
dangerismycat, line 174: Future improvement: also support types like arrays, which could call this function recursively

packages/gestalt-codemods/generic-codemods/README.md
```diff
@@ -0,0 +1,65 @@
+# Gestalt Generic Codemods
```
dangerismycat, line 1: We also have https://github.com/pinterest/gestalt/blob/master/packages/gestalt-codemods/README.md…how should these files co-exist? Should one link to the other? Both link to each other? Merge the two into one? Something else?

packages/gestalt-codemods/generic-codemods/utils.js
```diff
@@ -5,261 +5,278 @@ import {
...
-/**
- * throwErrorIfSpreadProps: Throws an error message if component contains spread props which are opaque to  codemods
- * E.g. <Box {...props} /> // error!
+ * throwErrorMessage: Throws an error message
  */
-const throwErrorIfSpreadProps = ({ file, JSXNode }: ThrowErrorIfSpreadType): void => {
-  if (
-    JSXNode.openingElement.attributes.some((attribute) => attribute.type === 'JSXSpreadAttribute')
-  ) {
+const throwErrorMessage = ({
```
dangerismycat, line 246: nit: this name doesn't quite cover what the function does. It makes it seem like this will always throw, when it's actually only going to throw conditionally

dangerismycat (approved):
I'm seeing some `*.output.js` files that are empty?

In general this is _awesome_. Thanks for digging deeper and finding these better solutions!

docs/pages/tooling.js
```diff
@@ -126,7 +126,14 @@ Most rules come with out-of-the-box autofixes, automating the adoption of Gestal
...
+Custom codemods can be found in this [directory](https://github.com/pinterest/gestalt/tree/master/packages/gestalt-codemods) under the corresponding upgrade version. Generic codemods (for common tasks like renaming components or props) can be found in this [directory](https://github.com/pinterest/gestalt/tree/master/packages/gestalt-codemods/generic-codemods). When generic codemods are use for an upgrade, the specific command is included in the pull request with the major version change.
+The generic codemods available are:
+- [renameComponent](https://github.com/pinterest/gestalt/blob/master/packages/gestalt-codemods/generic-codemods/renameComponent.js) to rename Gestalt components.
+- [modifyProp](https://github.com/pinterest/gestalt/blob/master/packages/gestalt-codemods/generic-codemods/modifyProp.js) to modify (rename or remove) Gestalt component props.
+- [modifyPropValue](https://github.com/pinterest/gestalt/blob/master/packages/gestalt-codemods/generic-codemods/modifyPropValue.js) to modify (rename, add, or remove) prop-value combinations in Gestalt component.
```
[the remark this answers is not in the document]
AlbertCarreras: > I'm seeing some `*.output.js` files that are empty?
> 
> In general this is _awesome_. Thanks for digging deeper and finding these better solutions!

DONE

packages/gestalt-codemods/generic-codemods/README.md
AlbertCarreras: MERGED, DONE

packages/gestalt-codemods/generic-codemods/utils.js
AlbertCarreras: you can't deep clone node paths (cos there are circular issues). I think the AST nodes are fine. I can try to stress test it. I'll merge first and stress test it to find edge cases.

PR #1970, opened 2022-03-07
Box: add elevation accent color option

ayeshakmaz:
#### What changed?

Add elevationAccent as a color option on Box

#### Why?
Now that we've added elevation options, we need to make that option available on Box as a background color (just the background elevation accent token)

Dark mode is coming soon

### Checklist


AlbertCarreras (approved):
LGTM

PR #1972, opened 2022-03-08
Link, Module, NumberField, RadioButton, SearchField, SelectList, Status,  Switch , TextArea, TextField: add visual diff testing and screenshot to description in VSCode (Flow extension)

AlbertCarreras:
#### What changed?

Link, Module, NumberField, RadioButton, SearchField, SelectList, Status,  Switch , TextArea, TextField: add visual diff testing and screenshot to description in VSCode (Flow extension)

#### Why?

Provide basic visual testing coverage for most Gestalt components and link screenshots to VSCode's Flow extension that shows component screenshot below the component description on hover for better in-editor documentation.

OKR: 
- Tooling & Infra
- Visual Diff testing 
- Every component in our library uses visual diff testing to guard against UI regression


docs/pages/visual-test/Switch-dark.js
```diff
@@ -0,0 +1,18 @@
+// @flow strict
+import { type Node } from 'react';
+import { Flex, Switch, Box, ColorSchemeProvider } from 'gestalt';
+
+export default function Screenshot(): Node {
+  return (
+    <ColorSchemeProvider colorScheme="dark">
+      <Box color="white" display="inlineBlock" padding={1}>
+        <Flex gap={4}>
+          <Switch onChange={() => {}} id="1" switched />
```
dangerismycat, line 10: Can you add labels for these so we can tell what each represents? This will be very helpful when viewing these in VS Code

dangerismycat (approved):
Thanks again for taking this on — tedious work, but it will be worth it!

docs/pages/visual-test/Switch-dark.js
AlbertCarreras: I though about it... I doubted about Labels because Switch doesn't have built-in labels... So I decided to show the raw component. But I agree it's confusing.

AlbertCarreras: DONE

PR #1973, opened 2022-03-08
Text: Add new semantic colors

ayeshakmaz:
#### What changed?

Add new values for Text that follow semantic design tokens

#### Why?

Instead of allowing all colors on Text, we're updating the color prop to use semantic color names that follow our new design tokens. The old values will be deprecated in a future PR.

### Links

- [Jira](https://jira.pinadmin.com/browse/GESTALT-3944)

### Checklist


christianvuerings:
@ayeshakmaz nit - what's the migration timeline for these? E.g. right now it's confusing as to when to use `red` or `error`.

image

ayeshakmaz:
@christianvuerings we're recommending no one us the old colors going forward, so people should be using error going forward. Next sprint I'll create a codemod and run it in Pinboard to convert current mappings over, but we need to reconcile the blue color first

AlbertCarreras (approved):
LGTM

AlbertCarreras:
LGTM but Flow is complaining...
Screen Shot 2022-03-08 at 9 38 01 PM

AlbertCarreras (changes requested):
The Flow issue

ayeshakmaz:
@AlbertCarreras let me know if this looks good to go now, thanks!

docs/pages/color_usage.js
```diff
@@ -43,7 +43,7 @@ export default function ColorUsagePage(): Node {
       >
         <ColorTile
           description="Primary base"
-          textColor="white"
+          textColor="light"
```
ayeshakmaz, line 46: Changes in this file make use of the new values as a sort of test

docs/pages/text.js
```diff
@@ -5,34 +5,39 @@ import PageHeader from '../components/PageHeader.js';
 import Page from '../components/Page.js';
 import GeneratedPropTable from '../components/GeneratedPropTable.js';
 import docgen, { type DocGen } from '../components/docgen.js';
+import MainSection from '../components/MainSection.js';
```
AlbertCarreras, line 8: cleanup changes, unrelated. (for review tracking)

docs/pages/text.js
```diff
@@ -41,63 +46,85 @@ export default function TextPage({ generatedDocGen }: {| generatedDocGen: DocGen
...
+    <Text color="warning" size="400">Warning</Text>
+    <Text color="error" size="400">Error</Text>
+    <Text color="shopping" size="400">Shopping</Text>
+    <Box color="primary" padding={1}>
+      <Text color="light" size="400">Light</Text>
+    </Box>
+    <Box color="infoWeak" padding={1}>
+      <Text color="dark" size="400">Dark</Text>
+    </Box>
+  </Flex>
```
ayeshakmaz, line 78: This is the main update on this page

packages/gestalt/src/Text.js
```diff
@@ -106,10 +115,22 @@ export default function Text({
   underline = false,
   weight = 'normal',
 }: Props): Node {
+  const semanticColors = [
```
AlbertCarreras, line 118: suggestion

i think we could prevent mismatches if we move `semanticColors` to textTypes.js and then spread it into `allowedColors`

packages/gestalt/src/Typography.css
```diff
@@ -31,6 +31,44 @@
   font-size: var(--font-size-600);
 }
 
+/* colors */
```
AlbertCarreras, line 34: nit, worth adding _semantic_ here? just a thought

packages/gestalt/src/textTypes.js
```diff
@@ -23,6 +23,15 @@ export const allowedColors = [
...
+  'success',
+  'error',
+  'warning',
+  'inverse',
+  'shopping',
+  'light',
+  'dark',
 ];
 
 export type Color =
```
AlbertCarreras, line 37: are we using this export type Color anywhere? maybe we could remove it. With autogenerated prop tables, we are inlining. could be worth checking

AlbertCarreras (approved):
LGTM

docs/pages/text.js
```diff
@@ -41,63 +46,85 @@ export default function TextPage({ generatedDocGen }: {| generatedDocGen: DocGen
...
+    </Text>
+  </Flex>
+</Box>
+        `}
+          />
+        </MainSection.Subsection>
+        <MainSection.Subsection
+          title="Sizes"
+          description={`
+          You can apply \`size\` options to define the size of the text. These font sizes follow those available through our [Design Tokens](/design_tokens#Font-size). If your text needs to be a [semantic heading (H1-H6)](/https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Heading_Elements), use [Heading](/heading) instead.
```
AlbertCarreras, line 122: ALERT link  is broken

AlbertCarreras: [semantic heading (H1-H6)](**/**https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Heading_Elements)

packages/gestalt/src/textTypes.js
ayeshakmaz: I'll look into this when we do this for Heading

PR #1974, opened 2022-03-08
Docs: Added design guidelines to Datapoint docs

ponori:
### Summary
Added design best practices & related components to the Datapoint docs.

#### What changed?
More design-specific guidance to the Datapoint docs.

### DIFF
--- a/docs/pages/datapoint.js
+++ b/docs/pages/datapoint.js
@@ -57,12 +57,16 @@ export default function DatapointPage({ generatedDocGen }: {| generatedDocGen: D
 <Datapoint size="lg" title="Saves" value="10,392" trend={{value: -12.193, accessibilityLabel: "Trending down"}} />
 `}
           />
+          {/*
+            The example below should also display a localized trend value, but given trend accepts only a number,
+            the value cannot be localized. Once the API is changed, we should add a localized trend to this example.
+            */}
           <MainSection.Card
             cardSize="md"
             type="do"
             description="Make sure to localize numeric values."
             defaultCode={`
-<Datapoint size="lg" title="Leistung" value="3.000,25" trend={{value: 7,5, accessibilityLabel: "Trending up"}}  />
+<Datapoint size="lg" title="Leistung" value="3.000,25" />
 `}
           />
           <MainSection.Card
@@ -98,7 +102,7 @@ export default function DatapointPage({ generatedDocGen }: {| generatedDocGen: D
       />
       <MainSection
         name="Localization"
-        description={`Be sure to localize the \`title\`, \`tooltipText\` and trend \`accessibilityLabel\` props. Note that localization can lengthen text by 20 to 30 percent.`}
+        description={`Be sure to localize the \`title\`, \`value\`, \`tooltipText\` and trend \`accessibilityLabel\` props. Note that localization can lengthen text by 20 to 30 percent.`}
       />
       <MainSection name="Variants">
         <MainSection.Subsection columns={2} title="Size">
@@ -162,6 +166,17 @@ export default function DatapointPage({ generatedDocGen }: {| generatedDocGen: D
           />
         </MainSection.Subsection>
       </MainSection>{' '}
+      <MainSection name="Related">
+        <MainSection.Subsection
+          description={`
+**[Table](/table)**
+When displaying a large volume of data values and trends (for instance, 10 or more), it’s recommended to display those values in tabular format.
+
+**[Status](/status)**
+Status is recommended in instances where information is more categorical or qualitative (such as health or phase).
+      `}
+        />
+      </MainSection>
     </Page>
   );
 }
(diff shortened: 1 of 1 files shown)


docs/pages/datapoint.js
```diff
@@ -39,9 +39,70 @@ export default function DatapointPage({ generatedDocGen }: {| generatedDocGen: D
...
+            type="do"
+            description="Be certain to use a positive `trendSentiment` only when the trend is overtly positive for the end user. Use a neutral `trendSentiment` in cases of ambiguity."
+            defaultCode={`
+<Datapoint size="lg" title="Total spend" value="$14,325" trend={{value: 5.6, accessibilityLabel: "Trending up"}} trendSentiment="neutral"  />
+`}
+          />
+          <MainSection.Card
+            cardSize="md"
+            type="don't"
+            description="Use Status' `subText` to display extraneous messaging."
```
AlbertCarreras, line 92: @ponori Does this one belong here? It's Status not Datapoint, how does it relate to Datapoint?

docs/pages/datapoint.js
```diff
@@ -39,9 +39,70 @@ export default function DatapointPage({ generatedDocGen }: {| generatedDocGen: D
           />
         </MainSection.Subsection>
       </MainSection>
+      <MainSection name="Best practices">
+        <MainSection.Subsection columns={2}>
+          <MainSection.Card
+            cardSize="md"
+            type="do"
+            description="Whenever possible, use the full number with separators to ensure clarity across all languages."
```
AlbertCarreras, line 47: What would be the opposite of this one? 

_Whenever possible, use the full number with separators to ensure clarity across all languages._

I'm not sure I understand this one.

docs/pages/datapoint.js
```diff
@@ -39,9 +39,70 @@ export default function DatapointPage({ generatedDocGen }: {| generatedDocGen: D
...
+`}
+          />
+          {/*
+            The example below should also display a localized trend value, but given trend accepts only a number,
+            the value cannot be localized. Once the API is changed, we should add a localized trend to this example.
+            */}
+          <MainSection.Card
+            cardSize="md"
+            type="do"
+            description="Make sure to localize numeric values."
```
AlbertCarreras, line 67: Would this one be better at the Localization section? 

I don't know if I see it as a best practice.

docs/pages/datapoint.js
```diff
@@ -39,9 +39,70 @@ export default function DatapointPage({ generatedDocGen }: {| generatedDocGen: D
...
+            description="Use Status' `subText` to display extraneous messaging."
+            defaultCode={`
+<Status type="problem" title="Error" subtext="Please try again" />
+`}
+          />
+        </MainSection.Subsection>
+      </MainSection>
+      <MainSection
+        name="Accessibility"
+        description={`Datapoint’s trend prop requires an \`accessibilityLabel\` to describe the trend icon (e.g., Trending up). `}
```
AlbertCarreras, line 101: nit 
trend prop
\`trend\` prop

AlbertCarreras (changes requested):
As per your feedback, I'll wait to those changes to R+

docs/pages/datapoint.js
ponori: I added this specifically because we typically think of localization around language as opposed to numbers. Happy to remove down the road when we're able to link to more detailed localization docs.

docs/pages/datapoint.js
ponori: Done!

docs/pages/datapoint.js
ponori: Added some more detail to this guideline to hopefully clarify.

docs/pages/datapoint.js
ponori: Thanks! Fixed!

docs/pages/datapoint.js
AlbertCarreras: Got it.

AlbertCarreras: I think I'd still repeat the same concept on the Location section. Kinda using the same words you just used, "We typically think of localization around language as opposed to numbers. In Datapoint, don't forget to localize the numbers as well.  

My main question is, how do we support ,  if we do, number localization in Pinboard?

docs/pages/datapoint.js
```diff
@@ -39,9 +39,76 @@ export default function DatapointPage({ generatedDocGen }: {| generatedDocGen: D
...
+            type="do"
+            description="Be certain to use a positive `trendSentiment` only when the trend is overtly positive for the end user. Use a neutral `trendSentiment` in cases of ambiguity."
+            defaultCode={`
+<Datapoint size="lg" title="Total spend" value="$14,325" trend={{value: 5.6, accessibilityLabel: "Trending up"}} trendSentiment="neutral"  />
+`}
+          />
+          <MainSection.Card
+            cardSize="md"
+            type="don't"
+            description="Be cautious of using `Datapoint` for large volumes of data. In cases where the content is tabular, use `Table`."
```
AlbertCarreras, line 92: nits

Be cautious of using Datapoint for large volumes of data. In cases where the content is tabular, use [Table](https://gestalt.pinterest.systems/table)."

AlbertCarreras (approved):
Awesome. Thanks

PR #1979, opened 2022-03-09
Docs: Add elevation guidelines page

ayeshakmaz:
#### What changed?

Add guidelines page for Elevation, update Box css to use new tokens

#### Why?

After adding our elevation tokens, we've added a guidelines page on how to create elevation in different scenarios 

Note: Dark mode support is coming next sprint.

### Links

- [Jira](https://jira.pinadmin.com/browse/GESTALT-3529)
- [Figma](https://www.figma.com/file/hkpUn0QPCn3uftkC7DmQVG/Elevation---Layout-Ref.?node-id=2%3A345)

### Checklist

### DIFF
--- a/docs/pages/elevation.js
+++ b/docs/pages/elevation.js
@@ -1,5 +1,6 @@
 // @flow strict
 import { type Node } from 'react';
+import { Box, Flex, Text } from 'gestalt';
 import MainSection from '../components/MainSection.js';
 import PageHeader from '../components/PageHeader.js';
 import Page from '../components/Page.js';
@@ -21,6 +22,18 @@ export default function ColorUsagePage(): Node {
           description="The default elevation level.
         Elevates messages temporarily appearing in front of other surfaces, such as modals and banners. "
         >
+          <Flex gap={4} direction="column">
+            <Box
+              width={300}
+              height={300}
+              rounding={3}
+              dangerouslySetInlineStyle={{
+                __style: { boxShadow: 'var(--elevation-floating' },
+              }}
+            />
+            <Text>$elevation-floating</Text>
+            <Text>10% (#000000) opacity / Blur: 8</Text>
+          </Flex>
           <MainSection.Subsection columns={2}>
             <MainSection.Card
               cardSize="md"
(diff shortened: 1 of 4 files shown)


docs/pages/elevation.js
```diff
@@ -0,0 +1,183 @@
...
+  return (
+    <Page title="Elevation">
+      <PageHeader
+        name="Elevation"
+        description="The elevation allows elements to be lifted from the background and can be defined by applying drop shadows or other visual cues, such as color and borders."
+        showSourceLink={false}
+      />
+      <MainSection
+        name="Elevation using shadows"
+        description="SHadows reflect a spatial relationship by indicating separation from the background or allowing content to be scrolled under another element. Elevation using shadows includes two levels: floating and raised."
```
AlbertCarreras, line 18: typo

SHadows > Shadows

AlbertCarreras:
This can be confusing
Screen Shot 2022-03-10 at 12 38 23 AM

docs/pages/elevation.js
```diff
@@ -0,0 +1,183 @@
...
+          title="Floating"
+          description="The default elevation level.
+        Elevates messages temporarily appearing in front of other surfaces, such as modals and banners. "
+        >
+          <Flex gap={4} direction="column">
+            <Box
+              width={300}
+              height={300}
+              rounding={3}
+              dangerouslySetInlineStyle={{
```
AlbertCarreras, line 30: This is still not supported in Box as an elevation Prop. Can it be confusing to add guidelines before we support elevation via props or a new component?

ayeshakmaz: Good call, I added a note that raised options are coming to Box soon. The rest are actually all possible in Box already

AlbertCarreras (approved):
LGTM, merge when designers give you the thumbs up. Maybe, I'd add a badge to make it obvious it shouldn't be used yet.

docs/pages/elevation.js
```diff
@@ -0,0 +1,166 @@
...
+import MainSection from '../components/MainSection.js';
+import PageHeader from '../components/PageHeader.js';
+import Page from '../components/Page.js';
+
+export default function ColorUsagePage(): Node {
+  return (
+    <Page title="Elevation">
+      <PageHeader
+        name="Elevation"
+        description="The elevation allows elements to be lifted from the background and can be defined by applying drop shadows or other visual cues, such as color and borders."
```
dangerismycat, line 13: typo: "Elevation allows elements…"

This would also be more readable if split into two sentences: "…from the background. It can be defined by…"

docs/pages/elevation.js
```diff
@@ -0,0 +1,166 @@
...
+        showSourceLink={false}
+      />
+      <MainSection
+        name="Elevation using shadows"
+        description="Shadows reflect a spatial relationship by indicating separation from the background or allowing content to be scrolled under another element. Elevation using shadows includes two levels: floating and raised."
+      >
+        <MainSection.Subsection
+          title="Floating"
+          description="The default elevation level.
+        Elevates messages temporarily appearing in front of other surfaces, such as modals and banners. "
```
dangerismycat, line 23: This sentence fragment is awkward and doesn't match style elsewhere. Consider: "The default elevation level, elevating messages temporarily…"

docs/pages/elevation.js
```diff
@@ -0,0 +1,166 @@
...
+        >
+          <Flex gap={4} direction="column">
```
dangerismycat, line 25: This is difficult to see on my older monitors. Can we make it more apparent in some way? Perhaps a color on the background, something like that?

docs/pages/elevation.js
```diff
@@ -0,0 +1,166 @@
...
+            <Text>10% (#000000) opacity / Blur: 8</Text>
+          </Flex>
+          <MainSection.Subsection columns={2}>
+            <MainSection.Card
+              cardSize="md"
+              type="do"
+              title="When to use"
+              description={`
+              - Highlighting an item that floats above other content, such as temporary messages or cards
+              - IconButton and other elements, serving as an affordance for scrolling behaviors
```
dangerismycat, line 37: This doesn't match the other "when to use" examples grammatically. What are we saying about IconButton? That this elevation is built-in, that it needs to be applied, something else?

docs/pages/elevation.js
```diff
@@ -0,0 +1,166 @@
...
+              `}
+            />
+            <MainSection.Card
+              cardSize="md"
+              type="don't"
+              title="When not to use"
+              description={`
+              - Content that doesn't need elevation or have its own setup (e.g. Pins, boards)
```
dangerismycat, line 45: This reads "Content that doesn't … have its own setup", which I think is the opposite of the intended meaning

docs/pages/elevation.js
```diff
@@ -0,0 +1,166 @@
...
+              `}
+            />
+          </MainSection.Subsection>
+        </MainSection.Subsection>
+        <MainSection.Subsection
+          title="Raised"
+          description={`
+          Presents a drop shadow on the edge of a top or bottom component, allowing surfaces to move behind when scrolled.
+
+          ⚠️ Note: Raised shadow options will be available automatically in Box soon.
```
dangerismycat, line 55: Given this note, should the Floating option be called out that it is available on Box? Given the presentation of the token and values, it makes it seem that the user has to implement the shadow themselves using dangerouslySetInlineStyle

docs/pages/elevation.js
```diff
@@ -0,0 +1,166 @@
+// @flow strict
+import { type Node } from 'react';
+import { Box, Flex, Text } from 'gestalt';
+import MainSection from '../components/MainSection.js';
+import PageHeader from '../components/PageHeader.js';
+import Page from '../components/Page.js';
+
+export default function ColorUsagePage(): Node {
+  return (
+    <Page title="Elevation">
```
dangerismycat, line 10: This doesn't include any information about elevation vis-a-vis dark mode. If that's coming in the future, should we add a note indicating that?

docs/pages/elevation.js
```diff
@@ -0,0 +1,166 @@
+// @flow strict
```
dangerismycat, line 1: Can/should this also include guidance on Box's docs as well? At a minimum the relevant prop descriptions should link to this page; ideally there would be specific guidance for those props as well.

ayeshakmaz: I'll make a separate ticket for this

docs/pages/elevation.js
ayeshakmaz: Given that we don't already support dark mode, I think we can leave that off for now. Plus dark mode is being added next sprint

docs/pages/elevation.js
ayeshakmaz: I don't know that adding a background color will help, as this shadow is meant to be on light backgrounds typically until dark mode is supported...

PR #1980, opened 2022-03-10
ActivationCard: Use Text instead of Heading

christianvuerings:
#### What changed?

Uses `Text` instead of `Heading` when rendering `ActivationCard`

#### Why?

* Using `Heading` results in an implicit accessibilityLevel to be set inside the `ActivationCard` component. That causes pinboard a11y tests to fail since we don't aren't able to override it from the outside.
* It doesn't make sense to use `Heading` since this is not describing a specific section on the page. It would make more sense to have `Heading` before the `ActivationCard` component.

#### Notes

* No visual differences

### Before

Screen Shot 2022-03-10 at 4 50 46 AM

### After

Screen Shot 2022-03-10 at 4 51 17 AM

### DIFF
--- a/packages/gestalt/src/ActivationCard.js
+++ b/packages/gestalt/src/ActivationCard.js
@@ -120,7 +120,9 @@ function CompletedCard({ dismissButton, message, status, statusMessage, title }:
         )}
         <Box>
           <Box>
-            <Heading size="400">{title}</Heading>
+            <Text size="400" weight="bold">
+              {title}
+            </Text>
           </Box>
           {message && (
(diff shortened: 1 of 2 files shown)


packages/gestalt/src/ActivationCard.js
```diff
@@ -120,7 +120,9 @@ function CompletedCard({ dismissButton, message, status, statusMessage, title }:
         )}
         <Box>
           <Box>
-            <Heading size="400">{title}</Heading>
+            <Text size="400" weight="bold">
```
ayeshakmaz, line 123: 🎉

ayeshakmaz (approved):

PR #1981, opened 2022-03-11
Docs: Add ability to write Markdown/MDX docs pages

rlingineni:
### Summary
This PR enables writing pages in [MDX](https://mdxjs.com/) format. 

We can use Markdown as a base, and then interweave it with JSX. Alternatively, pages can also be written in raw Markdown.

#### What changed?
- Updates NextJS configuration to use [MDX](https://nextjs.org/docs/advanced-features/using-mdx). This enables using MDX files.

- Updated  Routes definition to statically compile MD and MDX files into pages

- Added a `MarkdownPage` component that's the base layer

- Wrote a few "shortcodes" that hook into how certain elements are rendered to avoid writing entire react components

- Added a few props to the card and header components to remove/add unused spacing to make it look nicer. Used behind the scenes by the shortcodes when rendering

#### Why?
- Currently it seems like the Docs process is split between Dropbox and then converting to JSX. With this change, you can  [export raw Markdown](https://help.dropbox.com/files-folders/paper/export-docs) from Paper and have the entire text portion of the docs ready to go in an instant

- Would love to see a docs world where editing docs pages were inline. You can now create a file, preview, and edit directly from Github without ever compiling. Just like this [doc](https://github.com/johnpolacek/nextjs-mdx-blog-starter/blob/main/src/mdx/about.mdx)

- Composability. Writing in MDX means less syntax to worry about. We don't have to escape code tags for example. 

### Checklist


docs/components/PageHeader.js
```diff
@@ -29,6 +30,7 @@ const buildSourceLinkUrl = (componentName) =>
 export default function PageHeader({
   badge,
   defaultCode,
+  noMargin,
```
rlingineni, line 33: Just an explicit parameter for disabling the marginBottom

docs/docs.css
```diff
@@ -104,7 +104,7 @@ html[dir="rtl"] .Markdown ul + pre {
   scroll-margin-top: 75px;
 }
 
-.Markdown h3 {
+.Markdown .mdx-header:not() h3 {
```
rlingineni, line 107: This is hacky workaround around a color. Problem was that all of the H3s (###) when written in markdown would turn gray.  So I had to mark them with an "mdx-header" class to skip them.

If the headers don't absolutely have to be gray, and we could make them black, I don't need this workaround. Also, a code search for ### didn't reveal any h3s for me, so not sure if this part of the css is getting used

docs/markdown/redbox.md
```diff
@@ -0,0 +1,227 @@
+---
+title: Redbox
+description: You can use a redbox when you need a beautiful box that's red. Draws attention from users, and bulls. Look closer and you might find a DVD
+badge: pilot
+component: true
```
rlingineni, line 5: Just marking a component page as a component. Need to do this because of width constraints. A component page takes up full-width

docs/markdown/test.md
```diff
@@ -0,0 +1,116 @@
+---
+title: How to work with us
```
rlingineni, line 2: Added Front Matter support. We can expand this to add more properties via markdown

docs/markdown/testing.md
```diff
@@ -0,0 +1,111 @@
+---
```
rlingineni, line 1: Same file as above. Will delete in final merge. But just demonstrating that we have multiple markdown files at different routes

docs/next.config.js
```diff
@@ -14,14 +14,17 @@ module.exports = {
...
-      ...config.watchOptions,
-      poll: dev ? 500 : false,
-    },
-  }),
+  webpack: (config /*: WebpackConfig */, { dev } /*: {| dev: boolean |} */) /*: WebpackConfig */ =>
+    // These modules resolve by default in the server-side functions.
+    ({
+      resolve: {
+        // Explicitly tell webpack to ignore resolving "fs" for the client bundle
+        fallback: { fs: false, path: false },
```
rlingineni, line 22: Answer here: https://github.com/vercel/next.js/discussions/12124

To get the markdown files from disk, we use the `fs` module. Since we're allowed to use it in the `gestStaticProps` method, we're okay. 

However, webpack also tries to add it to the client and causes an error. This line should prevent that.

docs/components/MarkdownPage.js
```diff
@@ -0,0 +1,98 @@
...
+      <Box marginTop={12} marginBottom={4}>
+        <MainSection name={props.children} />
+      </Box>
+    ),
+    h3: (props) => (
+      <Box
+        ref={(node) => {
+          // eek, hacky. Getting around the .Markdown > h3 coloring css
+          if (node) {
+            node.classList.add('mdx-header');
```
rlingineni, line 39: Related to the mdx-hedaer css from above. Injecting the css class manually as an ovverride to box. Although, I might be able to use a regular div.

docs/pages/[id].js
```diff
@@ -0,0 +1,65 @@
...
+  return {
+    props: {
+      meta,
+      content: mdxSource,
+      pageSourceUrl: `https://github.com/pinterest/gestalt/tree/master/docs/pages/markdown/${id}.md`,
+    },
+  };
+}
+
+export async function getStaticPaths(): Promise<{|
```
Comment (rlingineni) [inline docs/pages/[id].js:50]: The only paths valid paths that should exist outside of traditional pages are in the "markdown" folder

flow-typed/npm/gray_matter_vx.x.x.js
```diff
@@ -0,0 +1,8 @@
+
+// @flow
+// gray-matter flow-typed version based on https://github.com/jonschlinkert/gray-matter
+
+declare module 'gray-matter' {
+  declare module.exports: any;
```
rlingineni, line 6: Not super proud of these flow typings, but they're all backed by Typescript definitions as well.

docs/markdown/how-to.md
```diff
@@ -0,0 +1,117 @@
...
+The goal of Gestalt is to create a shared library of design best practices, React components, and documentation. Therefore, the best contenders for addition to Gestalt are designs or components that benefit multiple teams and are used regularly throughout our products.
+
+## Requesting new components or changes
+
+**Please note:** The process below applies to Pinterest employees.
+We ask that any ideas that change the functionality of an existing Gestalt component or introduce a net-new component go through the following process:
+
+1. **Present work early in our [partnership meetings](#Meetings-and-events)**
+   What we build into Gestalt comes from the teams across Pinterest, so we’d love to pair with you early and often! Seeing early mocks, wireframes, or even product roadmaps and learning where your project may be going helps inform our team OKRs. The earlier we see the work, the better we're able to plan and the higher the likelihood we'll be able to help. Please [sign up](https://pinch.pinadmin.com/gestaltSignUp) for an Office Hours slot with your discussion topic.
+2. **Iterate on solutions**
```
AlbertCarreras, line 17: could we space this list items to make it more readable?

docs/markdown/redbox.md
AlbertCarreras: @rlingineni Are we still considering markdown for component pages or just for readme-like "getting started" pages?

docs/pages/[id].js
Comment (AlbertCarreras) [reply docs/pages/[id].js:50]: For context 
https://nextjs.org/docs/basic-features/data-fetching/get-static-paths

> If a page has [Dynamic Routes](https://nextjs.org/docs/routing/dynamic-routes) and uses getStaticProps, it needs to define a list of paths to be statically generated.

> When you export a function called getStaticPaths (Static Site Generation) from a page that uses dynamic routes, Next.js will statically pre-render all the paths specified by getStaticPaths.

> getStaticPaths must be used with getStaticProps. You cannot use it with [getServerSideProps](https://nextjs.org/docs/basic-features/data-fetching/get-server-side-props).

docs/pages/[id].js
```diff
@@ -0,0 +1,65 @@
+// @flow strict
```
Comment (AlbertCarreras) [inline docs/pages/[id].js:1]: can we add a top what/why/how comment on this page?

idk, something like

//* 
`pages/[id].js` renders markdown, statically generated files. 
Pages files must be within ./markdown folder 
*//

docs/utils/mdHelper.js
```diff
@@ -0,0 +1,29 @@
...
+): {| content?: string, meta: { [key: string]: string }, route: string, isMDX: boolean |} {
+  const docsDirectory = path.join(process.cwd(), 'markdown');
+  const fullPath = path.join(docsDirectory, `${route}.md`);
+
+  try {
+    const fileContents = fs.readFileSync(fullPath, 'utf8');
+    const { data, content } = matter(fileContents);
+
+    return { route, meta: data, content, isMDX: true };
+  } catch (ex) {
```
AlbertCarreras, line 17: we don't need the ex no?


`  } catch () {`

docs/pages/[id].js
```diff
@@ -0,0 +1,65 @@
...
+      <MDXRemote {...content} />
+    </MarkdownPage>
+  );
+}
+
+export async function getStaticProps(context: {| params: {| id: string |} |}): Promise<{|
+  props: {| meta: { [key: string]: string }, content: {||}, pageSourceUrl: string |},
+|}> {
+  const { id } = context.params;
+  const { meta, content } = getDocByRoute(id);
```
Comment (AlbertCarreras) [inline docs/pages/[id].js:38]: same for getDocByRoute. HOw do you feel about moving function within getStaticProps or closer to where it's called.

docs/components/MarkdownPage.js
```diff
@@ -0,0 +1,97 @@
...
+      newProps.removeMarginBottom = false;
+      return (
+        <MainSection.Card
+          {...newProps}
+          defaultCode={props.children || ""}
+          removeMarginBottom={!('removeMarginBottom' in props) ? true : props.removeMarginBottom}
+        />
+      );
+    },
+    Group: (props) => <Box marginBottom={12}>{props.children}</Box>,
```
AlbertCarreras, line 65: If we aren't doing components, we might not need these ones no?

Just basic markdown

AlbertCarreras: Could you also add a  "Create a markdown page" in Development section to get an idea of the easiness to the general user?
https://gestalt.pinterest.systems/development#Create-a-pull-request

docs/markdown/redbox.md
rlingineni: @AlbertCarreras  - wanted to keep this one around since it opens the door to supporting it in the future if we're open to it, and an example of how we can do it. A little easter egg. I can probably throw it out of the static paths, if we don't want to publish it. Maybe we can chat about in the next eng weekly, and if there is a world where this saves 10x the time of writing a regular page, then maybe it's worth keeping

docs/markdown/how-to.md
rlingineni: Got it. I'll add the spacing. 

Btw, just wanted to call something out here that I noticed. It seems that depending on the Markdown renderer used, lists with text following it are rendered differently. A little subtle.

For example, Github Renders the following as
1. **Bold Header**
Text underneath

In the [MDX renderer](https://mdxjs.com/playground/), it handles the same case to show up as.
1. **Bold Header** Text shows up here

Different [renderers](https://markdownlivepreview.com/) seem to handle in their own way.

The text continues on the same line after a bulleted item. A line break isn't inserted by default in the MDX markdown. This is most likely because a bold item is technically an inline item, and the renderer doesn't insert a break right after. 

A few ways to [fix this](https://gist.github.com/shaunlebron/746476e6e7a4d698b373):
1. Add two spaces after the bulleted header
2. Use an explicit `<br/>` or a `\` at the end of the line (which I did)
3. Use a Markdown Header instead (e.g #### H4). Ideal route. We'll have to figure out the margins here

I don't forsee it being a dealbreaker, thinking we can write around it, or tweak our writing to use headers.

docs/pages/[id].js
```diff
@@ -0,0 +1,65 @@
...
+      pageSourceUrl: `https://github.com/pinterest/gestalt/tree/master/docs/pages/markdown/${id}.md`,
+    },
+  };
+}
+
+export async function getStaticPaths(): Promise<{|
+  paths: {| params: {| id: string |} |}[],
+  fallback: boolean,
+|}> {
+  // get the paths that exist
```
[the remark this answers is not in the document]
Comment (rlingineni) [reply docs/pages/[id].js:54]: Improved the comment. 

There's two functions in that util folder.  In the future, if we wanted to do more with local files,  I think we can tweak these methods to be a generic 'FileReader' util.  One as a `getFilesInDirectory` and the other as a `getFileContents`, and pass the Markdown dir as an argument.

docs/utils/mdHelper.js
rlingineni: Good call out. 

A few things might go wrong here,

1. The file doesn't exist
2. The frontmatter parsing fails

Tbh, not sure how to handle showing the error. We return a 404, if the page happens to fail, and mark `isMDX` as `false`. Since these are server side rendered, I think if this were to fail, we'd catch it before it ships, since it's not client dependent, and perhaps that's okay.

docs/components/MarkdownPage.js
rlingineni: Agreed, probably don't need it if we're not doing components. However, if we do decide to extend a regular page with some interactivity, or we backtrack on component pages, this is pretty useful. Since they're super tiny functions, and pretty contained, I'm swaying in favor of keeping it

rlingineni: Good idea, I'll re-do the development page in Markdown, and try adding it to that

rlingineni:
@AlbertCarreras - thanks for the review! 

Good idea, it seems there's a bit of content to translate over for those pages, but they seem like great candidates to move over. I'll book some time next sprint to complete that transition

docs/docs.css
ayeshakmaz: I think the gray H3s were used on the old WHat's New changelog page. That gray styling be removed so this can be simpler

docs/components/MarkdownPage.js
ayeshakmaz: Yeah I think we can remove this and not worry about the gray

ayeshakmaz (commented):
This looks really great, especially on the more writing/instruction heavy pages. The MD version of the develop page is way more readable and clean. I think this would be a great change to start making for those tutorial/info pages, and like you said, we can work on the component pages in the future

docs/markdown/how-to.md
```diff
@@ -0,0 +1,131 @@
+---
```
dangerismycat, line 1: let's update these file names to match the page file names (and therefore routes) exactly

flow-typed/npm/gray_matter_vx.x.x.js
dangerismycat: Yikes. Can we get better type definitions for these?

docs/utils/mdHelper.js
dangerismycat: Perhaps we can use a Google Analytics event to log the error?

docs/pages/[id].js
```diff
@@ -0,0 +1,77 @@
...
+    },
+  };
+}
+
+export async function getStaticPaths(): Promise<{|
+  paths: {| params: {| id: string |} |}[],
+  fallback: boolean,
+|}> {
+  // get all the paths that exist within ./markdown folder
+  const paths = getAllMarkdownPosts();
```
Comment (dangerismycat) [inline docs/pages/[id].js:67]: It looks like getAllMarkdownPosts is async; do we need to await it here?

docs/utils/mdHelper.js
```diff
@@ -0,0 +1,36 @@
...
+  content?: string,
+  meta: { [key: string]: string },
+  route: string,
+  isMDX: boolean,
+|} {
+  const docsDirectory = path.join(process.cwd(), 'markdown');
+  const fullPath = path.join(docsDirectory, `${route}.md`);
+
+  try {
+    const fileContents = fs.readFileSync(fullPath, 'utf8');
```
dangerismycat, line 16: Do the file/dir reads in this file need to be sync?

docs/markdown/icons-in-md.md
```diff
@@ -0,0 +1,204 @@
...
+
+- Keep the design **simple and schematic**. Reduce the amount of graphic details by focusing on the basic characteristics of the object rather than creating a highly realistic image in order to speed up recognition. (Intricate details are difficult to distinguish at smaller sizes.)
+- Use the**5-second rule**: if it takes you more than 5 seconds to think of an appropriate icon for something, it is unlikely that an icon can effectively communicate that meaning.
+- [Test the icons for recognizability](https://www.nngroup.com/articles/icon-testing/): ask people what they expect the icons to stand for.
+- **Test the icons for memorability**: ask a repeat set of users if they can remember the icon’s meaning after being told what it represented a couple weeks earlier.
+
+
+## Icon sizing
+
+These are the guidelines for icon sizes (in px):
```
dangerismycat, line 108: Can this be a table?

docs/markdown/icons-in-md.md
```diff
@@ -0,0 +1,204 @@
...
+All touch targets for interactive icons need to be 44px or larger. Engineers can add padding to a touch target to meet the 44px requirement.
+
+
+
+## Icon color
+
+Icons can be created using the following color options. brandPrimary should only be used to represent the Pinterest logo, as it is not accessible. See the [design tokens](https://gestalt.netlify.app/design_tokens#Text-color) for more info.
+
+
+![Screen Shot 2022-05-18 at 1.42.18 PM.png](https://codahosted.io/docs/vlenTJHtyS/blobs/bl-bSZTauMmHr/c77129a1c3ece0e8b0ddca9761f701ab7182dd1b97dd1283c8caf291816e9ead43a0fb6741112ce4c49d582f257cd3ef8a327d2ef60f475febe38fc69d67fefe760e9baa738d7bcb19a8812417ede1ed6a70ff0d3da9667d57d79838360ef9b5117c03ce)
```
dangerismycat, line 132: An image definitely isn't ideal here…but perhaps this page isn't a good candidate for MDX anyway. We could potentially use actual components here?

docs/markdown/develop.md
```diff
@@ -0,0 +1,196 @@
+---
```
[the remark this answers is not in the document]
dangerismycat: This page looks great! A good candidate for MDX I think.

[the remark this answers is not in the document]
dangerismycat: Do we have a way of defining a max width for text blocks? This is also an issue on our current Development page, but we've solved this on other pages:
_problem_
Screen Shot 2022-07-08 at 5 17 12 PM
_fixed_
Screen Shot 2022-07-08 at 5 18 01 PM

docs/markdown/how-to.md
```diff
@@ -0,0 +1,131 @@
...
+
+The goal of Gestalt is to create a shared library of design best practices, React components, and documentation. Therefore, the best contenders for addition to Gestalt are designs or components that benefit multiple teams and are used regularly throughout our products.
+
+## Requesting new components or changes
+
+**Please note:** The process below applies to Pinterest employees.
+We ask that any ideas that change the functionality of an existing Gestalt component or introduce a net-new component go through the following process:
+
+1. **Present work early in our [partnership meetings](#Meetings-and-events)** \
+   What we build into Gestalt comes from the teams across Pinterest, so we’d love to pair with you early and often! Seeing early mocks, wireframes, or even product roadmaps and learning where your project may be going helps inform our team OKRs. The earlier we see the work, the better we're able to plan and the higher the likelihood we'll be able to help. Please [sign up](https://pinch.pinadmin.com/gestaltSignUp) for an Office Hours slot with your discussion topic.
```
dangerismycat, line 16: Can we get links to open in a new tab? In a way we can control on a callsite basis?

docs/pages/[id].js
```diff
@@ -0,0 +1,77 @@
+/*
+pages/[id].js renders a dynamic gestalt docs page with a with a page that isn't defined .
+
+The getStaticPaths() will look at the files in the ./markdown folder and try to render a page if it exists, or returns a 404.
```
Comment (dangerismycat) [inline docs/pages/[id].js:4]: Instead of a 404, which we don't handle gracefully, can we redirect to the home page for unknown paths?
Screen Shot 2022-07-08 at 5 24 55 PM

docs/markdown/redbox.md
```diff
@@ -0,0 +1,227 @@
+---
```
dangerismycat, line 1: Don't forget to remove these test pages before merging

docs/markdown/redbox.md
```diff
@@ -0,0 +1,227 @@
...
+title: Redbox
+description: You can use a redbox when you need a beautiful box that's red. Draws attention from users, and bulls. Look closer and you might find a DVD
```
dangerismycat, line 3: 😂

dangerismycat (approved):
Cool, this looks good. There's some cleanup to be done, but this looks promising enough to try out. Thanks for all the work you've put into this!

(apologies for the repeated comments, something was weird with my network connection)

docs/pages/[id].js
Comment (rlingineni) [reply docs/pages/[id].js:67]: Nope, it's sync. The caller can be sync

docs/utils/mdHelper.js
rlingineni: Good catch, will update them to be async

docs/utils/mdHelper.js
rlingineni: Added GA Logging, not sure the naming convention is right, but I didn't notice any precedent outside of button clicks atm

docs/pages/[id].js
Comment (rlingineni) [reply docs/pages/[id].js:4]: I don't think we can do a redirect at the moment. `GetStaticProps` handles the return value for `fallback` and it seems like 404 might be the way that it's handled.

https://nextjs.org/docs/api-reference/data-fetching/get-static-paths#getstaticpaths-return-values

However, it maybe possible to modify our 404 handler to redirect to the homepage perhaps?

docs/markdown/icons-in-md.md
rlingineni: Probably, the original doc didn't have it a table either

docs/markdown/icons-in-md.md
rlingineni: Yeah totally! We'll have to make sure the Code component renders better

docs/components/MarkdownPage.js
```diff
@@ -0,0 +1,169 @@
+// @flow strict
+
+// ignoring: since we do in fact want to render each component md block again
```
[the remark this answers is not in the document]
rlingineni: It's actually a disabling for the rule against inline components, inline with the next comment

docs/components/MarkdownPage.js
```diff
@@ -0,0 +1,169 @@
...
+    title: string,
+    badge: 'pilot' | 'deprecated',
+    component?: boolean,
+    description: string,
+  |},
+  pageSourceUrl?: string,
+|};
+
+export default function MarkdownPage({ children, meta, pageSourceUrl }: Props): Node {
+  const components = {
```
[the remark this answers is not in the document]
rlingineni: I was considering doing this, but most of them are one-line components, and the pages are static, so in theory, they should render just once. Once we get a few more components, I'll make this change in a later release.

docs/components/MarkdownPage.js
```diff
@@ -0,0 +1,169 @@
...
+      />
+    ),
+    Card: (props) => {
+      const newProps = { ...props };
+      newProps.description = undefined;
+      return <MainSection.Card {...newProps} />;
+    },
+    Code: (props: {| removeMarginBottom: boolean, children: string | null |}) => {
+      const newProps = { ...props };
+      newProps.children = null;
```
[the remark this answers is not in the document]
rlingineni: We need to override the children prop, because there'll be a linting error that empty components are self-closing. I've updated the override

docs/markdown/develop.md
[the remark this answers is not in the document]
rlingineni: Yes, we can mark as `fullwidth`:true in the header to adjust the MAX_WIDTH of the page

docs/markdown/how-to.md
rlingineni: We can, I'd probably make a custom `Link` component

flow-typed/npm/gray_matter_vx.x.x.js
rlingineni: Added

docs/utils/gAnalytics.js
```diff
@@ -0,0 +1,15 @@
+// @flow strict
+
+// `eventName` should be relatively generic.
+// pass `value` to provide more context, such as the current component
+export default function logGAEvent(event: string, value?: {| [key: string]: string |}) {
+  if (!window.gtag) {
```
AlbertCarreras, line 6: try 
  if (!window?.gtag) {

docs/markdown/testing/develop.md
```diff
@@ -0,0 +1,204 @@
+---
+title: Development
+component: false
+---
+
+## Set up your laptop
+
+- Install a code editor - we recommended **[VS Code](https://code.visualstudio.com/download)**
+- **Install nvm**
```
AlbertCarreras, line 9: 1. how are we setting the links here? 
Also I see different link behaviors
Kapture 2022-07-26 at 13 35 45

PR #1982, opened 2022-03-12
Dropdown, Module: Change badge prop to allow type

ayeshakmaz:
#### What changed?

After adding Badge types, this PR updates Dropdown and Module to support those Badge types by changing `badgeText` to `badge`, which is an object type that takes text and type as options

#### Why?

Since we have new options for Badge styling, we should make those options available inside the components that utilize Badge, which are Dropdown and Module currently. Since `position` is something the parent component controls, I opened it up to only support `text` and `type` options for specifying the Badge within these components. And since `type` is optional, the codemod only converts current instances of `badgeText="New"` to `badge={{ text: "New"}}`

#### Codemod

To convert Dropdown and Module:
`yarn codemod --parser=flow -t=packages/gestalt-codemods/46.0.0/new_badge_props.js relative/path/to/your/code`

To convert Module.Expandable:
Run `yarn codemod detectManualReplacement ~/path/to/your/code --component=Module --subcomponent=Expandable --prop=badgeText` to find locations of badgeText inside Module.Expandable

### Links

- [Jira](https://jira.pinadmin.com/browse/GESTALT-3980)

### Checklist


packages/gestalt/src/DropdownItem.js
```diff
@@ -13,7 +14,7 @@ type Props = {|
   /**
    * When supplied, will display a [Badge](https://gestalt.pinterest.systems/badge) next to the item's label. See the [Badges](https://gestalt.pinterest.systems/dropdown#Badges) variant to learn more.
    */
-  badgeText?: string,
+  badge?: BadgeType,
```
ayeshakmaz, line 17: Main change 1/3

packages/gestalt/src/DropdownLink.js
```diff
@@ -13,7 +14,7 @@ type Props = {|
   /**
    * When supplied, will display a [Badge](https://gestalt.pinterest.systems/badge) next to the item's label. See the [Badges](https://gestalt.pinterest.systems/dropdown#Badges) variant to learn more.
    */
-  badgeText?: string,
+  badge?: BadgeType,
```
ayeshakmaz, line 17: Main change 2/3

packages/gestalt/src/Module.js
```diff
@@ -30,15 +31,15 @@ export default function Module({
    *
    * Link: https://gestalt.pinterest.systems/text#static-badge
    */
-  badgeText?: string,
+  badge?: BadgeType,
```
ayeshakmaz, line 34: Main change 3/3

packages/gestalt/src/ModuleExpandable.js
```diff
@@ -32,7 +33,7 @@ type Props = {|
    * Array of modules displayed in a stack. Only one item can be expanded at a time. See [Expandable](https://gestalt.pinterest.systems/module#Expandable) variant to learn more.
    */
   items: $ReadOnlyArray<{|
-    badgeText?: string,
+    badge?: BadgeType,
```
ayeshakmaz, line 36: Main change 4/4

packages/gestalt/src/ModuleTitle.js
```diff
@@ -12,7 +12,7 @@ import Text from './Text.js';
  * https://gestalt.pinterest.systems/module
  */
 export default function ModuleTitle(props: {|
-  badgeText?: string,
+  badge?: BadgeType,
```
ayeshakmaz, line 15: Main change 5/5

packages/gestalt/src/OptionItem.js
```diff
@@ -21,7 +21,7 @@ export type OptionItemType = {|
 |};
 
 type Props = {|
-  badgeText?: string,
+  badge?: BadgeType,
```
ayeshakmaz, line 24: Main change 6/6

docs/pages/badge.js
```diff
@@ -274,7 +274,7 @@ function ModuleExample() {
   return (
     <Box column={12} maxWidth={800} padding={2}>
       <Module
-        badgeText="Beta"
+        badge={{text: "Beta"}}
```
dangerismycat, line 277: nittest of nits: this code doesn't get Prettier-ified but this should include spaces and use single quotes, like `{{ text: 'Beta' }}` (here and throughout)

docs/pages/module.js
```diff
@@ -292,7 +299,7 @@ function ModuleExample2() {
         </MainSection.Subsection>
 
         <MainSection.Subsection
-          title="ExpExpandable - Icon, Badge and IconButton"
```
dangerismycat, line 295: 🙀

packages/gestalt-codemods/46.0.0/new_badge_props.js
```diff
@@ -0,0 +1,95 @@
...
+    // Not Gestalt, bail
+    if (decl.source.value !== 'gestalt') {
+      return null;
+    }
+
+    // Find the local names of Dropdown and Module imports
+    localIdentifierNames = decl.specifiers
+      .filter((node) => ['Dropdown', 'Module'].includes(node.imported.name))
+      .map((node) => node.local.name);
+    return null;
```
dangerismycat, line 28: This return is useless, `.forEach` doesn't return anything

packages/gestalt-codemods/46.0.0/new_badge_props.js
```diff
@@ -0,0 +1,95 @@
...
+  });
+
+  // No Module or Dropdown imports, bail
+  if (!localIdentifierNames) {
```
dangerismycat, line 32: Given the `.filter` and `.map` calls, I believe `localIdentifierNames` will always at least be an empty array, which is truthy, so this check will never bail early. We need to check for array length instead

packages/gestalt/src/DropdownItem.jsdom.test.js
```diff
@@ -20,7 +20,7 @@ describe('Dropdown.Item', () => {
   test('adds badge', () => {
     render(
       <Dropdown.Item
-        badgeText="Beta Badge"
+        badge={{ text: 'Beta Badge' }}
```
dangerismycat, line 23: Should badges use sentence case? (I have no idea)

packages/gestalt/src/DropdownItem.js
dangerismycat: This type will need to be inlined everywhere it's used, as the codegen tool can't pick it up:
Screen Shot 2022-03-15 at 12 11 06 PM

dangerismycat (approved):

packages/gestalt-codemods/46.0.0/new_badge_props.js
ayeshakmaz: eslint complains if I remove it, I think it needs the arrow function to have a return regardless

packages/gestalt/src/DropdownItem.jsdom.test.js
ayeshakmaz: They probably should, all our docs do, just haven't worried about test values

packages/gestalt-codemods/46.0.0/new_badge_props.js
dangerismycat: Flow is likely complaining about an inconsistent return because other paths are returning `null`. Change those to just `return;` and Flow should be fine with removing the final return statement.

PR #1987, opened 2022-03-15
Callout: Update text to be readable in dark mode

ayeshakmaz:
#### What changed?

Use our new Text color for Callout to avoid the font color changing in dark mode

#### Why?

Previously the title and message of Callout were unreadable in dark mode. This updates the text color used to be one that does not change in dark mode, making the text readable

### Checklist

### DIFF
--- a/packages/gestalt/src/Callout.js
+++ b/packages/gestalt/src/Callout.js
@@ -157,7 +157,7 @@ function CalloutAction({
 /**
  * [Callout](https://gestalt.pinterest.systems/callout) is a banner displaying short messages with helpful information for a task on the page, or something that requires the user’s attention.
  *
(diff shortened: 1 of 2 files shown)


packages/gestalt/src/Callout.js
```diff
@@ -157,7 +157,7 @@ function CalloutAction({
 /**
  * [Callout](https://gestalt.pinterest.systems/callout) is a banner displaying short messages with helpful information for a task on the page, or something that requires the user’s attention.
  *
- * ⚠️ Please note: Callout is not currently supported in dark mode.
+ * ⚠️ Please note: Callout is not currently optimized for dark mode.
```
dangerismycat, line 160: progress! 😅

dangerismycat (approved):

PR #1989, opened 2022-03-16
Text: Add title prop

dangerismycat:
Currently in Text, we populate the `title` attribute on the element if the text is truncated using `lineClamp`, making the full text visible on hover. However, we only do this when `children` is a not Node, otherwise the title would show `[object Object]`.

This PR adds an optional prop `title` to Text, so that the title attribute can be set manually during truncation if `children` is a Node (e.g. when using Link).

### DIFF
--- a/docs/pages/text.js
+++ b/docs/pages/text.js
@@ -9,9 +9,11 @@ import MainSection from '../components/MainSection.js';
 
 export default function TextPage({ generatedDocGen }: {| generatedDocGen: DocGen |}): Node {
   return (
-    <Page title="Text">
-      <PageHeader name="Text" description={generatedDocGen?.description} />
+    <Page title={generatedDocGen?.displayName}>
+      <PageHeader name={generatedDocGen?.displayName} description={generatedDocGen?.description} />
+
       <GeneratedPropTable generatedDocGen={generatedDocGen} />
+
       <MainSection name="Variants">
         <MainSection.Subsection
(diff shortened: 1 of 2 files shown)


docs/pages/text.js
```diff
@@ -197,6 +206,63 @@ export default function TextPage({ generatedDocGen }: {| generatedDocGen: DocGen
...
         </MainSection.Subsection>
+
+        <MainSection.Subsection
+          title="Title"
+          description={`The \`title\` attribute on a \`<div>\` can be used to show the full text of a truncated string on hover. That attribute is populated automatically when the text is truncated using \`lineClamp\`, as long as \`children\` is a string.
+           If \`children\` is a \`React.Node\` (e.g. [when using Link](/link#Link-and-Text)), use the \`title\` prop to manually set the \`title\` attribute.`}
+        >
+          <MainSection.Card
+            defaultCode={`
+<Flex alignItems="center" direction="column" gap={2}>
```
ayeshakmaz, line 217: Seems like this could use some formatting 😅

ayeshakmaz (approved):
Looks good, just need to fix formatting issues

docs/pages/text.js
dangerismycat: whoa, Prettier killed my pretty formatting! 🙀

ayeshakmaz: !! strange!

PR #1993, opened 2022-03-16
InternalTextFieldIconButton: create component from InternalTextField

dangerismycat:
In preparation for future work using the icon button available on InternalTextField, this PR extracts that markup to a standalone component. This alters the API of InternalTextField a bit, and moves some ComboBox-specific logic back into ComboBox.


AlbertCarreras:
LGTM!

PR #1995, opened 2022-03-16
TextField: Add experimental show/hide password button

dangerismycat:
Currently TextField doesn't give users any way of viewing what they've typed so far when `type="password"`. This is a frustrating UX and is likely leading to failed logins. This PR adds an experimental "show/hide password" button to solve this problem.
Screen Shot 2022-03-17 at 5 32 06 PM
Screen Shot 2022-03-17 at 5 32 12 PM

### DIFF
--- a/docs/components/App.js
+++ b/docs/components/App.js
@@ -2,9 +2,11 @@
 import { useEffect, type Node } from 'react';
 import { ColorSchemeProvider, OnLinkNavigationProvider } from 'gestalt';
 import { useRouter } from 'next/router';
+import DocsExperimentProvider from './contexts/DocsExperimentProvider.js';
 import { AppContextProvider, AppContextConsumer } from './appContext.js';
 import { NavigationContextProvider } from './navigationContext.js';
 import AppLayout from './AppLayout.js';
+import DocsI18nProvider from './contexts/DocsI18nProvider.js';
 
 type Props = {|
   children?: Node,
@@ -55,7 +57,11 @@ export default function App({ children }: Props): Node {
           <ColorSchemeProvider colorScheme={colorScheme} id="gestalt-docs">
             <OnLinkNavigationProvider onNavigation={useOnNavigation}>
               <NavigationContextProvider>
-                <AppLayout>{children}</AppLayout>
+                <DocsExperimentProvider>
+                  <DocsI18nProvider>
+                    <AppLayout>{children}</AppLayout>
+                  </DocsI18nProvider>
+                </DocsExperimentProvider>
               </NavigationContextProvider>
             </OnLinkNavigationProvider>
           </ColorSchemeProvider>
--- a/docs/components/App.js
+++ b/docs/components/App.js
@@ -2,11 +2,9 @@
 import { useEffect, type Node } from 'react';
 import { ColorSchemeProvider, OnLinkNavigationProvider } from 'gestalt';
 import { useRouter } from 'next/router';
-import DocsExperimentProvider from './contexts/DocsExperimentProvider.js';
 import { AppContextProvider, AppContextConsumer } from './appContext.js';
 import { NavigationContextProvider } from './navigationContext.js';
 import AppLayout from './AppLayout.js';
-import DocsI18nProvider from './contexts/DocsI18nProvider.js';
 
 type Props = {|
   children?: Node,
@@ -50,6 +48,7 @@ export default function App({ children }: Props): Node {
     };
   }, [router.events]);
 
+  // NOTE: there are other Providers added in pages/_app.js
   return (
     <AppContextProvider>
       <AppContextConsumer>
@@ -57,11 +56,7 @@ export default function App({ children }: Props): Node {
           <ColorSchemeProvider colorScheme={colorScheme} id="gestalt-docs">
             <OnLinkNavigationProvider onNavigation={useOnNavigation}>
               <NavigationContextProvider>
-                <DocsExperimentProvider>
-                  <DocsI18nProvider>
-                    <AppLayout>{children}</AppLayout>
-                  </DocsI18nProvider>
-                </DocsExperimentProvider>
+                <AppLayout>{children}</AppLayout>
               </NavigationContextProvider>
             </OnLinkNavigationProvider>
           </ColorSchemeProvider>
--- a/docs/components/contexts/DocsExperimentProvider.js
+++ b/docs/components/contexts/DocsExperimentProvider.js
@@ -1,0 +1,39 @@
+// @flow strict
+import { type Node } from 'react';
+import { ExperimentProvider } from 'gestalt';
+
+/**
+ * To implement experimental behavior in the docs:
+ * - Add your experiment name here
+ * - Unless you want the experimental behavior live on the docs for everyone, REMOVE YOUR EXPERIMENT HERE before merging your PR!
+ * */
+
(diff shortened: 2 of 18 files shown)


docs/components/App.js
```diff
@@ -55,7 +57,11 @@ export default function App({ children }: Props): Node {
           <ColorSchemeProvider colorScheme={colorScheme} id="gestalt-docs">
             <OnLinkNavigationProvider onNavigation={useOnNavigation}>
               <NavigationContextProvider>
-                <AppLayout>{children}</AppLayout>
+                <DocsExperimentProvider>
```
dangerismycat, line 60: We have to establish these providers for the docs. The experiment provider is a way to pass experiment names while testing out a feature — a poor man's version of ModHeader for now, until we come up with a better solution. The i18n provider is necessary as we throw errors if the expected strings aren't provided.

docs/components/contexts/DocsExperimentProvider.js
```diff
@@ -0,0 +1,39 @@
...
+    }),
+    {},
+  );
+}
+
+type Props = {| children: Node |};
+
+export default function DocsExperimentProvider({ children }: Props): Node {
+  return (
+    <ExperimentProvider value={buildExperimentsObj(enabledExperiments)}>
```
dangerismycat, line 35: If more specific groups are needed, we can spread those into the object returned from the helper function, or possibly modify the function. Going with the simple solution for now for YAGNI reasons.

packages/gestalt/src/InternalTextFieldIconButton.js
```diff
@@ -4,31 +4,36 @@ import classnames from 'classnames';
...
 import TapArea from './TapArea.js';
+import Tooltip from './Tooltip.js';
 import styles from './InternalTextField.css';
 import { TAB, SPACE, ENTER } from './keyCodes.js';
 
 type Props = {|
-  accessibilityClearButtonLabel?: string,
   accessibilityHidden?: boolean,
-  hoverStyle: 'default' | 'none',
+  accessibilityLabel?: string,
```
dangerismycat, line 13: Some usability changes here:
- no need to be overly prescriptive about how the a11y label is used
- make hoverStyle and pogPadding optional with defaults
- added accessibilityPressed and role ("switch") per recommendation for this type of on/off switch button

packages/gestalt/src/TextField.js
```diff
@@ -117,18 +122,57 @@ const TextFieldWithForwardRef: React$AbstractComponent<Props, HTMLInputElement>
     placeholder,
     size = 'md',
     tags,
-    type = 'text',
+    type: typeProp = 'text',
     value,
   }: Props,
   ref,
 ): Node {
+  const [type, setType] = useState<Type>(typeProp);
```
dangerismycat, line 130: Yes, this is initializing a state variable with a prop value and then disregarding the prop value — often a code smell, I know. This is necessary to internalize the effective input type (password vs text) and not force the user to handle responding to clicks on the button.

packages/gestalt/src/contexts/ExperimentProvider.js
```diff
@@ -20,5 +20,5 @@ export default ExperimentProvider;
 
 export function useExperimentContext(experimentName: string): Experiment {
   const experiments = useContext(ExperimentContext) ?? {};
-  return experiments[experimentName];
+  return experiments[experimentName] ?? {};
```
dangerismycat, line 23: This allows the nice destructuring usage, which would otherwise throw if the experiment isn't defined:
```js
const { anyEnabled: myExpEnabled } = useExperimentContext('my_exp');
```

packages/gestalt-datepicker/src/DatePicker.jsdom.test.js
```diff
@@ -1,8 +1,20 @@
 // @flow strict-local
 import { useState } from 'react';
 import { render, screen, fireEvent } from '@testing-library/react';
+import { I18nProvider } from 'gestalt';
 import DatePicker from './DatePicker.js';
 
+const translations = {
```
dangerismycat, line 7: This is a bummer, but hopefully there aren't a lot of changes needed in the future here

packages/gestalt/src/TextField.jsdom.test.js
```diff
@@ -3,6 +3,8 @@ import { createRef } from 'react';
 import { fireEvent, render } from '@testing-library/react';
 import TextField from './TextField.js';
 
+jest.mock('./contexts/I18nProvider.js');
```
dangerismycat, line 6: I explored more global ways of applying this mock but came up short. We can revisit in the future, hopefully with more luck

packages/gestalt/src/TextField.js
rlingineni: This makes sense!

packages/gestalt/src/TapArea.js
```diff
@@ -45,16 +45,19 @@ type BaseTapArea = {|
...
 |};
+
 type TapAreaType = {|
   ...BaseTapArea,
   accessibilityControls?: string,
   accessibilityExpanded?: boolean,
   accessibilityHaspopup?: boolean,
-  role?: 'button',
+  accessibilityPressed?: boolean,
+  role?: 'button' | 'switch',
```
ayeshakmaz, line 60: For an item with role="switch", you'll also need to supply the aria-checked attribute https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/switch_role

packages/gestalt/src/InternalTextFieldIconButton.js
ayeshakmaz: I think here accessibilityPressed needs to be accessibilityChecked for role="switch" https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/switch_role

packages/gestalt/src/TapArea.js
dangerismycat: Ah, good catch. Just added

dangerismycat: Update: just _replaced_ 😅

docs/components/App.js
dangerismycat: (This now refers to the providers added in `_app.js`

packages/gestalt/src/TextField.js
```diff
@@ -117,18 +122,56 @@ const TextFieldWithForwardRef: React$AbstractComponent<Props, HTMLInputElement>
...
```
AlbertCarreras, line 130: Should  we add this comment the code to  give kept the context?

AlbertCarreras:
I was playing with it locally, and it seems to  me weird that the eye icon is display on password mode, but the hidden-eye icon is display on text  mode. 

Screen Shot 2022-03-30 at 6 39 16 PM
png">
Screen Shot 2022-03-30 at 6 42 24 PM

TikTok
Screen Shot 2022-03-30 at 6 42 03 PM

Screen Shot 2022-03-30 at 6 41 48 PM


Can we add some context from where we made our decision, just for context?

dangerismycat:
> Just for context, can we add some information/context from how we made this decision, decided to go this way?

@AlbertCarreras Good point. I went with what made sense to me, which designers approved but may not have been paying attention to which icon is used for a given state. I'll ask in the team channel to double-check.

PR #1996, opened 2022-03-17
PageHeader: PageHeader redesign v.2 (Part I: core changes)

AlbertCarreras:
#### What changed?

PageHeader now supports:
- Badge in title
- Helper button in title
- Thumbnail in title
- Second block for secondary components such as Datapoint or form fields
- Responsiveness to lg, md, sm, and mobile breakpoints
- Full vis diff test coverage

Follow up PRs:
- II: Improve unit testing
- III: Update Docs with updated documentation See preview here: https://deploy-preview-2048--gestalt.netlify.app/pageheader 

##### Before
Screen Shot 2022-03-18 at 8 52 53 PM

##### After
Screen Shot 2022-03-18 at 8 52 42 PM

#### Why?

Increase support to m10n teams needs and increase PageHeader adoptions

### Breaking Changes

PageHeader V.2 comes with breakpoint responsiveness. Under sm breakpoint, primary and secondary actions get displayed as options in a Dropdown.

`primaryAction` & `secondaryAction` have new Flow types:

BEFORE
Screen Shot 2022-04-06 at 6 39 11 PM

AFTER
Screen Shot 2022-04-06 at 6 38 52 PM

Run the following codemod to locate PageHeader with `primaryAction` & `secondaryAction` props so manual changes can be done.  

`yarn codemod detectManualReplacement ~/path/to/your/code --component=PageHeader --prop=primaryAction`
`yarn codemod detectManualReplacement ~/path/to/your/code --component=PageHeader --prop=secondaryAction`

### Links

- [TDD](https://paper.dropbox.com/doc/Gestalt-doc-PageHeader-ZF8W7ZGG7fXBzl8mGbZe7)
- [Figma](https://www.figma.com/file/weQzP4F2URMJFWQ1q5kaiU/Gestalt-PageHeader-2022?node-id=163%3A2503)

### Checklist


docs/pages/pageheader.js
```diff
@@ -658,7 +640,10 @@ function SecondaryActionExample() {
 }
 
 export async function getServerSideProps(): Promise<{| props: {| generatedDocGen: DocGen |} |}> {
+  const docGen = await docgen({ componentName: 'PageHeader' });
+  docGen.props.dropdownItems.flowType.raw =
+    '$ReadOnlyArray<React.Element<typeof DropdownItem | typeof DropdownLink>>>';
```
dangerismycat, line 645: typo: one too many `>` at the end

packages/gestalt/src/Dropdown.js
```diff
@@ -138,6 +138,7 @@ export default function Dropdown({
   const [hoveredItem, setHoveredItem] = useState<number>(0);
 
   const dropdownChildrenArray = Children.toArray(children);
+
```
dangerismycat, line 141: reverting this change will make for cleaner git history, making it clear when material changes were made to this component

packages/gestalt/src/DropdownItem.js
```diff
@@ -29,7 +29,8 @@ type Props = {|
   dataTestId?: string,
   /**
    * Callback when the user selects an item using the mouse or keyboard.
-   */ onSelect: ({|
+   */
```
dangerismycat, line 32: ditto here

packages/gestalt/src/PageHeaderSubcomponents.js
```diff
@@ -0,0 +1,200 @@
...
+export function PageHeaderBadge({
+  badgeText,
+  badgeTooltipText,
+}: {|
+  badgeText: string,
+  badgeTooltipText?: string,
+|}): Node {
+  return (
+    <Fragment>
+      {badgeText && !badgeTooltipText ? (
```
dangerismycat, line 56: Consider simplifying these checks by changing the conditional render in PageHeader to look for `badge.badgeText` specifically, since that's the actual determinant for if this renders

(also badgeText is a required prop, so these checks are especially unnecessary)

packages/gestalt/src/PageHeaderSubcomponents.js
```diff
@@ -0,0 +1,200 @@
...
+      event:
+        | SyntheticMouseEvent<HTMLButtonElement>
+        | SyntheticKeyboardEvent<HTMLButtonElement>
+        | SyntheticMouseEvent<HTMLAnchorElement>
+        | SyntheticKeyboardEvent<HTMLAnchorElement>,
+      dangerouslyDisableOnNavigation: () => void,
+    |}) => void,
+  |},
+|}): Node {
+  return helperIconButton ? (
```
dangerismycat, line 85: This check is redundant — it's already happening in PageHeader. This subcomponent can always assume `helperIconButton` will be defined.

packages/gestalt/src/PageHeaderSubcomponents.js
```diff
@@ -0,0 +1,200 @@
...
+        </Tooltip>
+      ) : null}
+    </Fragment>
+  );
+}
+
+export function PageHeaderHelperIconButton({
+  helperIconButton,
+}: {|
+  helperIconButton?: {|
```
dangerismycat, line 71: Rather than pass in the whole object, consider passing in each item. From this component's perspective, it's concerned about the individual props within `helperIconButton`, not the object as a whole

packages/gestalt/src/PageHeaderSubcomponents.js
```diff
@@ -0,0 +1,200 @@
...
+      size="xs"
+    />
+  ) : null;
+}
+
+export function PageHeaderSubtext({
+  subtext,
+  helperLink,
+}: {|
+  subtext?: string,
```
dangerismycat, line 104: This prop should be required

packages/gestalt/src/PageHeaderSubcomponents.js
```diff
@@ -0,0 +1,200 @@
...
+    accessibilityLabel: string,
+    href: string,
+    onClick: ({|
+      event: SyntheticMouseEvent<HTMLAnchorElement> | SyntheticKeyboardEvent<HTMLAnchorElement>,
+      dangerouslyDisableOnNavigation: () => void,
+    |}) => void,
+  |},
+|}): Node {
+  return (
+    <Box wrap display="none" smDisplay="block">
```
dangerismycat, line 116: `wrap` doesn't do anything without `display="flex"`

packages/gestalt/src/PageHeaderSubcomponents.js
```diff
@@ -0,0 +1,200 @@
...
+      <Text inline>
+        {subtext}{' '}
```
dangerismycat, line 118: Consider using `gap` for spacing instead of whitespace, since it will be conditionally applied only when helperLink exists

packages/gestalt/src/PageHeaderSubcomponents.js
```diff
@@ -0,0 +1,200 @@
...
+        {helperLink ? (
+          <Link
+            inline
+            accessibilityLabel={helperLink.accessibilityLabel}
+            href={helperLink.href}
+            onClick={helperLink.onClick}
+            target="blank"
+          >
+            <span style={{ textDecoration: 'underline' }}>{helperLink.text}</span>
```
dangerismycat, line 127: `<Text underline>`?

packages/gestalt/src/PageHeaderSubcomponents.js
```diff
@@ -0,0 +1,200 @@
...
+          </Link>
+        ) : null}
+      </Text>
+    </Box>
+  );
+}
+
+export function PageHeaderActionBlock({
```
dangerismycat, line 135: This API is overly permissive and will allow devs to unwittingly create bad UIs. We need to guard against several scenarios:
- Only secondary action with no primary action (I could be wrong, we may want to allow this)
- Only actions with no dropdown items
- Only dropdown items with no actions
- Mismatch between actions and dropdown items
- No label provided

We should check for these scenarios and throw errors if needed.

I wonder if tighter coupling of actions and dropdown items in the API would make that connection better. Perhaps each action takes both a button and dropdown item in the same prop, to ensure that devs provide both (two required fields on the value of the optional prop)

packages/gestalt/src/PageHeaderSubcomponents.js
```diff
@@ -0,0 +1,200 @@
...
+              {dropdownItems}
+            </Dropdown>
+          )}
+        </Box>
+      ) : null}
+    </Fragment>
+  );
+}
+
+export function PageHeaderItemsBlock({ items }: {| items?: $ReadOnlyArray<Node> |}): Node {
```
dangerismycat, line 191: This component should assume items will be defined, with a check at the callsite in PageHeader (probably checking to make sure items has length)

dangerismycat (changes requested):
Is the 1px height difference in the existing snapshots expected?

It also looks like a number of snapshots are adding significant whitespace below the component — is that expected?
Screen Shot 2022-03-24 at 2 59 54 PM

The alignment with the badge and icon button looks off…is this intended?
Screen Shot 2022-03-24 at 3 02 01 PM

Can we set a width on the parent of this example (and possibly others) to show how the component will display on large viewports?
Screen Shot 2022-03-24 at 3 06 01 PM
These also look pretty odd:
Screen Shot 2022-03-24 at 3 31 27 PM

AlbertCarreras:
> Is the 1px height difference in the existing snapshots expected?
> 
> It also looks like a number of snapshots are adding significant whitespace below the component — is that expected? Screen Shot 2022-03-24 at 2 59 54 PM
> 

Yes, that's because I need space below the PageHeader to capture the Dropdown on xs breakpoint


> The alignment with the badge and icon button looks off…is this intended? Screen Shot 2022-03-24 at 3 02 01 PM
> 
> Can we set a width on the parent of this example (and possibly others) to show how the component will display on large viewports? Screen Shot 2022-03-24 at 3 06 01 PM These also look pretty odd: Screen Shot 2022-03-24 at 3 31 27 PM

packages/gestalt/src/Dropdown.js
AlbertCarreras: DONE

packages/gestalt/src/DropdownItem.js
AlbertCarreras: DONE

docs/pages/pageheader.js
AlbertCarreras: DONE

packages/gestalt/src/PageHeaderSubcomponents.js
AlbertCarreras: DONE

packages/gestalt/src/PageHeaderSubcomponents.js
AlbertCarreras: DONE

packages/gestalt/src/PageHeaderSubcomponents.js
AlbertCarreras: DONE

packages/gestalt/src/PageHeaderSubcomponents.js
AlbertCarreras: Underline breaks with inline. 'Cos we changed to Flex, then we can use <Text underline>

AlbertCarreras: DONE

packages/gestalt/src/PageHeaderSubcomponents.js
AlbertCarreras: DONE

packages/gestalt/src/PageHeaderSubcomponents.js
AlbertCarreras: @dangerismycat 

Decided to go the error route. 

"Mismatch between actions and dropdown items" it's hard cos Action can be IconButton with Dropdown and several items there. So it isn't one to one situation.

packages/gestalt/src/PageHeader.css
```diff
@@ -0,0 +1,9 @@
+.pageHeader {
```
ayeshakmaz, line 1: For this, I'd recommend using
border-bottom: 1px solid var(--color-border-container)

Then you wont need separate CSS for dark mode

AlbertCarreras: DONE

packages/gestalt/src/PageHeaderSubcomponents.js
AlbertCarreras: Just remembered why I didn't use Flex
Screen Shot 2022-03-30 at 8 03 37 PM

packages/gestalt/src/PageHeaderComponents.js
```diff
@@ -0,0 +1,210 @@
...
+      >
+        <Heading size="500" lineClamp={1} accessibilityLevel={1}>
+          {title}
+        </Heading>
+      </Box>
+    </Fragment>
+  );
+}
+
+export function PageHeaderThubnail({ thumbnail }: {| thumbnail?: Element<typeof Image> |}): Node {
```
dangerismycat, line 37: typo: PageHeaderThumbnail (missing 'm')

also: the prop should be required

packages/gestalt/src/PageHeaderComponents.js
```diff
@@ -0,0 +1,210 @@
...
+        </Tooltip>
+      ) : null}
+    </Fragment>
+  );
+}
+
+export function PageHeaderHelperIconButton({
+  helperIconButton,
+}: {|
+  helperIconButton: {|
```
dangerismycat, line 71: This is still a weird pattern. It's basically like
```js
<Component componentProps={componentProps} />
```
From the child component's perspective, that makes no sense — its props are needlessly wrapped in a redundant object that just repeats the component's name. This should really be
```js
<PageHeaderHelperIconButton
  accessibilityControls={accessibilityControls}
  accessibilityExpanded={accessibilityExpanded}
  onClick={onClick}
/>
```

packages/gestalt/src/PageHeaderComponents.js
```diff
@@ -0,0 +1,210 @@
...
+      />
+    </Box>
+  );
+}
+
+export function PageHeaderSubtext({
+  subtext,
+  helperLink,
+}: {|
+  subtext?: string,
```
dangerismycat, line 106: `subtext` should be required

packages/gestalt/src/PageHeaderSubcomponents.js
dangerismycat: Are you referring to "Learn more" wrapping? You should be able to fix that by wrapping in `<Flex.Item flex="none">`

packages/gestalt/src/PageHeaderComponents.js
```diff
@@ -0,0 +1,210 @@
...
+}: {|
+  primaryAction: ActionType,
+  secondaryAction?: ActionType,
+  dropdownItems?: $ReadOnlyArray<Node>,
+  dropdownAccessibilityLabel?: string,
+|}): Node {
+  const [open, setOpen] = useState(false);
+  const anchorRef = useRef(null);
+
+  if (!dropdownItems || !dropdownAccessibilityLabel) {
```
dangerismycat, line 153: I think we can improve our guidance here…what if the user provides `primaryAction` and `secondaryAction` but only one dropdown item? What if they provide `primaryAction` and three dropdown items? Or what if the order of the dropdown items doesn't match `[primary, secondary]`? This API gives the user a lot of ways to shoot themselves in the foot, which isn't great. Consider how the dropdown items can be more tightly coupled to their respective actions — perhaps as part of the action props rather than a separate prop.

packages/gestalt/src/PageHeaderComponents.js
```diff
@@ -0,0 +1,210 @@
...
+    </Fragment>
+  );
+}
+
+export function PageHeaderItemsBlock({ items }: {| items: $ReadOnlyArray<Node> |}): Node {
+  return (
+    <Box display="none" lgDisplay="block" overflow="hidden">
+      <Flex gap={2}>
+        {items?.[0] ? items?.[0] : null}
+        {items?.[1] ? items?.[1] : null}
```
dangerismycat, line 206: `items` is required, no need for optional chaining here

dangerismycat (changes requested):
I think the actions-dropdown items API still needs work to make sure that users are guided towards the correct path rather than given the freedom to give incorrect options

packages/gestalt/src/PageHeaderComponents.js
AlbertCarreras: DONE

packages/gestalt/src/PageHeaderComponents.js
AlbertCarreras: DONE

packages/gestalt/src/PageHeaderComponents.js
AlbertCarreras: DONE

packages/gestalt/src/PageHeaderComponents.js
AlbertCarreras: DONE

packages/gestalt/src/PageHeaderComponents.js
AlbertCarreras: DONE

packages/gestalt/src/PageHeader.js
```diff
@@ -4,25 +4,83 @@ import Box from './Box.js';
...
    */
   maxWidth?: Dimension,
   /**
-   * The primary action of the page. Can be [Button](https://gestalt.pinterest.systems/button), [Link](https://gestalt.pinterest.systems/link), [Tooltip](https://gestalt.pinterest.systems/tooltip) surrounding IconButton or a combination of IconButton, Tooltip and [Dropdown](https://gestalt.pinterest.systems/dropdown).
+   * The primary action of the page. Can be [Button](https://gestalt.pinterest.systems/button), [Link](https://gestalt.pinterest.systems/link), [Tooltip](https://gestalt.pinterest.systems/tooltip) surrounding IconButton or a combination of IconButton, Tooltip and [Dropdown](https://gestalt.pinterest.systems/dropdown). Primary and secondary actions are replaced with a [Dropdown](https://gestalt.netlify.app/dropdown) under the [sm breakpoint](https://gestalt.netlify.app/screen_sizes#Web-(px)). `primaryAction` takes both the main component and its equivalent using Dropdown subcomponents.
    */
-  primaryAction?: Element<typeof Button | typeof IconButton | typeof Link | typeof Tooltip>,
+  primaryAction?: {| component: ActionType, dropdownItems: $ReadOnlyArray<Node> |},
```
dangerismycat, line 79: I like this tighter coupling of the action and dropdown item. However: do we have a need to allow multiple dropdown items for a given action? Can we accept a Node rather than an array of Nodes for that field?

packages/gestalt/src/PageHeaderComponents.js
```diff
@@ -0,0 +1,212 @@
...
+    <Fragment>
+      <Box display="block" smDisplay="none">
+        <Heading size="400" lineClamp={1} accessibilityLevel={1}>
+          {title}
+        </Heading>
+      </Box>
+      <Box
+        display="none"
+        smDisplay="block"
+        dangerouslySetInlineStyle={{ __style: hasThumbnail ? { marginTop: '-5px' } : {} }}
```
dangerismycat, line 33: This seems like a leaky abstraction: this component is too aware of the context it's used in. Consider moving this margin to the markup surrounding PageHeaderTitle, which reduces the scope of this component and allows removing the hasThumbnail prop.

packages/gestalt/src/PageHeaderComponents.js
```diff
@@ -0,0 +1,212 @@
...
+export function PageHeaderBadge({
+  badgeText,
+  badgeTooltipText,
+}: {|
+  badgeText: string,
+  badgeTooltipText?: string,
+|}): Node {
+  return (
+    <Fragment>
+      {badgeText && !badgeTooltipText ? (
```
dangerismycat, line 62: badgeText is required, so these checks are superfluous

packages/gestalt/src/PageHeaderComponents.js
```diff
@@ -0,0 +1,212 @@
...
+}
+
+export function PageHeaderActionBlock({
+  primaryAction,
+  secondaryAction,
+  dropdownAccessibilityLabel = '',
+}: {|
+  primaryAction: {| component: ActionType, dropdownItems: $ReadOnlyArray<Node> |},
+  secondaryAction?: {| component: ActionType, dropdownItems: $ReadOnlyArray<Node> |},
+  dropdownAccessibilityLabel?: string,
```
dangerismycat, line 156: Can we add an error check somewhere to ensure that the user always passes dropdownAccessibilityLabel when actions are provided?

packages/gestalt/src/PageHeader.js
```diff
@@ -31,48 +89,90 @@ type Props = {|
...
+                  </Flex>
+                </Flex>
               </Box>
-            </Flex>
-          )}
+            </Flex.Item>
+            <Flex.Item minWidth={0} flex="none">
+              <Flex gap={4} alignItems={subtext ? undefined : 'center'} height="100%">
+                {items && items.length !== 0 ? <PageHeaderItemsBlock items={items} /> : null}
+                {primaryAction ? (
```
dangerismycat, line 164: Is it a valid assumption that we don't allow a user to pass only a secondary action? What if the page's primary CTA is elsewhere and not in the header?

packages/gestalt/src/PageHeaderComponents.js
```diff
@@ -0,0 +1,212 @@
...
+      </Box>
+    </Fragment>
+  );
+}
+
+export function PageHeaderItemsBlock({ items }: {| items: $ReadOnlyArray<Node> |}): Node {
+  return (
+    <Box display="none" lgDisplay="block" overflow="hidden">
+      <Flex gap={2}>
+        {items?.[0] ? items[0] : null}
```
dangerismycat, line 207: This was marked as "done" but there's still unnecessary optional chaining here for the required prop `items`

packages/gestalt/src/PageHeaderComponents.js
```diff
@@ -0,0 +1,212 @@
...
+    |}) => void,
+  |},
+|}): Node {
+  return (
+    <Box display="none" smDisplay="block">
+      <Text inline>
+        {subtext}
+        {helperLink ? (
+          <Fragment>
+            {' '}
```
dangerismycat, line 130: Did you try using Flex with `<Flex.Item flex="none">` around this block to allow using `gap` and avoid this hard-coded whitespace?

packages/gestalt/src/PageHeaderComponents.js
```diff
@@ -0,0 +1,212 @@
...
+    event:
+      | SyntheticMouseEvent<HTMLButtonElement>
+      | SyntheticKeyboardEvent<HTMLButtonElement>
+      | SyntheticMouseEvent<HTMLAnchorElement>
+      | SyntheticKeyboardEvent<HTMLAnchorElement>,
+    dangerouslyDisableOnNavigation: () => void,
+  |}) => void,
+|}): Node {
+  return (
+    <Box dangerouslySetInlineStyle={{ __style: { marginTop: '-1px' } }} rounding="circle">
```
dangerismycat, line 93: Applying negative margins on the outermost element of a sub-component is a bit of an anti-pattern…consider applying this where the sub-component is used instead.

packages/gestalt/src/PageHeader.js
```diff
@@ -31,48 +89,90 @@ type Props = {|
...
-          </Box>
+  const { text: badgeText, tooltipText: badgeTooltipText } = badge || {};
 
-          {primaryAction && (
-            <Flex alignItems="center" gap={2} flex="none">
-              {secondaryAction}
-              {/* 48px height needed to maintain proper sizing when action is a Link */}
-              <Box alignItems="center" display="flex" marginTop={4} marginBottom={4} height="48px">
-                {primaryAction}
+  return (
```
dangerismycat, line 123: It looks like the breakpoints still need some work:
Screen Shot 2022-04-04 at 10 50 36 AM

dangerismycat (changes requested):

packages/gestalt/src/PageHeader.js
AlbertCarreras: If we are using IconButton as primary icon to open a Dropdown, we need an Array of Nodes to pass all the options in the Dropdown.
Screen Shot 2022-04-04 at 3 37 43 PM

packages/gestalt/src/PageHeaderComponents.js
AlbertCarreras: DONE

packages/gestalt/src/PageHeaderComponents.js
AlbertCarreras: I'm gonna add a default `dropdownAccessibilityLabel` with i18n provider, that should cover it. I could triple check in case there's no default, but we don't do that anywhere in the code.

packages/gestalt/src/PageHeader.js
AlbertCarreras: Not allowed based on designs.

packages/gestalt/src/PageHeaderComponents.js
AlbertCarreras: DONE

packages/gestalt/src/PageHeaderComponents.js
AlbertCarreras: Screen Shot 2022-04-04 at 3 50 23 PM

packages/gestalt/src/PageHeaderComponents.js
AlbertCarreras: DONE

packages/gestalt/src/PageHeader.js
AlbertCarreras: Breakpoints aren't based on container width but window width.

Screen Shot 2022-04-04 at 4 01 58 PM
 be used

packages/gestalt/src/PageHeaderComponents.js
AlbertCarreras: Thumbnails are hidden on small breakpoints, so I need to apply the margin only on that case.

packages/gestalt/src/PageHeader.js
dangerismycat: Right, but that can be addressed where the Node(s) from the actions are passed to Dropdown. Is there a need for this field to accept an array? Is there a situation where the user needs to pass multiple dropdown items for a given action?

cypress/integration/accessibility_PageHeader_spec.js
```diff
@@ -12,6 +12,10 @@ describe('PageHeader Accessibility check', () => {
           id: 'heading-order',
           enabled: false,
         },
+        {
+          id: 'color-contrast',
```
dangerismycat, line 16: Why does this rule need to be disabled? Can we document that reason in the code?

packages/gestalt/src/PageHeaderComponents.js
```diff
@@ -0,0 +1,204 @@
...
+    </Fragment>
+  );
+}
+
+export function PageHeaderItemsBlock({ items }: {| items: $ReadOnlyArray<Node> |}): Node {
+  return (
+    <Box display="none" lgDisplay="block" overflow="hidden">
+      <Flex gap={2}>
+        {items[0] ? items[0] : null}
+        {items?.[1] ? items[1] : null}
```
dangerismycat, line 200: This optional chaining also isn't necessary — if `items` is defined on the previous line, it's defined here too

Though really: why does this need to be so verbose? This can be much more functional:
```js
<Flex gap={2}>
  {items.slice(0, 2).map((item, i) => <Flex.Item key={i}>{item}</Flex.Item>)}
</Flex>
```

packages/gestalt/src/PageHeaderComponents.js
dangerismycat: Can you be more specific about what isn't matching the design spec here? It's difficult for me to offer suggestions based on an out-of-context screenshot

cypress/integration/accessibility_PageHeader_spec.js
AlbertCarreras: Screen Shot 2022-04-04 at 7 28 39 PM

packages/gestalt/src/PageHeaderComponents.js
AlbertCarreras: DONE

packages/gestalt/src/PageHeaderComponents.js
```diff
@@ -0,0 +1,205 @@
...
+    onClick: ({|
+      event: SyntheticMouseEvent<HTMLAnchorElement> | SyntheticKeyboardEvent<HTMLAnchorElement>,
+      dangerouslyDisableOnNavigation: () => void,
+    |}) => void,
+  |},
+|}): Node {
+  return (
+    <Box display="none" smDisplay="block">
+      <Text inline>
+        {subtext}{' '}
```
AlbertCarreras, line 120: image

 vs 

Screen Shot 2022-04-04 at 7 45 42 PM

The wrap is different using Flex. It's not wrapping as a single line of Text.

dangerismycat: Ah I see what you're saying 👍

dangerismycat (approved):

PR #2000, opened 2022-03-18
ExperimentProvider: create new Context

dangerismycat:
[plan for experimentation in Gestalt](https://paper.dropbox.com/doc/Experimentation-in-Gestalt--BddgXa8eh85dBvzWzoUx3ki0Ag-SG2FCNEsklOe5lJ7gznTa)

tl;dr of that ⬆️ doc: we need a better way of running experiments on Gestalt components. Undocumented props (`_expProp`) aren't great.

This PR creates a new ExperimentProvider which I'll implement in Pinboard and start playing around with. Similar to DeviceTypeProvider, this is an alpha and the API may change considerably in the near future, so I'm leaving it undocumented for now. I'll add documentation when our story around experimentation in Gestalt components is a bit more solidified.

### DIFF
--- a/packages/gestalt/src/contexts/ExperimentProvider.js
+++ b/packages/gestalt/src/contexts/ExperimentProvider.js
@@ -1,0 +1,24 @@
+// @flow strict
+import { type Context, createContext, useContext } from 'react';
+
+type Experiment = {|
+  anyEnabled: boolean,
+  group: string,
+|};
+
+type ExperimentContextType = {|
+  [experimentName: string]: Experiment,
+|};
+
+const ExperimentContext: Context<ExperimentContextType> = createContext<ExperimentContextType>({});
+
+/**
+ * *ALPHA - DO NOT USE YET - MAY HAVE BREAKING CHANGES IN THE NEAR FUTURE*
+ */
+const ExperimentProvider = ExperimentContext.Provider;
+export default ExperimentProvider;
+
+export function useExperimentContext(experimentName: string): Experiment {
+  const experiments = useContext(ExperimentContext) ?? {};
+  return experiments[experimentName];
+}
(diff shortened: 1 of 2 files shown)


packages/gestalt/src/contexts/ExperimentProvider.js
```diff
@@ -0,0 +1,24 @@
...
+  [experimentName: string]: Experiment,
+|};
+
+const ExperimentContext: Context<ExperimentContextType> = createContext<ExperimentContextType>({});
+
+/**
+ * *ALPHA - DO NOT USE YET - MAY HAVE BREAKING CHANGES IN THE NEAR FUTURE*
+ */
+const ExperimentProvider = ExperimentContext.Provider;
+export default ExperimentProvider;
```
AlbertCarreras, line 19: For context, both Color and LinkNavigation providers, follow the pattern of custom-named props. If we export directly the prop is just value, which is the default provider prop.

https://github.com/pinterest/gestalt/pull/1996/files#diff-a7d0620f7dbfed6bca89e822ad18205902f10845a7039f4fa28028b464947faa

https://github.com/pinterest/gestalt/pull/1996/files#diff-a7d0620f7dbfed6bca89e822ad18205902f10845a7039f4fa28028b464947faa

Any thoughts?

AlbertCarreras: Not concerned about this implementation per se, just the consistency of implementations.

AlbertCarreras (approved):
Cos' it's just a context and undocumented, LGTM. We can iterate API etc as we move on.

packages/gestalt/src/contexts/ExperimentProvider.js
dangerismycat: I think OnLinkNavigationProvider probably could've used the same pattern of exporting the Provider directly, since onNavigation is passed directly through to `value`. ColorSchemeProvider is different, since it contains additional logic and markup. Unless I'm misunderstanding what you're asking about?

AlbertCarreras: You're right that it's not strictly necessary;

I was just referring to the pattern and the benefit of having autoGenerated prop tables for free based on the props.

PR #2001, opened 2022-03-18
I18nProvider: create new Context

dangerismycat:
Consider this PR more of a discussion than a statement.

We've talked in the past about the annoyance of certain props that exist only to get translations for common strings, like Sheet's `accessibilityDismissButtonLabel`, SearchField's `accessibilityClearButtonLabel`, etc. This PR creates a new Context to attempt to solve this issue: we provide translations for these strings at the root level of the app, then read that context to get the string value within the component. This will allow us to remove existing props and obviate the need for new ones for new features/components.

The shape of the Context value is just a suggestion. I think most of these annoying props are at the base level of component APIs (not a field on a prop), but if that assumption doesn't prove true then we'll need to revisit the Context value shape to accommodate those nested translations.


packages/gestalt/src/contexts/I18nProvider.js
```diff
@@ -0,0 +1,23 @@
+// @flow strict
+import { type Context, createContext, useContext } from 'react';
+
+type PropTranslations = {|
+  [propName: string]: string,
```
christianvuerings, line 5: Please be more strict around which translations are required for which components

liuyenwei:
> This PR creates a new Context to attempt to solve this issue: we provide translations for these strings at the root level of the app, then read that context to get the string value within the component. This will allow us to remove existing props and obviate the need for new ones for new features/components.

Is the idea here that we would insert _all_ translations at the root of the app for the various accessibilityLabels we have in the app?  i.e. let's say a certain tree has 4 `Sheet` components, each one currently needing its own `accessibilityDismissButtonLabel` prop.  Would this approach mean that we use the same label for each one of those instances or would we inject all four different translations at the root of the tree?  

Based on the changes in this PR, my assumption is the former (all `Sheets` use the same `accessibilityDismissButtonLabel` value) but I wonder if that would still allow us to provide a good experience from an accessibility perspective?

packages/gestalt/src/contexts/I18nProvider.js
```diff
@@ -0,0 +1,40 @@
...
+ * - When consuming useI18nContext in component, throw an error if any translation isn't available
+ */
+
+type TextTranslations = {|
+  accessibilityHidePasswordLabel: string,
+  accessibilityShowPasswordLabel: string,
+|};
+
+type I18nContextType = ?{|
+  Text: TextTranslations,
```
AlbertCarreras, line 18: Just a thought. A single type might be easier to scan if this keeps growing.  Is there any particular reason to break down in cmp types rather than inlining?

dangerismycat: Can you describe your idea for inlining?

AlbertCarreras (approved):
LGTM, added nit comments. We can iterate on it.

dangerismycat:
> > This PR creates a new Context to attempt to solve this issue: we provide translations for these strings at the root level of the app, then read that context to get the string value within the component. This will allow us to remove existing props and obviate the need for new ones for new features/components.
> 
> Is the idea here that we would insert _all_ translations at the root of the app for the various accessibilityLabels we have in the app? i.e. let's say a certain tree has 4 `Sheet` components, each one currently needing its own `accessibilityDismissButtonLabel` prop. Would this approach mean that we use the same label for each one of those instances or would we inject all four different translations at the root of the tree?
> 
> Based on the changes in this PR, my assumption is the former (all `Sheets` use the same `accessibilityDismissButtonLabel` value) but I wonder if that would still allow us to provide a good experience from an accessibility perspective?

That's the idea, yeah. This wouldn't replace every accessibility label, just the ones that will always be the same (e.g. "Close modal", "Show password", etc). The goal is to reduce the prop burden at each individual callsite where possible.

There's also nothing precluding us from overriding these values and treating the Context translations as a default, allowing for more specific labels where helpful. If the component both registers a label with this Context and accepts an optional prop value, we could make (translated) "Close modal" the default if no prop is provided, but allow setting "Close PinBuilder modal" at the callsite.

AlbertCarreras:
> > This PR creates a new Context to attempt to solve this issue: we provide translations for these strings at the root level of the app, then read that context to get the string value within the component. This will allow us to remove existing props and obviate the need for new ones for new features/components.
> 
> Is the idea here that we would insert _all_ translations at the root of the app for the various accessibilityLabels we have in the app? i.e. let's say a certain tree has 4 `Sheet` components, each one currently needing its own `accessibilityDismissButtonLabel` prop. Would this approach mean that we use the same label for each one of those instances or would we inject all four different translations at the root of the tree?
> 
> Based on the changes in this PR, my assumption is the former (all `Sheets` use the same `accessibilityDismissButtonLabel` value) but I wonder if that would still allow us to provide a good experience from an accessibility perspective?

Not all accessibility labels would go here. Just some accessibility labels that  could all be the same text. For example, we are planning on adding an appendix to the accessibilityLabels in  Links that are external that would say "opens in an external link". That labels should be identical in all Links, and the regular engineers shouldn't worry about adding an extra prop. This solutions let us handle a few cases like this were we can built the labels in for free, and handle their translations.

liuyenwei:
> That's the idea, yeah. This wouldn't replace every accessibility label, just the ones that will always be the same (e.g. "Close modal", "Show password", etc). The goal is to reduce the prop burden at each individual callsite where possible.

> There's also nothing precluding us from overriding these values and treating the Context translations as a default, allowing for more specific labels where helpful. If the component both registers a label with this Context and accepts an optional prop value, we could make (translated) "Close modal" the default if no prop is provided, but allow setting "Close PinBuilder modal" at the callsite.

> Not all accessibility labels would go here. Just some accessibility labels that could/should all be the same. For example, we are planning on adding an appendix to the accessibilityLabels in Links that are external that would say "opens in an external link". That labels "apendix" should be identical in all Links, and the regular engineers shouldn't worry about adding an extra prop. This solutions let us handle a few cases like this were we can built the labels in for free, and handle their translations.

Gotcha, thanks for the clarification.  I like the idea of treating these more as "default" values and allowing callsites to override since that provides consumers the flexibility of giving a component a more specific label. 

The approach here should be fine with how we handle translations in Pinboard.  We're actively trying to move away from front-loading translations in the App (part of the reason for my question) but given the total number of translations this would likely have, I think that's okay.

packages/gestalt/src/contexts/I18nProvider.js
liuyenwei: suggestion - you might be able to do something like this so you can refine valid inputs/outputs for this:
```
type TextTranslations = {|
  Text: {| 
      accessibilityHidePasswordLabel: string,
  |},
  Sheet: {|
      somethingSomethingLabel: string
   |}
|}

type ValidComponent = $Keys<TextTranslations>
...
function useI18nContext<C: ValidComponent>(componentName: C): $ElementType<TextTranslations, C> 
```

i _feel_ like that should work but don't quote me on that 😅

dangerismycat: nm, figured it out!

packages/gestalt/src/contexts/I18nProvider.js
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  // Missing translation for a given string
+  const translationPropNames = Object.keys(componentTranslations);
+  const missingTranslations = translationPropNames.filter(
+    (propName) => typeof componentTranslations[propName] !== 'string',
+  );
+  if (missingTranslations.length > 0) {
+    const multipleMissing = missingTranslations.length > 1;
+    throw new Error(
```
christianvuerings, line 57: @dangerismycat could you add flow tests & jest unit tests for the i18n provider?

dangerismycat: I assume you mean the Hook — will do. I hadn't initially because it was so simple but agreed that now it has enough logic to test.

PR #2006, opened 2022-03-21
Text: fix pesky Flow issue with colors

dangerismycat:
🤞


dangerismycat:
abandoning, fixed by https://github.com/pinterest/gestalt/pull/2007

PR #2007, opened 2022-03-21
Internal: Fix flaky flow issue on Text

christianvuerings:
### Summary

Fixes a flaky flow issue on `Text`.

### Why

To unblock other developers

### Background

Flow doesn't understand the following line:
```
const colorName = semanticColors.includes(color) ? `${color}Text` : color;
```
even though `warning` is included in `semanticColors`, so the colorName would always be `warningText`, it still thinks `colorName` can be `warning`, which is not the case.

image

The fix is to manually remove `success` / `warning` / `subtle` / `light` / `error` / `dark` in the `if` check.

Ideally though, we should finish the migration and remove the literal colors

### DIFF
--- a/packages/gestalt/src/Text.js
+++ b/packages/gestalt/src/Text.js
@@ -122,7 +122,15 @@ export default function Text({
 }: Props): Node {
   let colorClass = null;
   const colorName = semanticColors.includes(color) ? `${color}Text` : color;
-  if (allowedColors.includes(color)) {
+  if (
+    allowedColors.includes(color) &&
+    colorName !== 'dark' &&
+    colorName !== 'error' &&
+    colorName !== 'light' &&
+    colorName !== 'subtle' &&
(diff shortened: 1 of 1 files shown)


packages/gestalt/src/Text.js
```diff
@@ -122,7 +122,15 @@ export default function Text({
 }: Props): Node {
   let colorClass = null;
   const colorName = semanticColors.includes(color) ? `${color}Text` : color;
-  if (allowedColors.includes(color)) {
+  if (
+    allowedColors.includes(color) &&
+    colorName !== 'dark' &&
```
christianvuerings, line 127: Note that `!['dark', 'error', 'light', 'subtle', 'success', 'warning'].includes(colorName)` doesn't work

ayeshakmaz (approved):
Thank you!!

PR #2011, opened 2022-03-22
Docs: implemented OnLinkNavigationProvider in Docs to extend next/router to all Gestalt components used in Docs

AlbertCarreras:
#### What changed?

Docs: implemented OnLinkNavigationProvider in Docs to extend next/router to all Gestalt components used in Docs

#### Why?

Prevent homepage illustration cards and upcoming component overview illustration cards from doing full refresh on redirect


docs/components/App.js
```diff
@@ -13,6 +13,25 @@ type Props = {|
 export default function App({ children }: Props): Node {
   const router = useRouter();
 
+  // $FlowIssue[prop-missing]
+  const isLeftClickEvent = (event) => event.button === 0; // ignore everything but left clicks
```
dangerismycat, line 17: Can we add a type for `event` to fix the Flow error?

dangerismycat (approved):

docs/components/App.js
AlbertCarreras: There's a bigger issue here with the event in the provider. 
Leaving FlowIssue as I think I'd have to touch the event in the provider and this can easily get out of scope.
Screen Shot 2022-03-23 at 4 41 28 PM

AlbertCarreras: There's a bigger issue here with the event in the provider. 
Leaving FlowIssue as I think I'd have to touch the event in the provider and this can easily get out of scope.
Screen Shot 2022-03-23 at 4 41 28 PM

PR #2014, opened 2022-03-23
I18nProvider: s/Text/TextField/

dangerismycat:
🤦


packages/gestalt/src/contexts/I18nProvider.js
```diff
@@ -8,17 +8,15 @@ import { type Context, createContext, useContext } from 'react';
  * - Add component translation object to initialContext using `null` for all values
  */
 
-type TextTranslations = {|
```
dangerismycat, line 11: This only made sense as a separate type when I thought I needed to use it elsewhere in this file, but since that's not the case it's a lot cleaner to inline it

PR #2031, opened 2022-03-31
TextArea, TextField: Add support for readonly fields

ayeshakmaz:
#### What changed?

In order to show users informations in forms without offering the ability to edit, we added `readonly` as an option for TextField and TextArea

#### Why?

Sometimes fields are not interactive, but also not disabled because they need to showcase info to the user (and disabled fields don't pass contrast)
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/textarea#attr-readonly

### Links

- [Jira](https://jira.pinadmin.com/browse/GESTALT-4132)

### Checklist


packages/gestalt/src/TextArea.js
```diff
@@ -75,15 +75,19 @@ type Props = {|
    */
   placeholder?: string,
   /**
-   * Ref that is forwarded to the underlying input element. See the [ref example](https://gestalt.pinterest.systems/textArea#refExample) for more details.
+   * Indicate if the input is currently readonly. See the [readonly example](https://gestalt.pinterest.systems/textarea#Readonly) for more details.
+   */
+  readonly?: boolean,
```
ayeshakmaz, line 80: New prop here

AlbertCarreras:
readonly & disabled, disabled, prevails. 

Screen Shot 2022-03-31 at 3 50 44 PM

AlbertCarreras (approved):
LGTM, but let's wait for Ryan's

ayeshakmaz:
@AlbertCarreras yeah, the html version will always prioritize the most restrictive option passed, so I think we're fine to leave as is. If people add both, then I'd assume the intention was to make it disabled

packages/gestalt/src/TextField.js
```diff
@@ -73,6 +73,10 @@ type Props = {|
    * Placeholder text shown the the user has not yet input a value.
    */
   placeholder?: string,
+  /**
+   * Indicate if the input is readonly. See the [disabled example](https://gestalt.pinterest.systems/textfield#Readonly) for more details.
```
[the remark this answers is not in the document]
ayeshakmaz: Fixed, thanks!

docs/pages/textarea.js
```diff
@@ -11,7 +9,26 @@ import docgen, { type DocGen } from '../components/docgen.js';
 export default function DocsPage({ generatedDocGen }: {| generatedDocGen: DocGen |}): Node {
   return (
     <Page title="TextArea">
```
dangerismycat, line 11: super nit: I've been using `{generatedDocGen?.displayName}` for the Page `title` and PageHeader `name`. It doesn't matter, but it's kinda neat.

docs/pages/textarea.js
```diff
@@ -36,219 +53,270 @@ export default function DocsPage({ generatedDocGen }: {| generatedDocGen: DocGen
...
-}
-`}
-      />
-      <Example
-        id="errorMessageExample"
-        name="Example: Error message"
-        description={`
+      <MainSection name="Variants">
+        <MainSection.Subsection
+          title="Basic"
```
dangerismycat, line 58: This title never really conveyed what was going on. The description talks about the width, so perhaps "Width" instead?

docs/pages/textarea.js
```diff
@@ -36,219 +53,270 @@ export default function DocsPage({ generatedDocGen }: {| generatedDocGen: DocGen
...
+          value={value}
+        />
+      </Box>
+    );
+  }
+      `}
+          />
+        </MainSection.Subsection>
+        <MainSection.Subsection
+          title="Readonly"
```
dangerismycat, line 108: nit: "Read-only", since we're talking about the behavior and not necessarily the prop name (e.g. we use "Error message" and not "errorMessage" for the section about `errorMessage`)

docs/pages/textarea.js
```diff
@@ -36,219 +53,270 @@ export default function DocsPage({ generatedDocGen }: {| generatedDocGen: DocGen
...
+          description={`
+      A \`TextArea\` can be in read-only mode in order to present information to the user without allowing them to edit the content. Typically they are used to show content or information that the user does not have permission or access to edit.
```
dangerismycat, line 110: nit: no article, just "TextArea can be…"

Also there's a mismatch between the singular count of TextArea (regardless of if the article is used) and the plural "them" and "they"

docs/pages/textarea.js
```diff
@@ -36,219 +53,270 @@ export default function DocsPage({ generatedDocGen }: {| generatedDocGen: DocGen
...
+    `}
+        >
+          <MainSection.Card
+            defaultCode={`
+  function Example(props) {
+    const [value, setValue] = React.useState('To keep shopping inspirational and actionable, we set high standards for our Merchants. Your website was not approved due to fuzzy, low quality images.');
```
dangerismycat, line 116: nit: this doesn't really seem like a realistic use case for read-only TextArea, but I also don't have a better one so feel free to run with this for now if you like. but it doesn't strike me as something we'd display in a TextArea

docs/pages/textarea.js
```diff
@@ -36,219 +53,270 @@ export default function DocsPage({ generatedDocGen }: {| generatedDocGen: DocGen
...
+      </Box>
+    );
+  }
+      `}
+          />
+        </MainSection.Subsection>
+        <MainSection.Subsection
+          title="Disabled"
+          description={`
+      A \`TextArea\` can be disabled to indicate the user is unable to interact with it, either by mouse or keyboard. Disabled fields do not need to pass contrast requirements, so do not use a \`disabled\` TextArea to present information to the user (use \`readonly\` instead).
```
dangerismycat, line 135: Same nit here about articles with components

docs/pages/textarea.js
```diff
@@ -36,219 +53,270 @@ export default function DocsPage({ generatedDocGen }: {| generatedDocGen: DocGen
...
+      </Box>
+    );
+  }
+      `}
+          />
+        </MainSection.Subsection>
+        <MainSection.Subsection
+          title="Error message"
+          description={`
     A TextArea can display its own error message.
```
dangerismycat, line 161: ditto here and throughout with nits about articles, but this isn't new copy so feel free to ignore

docs/pages/textarea.js
```diff
@@ -36,219 +53,270 @@ export default function DocsPage({ generatedDocGen }: {| generatedDocGen: DocGen
...
-      setTags([
-        ...tags,
-        // Avoid creating a tag on content on the last line, and filter out
-        // empty tags
-        ...tagInput.splice(0, tagInput.length - 1).filter(val => val !== ''),
-      ]);
-    }
-    setValue(tagInput[tagInput.length - 1]);
-  }
+            const onChangeTagManagement = ({ value }) => {
```
dangerismycat, line 240: this looks like a regression:
Screen Shot 2022-03-31 at 4 29 20 PM

docs/pages/textarea.js
```diff
@@ -36,219 +53,270 @@ export default function DocsPage({ generatedDocGen }: {| generatedDocGen: DocGen
...
+                  </Popover>
+                )}
+              </Box>
+            );
+          }
+          `}
+          />
+        </MainSection.Subsection>
+        <MainSection.Subsection
+          title="With Tags"
```
dangerismycat, line 225: IMO this is referring to tags as a concept and not the Tag component specifically, so sentence case would make sense

docs/pages/textarea.js
```diff
@@ -36,219 +53,270 @@ export default function DocsPage({ generatedDocGen }: {| generatedDocGen: DocGen
...
+        <MainSection.Subsection
+          title="Autofocus"
+          description={`
+        \`TextArea\` intentionally lacks support for autofocus. Generally speaking,
+        autofocus interrupts normal page flow for screen readers making it an
+        anti-pattern for accessibility.
+      `}
+        />
+        <MainSection.Subsection
+          title="onSubmit"
```
dangerismycat, line 307: Perhaps "Form submission"?

docs/pages/textfield.js
```diff
@@ -446,24 +446,55 @@ function Example(props) {
...
+      />
+    </Box>
+  );
+}
+`}
+          />
+        </MainSection.Subsection>
+
+        <MainSection.Subsection
+          title="Readonly"
```
dangerismycat, line 480: ditto here on the hyphen

packages/gestalt/src/InternalTextField.js
```diff
@@ -163,6 +165,7 @@ const InternalTextFieldWithForwardRef: React$AbstractComponent<Props, HTMLInputE
         // https://stackoverflow.com/questions/14447668/input-type-number-is-not-showing-a-number-keypad-on-ios
         pattern={type === 'number' ? '\\d*' : undefined}
         placeholder={placeholder}
+        readOnly={readonly}
```
dangerismycat, line 168: I was going to leave this comment below, but this is a perfect example: React thinks this should be camelCase, so I think we should too 😅

dangerismycat (approved):
The docs copy looks great. You can take or leave the nits — the only thing I feel strongly about is the casing of the new prop

docs/pages/textarea.js
ayeshakmaz: Yeah, it's the example from the person in Slack, but they have it in an error modal which makes it feel more "correct". I'll go with this for now and see if we can find a better product example.

PR #2036, opened 2022-04-01
I18nProvider: provide default strings, downgrade throwing errors to console

dangerismycat:
While the previous approach worked fine in the product, I was unable to successfully mock the implemented provider (with translations) unless included in the `render`/`shallow` call in each component using TextField, or whose children use TextField. That's impractical for now and unreasonable for the future (every component with a string in I18nProvider would have to have such a mock), so I'm backing off throwing errors for missing translations for now. We can perhaps revisit in the future but it's not worth blocking our merge/bump pipeline for this.

Also, after discussions with @jennyscript, I warmed to the idea of providing default English strings within Gestalt. This replicates the behavior of Pinboard's i18n function, so it's not actually different from behavior on non-Gestalt components.

### DIFF
--- a/packages/gestalt/src/contexts/I18nProvider.js
+++ b/packages/gestalt/src/contexts/I18nProvider.js
@@ -23,6 +23,13 @@ export const initialContext = {
   },
 };
 
+const defaultTranslations: I18nContextType = {
+  TextField: {
+    accessibilityHidePasswordLabel: 'Hide password',
+    accessibilityShowPasswordLabel: 'Show password',
+  },
+};
+
 const I18nContext: Context<I18nContextType> = createContext<I18nContextType>(initialContext);
 
 /**
@@ -39,7 +46,7 @@ export function useI18nContext<C: ValidComponent>(
   const propsTranslations = useContext(I18nContext);
   const componentTranslations = propsTranslations[componentName];
 
-  // No translation provided for the entire component
+  // No translation provided for the entire component even in initialContext
   if (!componentTranslations) {
     throw new Error(
       `${componentName} translations not available — please add translations to I18nProvider`,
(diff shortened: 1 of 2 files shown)


packages/gestalt/src/contexts/I18nProvider.js
```diff
@@ -57,15 +64,28 @@ export function useI18nContext<C: ValidComponent>(
...
-      )} ${
-        multipleMissing ? 'are' : 'is'
-      } missing translations — please add translations to I18nProvider`,
-    );
+    const isNotProd = process.env.NODE_ENV !== 'production';
+
+    if (isNotProd) {
+      // Ideally we would throw an actual error here (at least in dev), but this wasn't working with the mocks in Pinboard
+      // eslint-disable-next-line no-console
+      console.error(
```
dangerismycat, line 75: This is all the same, just putting the error in the console instead of throwing

packages/gestalt/src/contexts/I18nProvider.js
```diff
@@ -57,15 +64,28 @@ export function useI18nContext<C: ValidComponent>(
...
+        `${componentName} prop${multipleMissing ? 's' : ''} ${missingAndNullTranslations.join(
+          ', ',
+        )} ${
+          multipleMissing ? 'are' : 'is'
+        } missing translations — please add translations to I18nProvider`,
+      );
+    }
+    const defaultComponentTranslations = defaultTranslations[componentName];
+    Object.keys(defaultComponentTranslations).forEach((propName) => {
```
dangerismycat, line 84: This is the new bit, using the default translations where needed

ayeshakmaz (approved):
LGTM

PR #2025, opened 2022-03-29
Internal: fix flow issue in Iconography and SVGs docs

christianvuerings:
#### What changed?

Fixes a tricky flow issue in the `Iconography and SVGs` docs page

PR #2046, opened 2022-04-05
Internal: Fix flow issue without .map() in Iconography and SVGs docs

christianvuerings:
#### What changed?

Fixes the flaky flow issue without using a `.map` as we did in #2025

### DIFF
--- a/docs/pages/iconography_and_svgs.js
+++ b/docs/pages/iconography_and_svgs.js
@@ -5,28 +5,29 @@ import MainSection from '../components/MainSection.js';
 import Page from '../components/Page.js';
 import PageHeader from '../components/PageHeader.js';
 
+const { icons } = Icon;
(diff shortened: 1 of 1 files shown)


docs/pages/iconography_and_svgs.js
```diff
@@ -5,28 +5,29 @@ import MainSection from '../components/MainSection.js';
...
         <Box padding={2}>
-          <Icon color="darkGray" accessibilityLabel={String(iconName)} icon={iconName} />
+          <Icon color="darkGray" accessibilityLabel={iconName} icon={iconName} />
         </Box>
       </TapArea>
     </Tooltip>
   );
 }
 
+function findIcon(icon?: string): ?IconName {
```
dangerismycat, line 23: It seems like this is the crux of the fix to remove the stringification and .map call?

dangerismycat (approved):

PR #2047, opened 2022-04-05
Font weight: implement experiment to change font weight from bold to semibold 

AlbertCarreras:
#### What changed?

- Create new token `--font-weight-semibold` with value `600` (alternative to `--font-weight-bold` with value `700`).
- Experiments `web_gestalt_semibold_weight`/`mweb_gestalt_semibold_weight` to be used with `ExperimentProvider`
- Experiments used conditionally with "deviceType" from DeviceTypeProvider (following pattern from https://github.com/pinterest/gestalt/pull/2050/files)
- Implement `web_gestalt_semibold_weight`/`mweb_gestalt_semibold_weight` experiment within Text, Heading and other cmps (Badge, AvatarFoundations).

BEFORE
Screen Shot 2022-04-06 at 4 13 59 PM

AFTER (experiment enabled)
Screen Shot 2022-04-06 at 4 14 29 PM

#### Why?

Context: In 2019 when Lego was being introduced into the web platform, the semibold weight did not display with enough contrast and so the decision was to switch over to bold on web. Whatever was causing that lack of contrast no longer seems to be the case, and so there's an opportunity to align type treatments across web and native mobile.

_Acceptance criteria_

Assess whether it makes sense to change our "bold" weight to semibold on web
Change the token values from a weight of 700 to 600
Launch these changes as an experiment if possible to see if there are any negative impact to metrics.

### Links

- [Jira](https://jira.pinadmin.com/browse/GESTALT-3892)

### Checklist


docs/components/contexts/DocsExperimentProvider.js
```diff
@@ -8,7 +8,7 @@ import { ExperimentProvider } from 'gestalt';
  * - Unless you want the experimental behavior live on the docs for everyone, REMOVE YOUR EXPERIMENT HERE before merging your PR!
  * */
 
-const enabledExperiments = [];
+const enabledExperiments = ['gestalt_semibold_weight'];
```
AlbertCarreras, line 11: To be removed before merging, just to allow reviews to see effect in Docs

dangerismycat (commented):
Cool, this looks great. Some larger changes needed:
- After discussion with PJ, we'd like to run this as two experiments, one on desktop and one on mweb. That way we can easily see differences in user behavior between the different platforms. It looks like you haven't created your experiment in Helium yet, so instead create two and prefix them with the platform: `web_gestalt_semibold_weight`, `mweb_gestalt_semibold_weight`. (In the future you may want to create the experiments first so you don't have to update the Gestalt references if you need to change the exp name for some reason, but that's up to you — it doesn't matter much)
- Check out #2050 - for these platform-specific experiments, we'll need to use device type in the logic. (I'm implementing DeviceTypeProvider in Pinboard right now)

packages/gestalt/src/Text.js
```diff
@@ -106,6 +108,25 @@ export default function Text({
 }: Props): Node {
   const colorClass = semanticColors.includes(color) && colors[`${color}Text`];
 
+
+
+  let inSemiBoldExp;
+
+  const deviceType = useDeviceType();
```
dangerismycat, line 115: Since you're repeating this logic four times, it could be worth pulling into a temporary custom hook for the duration of the experiment

dangerismycat (approved):

packages/gestalt/src/Text.js
AlbertCarreras: DONE

PR #2069, opened 2022-04-14
Internal: Bump peer dependencies to support React 18

liuyenwei:
### Summary
Updates the react/react-dom peer dependencies for gestalt and gestalt-datepicker  to support 18.x.  We currently aren't leveraging any React 18 specific APIs so this change is safe.  In the future, if we start using some of the React 18 APIs, we'll need to think about backwards compatibility or dropping support for older versions of React.  

Note: this does not update the React version for docs.


AlbertCarreras (approved):
Thank you!

PR #2073, opened 2022-04-19
Link: design changes (part I: inline/underline)

AlbertCarreras:
#### What changed?

Link: design changes (part I: inline/underline)

**Main changes** 
`hoverStyle` props becomes `underline` with new values `"auto" | "none" | "hover" | "always"`

The "auto" lets the component internally manage the underline style based on `inline`

`underline= "none" | "hover" | "always"` is an API already existing in some DS. See [Material UI](https://mui.com/components/links/#underline)[#underline](https://mui.com/components/links/#underline)

**Value equivalencies**
| before    | after  |
|-----------|--------|
| none      | none   |
| underline | hover  |
|           | always |\]
|           | auto |\]

**New behavior**

When `inline` is supplied, we set the underline style internally to `always`.
When `inline` is not supplied, we set the underline style internally to `hover`.
When `underline` is supplied, we override the underline style internally managed by the component.

#### Why?

This is the step 1 of the [Link standardization project](https://paper.dropbox.com/doc/Gestalt-Link-Standardization--BgHgy3V31Efv1FE0GiNiKd6RAg-Opk3rapNVzU43gEfjedvb).

### Breaking Changes

Link comes with API changes and new behaviors

`hoverStyle` becomes `underline`
`inline` sets `underline="always"` so we must set it to `underline="hover"` for safety

Run the following codemods

**1st**
 `yarn codemod modifyPropValue ~/path/to/your/code
 --component=Link
 --previousProp=hoverStyle
 --nextProp=underline
 --previousValue=underline
 --nextValue=hover`

**2nd**
 `yarn codemod modifyProp ~/path/to/your/code
 --component=Link
 --previousProp=hoverStyle
 --nextProp=underline`

**3rd**
WIP Custom codemod that replaces
`<Link inline>` (w/o `underline` prop) to `<Link inline underline="hover">`

run `yarn codemod --parser=flow -t=packages/gestalt-codemods/51.0.0/link-add-underline-hover.js relative/path/to/your/code`

### Links

- [TDD](https://paper.dropbox.com/doc/Link--BgHiqMVz~DGCTqGCYy0GzE~_Ag-wVPcHLCkPfUVn8mIDtIN9)
- [Figma](https://www.figma.com/file/oUxggN7E1NlZftV4G4UfJD/Links-exploration?node-id=132%3A95

### Checklist


packages/gestalt/src/Link.js
```diff
@@ -30,13 +31,9 @@ type Props = {|
    */
   accessibilitySelected?: boolean,
   /**
-   * Link is a wrapper around components (or children), most commonly text, so that they become hyperlinks. See the [Link and Text](https://gestalt.pinterest.systems/link#Link-and-Text) variant for more information.
```
dangerismycat, line 33: You don't think this link is still useful?

packages/gestalt/src/Link.js
```diff
@@ -81,20 +78,26 @@ type Props = {|
...
 - 'blank' opens the anchor in a new window.
 - 'self' opens an anchor in the same frame.
+
+See the [target variant](https://gestalt.pinterest.systems/link#Target) to learn more.
    */
   target?: null | 'self' | 'blank',
+  /**
+   * When `underline` is supplied, we override the underline style internally managed by the component. See the [underline variant](https://gestalt.pinterest.systems/link#Underline) to learn more.
+   */
+  underline?: 'none' | 'always' | 'hover',
```
dangerismycat, line 96: Given that this is an override for something we determine internally, would it make sense to provide an "auto" option? I'm thinking of a use case where the user would conditionally override the default underline style, and something like
```js
<Link underline={condition ? 'none' : 'auto'}>
```
communicates a bit better than
```js
<Link underline={condition ? 'none' : undefined}>
```
even though they're functionally the same.

(And "auto" feels more right than "default", since we're not specifying a default value — the default value is different according to the situation — but explicitly asking for the automatic selection behavior)

docs/pages/link.js
```diff
@@ -8,44 +8,281 @@ import GeneratedPropTable from '../components/GeneratedPropTable.js';
...
+    </Text>
+  </Text>
+</Flex>;
+`}
+          />
+        </MainSection.Subsection>
+        <MainSection.Subsection columns={2}>
+          <MainSection.Card
+            type="do"
+            description={`\`isExternal\` PROP COMING SOON.
```
dangerismycat, line 239: 😅 Let's remove this do/don't pair for now?

docs/pages/link.js
```diff
@@ -68,9 +305,134 @@ export default function DocsPage({ generatedDocGen }: {| generatedDocGen: DocGen
...
+        </MainSection.Subsection>
+
+        <MainSection.Subsection
+          title="Keyboard navigation"
+          description="Give Link keyboard focus with the tab key (or shift + tab when tabbing backwards). Activate Link with the enter/return key."
+        />
         <MainSection.Subsection
           title="Accessible Tab Link"
-          description={`Use \`accessibilitySelected\` and \`role="Tab"\` when using Link as a tab.`}
+          description={`Use \`accessibilitySelected\` and \`role="Tab"\` when using Link as a tab. However, don't use Link to create custom tabs, use [Tabs](/tabs) instead.`}
```
dangerismycat, line 435: nit: `role="tab"` (lowercase)

Though really this example needs to be improved, as we're both showing a use case that should use a different component, and a use case that doesn't actually match what the "tab" role is supposed to be used for.

We can consider an update out of scope for this PR, but please put it on your list of Link improvements to come up with an example of [an actual "tab" role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Tab_Role) for this one.

docs/pages/link.js
```diff
@@ -109,74 +471,210 @@ function TabExample() {
...
+1. \`color="default"\`
+    Default text color for links used across Pinterest UI.
+2. \`color="subtle"\`
+    Intended for links placed within a subtle inline text (e.g. additional link on a sub-header).
+3. \`color="inverse"\`
+    For links over a dark-colored background. Make sure it has a 4.5:1 [contrast ratio](/accessibility#Visuals) between foreground and background.
+4. \`color="shopping"\`
+    Used on links related to shopping products or surfaces when a color is needed to highlight a link. Please note: This color should only be used in a shopping context.
+5. Status: \`color="error"\`, \`color="warning"\`, \`color="success"\`
+    Used purposeful, applies only on links within a status message. See [status colors](/color_usage#Status-text-colors) for reference.
```
dangerismycat, line 514: nit: "Used _purposefully_"

packages/gestalt/src/Link.js
AlbertCarreras: actually I like that.

packages/gestalt/src/Link.js
AlbertCarreras: DONE

packages/gestalt/src/Link.js
AlbertCarreras: DONE

docs/pages/link.js
AlbertCarreras: DONE

docs/pages/link.js
AlbertCarreras: DONE typo

packages/gestalt/src/Link.js
```diff
@@ -144,6 +147,12 @@ const LinkWithForwardRef: AbstractComponent<Props, HTMLAnchorElement> = forwardR
 
   const { isFocusVisible } = useFocusVisible();
 
+  let underlineStyle = inline ? 'always' : 'hover';
+
+  if (underline && underline !== 'auto') {
```
dangerismycat, line 152: You can remove the `underline !== 'auto'` check if you remove the default value in the destructuring above (it's not necessary)

dangerismycat (approved):

docs/pages/link.js
AlbertCarreras: DONE

packages/gestalt/src/Link.js
AlbertCarreras: @dangerismycat I wanted to add the underline="auto" so it's reflected in the docs, as the natural state of  the 
 component is to manage internally  `underline` (which is basically the `auto`)

dangerismycat: You can also accomplish that with a comment alongside the prop definition. I think we do this in Box and Flex where the defaults aren't defined in JS

PR #2080, opened 2022-04-21
Docs: Cleanup layout and shaded color examples

ayeshakmaz:
#### What changed?

Small cleanups to adjust layout for Best Practices and shade colors for combination color examples

#### Why?

Try to provide an ideal background color when showcasing color example of a component using CombinationNew

### DIFF
--- a/docs/components/CombinationNew.js
+++ b/docs/components/CombinationNew.js
@@ -5,7 +5,6 @@ import MainSectionCard from './MainSectionCard.js';
 
 type Props = {
   children: (props: { [key: string]: any, ... }, index?: number) => Node, // flowlint-line unclear-type:off
-  shaded?: boolean,
   hideTitle?: boolean,
   ...
 };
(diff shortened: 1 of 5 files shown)


docs/components/CombinationNew.js
```diff
@@ -5,7 +5,6 @@ import MainSectionCard from './MainSectionCard.js';
 
 type Props = {
   children: (props: { [key: string]: any, ... }, index?: number) => Node, // flowlint-line unclear-type:off
-  shaded?: boolean,
```
ayeshakmaz, line 8: prop was never used at this level

docs/components/MainSectionCard.js
```diff
@@ -15,6 +15,7 @@ type Props = {|
   defaultCode?: string,
   description?: string,
   iframeContent?: string,
+  shadeColor?: 'tertiary' | 'darkWash' | 'lightWash' | 'default',
   shaded?: boolean,
```
ayeshakmaz, line 19: leaving shaded for backwards compatibility and ease-of-use, shadeColor is like a power-up option

PR #2085, opened 2022-04-26
Toast: fix for darkMode

zifahm:
### Summary
Toast was not working in darMode. The color white was redering as black if darkMode is set.

#### What changed?
changed default color to darkGray. And the Box Component takes care of it during darkMode.

### DIFF
--- a/packages/gestalt/src/Toast.js
+++ b/packages/gestalt/src/Toast.js
@@ -1,12 +1,12 @@
 // @flow strict
 import { type Element, type Node } from 'react';
 import Box from './Box.js';
+import { useColorScheme } from './contexts/ColorSchemeProvider.js';
 import Flex from './Flex.js';
 import Mask from './Mask.js';
 import Text from './Text.js';
 import styles from './Toast.css';
 import useResponsiveMinWidth from './useResponsiveMinWidth.js';
-import { useColorScheme } from './contexts/ColorSchemeProvider.js';
 
 const TOAST_MAX_WIDTH_PX = 500;
 const TOAST_WIDTH_PX = 330;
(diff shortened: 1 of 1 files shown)


ayeshakmaz:
Hi there @zifahm - I apologize that the docs are confusing, but our Toast was actually working properly in dark mode. The key is to look at this example: https://gestalt.pinterest.systems/toast#displayExample
When you use that example to show the Toast while in darkMode on the docs, you'll see it works appropriately. 

I will add a note to the docs to showcase that the other examples are not currently picking up dark mode properly. Thanks for bringing this up!

zifahm:
Hi @ayeshakmaz , when I tried to reporduce the same example in codesanbox with colorSchema set to dark, toast is not picking it up. Could you help me with how would I make it work?

https://codesandbox.io/s/example-how-to-display-forked-12evnw?file=/example.js

ayeshakmaz:
@zifahm I believe if you add an "id" to your ColorSchemeProvider, it works well and produces the right Toast. I'll double check on the docs and see if the guidance around "id" prop needs to be updated. Thanks for the sandbox!

zifahm:
Hi, @ayeshakmaz 

I have a problem if I pass an id for the ColorSchemaProvider and setting to darkmode, my other models or layers does not pick up.

But if I dont pass an Id there isnt any problem, but here now as you suggested above toast doesnt pick it up if it dark mode is set.

It's like, if id is present model or layer fails , if id is'nt passed toast fails.

How would one solve this problem?

ayeshakmaz:
Hi @zifahm, are you able to reproduce this issue in a code sandbox by chance?

zifahm:
Hi @ayeshakmaz 
sure : https://codesandbox.io/s/overlaying-content-forked-k46v71?file=/example.js

So if you open the layer first, you see its `lightGray` and then if you remove the id for colorSchema its goes back to being dark because darkMode is set. This is true for toast as well. 

Right now I'm doing a patch package on toast with the same code fix of this PR

ayeshakmaz:
Hi @zifahm we recently made an update to Toast in dark mode, so I believe this issue should be resolved now. Let me know if you still see it after version 59.0.0

PR #2086, opened 2022-04-26
Video: fix on Flow type

AlbertCarreras:
#### What changed?

Video: fix on Flow type

#### Why?

The Flowtype provided doesn't match the actual Flowtype for ref in Classes. Moving ref type signature to a manual definition.

### DIFF
--- a/docs/pages/video.js
+++ b/docs/pages/video.js
@@ -4,7 +4,6 @@ import GeneratedPropTable from '../components/GeneratedPropTable.js';
 import Page from '../components/Page.js';
 import PageHeader from '../components/PageHeader.js';
 import docgen, { type DocGen } from '../components/docgen.js';
-import deepCloneReplacingUndefined from '../utils/deepCloneReplacingUndefined.js';
 import MainSection from '../components/MainSection.js';
 
 export default function DocsPage({ generatedDocGen }: {| generatedDocGen: DocGen |}): Node {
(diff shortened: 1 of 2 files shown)


docs/pages/video.js
```diff
@@ -366,7 +365,17 @@ function Example () {
 export async function getStaticProps(): Promise<{| props: {| generatedDocGen: DocGen |} |}> {
   const generatedDocGen = await docgen({ componentName: 'Video' });
 
+  generatedDocGen.props.ref = {
+    required: false,
+    defaultValue: null,
+    flowType: {
+      name: 'Ref',
+      raw: 'Ref<typeof Video>',
```
dangerismycat, line 373: Should "Video" be capitalized here? Are we actually referring to the `<video>` element?

docs/pages/video.js
```diff
@@ -366,7 +365,17 @@ function Example () {
...
+    },
+    description: 'Ref on the Gestalt Video component.',
+  };
+
   return {
-    props: { generatedDocGen: deepCloneReplacingUndefined(generatedDocGen) },
```
dangerismycat, line 370: Just making sure you double-check that the page loads with no errors. I don't recall the exact error this was necessary for, but it was something about undefined values in the JSON

dangerismycat (approved):

docs/pages/video.js
AlbertCarreras: It's the Video class component. I think the correct type is more like `{ current: Node}`. I might fix it in the future. I wanted to document it. as it's not a forwardRef.

PR #2089, opened 2022-04-27
TextArea: Added default value to labelDisplay prop

ponori:
labelDisplay prop was missing the "visible" default value.


dangerismycat (approved):

dangerismycat:
Tiny changes like this are just patches…if you're not sure of the semver label to use, feel free to leave it blank and the reviewing eng can add it

PR #2097, opened 2022-05-02
DatePicker: add `name` prop

AlbertCarreras:
#### What changed?

Adding `name` prop which gets passed to the internal TextField component used by DatePicker.

#### Why?

To keep consistent pattern across form fields and enable the consistent use of form-related API such as [formData](https://developer.mozilla.org/en-US/docs/Web/API/FormData)

### DIFF
--- a/packages/gestalt-datepicker/src/DatePicker.js
+++ b/packages/gestalt-datepicker/src/DatePicker.js
@@ -88,6 +88,10 @@ type Props = {|
    */
   minDate?: Date,
   /**
+   * A unique name for the input.
+   */
+  name?: string,
+  /**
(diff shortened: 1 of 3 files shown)


packages/gestalt-datepicker/src/DatePicker.js
```diff
@@ -87,6 +87,10 @@ type Props = {|
    * Disable dates outside a min date. See the [delimited selection period example](https://gestalt.pinterest.systems/datepicker#maxMinDates) to learn more.
    */
   minDate?: Date,
+  /**
+   * A unique name for the input.
+   */
+  name: string,
```
dangerismycat, line 93: Should this be required? If so, can you add a note about the relevant codemod in the PR summary?

dangerismycat (approved):

packages/gestalt-datepicker/src/DatePicker.js
AlbertCarreras: Actually, this was a bug. it's just optional all the way

PR #2104, opened 2022-05-05
Internal: upgrade to React 18

dangerismycat:
This PR upgrades Gestalt and the docs site to use React 18. The only change to use any React 18 APIs is in the CodeSandbox handler, where `render` is replaced with `createRoot`. Since no other changes are made (yet), this should continue to be fully backwards-compatible with React 17.

I'm marking this as a `major` change, but don't expect there to be any real breaking changes.

This also introduces some console spam on unit test runs, which I'll address by upgrading a few packages in a follow-up PR.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -126,8 +126,9 @@
         "scripts/templates/accessibility_ComponentName_spec.js"
       ],
       "rules": {
+        "flowtype/require-valid-file-annotation": "off",
         "jest/expect-expect": "off",
-        "flowtype/require-valid-file-annotation": "off"
+        "jest/valid-expect": "off"
       }
     }
   ]
--- a/cypress/integration/page_header_spec.js
+++ b/cypress/integration/page_header_spec.js
@@ -10,7 +10,10 @@ describe('Page Headers', () => {
   it('navigates to the development page', () => {
     cy.contains('a', 'Development').click();
     cy.url().should('match', /development/);
-    cy.get('.docSearch-content h1').invoke('text').invoke('trim').should('equal', 'Development');
+    cy.get('.docSearch-content h1').should(($h1) => {
+      const trimmedText = $h1.text().trim();
+      expect(trimmedText).to.equal('Development');
+    });
     cy.url().should('include', Cypress.config('baseUrl'));
   });
 });
--- a/docs/components/handleCodeSandbox.js
+++ b/docs/components/handleCodeSandbox.js
@@ -78,9 +78,12 @@ const handleCodeSandbox = async ({ code, title }: {| code: string, title: string
       },
       'index.js': {
         content: `import React from "react";
-import { render } from "react-dom";
+import { createRoot } from "react-dom/client";
 import Example from "./example";
-render(<Example />, document.querySelector("#root"));`,
+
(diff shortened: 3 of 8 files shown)


.eslintrc.json
```diff
@@ -126,8 +126,9 @@
         "scripts/templates/accessibility_ComponentName_spec.js"
       ],
       "rules": {
+        "flowtype/require-valid-file-annotation": "off",
         "jest/expect-expect": "off",
-        "flowtype/require-valid-file-annotation": "off"
+        "jest/valid-expect": "off"
```
dangerismycat, line 131: Jest assumes that any `expect` is its own global, which isn't true in our Cypress tests (see page_header_spec for the change that necessitated this suppression)

cypress/integration/page_header_spec.js
```diff
@@ -10,7 +10,10 @@ describe('Page Headers', () => {
   it('navigates to the development page', () => {
     cy.contains('a', 'Development').click();
     cy.url().should('match', /development/);
-    cy.get('.docSearch-content h1').invoke('text').invoke('trim').should('equal', 'Development');
+    cy.get('.docSearch-content h1').should(($h1) => {
```
dangerismycat, line 13: Without this change, Cypress would look at the selector before the page had a chance to re-render with the updated body content after the navigation. Cypress would see the old content and fail. Though not obvious, this syntax basically makes Cypress await here.

docs/components/handleCodeSandbox.js
```diff
@@ -78,9 +78,12 @@ const handleCodeSandbox = async ({ code, title }: {| code: string, title: string
       },
       'index.js': {
         content: `import React from "react";
-import { render } from "react-dom";
+import { createRoot } from "react-dom/client";
```
dangerismycat, line 81: Netlify takes care of this under the hood for the docs site, so this was the only place that needed to be updated to the new API

packages/gestalt/src/__snapshots__/Avatar.test.js.snap
```diff
@@ -577,9 +577,7 @@ exports[`Avatar renders with an empty name shows default icon 1`] = `
           fill="#111"
           fontSize="40px"
           textAnchor="middle"
-        >
-          
-        </text>
+        />
```
dangerismycat, line 580: Apparently the new renderer version has slightly different behavior here? ¯\_(ツ)_/¯

.eslintrc.json
liuyenwei: should we disable the invalid use case inline vs turning this off globally?

dangerismycat: Normally I would agree, but this is applying a Jest rule to Cypress tests, where it isn't relevant. Really we could/should turn off all the Jest rules for the Cypress tests, but I was taking a reverse-YAGNI approach to turning those rules off 😅

ayeshakmaz (approved):

.eslintrc.json
liuyenwei: ah i missed that this was scoped to cypress/**/*.js

liuyenwei (approved):

PR #2105, opened 2022-05-06
Tooltip: minor refactor

AlbertCarreras:
#### What changed?

Tooltip: minor refactor

### DIFF
--- a/docs/pages/scrollboundarycontainer.js
+++ b/docs/pages/scrollboundarycontainer.js
@@ -121,48 +121,45 @@ function ScrollBoundaryContainerExample() {
               idealDirection="right"
               text="See stats about different types of content created by you and/or others on Pinterest. Filter to get more details on your organic (not an ad) and paid (promoted as an ad) content."
             >
-              <Icon
-                icon="info-circle"
-                accessibilityLabel="Information"
-                color="default"
-              />
+              <TapArea>
+                <Icon
+                  icon="info-circle"
+                  accessibilityLabel="Information"
+                  color="default"
+                />
+              </TapArea>
             </Tooltip>
           </Flex>
-          <Box
-            display="flex"
-            direction="column"
-          >
-          <Box paddingY={1}>
-              <RadioButton
-                checked={content === "all"}
-                id="allcontent"
-                label="All"
-                name="content"
-                onChange={() => setContent( "all" )}
-                value="all"
-              />
-            </Box>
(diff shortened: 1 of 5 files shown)


packages/gestalt/src/Contents.js
```diff
@@ -213,51 +213,48 @@ class Contents extends Component<Props, State> {
 
     return (
       <div
-        className={styles.container}
```
AlbertCarreras, line 216: Merge this and the following div into 1

docs/pages/scrollboundarycontainer.js
```diff
@@ -121,48 +121,49 @@ function ScrollBoundaryContainerExample() {
               idealDirection="right"
               text="See stats about different types of content created by you and/or others on Pinterest. Filter to get more details on your organic (not an ad) and paid (promoted as an ad) content."
             >
-              <Icon
```
AlbertCarreras, line 124: Example better a11y

docs/pages/scrollboundarycontainer.js
```diff
@@ -121,48 +121,49 @@ function ScrollBoundaryContainerExample() {
...
-                checked={content === "paid"}
-                id="paid"
-                label="Paid and earned"
-                name="content"
-                onChange={() => setContent( "paid" )}
-                value="paid"
-              />
-            </Box>
-          </Box>
+          <Flex direction="column">
```
dangerismycat, line 133: looks like the left side is getting cut off a bit

Screen Shot 2022-05-09 at 3 38 10 PM

docs/pages/scrollboundarycontainer.js
```diff
@@ -121,48 +121,49 @@ function ScrollBoundaryContainerExample() {
...
+            <Fieldset legend="Select content type" legendDisplay="hidden">
+              <Box paddingY={1}>
```
dangerismycat, line 135: should this use Flex and `gap`, or does that not work well with Fieldset?

dangerismycat (approved):

docs/pages/scrollboundarycontainer.js
AlbertCarreras: Cant repro this

docs/pages/scrollboundarycontainer.js
AlbertCarreras: DONE. good catch

PR #2109, opened 2022-05-09
Icon: Add info color option

ayeshakmaz:
#### What changed?

Add option to create info icons (used in Callout, Banner etc.) to avoid contamination with Shopping option

#### Why?

Create separation so we can make updates to either in the future

### Checklist


AlbertCarreras (approved):
LGTM after passing snapshots

PR #2115, opened 2022-05-12
Datapoint: add prop to set z-index on integrated tooltip

dangerismycat:
[context from Slack](https://pinterest.slack.com/archives/C13KLG5P0/p1652206272829709)

Screen Shot 2022-05-11 at 5 31 16 PM

It should be no surprise that the same issues that caused us to add `zIndex` to Tooltip mean that we'll need to be able to set that prop where we integrate tooltip into other components, like Datapoint. This PR adds a prop to Datapoint to do just that. (We'll likely need to do the same for IconButton and any other places we integrate Tooltip into other components)

### DIFF
--- a/packages/gestalt/src/Datapoint.js
+++ b/packages/gestalt/src/Datapoint.js
@@ -1,12 +1,13 @@
 // @flow strict
 import { type Node } from 'react';
+import Badge from './Badge.js';
 import DatapointTrend from './DatapointTrend.js';
 import Flex from './Flex.js';
 import Icon from './Icon.js';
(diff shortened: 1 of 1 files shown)


rlingineni (commented):
Looks good to me, seems like a passthrough prop

Minor thought, let's say in the future, someone wanted to pass more props, or e.g the ideal direction wasn't "up". 

Would it make sense to instead create a larger prop called `tooltipOptions` and then the `zIndex`, `idealDirection` and more would be sub-props?

AlbertCarreras (approved):
LGTM

PR #2117, opened 2022-05-12
ComboBox: add zIndex prop

AlbertCarreras:
#### What changed?

ComboBox: add zIndex prop

#### Why?

Follow pattern supported in other components such as Dropdown
Prevent issues like https://codesandbox.io/s/compassionate-field-32f0oy?file=/example.js

### DIFF
--- a/packages/gestalt/src/ComboBox.js
+++ b/packages/gestalt/src/ComboBox.js
@@ -26,6 +26,7 @@ import handleContainerScrolling, {
   KEYS,
   type DirectionOptionType,
 } from './utils/keyboardNavigation.js';
+import { type Indexable } from './zIndex.js';
 
 type Size = 'md' | 'lg';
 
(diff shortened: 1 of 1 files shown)


packages/gestalt/src/ComboBox.js
```diff
@@ -144,6 +144,10 @@ type Props = {|
    * List of tags to display in the component. See [tags](https://gestalt.pinterest.systems/combobox#Tags) variant to learn more.
    */
   tags?: $ReadOnlyArray<Element<typeof Tag>>,
+  /**
+   * An object representing the zIndex value of the ComboBox list box. Learn more about [zIndex classes](https://gestalt.pinterest.systems/zindex_classes)
+   */
+  zIndex?: {| index(): number |},
```
dangerismycat, line 150: You can just use Indexable here (as we do with all other zIndex references)

dangerismycat (approved):

packages/gestalt/src/ComboBox.js
AlbertCarreras: DONE

PR #2122, opened 2022-05-17
Docs: Unified Switch docs

ponori:
Adding design guidelines for Switch component

### DIFF
--- a/docs/pages/switch.js
+++ b/docs/pages/switch.js
@@ -64,18 +64,24 @@ export default function DocsPage({ generatedDocGen }: {| generatedDocGen: DocGen
             type="do"
             description="Use a label to give the Switch context when possible."
             defaultCode={`
-<Flex gap={2} alignItems="center" width="100%" maxWidth={300}>
-  <Flex.Item flex="grow">
-    <Label htmlFor="showsecretboards">
-      <Text weight="bold">Make this board secret</Text>
-    </Label>
-  </Flex.Item>
-  <Switch
-    onChange={() => {}}
-    id="showsecretboards"
-    switched={true}
-  />
-</Flex>
+function SwitchExample() {
+  const [switched, setSwitched] = React.useState(true);
+
+  return (
+    <Flex gap={2} alignItems="center" width="100%" maxWidth={300}>
+      <Flex.Item flex="grow">
+        <Label htmlFor="makesecret">
+          <Text weight="bold">Make this board secret</Text>
+        </Label>
+      </Flex.Item>
+      <Switch
+        onChange={() => setSwitched(!switched)}
+        id="makesecret"
+        switched={switched}
+      />
+    </Flex>
+  );
+}
 `}
           />
           <MainSection.Card
@@ -83,16 +89,22 @@ export default function DocsPage({ generatedDocGen }: {| generatedDocGen: DocGen
             type="don't"
             description="Truncate label text, instead allow it to wrap to form another line. "
             defaultCode={`
-<Flex gap={2} alignItems="center">
-  <Switch
-    onChange={() => {}}
-    id="setboard"
-    switched={false}
-  />
-  <Label htmlFor="setboard">
-    <Text>Set board to...</Text>
-  </Label>
-</Flex>
+function SwitchExample() {
+  const [switched, setSwitched] = React.useState(false);
+
+  return (
+    <Flex gap={2} alignItems="center">
+      <Switch
+        onChange={() => setSwitched(!switched)}
+        id="setboard"
+        switched={switched}
+      />
+      <Label htmlFor="setboard">
+        <Text>Set board to...</Text>
+      </Label>
+    </Flex>
+  );
+}
 `}
           />
           <MainSection.Card
@@ -124,25 +136,25 @@ export default function DocsPage({ generatedDocGen }: {| generatedDocGen: DocGen
 <Flex direction="column" gap={2} width="100%" maxWidth={300}>
   <Heading size="300">On Pinterest notifications</Heading>
   <Flex direction="column" gap={4}>
-    <Flex>
+    <Flex alignItems="center">
       <Flex.Item flex="grow">
         <Text>Comments</Text>
       </Flex.Item>
       <Icon icon="check-circle" color="default" />
     </Flex>
-    <Flex>
+    <Flex alignItems="center">
       <Flex.Item flex="grow">
         <Text>Mentions</Text>
       </Flex.Item>
       <Icon icon="check-circle" color="default" />
     </Flex>
-    <Flex>
+    <Flex alignItems="center">
       <Flex.Item flex="grow">
         <Text>Views</Text>
       </Flex.Item>
       <Icon icon="circle-outline" />
     </Flex>
-    <Flex>
+    <Flex alignItems="center">
       <Flex.Item flex="grow">
         <Text>Saves</Text>
       </Flex.Item>
@@ -161,7 +173,7 @@ export default function DocsPage({ generatedDocGen }: {| generatedDocGen: DocGen
   <Heading size="300">Tune your home feed</Heading>
   <Text>Turn off any boards that you don't want us to use for your home feed recommendations</Text>
   <Flex direction="column" gap={4}>
-    <Flex>
+    <Flex alignItems="center">
       <Flex.Item flex="grow">
         <Label htmlFor="boardfood">
           <Flex direction="column" gap={1}>
@@ -176,7 +188,7 @@ export default function DocsPage({ generatedDocGen }: {| generatedDocGen: DocGen
         switched={false}
       />
     </Flex>
-    <Flex>
+    <Flex alignItems="center">
       <Flex.Item flex="grow">
         <Label htmlFor="boardoutfits">
           <Flex direction="column" gap={1}>
@@ -191,7 +203,7 @@ export default function DocsPage({ generatedDocGen }: {| generatedDocGen: DocGen
         switched={true}
       />
     </Flex>
-    <Flex>
+    <Flex alignItems="center">
       <Flex.Item flex="grow">
         <Label htmlFor="boardhomedecor">
           <Flex direction="column" gap={1}>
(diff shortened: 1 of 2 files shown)


docs/pages/switch.js
```diff
@@ -57,6 +57,196 @@ export default function DocsPage({ generatedDocGen }: {| generatedDocGen: DocGen
           />
         </MainSection.Subsection>
       </MainSection>
+      <MainSection name="Best practices">
+        <MainSection.Subsection columns={2}>
```
ayeshakmaz, line 61: Be sure to wrap every 2 cards in their own subsection (see IconButton for example)

docs/pages/switch.js
```diff
@@ -57,6 +57,196 @@ export default function DocsPage({ generatedDocGen }: {| generatedDocGen: DocGen
...
+      <Icon icon="check-circle" color="default" accessibilityLabel="checked circle" />
+    </Flex>
+  </Flex>
+</Flex>
+`}
+          />
+          <MainSection.Card
+            cardSize="md"
+            type="do"
+            description="Use a standalone switch if the surrounding context is clear without an associated label. However, be sure to include a visually hidden label for screen readers."
```
ayeshakmaz, line 170: Apologies, I think I missed this in the dropbox

We may want to do "Use a standalone switch if the surrounding context is clear without an immediately adjacent label. However, be sure to pair the Label with the Switch using the `htmlFor` and `id` props." instead

docs/pages/switch.js
```diff
@@ -57,6 +57,196 @@ export default function DocsPage({ generatedDocGen }: {| generatedDocGen: DocGen
           />
         </MainSection.Subsection>
       </MainSection>
+      <MainSection name="Best practices">
+        <MainSection.Subsection columns={2}>
+          <MainSection.Card
+            cardSize="md"
+            type="do"
+            description="Use a label to give the Switch context when possible."
```
dangerismycat, line 65: nit: no article before "Switch"

docs/pages/switch.js
```diff
@@ -57,6 +57,196 @@ export default function DocsPage({ generatedDocGen }: {| generatedDocGen: DocGen
...
+`}
+          />
+        </MainSection.Subsection>
+      </MainSection>
+      <MainSection
+        name="Accessibility"
+        description={`
+    Switches should have labels that can be read by screen readers, and that can be clicked or tapped to make it easier for users to select and deselect. Therefore, make sure to supply the \`label\` prop. If that is not possible, make sure your stand-alone Label has an \`htmlFor\` prop that matches the \`id\` on the Switch. Test that the Switch and label are properly connected by clicking or tapping on the label and confirming that it activates the Switch next to it.
+
+      If Switch is labeled by content elsewhere on the page, or a more complex label is needed, the \`labelDisplay\` prop can be used to visually hide the label. In this case, it is still available to screen reader users, but will not appear visually on the screen. See the [Label visibility example](https://gestalt.netlify.app/checkbox#Label-visibility) for more detail.`}
```
ayeshakmaz, line 232: Small tweaks: 

Switches should have [Labels](/Label) that can be read by screen readers, and that can be clicked or tapped to make it easier for users to select and deselect. Make sure Label has an \`htmlFor\` prop that matches the \`id\` on the Switch. Test that the Switch and Label are properly connected by clicking or tapping on the label and confirming that it activates the Switch next to it.
    
[delete second paragraph, doesn't apply to Switch]

docs/pages/switch.js
```diff
@@ -57,6 +57,196 @@ export default function DocsPage({ generatedDocGen }: {| generatedDocGen: DocGen
...
+      />
+    </Flex>
+  );
+}
+`}
+          />
+          <MainSection.Card
+            cardSize="md"
+            type="don't"
+            description="Truncate label text, instead allow it to wrap to form another line. "
```
dangerismycat, line 90: grammar nit: a comma isn't the right separator here. Either a period and start a new sentence with "Instead", or use a semicolon and add a comma after "instead"

docs/pages/switch.js
```diff
@@ -57,6 +57,196 @@ export default function DocsPage({ generatedDocGen }: {| generatedDocGen: DocGen
...
+    Switches should have labels that can be read by screen readers, and that can be clicked or tapped to make it easier for users to select and deselect. Therefore, make sure to supply the \`label\` prop. If that is not possible, make sure your stand-alone Label has an \`htmlFor\` prop that matches the \`id\` on the Switch. Test that the Switch and label are properly connected by clicking or tapping on the label and confirming that it activates the Switch next to it.
+
+      If Switch is labeled by content elsewhere on the page, or a more complex label is needed, the \`labelDisplay\` prop can be used to visually hide the label. In this case, it is still available to screen reader users, but will not appear visually on the screen. See the [Label visibility example](https://gestalt.netlify.app/checkbox#Label-visibility) for more detail.`}
+      >
+        <MainSection.Subsection
+          title="Keyboard navigation"
+          columns={2}
+          description={`
+    Switch has conventional keyboard support.
+    - Users relying on the keyboard expect to move focus to the Switch by using the tab key or shift+tab when moving backwards.
```
ayeshakmaz, line 239: Maybe add - Once focused, the Space key toggles the Switch

docs/pages/switch.js
```diff
@@ -141,6 +331,50 @@ function SwitchExample() {
...
       </MainSection>
+
+      <MainSection name="Writing">
+        <MainSection.Subsection columns={2}>
+          <MainSection.Card
+            cardSize="md"
+            type="do"
+            description={`
+- Be clear and brief with Switch labels so they can be easily understood.
+- When possible, use verb to clarify the action. Something like “set…” or “show…”
```
ayeshakmaz, line 342: "use verbs..."

docs/pages/switch.js
```diff
@@ -141,6 +331,50 @@ function SwitchExample() {
...
+- If possible, be clear whether the setting is activated or deactivated.
+- Use sentence case for labels
```
ayeshakmaz, line 344: super nit: missing period

docs/pages/switch.js
```diff
@@ -141,6 +331,50 @@ function SwitchExample() {
...
+`}
+          />
+          <MainSection.Card
+            cardSize="md"
+            type="don't"
+            description={`
+- Use vague language out of context, like “turn on” or “turn off” repeating the state of the switch is redundant and can clutter the interface.
+- Don’t use “you,” “your,” or “my” to describe and action. Instead describe switches objectively.
```
ayeshakmaz, line 352: "an action [...] Switches..."

docs/pages/switch.js
```diff
@@ -65,7 +255,7 @@ export default function DocsPage({ generatedDocGen }: {| generatedDocGen: DocGen
         >
           <MainSection.Card
             defaultCode={`
-        function SwitchExample() {
+        function SwitchExample4() {
```
dangerismycat, line 258: nit: this is unnecessary — the function components in different examples can all have the same name. only the ids need to be unique

docs/pages/switch.js
```diff
@@ -57,6 +57,196 @@ export default function DocsPage({ generatedDocGen }: {| generatedDocGen: DocGen
...
+    Switch has conventional keyboard support.
+    - Users relying on the keyboard expect to move focus to the Switch by using the tab key or shift+tab when moving backwards.
+    - Setting \`disabled\` will prevent Switch from receiving keyboard focus or input
+`}
+        />
+      </MainSection>
+
+      <MainSection
+        name="Localization"
+        description={`Be sure to localize \`label\` and any \`subtext\`. Be mindful of label length so that it doesn’t truncate in languages with lengthier character counts.`}
```
dangerismycat, line 247: nit: "any" isn't needed before "subtext", and implies that there could be multiple subtexts

docs/pages/switch.js
dangerismycat, line 239: These need to be consistent: either end both with a period or without, but not mixed

docs/pages/switch.js
```diff
@@ -57,6 +57,196 @@ export default function DocsPage({ generatedDocGen }: {| generatedDocGen: DocGen
...
+<Flex gap={2} alignItems="center">
+  <Switch
+    onChange={() => {}}
+    id="showsecretboards"
+    switched={false}
+    disabled={true}
+  />
+  <Label htmlFor="showsecretboards">
+    <Flex direction="column" gap={1}>
+      <Text color="subtle">Show secret boards</Text>
```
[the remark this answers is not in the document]
ponori: Done!

docs/pages/switch.js
dangerismycat, line 170: Keep in mind that the code blocks aren't available for Dos/Don'ts, so the reader won't be able to tell that this is implementing a separate Label. There's nothing visually indicating that this is using Label instead of `label`, so this doesn't really communicate anything. Given that we mention this in the Accessibility section and the "With a label" variant, this Don't may not be necessary?

docs/pages/switch.js
dangerismycat, line 342: Similar to above, these either all need to end with periods or all not (including this one with the ellipsis)

docs/pages/switch.js
ponori: Done!

docs/pages/switch.js
ponori: This totally works for me. @khalvy, whaddaya think?

ayeshakmaz: maybe just convert this example to the "With a label example"

docs/pages/switch.js
ponori: Done!

docs/pages/switch.js
ponori: Nice! Done!

docs/pages/switch.js
ponori: Done!

docs/pages/switch.js
ponori: Done!

docs/pages/switch.js
ponori: Done!

docs/pages/switch.js
ponori: Done!

docs/pages/switch.js
ponori: Done!

docs/pages/switch.js
ponori: Done!

docs/pages/switch.js
ponori: Done!

docs/pages/switch.js
ponori: Done!

docs/pages/switch.js
ponori: I think think will be helpful from a design perspective. I'd argue to keep.

docs/pages/switch.js
ponori: Done!

docs/pages/switch.js
ayeshakmaz: works for me!

ayeshakmaz: Switch doesnt actually have a label prop so, this is mainly just stating that the label doesn't have to be right next to the switch, and just to make sure to link them properly

ponori: > maybe just convert this example to the "With a label example"

Sorry, @ayeshakmaz, could you elaborate?

ayeshakmaz: oh I meant instead of having this example as a best practice, use this code for the "with a label" variant example

dangerismycat: If you think it's self-explanatory for designers, 👍

ponori: Oh, nice! I like it! Done!

ponori: Yeah, @dangerismycat, let's see what we hear and adjust if necessary. @khalvy, just wanted to confirm if you're cool with this change.

ayeshakmaz (approved):

PR #2123, opened 2022-05-18
RadioGroup: New component to ensure accessible radio button groups

ayeshakmaz:
#### What changed?

Added a new component RadioGroup, along with RadioGroup.RadioButton (whose api is identical to the stand alone RadioButton) in order to create visually consistent and accessible groups of radio buttons. 

#### Why?

Radio buttons cannot exist on their own, so using RadioGroup creates more accessible forms by wrapping it in a FIeldset automatically and supplying a legend for the entire group. RadioGroup also creates consistent spacing between each button, allowing us to create a more consistent UI.

Note: I haven't replaced all old RadioButton instances in our docs to use this new cmp. Should be done when we deprecate RadioButton

### Links

- [Jira](https://jira.pinadmin.com/browse/GESTALT-3604)
- [Docs](https://paper.dropbox.com/doc/Gestalt-doc-RadioGroup--Bhpqab6iB9DgnOu0cdhCW4CDAg-fbzi8kKP4PigLOka4UEB5)

### Checklist


docs/pages/radiobutton.js
```diff
@@ -13,335 +11,18 @@ export default function DocsPage({ generatedDocGen }: {| generatedDocGen: DocGen
...
-            onChange={() => setFavoriteFood( 'sushi' )}
-            value="sushi"
-          />
-        </Flex>
-      </Fieldset>
-    </Flex>
-  );
-}
-`}
+      <SlimBanner
```
ayeshakmaz, line 14: I moved all the docs over to RadioGroup to promote adoption and utilize the new docs

packages/gestalt/src/RadioGroup.js
```diff
@@ -0,0 +1,64 @@
+// @flow strict
+import { type Node } from 'react';
+import Fieldset from './Fieldset.js';
+import Flex from './Flex.js';
+import RadioGroupButton from './RadioGroupButton.js';
+import { RadioGroupContextProvider } from './RadioGroupContext.js';
+
+type Props = {|
+  /**
+   * The individual RadioGroup.RadioButtons for this group
```
dangerismycat, line 10: We should probably mention that children can be grouped if necessary. Also worth mentioning any other components that would be valid in here, like Label or any others (and specifically call out which components should _not_ by valid). Something like

> A collection of RadioGroup.RadioButton representing the available options, as well as any Labels or ____ if needed. Other components such as ____ should _not_ be included. Note that children can be grouped into organizational components if desired.

packages/gestalt/src/RadioGroup.js
```diff
@@ -0,0 +1,64 @@
...
+   *
+   */
+  children: Node,
+  /**
+   * A unique identifier for this RadioGroup. `id` must be specified when an `errorMessage` is added.
+   *
+   */
+  id?: string,
```
dangerismycat, line 18: Should we just make this always required? Even if not strictly necessary, it would eliminate the possibility of forgetting it with errorMessage.

packages/gestalt/src/RadioGroup.js
```diff
@@ -0,0 +1,64 @@
...
+  /**
+   * The legend of the radio group that describes what is being selected.
```
dangerismycat, line 20: Consider not repeating the word "legend" in the description — how would we describe this to someone who doesn't know what a legend is?

packages/gestalt/src/RadioGroup.js
```diff
@@ -0,0 +1,64 @@
...
+  errorMessage?: string,
+  /**
+   * Whether the legend should be visible or not. If hidden, the legend is still available for screen reader users, but does not appear visually. See the [legend visibility](https://gestalt.pinterest.systems/radiogroup#Legend-visibility) variant to learn more.
+   *
+   */
+  legendDisplay?: 'visible' | 'hidden',
+|};
+
+/**
+ *  RadioGroups are used for selecting only 1 item from a list of 2 or more items. If you need multiple selection or have only one option, use [Checkbox](https://gestalt.pinterest.systems/checkbox). If you need, to provide a binary on/off choice that takes effect immediately, use [Switch](https://gestalt.pinterest.systems/switch).
```
dangerismycat, line 42: Don't forget to make RadioGroup a link for VS Code users.

typo: extra comma after "If you need"

packages/gestalt/src/RadioGroup.js
```diff
@@ -0,0 +1,64 @@
...
+ *
+ */
+export default function RadioGroup({
+  children,
+  errorMessage,
+  id,
+  legend,
+  legendDisplay,
+  direction = 'column',
```
dangerismycat, line 51: super nit: alphabetize these please 🙏

packages/gestalt/src/RadioGroupButton.js
```diff
@@ -0,0 +1,192 @@
...
+   */
+  label?: string,
+  /**
+   * The name given for all radio buttons in a single group.
+   */
+  name?: string,
+  /**
+   * Callback triggered when the user interacts with the input.
+   */
+  onChange: AbstractEventHandler<SyntheticInputEvent<HTMLInputElement>, {| checked: boolean |}>,
```
dangerismycat, line 42: Since we don't define AbstractEventHandler anywhere in the docs, we've been inlining that type definition where it's used

packages/gestalt/src/RadioGroupButton.js
```diff
@@ -0,0 +1,192 @@
...
+   */
+  subtext?: string,
+  /**
+   * The value of the input.
+   */
+  value: string,
+|};
+
+/**
+ *  Use [RadioGroup.RadioButtons](https://gestalt.pinterest.systems/radiogroup) when you have a few options that a user can choose from. Never use radio buttons if the user can select more than one option from a list.
```
dangerismycat, line 62: Note that you can add an anchor to the URL to go directly to the sub-component.

Also: this description reads like the one for RadioGroup, and should probably also mention that this sub-component is only to be used within RadioGroup

packages/gestalt/src/RadioGroupButton.js
```diff
@@ -0,0 +1,192 @@
...
+              </Box>
+            )}
+          </Box>
+        </Label>
+      )}
+    </Box>
+  );
+});
+
+RadioGroupButtonWithForwardRef.displayName = 'RadioGroupButton';
```
dangerismycat, line 190: The displayName should match the exported (from Gestalt) sub-component name, otherwise dev tools markup is confusing

packages/gestalt/src/RadioGroupContext.js
```diff
@@ -0,0 +1,24 @@
+// @flow strict
+import { createContext, useContext, type Context } from 'react';
+
+type RadioGroupContextType = {|
+  parentName: string,
+  hasError?: boolean,
```
dangerismycat, line 6: We don't seem to be using this field, and the Hook below doesn't even return it. Can we remove it?

packages/gestalt/src/RadioGroupContext.js
```diff
@@ -0,0 +1,24 @@
...
+};
+
+const context: Context<RadioGroupContextType> =
+  createContext<RadioGroupContextType>(initialContextState);
+
+const RadioGroupContextProvider = context.Provider;
+
+function useRadioGroupContext(): RadioGroupContextType {
+  const { parentName } = useContext(context);
+  return { parentName };
```
dangerismycat, line 21: This should probably return the whole Context rather than just this one field

docs/pages/radiogroup.js
```diff
@@ -0,0 +1,695 @@
...
+import Page from '../components/Page.js';
+import { multipledocgen, type DocGen } from '../components/docgen.js';
+
+export default function DocsPage({
+  generatedDocGen,
+}: {|
+  generatedDocGen: {| [string]: DocGen |},
+|}): Node {
+  return (
+    <Page title="RadioGroup">
```
dangerismycat, line 16: nit: the title here and name for PageHeader can be autogen as well (`generatedDocGen.RadioGroup.displayName`)

docs/pages/radiogroup.js
```diff
@@ -0,0 +1,695 @@
...
+              value="Prefer not to state"
+            />
+          </RadioGroup>
+        );
+      }
+      `}
+      />
+
+      <GeneratedPropTable generatedDocGen={generatedDocGen.RadioGroup} />
+      <GeneratedPropTable
```
dangerismycat, line 65: Check out our other sub-component docs; wrapping this in a MainSection and a Subsection allows us to display the sub-component's description and improves the formatting a bit:
```js
<MainSection name="Subcomponents">
  <MainSection.Subsection
    title={generatedDocGen?.FlexItem?.displayName}
    description={generatedDocGen?.FlexItem?.description}
  >
    <GeneratedPropTable
      generatedDocGen={generatedDocGen.FlexItem}
      id={generatedDocGen?.FlexItem?.displayName}
      name={generatedDocGen?.FlexItem?.displayName}
    />
  </MainSection.Subsection>
</MainSection>
```

docs/pages/radiogroup.js
```diff
@@ -0,0 +1,695 @@
...
+      />
+
+      <MainSection name="Usage guidelines">
+        <MainSection.Subsection columns={2}>
+          <MainSection.Card
+            cardSize="md"
+            type="do"
+            title="When to use"
+            description={`
+          - In a list, form or a Table, to present users with multiple, related options where only one option can be selected.
```
dangerismycat, line 79: The table usage can be a bit more abstract than just our component specifically, so more like "In a list, form, or table, to present…"

docs/pages/radiogroup.js
```diff
@@ -0,0 +1,695 @@
...
+          - When selection doesn’t take immediate effect and requires Form submission.
```
dangerismycat, line 80: Form doesn't need to be capitalized, not referring to a specific component

docs/pages/radiogroup.js
```diff
@@ -0,0 +1,695 @@
...
+          />
+          <MainSection.Card
+            cardSize="md"
+            type="don't"
+            title="When not to use"
+            description={`
+          - Situations where users can select multiple options. Use [Checkbox](/checkbox) instead.
+          - When there is only one item to select or deselect. Use Checkbox instead.
+          - When a selection takes immediate effect, especially on mobile. Use [Switch](/switch) instead.
+          - When visually, it’s hard to tell that a RadioGroup turns something on or off. Use Switch instead.
```
dangerismycat, line 91: This is a little awkward. Consider something like "When it is visually difficult to observe that RadioGroup turns something on or off."

docs/pages/radiogroup.js
```diff
@@ -0,0 +1,695 @@
...
+      </MainSection>
+      <MainSection name="Best Practices" />
+      <MainSection name="Accessibility">
+        <MainSection.Subsection
+          title="Labels"
+          description="Each RadioButton in a RadioGroup should have a label that can be read by screen readers, and that can be clicked or tapped to make it easier for users to select and deselect options. Therefore, make sure to supply the label prop. If that’s not possible, make sure your stand-alone Label has an `htmlFor` prop that matches the `id` of the RadioButton. Test that a RadioButton and label are properly connected by clicking or tapping on the label and confirming that it activates the RadioButton next to it."
+        />
+        <MainSection.Subsection
+          title="Legends"
+          description={`Each RadioGroup should have a legend that clearly delineates what is being chosen. If you cannot use the provided legend styling, \`legendDisplay\` can be set to \`hidden\`, and an alternative legend can be displayed.`}
```
dangerismycat, line 104: Does the alternate legend need to have a connection to the RadioGroup at all? A unique id or something similar?

docs/pages/radiogroup.js
```diff
@@ -0,0 +1,695 @@
...
+        />
+        <MainSection.Subsection
+          title="Keyboard interaction"
+          description={`After focus has been set on the first RadioButton inside a RadioGroup, the arrow keys are used to cycle focus between the various options. Clicking or tapping the label of RadioButton should also focus that particular RadioButton. `}
+        />
+        <MainSection.Subsection
+          title="Error message"
+          description={`If RadioGroup has an \`errorMessage\`, the \`id\` on RadioGroup must also be set to properly link the error message to RadioGroup.
```
dangerismycat, line 112: Again, I think `id` should just be required regardless. Completely obviates this type of error.

docs/pages/radiogroup.js
```diff
@@ -0,0 +1,695 @@
...
+          checked={favorite === 'plants'}
+          id="favoritePlants"
+          label="Plants"
+          name="favorite"
+          onChange={() => setFavorite( 'plants' )}
+          value="plants"
+        />
+      </RadioGroup>
+
+      <RadioGroup legend="What is your favorite snack?" errorMessage="Please select one" direction="row" id="directionExample>
```
dangerismycat, line 161: id is missing a closing `"`

docs/pages/radiogroup.js
```diff
@@ -0,0 +1,695 @@
...
+
+  return (
+    <Flex gap={8}>
+      <RadioGroup legend="What is your favorite pet?">
+        <RadioGroup.RadioButton
+          checked={favorite === 'dogs'}
+          id="favoriteDog"
+          label="Dogs"
+          name="favorite"
+          onChange={() => setFavorite( 'dogs' )}
```
dangerismycat, line 140: nit: spaces around the arg don't match our prettier config (here and throughout)

docs/pages/radiogroup.js
```diff
@@ -0,0 +1,695 @@
...
+          description={`If RadioGroup has an \`errorMessage\`, the \`id\` on RadioGroup must also be set to properly link the error message to RadioGroup.
+    `}
+        />
+      </MainSection>
+      <MainSection
+        name="Localization"
+        description={`Be sure to localize \`errorMessage\`, \`subtext\`, \`label\`, and \`legend\`. Be mindful of label length so that it doesn’t truncate in languages with lengthier character counts.`}
+      />
+
+      <MainSection name="Variants">
```
dangerismycat, line 121: Do we want more realistic examples? Most of these aren't Pinterest-specific

docs/pages/radiogroup.js
```diff
@@ -0,0 +1,695 @@
...
+      </RadioGroup>
+    </Flex>
+  );
+}
+          `}
+          />
+        </MainSection.Subsection>
+        <MainSection.Subsection
+          title="Size"
+          description="RadioButtons can be either `sm` (16px) or `md` (24px) which is the default."
```
dangerismycat, line 195: typo: comma needed after "(24px)"

docs/pages/radiogroup.js
```diff
@@ -0,0 +1,695 @@
...
+  generatedDocGen: {| [string]: DocGen |},
+|}): Node {
+  return (
+    <Page title="RadioGroup">
+      <PageHeader
+        name="RadioGroup"
+        description={generatedDocGen.RadioGroup?.description}
+        defaultCode={`
+      function RadioButtonExample() {
+        const [favorite, setFavorite] = React.useState(undefined);
```
dangerismycat, line 22: super nit: initializing with `undefined` is redundant (here and throughout)

docs/pages/radiogroup.js
```diff
@@ -0,0 +1,695 @@
...
+          id="checked"
+          label="Checked"
+          name="stateExample"
+          onChange={() => setFavorite( 'checked' )}
+          value="checked"
+        />
+        <RadioGroup.RadioButton
+          checked={false}
+          id="uncheckedDisabled"
+          label="Unchecked but disabled"
```
dangerismycat, line 297: nit: not sure if "but" is the right word for these two. "and" seems more appropriate

docs/pages/radiogroup.js
```diff
@@ -0,0 +1,695 @@
...
+          value="checkedDisabled"
+          disabled
+        />
+      </RadioGroup>
+  );
+}
+        `}
+          />
+        </MainSection.Subsection>
+        <MainSection.Subsection title="With subtext">
```
dangerismycat, line 318: Can this one have a description to clarify when/how subtext should be used?

docs/pages/radiogroup.js
```diff
@@ -0,0 +1,695 @@
...
+        />
+    </RadioGroup>
+  );
+}
+        `}
+          />
+        </MainSection.Subsection>
+        <MainSection.Subsection
+          title="With Image"
+          description="When including images, you can use the subtext property to clearly describe the information being presented by the image, or use the image's alt text to provide more context."
```
dangerismycat, line 361: "subtext" and "alt" should have backticks

docs/pages/radiogroup.js
```diff
@@ -0,0 +1,695 @@
...
+    </RadioGroup>
+  );
+}
+          `}
+          />
+        </MainSection.Subsection>
+        <MainSection.Subsection
+          title="Legend visibility"
+          description={`
+            By default, the legend is visible above the items in the RadioGroup. However, if the form items are labeled by content elsewhere on the page, or a more complex legend is needed, the \`legendDisplay\` prop can be used to visually hide the legend. In this case, it is still available to screen reader users, but will not appear visually on the screen.
```
dangerismycat, line 454: "legend" needs backticks throughout, as it's referring to the prop.

Also: is the "?" after "Primary company account goal" expected?

docs/pages/radiogroup.js
```diff
@@ -0,0 +1,695 @@
...
+      </RadioGroup>
+    </Flex>
+  );
+}
+        `}
+          />
+        </MainSection.Subsection>
+        <MainSection.Subsection
+          title="Using ref"
+          description={`The innermost \`RadioButton\` element can be accessed via \`ref\``}
```
dangerismycat, line 511: "innermost" isn't clear here. Are we referring to the underlying `<input>`?

Also this needs a period at the end.

docs/pages/radiogroup.js
```diff
@@ -0,0 +1,695 @@
...
+      </RadioGroup>
+    </Flex>
+  );
+}`}
+          />
+        </MainSection.Subsection>
+        <MainSection.Subsection
+          title="Adding a Popover"
+          description={`
+    A \`RadioButton\` with an anchor ref to a Popover component doesn't pass the correct positioning to the Popover. Instead set the anchor ref to the parent container.
```
dangerismycat, line 559: nit: no article before "RadioButton"

docs/pages/radiogroup.js
```diff
@@ -0,0 +1,695 @@
...
+                  href={
+                    option === "cat"
+                      ? "https://www.pinterest.com/search/pins/?q=cats"
+                      : "https://www.pinterest.com/search/pins/?q=dogs"
+                  }
+                  target='blank'
+                  underline="always"
+                >
+                  { option === "cat"
+                      ? "Check cats on Pinterest!"
```
dangerismycat, line 622: nit: should this be "Check _out_ cats…"?

docs/pages/radiogroup.js
```diff
@@ -0,0 +1,695 @@
...
+          />
+        </MainSection.Subsection>
+      </MainSection>
+      <MainSection name="Writing">
+        <MainSection.Subsection columns={2}>
+          <MainSection.Card
+            cardSize="md"
+            type="do"
+            description={`
+            - Be brief with RadioGroup button labels so they are easily scanned
```
dangerismycat, line 644: These should all end in periods

docs/pages/radiogroup.js
```diff
@@ -0,0 +1,695 @@
...
+            - Error messages should be simple, clear and direct without negative, overly clever and technical language
+            - A good error message: “To continue you must select one item from this list”`}
+          />
+          <MainSection.Card
+            cardSize="md"
+            type="don't"
+            description={`
+            - Include lengthy text labels that make it hard for a user to scan a list of choices
+            - Write error messages that are overly-technical, long, negative, and too clever
+            - A not-so-great error message: “Hey there, nice try, but not selecting something is baaaad. Bad as in bad. Per error code i-five, you must select a choice from this boolean”`}
```
dangerismycat, line 654: "select a choice from this boolean"?

docs/pages/radiogroup.js
```diff
@@ -0,0 +1,695 @@
...
+        <MainSection.Subsection
+          description={`
+    **[CheckBox](/checkbox)**
+    Use when presenting a user with a list of choices where multiple options can be selected.
+    `}
+        />
+        <MainSection.Subsection
+          description={`
+    **[Switch](/Switch)**
+    Use for single-cell options that can be turned on or off. Examples include a list of settings that take effect immediately without having to confirm Form submission.
```
dangerismycat, line 669: "Form" doesn't need to be capitalized

docs/pages/radiogroup.js
ayeshakmaz: it's supposed to be an example fo technical language used

packages/gestalt/src/RadioGroupButton.js
ayeshakmaz: is there an example of that?

docs/pages/visual-test/RadioGroup-light.js
```diff
@@ -0,0 +1,26 @@
+// @flow strict
+import { type Node } from 'react';
+import { RadioGroup, ColorSchemeProvider } from 'gestalt';
+
+export default function Snapshot(): Node {
+  return (
+    <ColorSchemeProvider colorScheme="light">
```
AlbertCarreras, line 7: This snap should be updated. The image is incorrect. Also can we add 

`      <Box color="white" display="inlineBlock" padding={1}>
`

wrapping the component so the image has some padding?

I'd would strongly suggest using a simple example rather than testing as this images are displayed in VSCOde to all engineers

docs/pages/visual-test/RadioGroup-dark.js
```diff
@@ -0,0 +1,28 @@
+// @flow strict
+import { type Node } from 'react';
+import { Box, RadioGroup, ColorSchemeProvider } from 'gestalt';
+
+export default function Snapshot(): Node {
+  return (
+    <ColorSchemeProvider colorScheme="dark">
+      <Box color="white">
+        <RadioGroup legend="testing" id="testing-example">
```
AlbertCarreras, line 9: Same comment as below re padding and example

docs/components/componentOverviewGeneralComponentsList.js
```diff
@@ -238,8 +238,8 @@ const GENERAL: ListItemType = [
   },
   {
     svg: <RadioButton />,
-    name: 'RadioButton',
-    description: 'se RadioButtons when you have a few options that a user can choose from.',
+    name: 'RadioGroup',
```
AlbertCarreras, line 241: Should we use `Fieldset`'s  svg instead? It's actually more representative. We can remove the RadioButton svg if so

docs/pages/radiobutton.js
```diff
@@ -13,335 +11,18 @@ export default function DocsPage({ generatedDocGen }: {| generatedDocGen: DocGen
...
-        size={['sm', 'md']}
-        id="radio-state-combos"
-        hasCheckerboard={false}
-        labelPrefix="radio-state-combos"
-      >
-        {(props, i) => (
-          <RadioButton id={`radio-state-combos-${i}`} onChange={() => {}} value="" {...props} />
-        )}
-      </Combination>
+      <GeneratedPropTable generatedDocGen={generatedDocGen} />
```
AlbertCarreras, line 25: LGTM!

docs/pages/radiogroup.js
AlbertCarreras: +1

docs/pages/radiogroup.js
AlbertCarreras: Screen Shot 2022-05-18 at 7 40 48 PM

All other cmps have the Subcomponent section right before Variants

packages/gestalt/src/RadioGroup.js
AlbertCarreras, line 10: Suggestion. Can we add the same SlimBanner content into the description on RadioButton so engs see the info?

And yes the info we'll be redundant but we can override the description for RadioButton easily so it's not duplicated in the docs. I feel passing the info to VScode can be helpful as well.

docs/pages/radiobutton.js
```diff
@@ -1,347 +1,31 @@
...
-          ref={ref}
-          size={size}
-        />
-    </Flex>
-  );
-}`}
+      <SlimBanner
+        type="error"
+        iconAccessibilityLabel="Info"
+        message="The stand alone RadioButton is soon to be deprecated, use RadioGroup and RadioGroup.RadioButton instead."
```
AlbertCarreras, line 20: "stand alone" or "standalone"?

ayeshakmaz: ahh good catch

docs/pages/radiogroup.js
```diff
@@ -0,0 +1,986 @@
...
+          />
+          <MainSection.Card
+            cardSize="md"
+            type="don't"
+            title="When not to use"
+            description={`
+          - Situations where users can select multiple options. Use [Checkbox](/checkbox) instead.
+          - When there is only one item to select or deselect. Use Checkbox instead.
+          - When a selection takes immediate effect, especially on mobile. Use [Switch](/switch) instead.
+          - When it is visually difficult to observe that RadioGroup turns something on or off. Use Switch instead.
```
AlbertCarreras, line 98: Switch with link, doesnt have one here

docs/pages/radiogroup.js
```diff
@@ -0,0 +1,986 @@
...
+          - When selection doesn’t take immediate effect and requires form submission.
+        `}
+          />
+          <MainSection.Card
+            cardSize="md"
+            type="don't"
+            title="When not to use"
+            description={`
+          - Situations where users can select multiple options. Use [Checkbox](/checkbox) instead.
+          - When there is only one item to select or deselect. Use Checkbox instead.
```
AlbertCarreras, line 96: Checkbox with link, doesnt have one here, unless your avoiding the second link on purpose

docs/pages/radiogroup.js
```diff
@@ -0,0 +1,986 @@
...
+            checked={favorite === "daily"}
+            id="daily"
+            name="budget"
+            onChange={() => setFavorite('daily')}
+            value="daily"
+          />
+          <Label htmlFor="daily">
+            <Flex alignItems="center">
+              <Text>Daily</Text>
+              <IconButton size="sm" icon="info-circle" iconColor="darkGray" accessibilityLabel="info" tooltip={{text: "Sets a cap for the amount your campaign can spend each day"}}/>
```
AlbertCarreras, line 206: Should this be an Icon wrapped in a TapArea? It's the Datapoint pattern 

```
<Tooltip
            accessibilityLabel=""
            idealDirection="up"
            text={tooltipText}
            zIndex={tooltipZIndex}
          >
            {/* Interactive elements require an a11yLabel on them or their children.
            That's why we set`accessibilityLabel` on `TapArea` instead of `Tooltip` */}
            <TapArea accessibilityLabel={tooltipText} rounding="circle" tapStyle="none">
              <Icon accessibilityLabel="" size={16} icon="info-circle" color="subtle" />
            </TapArea>
          </Tooltip>
        )}
```

docs/pages/radiogroup.js
ayeshakmaz: we link it right above, so I think in our guidelines we say only link once within a block

docs/pages/radiogroup.js
ayeshakmaz: I think an IconButton is also fine here

docs/pages/radiogroup.js
```diff
@@ -0,0 +1,986 @@
...
+  );
+}
+        `}
+          />
+        </MainSection.Subsection>
+      </MainSection>
+      <MainSection name="Accessibility">
+        <MainSection.Subsection
+          title="Labels"
+          description="Each RadioButton in a RadioGroup should have a label that can be read by screen readers, and that can be clicked or tapped to make it easier for users to select and deselect options. Therefore, make sure to supply the label prop. If that’s not possible, make sure your stand-alone Label has an `htmlFor` prop that matches the `id` of the RadioButton. Test that a RadioButton and label are properly connected by clicking or tapping on the label and confirming that it activates the RadioButton next to it."
```
AlbertCarreras, line 341: nit: check we are consistent on how we write stand-alone, stand alone, standalone

docs/pages/radiogroup.js
```diff
@@ -0,0 +1,986 @@
...
+          title="Legends"
+          description={`Each RadioGroup should have a \`legend\` that clearly delineates what is being chosen. If you cannot use the provided legend styling, \`legendDisplay\` can be set to \`hidden\`, and an alternative legend can be displayed. See the [legend visibility](#Legend-visibility) variant for an example.`}
+        />
+        <MainSection.Subsection
+          title="Keyboard interaction"
+          description={`After focus has been set on the first RadioButton inside a RadioGroup, the arrow keys are used to cycle focus between the various options. Clicking or tapping the label of RadioButton should also focus that particular RadioButton. All RadioGroup.RadioButtons within a RadioGroup should share the same \`name\` to ensure keyboard accessibility, but that \`name\` needs to be unique from other RadioGroup buttons on the page.`}
+        />
+        <MainSection.Subsection
+          title="Error message"
+          description={`If RadioGroup has an \`errorMessage\`, the \`id\` on RadioGroup must also be set to properly link the error message to RadioGroup.
```
AlbertCarreras, line 353: Can we lint to errorMessage Variant ?

ayeshakmaz: Removing this, doesn't apply anymore

docs/pages/radiogroup.js
```diff
@@ -0,0 +1,986 @@
...
+      </RadioGroup>
+    </Flex>
+  );
+}
+        `}
+          />
+        </MainSection.Subsection>
+        <MainSection.Subsection
+          title="States"
+          description="Remember that disabled RadioButtons cannot be accessed by the keyboard and therefore should not contain any necessary info to complete the choice presented."
```
AlbertCarreras, line 505: suggestion

"Remember that" or just "Disabled RadioButtons cannot be..."?

docs/pages/radiogroup.js
```diff
@@ -0,0 +1,986 @@
...
+          title="With an error"
+          description="Use `errorMessage` to show an error message below the radio options."
+        >
+          <MainSection.Card
+            defaultCode={`
+function RadioButtonExample() {
+  const [availability, setAvailability] = React.useState();
+
+  return (
+    <RadioGroup legend="Which time slot works best for you?" errorMessage="Please select one" id="VariantWithErrorMessage">
```
AlbertCarreras, line 656: nit error could disappear with selection
` errorMessage={!availability && "Please select one"}  `

docs/pages/radiogroup.js
```diff
@@ -0,0 +1,986 @@
...
+          description="The `label` on RadioGroup.RadioButton can be replaced with a custom [Label](/Label), as demonstrated below. Ensure the `htmlFor` property matches the `id` on the RadioButton."
+        >
+          <MainSection.Card
+            defaultCode={`
+        function RadioButtonExample() {
+          const [favorite, setFavorite] = React.useState();
+
+          return (
+              <RadioGroup legend="Campaign budget" id="bestPracticeBudget">
+                <Flex direction="row" gap={2} alignItems="center" justifyContent="start">
```
AlbertCarreras, line 701: `justifyContent="start"` is not needed

docs/pages/radiogroup.js
```diff
@@ -0,0 +1,986 @@
...
+                    checked={favorite === "daily"}
+                    id="daily-label-ex-custom"
+                    name="budget-custom-label"
+                    onChange={() => setFavorite('daily')}
+                    value="daily"
+                  />
+                  <Label htmlFor="daily-label-ex-custom">
+                    <Flex alignItems="center">
+                      <Text>Daily</Text>
+                      <IconButton size="sm" icon="info-circle" iconColor="darkGray" accessibilityLabel="info" tooltip={{text: "Sets a cap for the amount your campaign can spend each day"}}/>
```
AlbertCarreras, line 712: Same as previous comment re TapArea + Icon to follow DataPoint pattern

docs/pages/radiogroup.js
```diff
@@ -0,0 +1,986 @@
...
+          value="interest"
+        />
+      </RadioGroup>
+    </Flex>
+  );
+}
+        `}
+          />
+        </MainSection.Subsection>
+        <MainSection.Subsection
```
AlbertCarreras, line 795: I didn't include this sectoin when we rewrote Checkbox. Thoughts about removing it?

docs/pages/radiogroup.js
```diff
@@ -0,0 +1,986 @@
...
+            idealDirection="right"
+            onDismiss={() => setOpen(false)}
+            positionRelativeToAnchor={false}
+            shouldFocus={false}
+            size="md"
+          >
+            <Box padding={3}>
+              <Text
+                color="default"
+                weight="bold"
```
AlbertCarreras, line 896: Bold + underline shouldn't go together, check with Cintia?

docs/pages/radiogroup.js
```diff
@@ -0,0 +1,986 @@
...
+          checked={favorite === 'plants'}
+          id="favoritePlants"
+          label="Plants"
+          name="favorite"
+          onChange={() => setFavorite('plants')}
+          value="plants"
+        />
+      </RadioGroup>
+
+      <RadioGroup legend="What is your favorite snack?" errorMessage="Please select one" direction="row" id="directionExample">
```
AlbertCarreras, line 397: Should we use the recommended error message? 

`To continue you must select one item from this list`

docs/pages/radiogroup.js
```diff
@@ -0,0 +1,986 @@
...
+          <MainSection.Card
+            defaultCode={`
+function RadioButtonExample() {
+  const [favorite, setFavorite] = React.useState();
+  const [favoriteFood, setFavoriteFood] = React.useState();
+
+  return (
+    <Flex gap={8}>
+      <RadioGroup legend="What is your favorite snack?" errorMessage="Please select one" id="sizeExample">
+        <RadioGroup.RadioButton
```
AlbertCarreras, line 442: should we use the recommended one? `To continue you must select one item from this list`

docs/pages/radiogroup.js
```diff
@@ -0,0 +1,986 @@
...
+        />
+        <RadioGroup.RadioButton
+          checked={favoriteFood === 'curry'}
+          id="favoriteCurry"
+          label="Curry"
+          name="favoriteFood"
+          onChange={() => setFavoriteFood('curry')}
+          value="curry"
+        />
+        <RadioGroup.RadioButton
```
AlbertCarreras, line 414: Any thoughts on why it's saying 3 out of "6
Screen Shot 2022-05-19 at 3 46 28 PM
"

packages/gestalt/src/RadioGroup.js
```diff
@@ -0,0 +1,64 @@
...
+ *  [RadioGroups](https://gestalt.pinterest.systems/radiogroup) are used for selecting only 1 item from a list of 2 or more items. If you need multiple selection or have only one option, use [Checkbox](https://gestalt.pinterest.systems/checkbox). If you need to provide a binary on/off choice that takes effect immediately, use [Switch](https://gestalt.pinterest.systems/switch).
+ *
+ */
+export default function RadioGroup({
+  children,
+  direction = 'column',
+  errorMessage,
+  id,
+  legend,
+  legendDisplay,
```
AlbertCarreras, line 51: should we set a default so it shows up on the prop table? It's the default from Fieldset but it doesnt get reflected if not specified here too

packages/gestalt/src/RadioGroup.test.js
```diff
@@ -0,0 +1,33 @@
...
+  it('renders', () => {
+    const tree = create(
+      <RadioGroup legend="testing" id="testing-example">
+        <RadioButton id="choice-1" onChange={() => {}} name="choice" value="choice-1" />
+      </RadioGroup>,
+    ).toJSON();
+    expect(tree).toMatchSnapshot();
+  });
+
+  it('renders a hidden legend', () => {
```
AlbertCarreras, line 16: snap for non default direction?

packages/gestalt/src/RadioGroupButton.js
AlbertCarreras: @ayeshakmaz  This one is 

`onChange: ({| event: SyntheticInputEvent<HTMLInputElement>, checked: boolean |}) => void,
`

packages/gestalt/src/RadioGroupButton.js
```diff
@@ -0,0 +1,192 @@
...
+    subtext,
+    value,
+    size = 'md',
+  }: Props,
+  ref,
+): Node {
+  const [focused, setFocused] = useState(false);
+  const [hovered, setHover] = useState(false);
+
+  const handleChange: (event: SyntheticInputEvent<HTMLInputElement>) => mixed = (event) =>
```
AlbertCarreras, line 89: mixed? does `void` work?

packages/gestalt/src/RadioGroupButton.js
```diff
@@ -0,0 +1,192 @@
...
+
+  const styleSize = size === 'sm' ? controlStyles.sizeSm : controlStyles.sizeMd;
+
+  const bgStyle = disabled && !checked ? styles.BgDisabled : styles.BgEnabled;
+
+  const { isFocusVisible } = useFocusVisible();
+
+  const { parentName } = useRadioGroupContext();
+  if (parentName !== 'RadioGroup') {
+    throw new Error(`RadioGroup.RadioButton must be used within a RadioGroup.`);
```
AlbertCarreras, line 124: can we add link here? 

    throw new Error(`RadioGroup.RadioButton must be used within a [RadioGroup](https://gestalt.pinterest.systems/radiogroup).`);

packages/gestalt/src/RadioGroupButton.js
```diff
@@ -0,0 +1,192 @@
...
+  }
+
+  return (
+    <Box
```
AlbertCarreras, line 128: Not looking at this as is same logic as RadioButton

packages/gestalt/src/RadioGroupButton.js
```diff
@@ -0,0 +1,192 @@
...
+  /**
+   * The value of the input.
+   */
+  value: string,
+|};
+
+/**
+ *  Use [RadioGroup.RadioButtons](https://gestalt.pinterest.systems/radiogroup#RadioGroup.RadioButton) to present an option for selection to the user within a RadioGroup. They should not be used outside of RadioGroup or when the user can select more than one option from a list.
+ *
+ * ![RadioButton light mode](https://raw.githubusercontent.com/pinterest/gestalt/master/cypress/integration/visual-test/__image_snapshots__/RadioButton%20%230.png)
```
AlbertCarreras, line 64: We should remove this ones and replace them with RadioGroup

packages/gestalt/src/RadioGroup.css
```diff
@@ -0,0 +1,3 @@
+.customClass {
```
AlbertCarreras, line 1: @ayeshakmaz Is this the default set up by the template? Remove and `run yarn run flow-generate:css` to remove 
packages/gestalt/src/RadioGroup.css.flow

AlbertCarreras (approved):
LGTM! Thanks!

Some are nits, but some are strong suggestions

packages/gestalt/src/RadioGroup.css
AlbertCarreras: not a nit, Should be removed if unused

packages/gestalt/src/RadioGroupButton.js
AlbertCarreras: not a nit, Should be removed if unused

packages/gestalt/src/RadioGroup.js
AlbertCarreras: not a nit

docs/pages/radiogroup.js
AlbertCarreras: just curious...

docs/pages/radiogroup.js
AlbertCarreras: I think these will be replaced with @ponori's  new help/info button components, so we'll be consistent with those and we'll clean up this IconButtons. I prefer TapAreas, take less space, dont have a fake click with no event... But cos we don't have a consistent way of doing this helper/info buttons, I guess we can go either way

docs/pages/radiogroup.js
ayeshakmaz: Given it's not in a flow, I think Please select one is fine

docs/pages/radiogroup.js
ayeshakmaz: I'll leave as is, since we want people to visually see what the error looks like, and there's no way to de-select a radio after it's been checked

packages/gestalt/src/RadioGroupButton.js
ayeshakmaz: Ahhh I see, thank you!

docs/pages/radiogroup.js
ayeshakmaz: I think given best practices are realistic, and the others are at least realistic uses of radio buttons, then it's probably fine for now

docs/pages/radiogroup.js
ayeshakmaz: Yeah probably not worth worrying about with InfoButton on the horizon

docs/pages/radiogroup.js
ayeshakmaz: same here

docs/pages/radiogroup.js
ayeshakmaz: Oh the same name was being used for two different examples - sorry updated without commenting

docs/pages/radiogroup.js
dangerismycat: I more meant that one can't really select a choice from a true or false value, but I also think it's funny so ¯\_(ツ)_/¯ 😂

docs/pages/radiogroup.js
dangerismycat: It is a little weird that it still shows the compressed click behavior though. Agreed that it shouldn't really matter (esp since readers can't see the code), but the "different colors on buttons with icons" convo from the Menu button example shows that people do pick up on these small details in the docs

docs/pages/radiogroup.js
dangerismycat: Agreed that this is fine — we do similar things with error examples on other components

docs/pages/radiogroup.js
dangerismycat: I'm fine with removing it, since it's not a prop we encourage people to use unless their use case really, really calls for it

dangerismycat (approved):

docs/pages/radiogroup.js
ayeshakmaz: I updated the color for now to match datapoint, was thinking the migration from IconButton -> InfoButton will be easier than from a Tap Area surrounding an Icon, but we can address this when InfoButton comes along.

AlbertCarreras (approved):

PR #2126, opened 2022-05-19
Docs: adding slimBanner prop into Docs' PageHeader to display information related to the component. 

AlbertCarreras:
#### What changed?

Docs: adding slimBanner prop into Docs' PageHeader to display information related to the component. 

##BEFORE
Screen Shot 2022-05-19 at 5 32 15 PM

##AFTER
Screen Shot 2022-05-19 at 5 29 36 PM

#### Why?

New comms pattern re components

### DIFF
--- a/docs/components/PageHeader.js
+++ b/docs/components/PageHeader.js
@@ -1,20 +1,9 @@
 // @flow strict
-import { type Node } from 'react';
-import { Badge, Box, Flex, Heading, Text, Tooltip } from 'gestalt';
+import { type Node, type Element } from 'react';
+import { Badge, Box, Flex, Heading, Text, Tooltip, SlimBanner } from 'gestalt';
 import Markdown from './Markdown.js';
 import MainSection from './MainSection.js';
 import trackButtonClick from './buttons/trackButtonClick.js';
-
-type Props = {|
-  badge?: 'pilot' | 'deprecated',
-  defaultCode?: string,
-  description?: string,
-  fileName?: string, // only use if name !== file name
-  folderName?: string, // only use if name !== file name and the link should point to a directory
-  name: string,
-  shadedCodeExample?: boolean,
-  showSourceLink?: boolean,
-|};
 
 const buildSourceLinkPath = (componentName) => {
   const packageName = componentName === 'DatePicker' ? 'gestalt-datepicker' : 'gestalt';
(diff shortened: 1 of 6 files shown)


docs/components/PageHeader.js
```diff
@@ -52,49 +54,50 @@ export default function PageHeader({
...
-            <a
-              href={sourceLink}
-              onClick={() => trackButtonClick('View source on GitHub', sourcePathName)}
-              target="blank"
-            >
-              <Text underline>View source on GitHub</Text>
-            </a>
-          )}
+          {description && <Markdown text={description} />}
+          {slimBanner ? slimBanner : null}
```
dangerismycat, line 89: nit: add `null` as a default value for slimBanner and this can just be
```js
{slimBanner}
```

docs/components/PageHeader.js
```diff
@@ -52,49 +54,50 @@ export default function PageHeader({
...
-            {badge === 'pilot' ? (
-              <Tooltip
-                inline
-                text={`This is the initial version of ${name}, and additional (non-breaking) functionality is planned for the future. Any feedback is greatly appreciated!`}
+      <Flex direction="column" gap={defaultCode ? 8 : 0}>
+        <Flex direction="column" gap={2}>
+          <Flex alignItems="baseline" direction="row" gap={2} justifyContent="between" wrap>
+            <Heading>
+              {name}{' '}
+              {badge === 'pilot' ? (
```
dangerismycat, line 62: unrelated nit, ignore if you want: this could be a bit simpler/DRYer with a map:
```js
const badgeMap = {
  pilot: { text: ... , tooltipText ... },
  deprecated: { text: ... , tooltipText ... },
};

...

{badge && (
  <Tooltip inline text={badges[badge].tooltipText}>
    <Badge position="top" text={badges[badge].text} />
  </Tooltip>
)}
```
)}

docs/pages/tooltip.js
```diff
@@ -822,7 +837,11 @@ Toast provides feedback on an interaction. One example of Toast is the confirmat
 }
 
 export async function getServerSideProps(): Promise<{| props: {| generatedDocGen: DocGen |} |}> {
+  const generatedDocGen = await docgen({ componentName: 'Tooltip' });
+
+  generatedDocGen.description = generatedDocGen.description.split('\n')[0];
```
dangerismycat, line 842: This is a bit brittle, as it relies on the comment formatting rather than the content. Can we instead take an approach similar to how we read default values in comments, such that we can display only the description on the docs, and the description + banner content in IDEs?

dangerismycat (approved):

docs/pages/tooltip.js
AlbertCarreras: Let me see... the problem is that we want the Alert text on the description... I'll try to see if I can parse somehow.

docs/components/PageHeader.js
AlbertCarreras: DONE

docs/components/PageHeader.js
AlbertCarreras: DONE

PR #2129, opened 2022-05-21
Video: Add startTime prop to explicitly set start time on video node

cmswalker:
Video: Add startTime prop to explicitly set start time on video node currentTime property

### Summary
Adding starTime prop to support setting start times for videos

#### What changed?
New prop that sets currentTime via `seek()`. Can be called before video plays/with autoplay/or any time in between

#### Why?

Modifying `currentTime` manually is an anticipated action for the video api.

Being able to load a video from a set start time [is important](https://youtu.be/2kWUbrx4EMw?t=11). You can embed a querystring in the url such as `?t=10` to send to someone else, the consumer can then pull that timestamp out and provide it to the video player to start at that point.

This is also a requirement for using [key moments](https://youtu.be/rsFLIewWSrI?t=271) support via google SEO

### DIFF
--- a/packages/gestalt/src/Video.js
+++ b/packages/gestalt/src/Video.js
@@ -216,6 +216,10 @@ type Props = {|
    */
   src: Source,
   /**
+   * Set the current play time in seconds the video will start from. See [MDN Web Docs: HTMLMediaElement.currentTime](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/currentTime)
+   */
+  startTime?: number,
+  /**
    * Specifies the volume of the video audio: 0 for muted, 1 for max. See the [video controls variant](https://gestalt.pinterest.systems/video#Video-controls) to learn more.
    */
   volume: number,
@@ -331,6 +335,7 @@ export default class Video extends PureComponent<Props, State> {
   player: ?HTMLDivElement;
 
   static defaultProps: {|
+    startTime: number,
     disableRemotePlayback: boolean,
     backgroundColor: BackgroundColor,
     playbackRate: number,
@@ -338,6 +343,7 @@ export default class Video extends PureComponent<Props, State> {
     preload: 'auto' | 'metadata' | 'none',
     volume: number,
   |} = {
+    startTime: 0,
     disableRemotePlayback: false,
     // eslint-disable-next-line react/default-props-match-prop-types
     backgroundColor: 'black',
@@ -363,7 +369,7 @@ export default class Video extends PureComponent<Props, State> {
    */
 
   componentDidMount() {
-    const { captions, playbackRate, volume, playing, autoplay } = this.props;
+    const { captions, playbackRate, volume, playing, autoplay, startTime } = this.props;
     // Set up event listeners to catch backdoors in fullscreen
     // changes such as using the ESC key to exit
     if (typeof document !== 'undefined') {
@@ -375,6 +381,10 @@ export default class Video extends PureComponent<Props, State> {
     this.setVolume(volume);
     // Set the initial playback rate
     this.setPlaybackRate(playbackRate);
+
+    if (startTime) {
+      this.seek(startTime);
+    }
 
     if (!autoplay && playing) {
       this.play();
--- a/packages/gestalt/src/Video.test.js
+++ b/packages/gestalt/src/Video.test.js
@@ -136,3 +136,18 @@ test('Video with objectFit', () => {
   ).toJSON();
   expect(tree).toMatchSnapshot();
 });
+
+test('Video with startTime', () => {
+  const tree = create(
+    <Video
+      {...A11Y_LABELS}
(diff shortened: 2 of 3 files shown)


cmswalker:
An alternative could be to provide `currentTime` prop directly for manipulating video ref time that way so users no longer have to manage this outside of the video component

packages/gestalt/src/Video.js
```diff
@@ -376,6 +382,10 @@ export default class Video extends PureComponent<Props, State> {
     // Set the initial playback rate
     this.setPlaybackRate(playbackRate);
 
+    if (startTime) {
```
dangerismycat, line 385: We probably also want to add logic in componentDidUpdate to listen for any changes to `startTime`, yes? Or would we want to ignore those changes? I could see a use case where the start time is changed outside of Video, and we'd want to jump to that new start time — but let me know if that's not the case.

packages/gestalt/src/Video.test.js
```diff
@@ -136,3 +136,18 @@ test('Video with objectFit', () => {
   ).toJSON();
   expect(tree).toMatchSnapshot();
 });
+
+test('Video with startTime', () => {
```
dangerismycat, line 140: Thanks for adding tests! I'm not sure if a snapshot is very valuable for this test, though. What would be useful is a test that passes a `startTime` and verifies that the `<video>` element seeks to that time.

packages/gestalt/src/Video.js
cmswalker: Yeah, I thought about this for a bit. It's probably the best way to avoid outside components updating the `currentTime` if they need to do it once video play is underway. Otherwise we'd need to expose `seek` API for the wrapper

packages/gestalt/src/Video.test.js
cmswalker: Yeah I can kill this one if thats easier, I noticed some of this testing isn't possible without a headed browser testing suite which is why I moved it here along with some of the other assertions like `objectFit` where they're more like contract tests

packages/gestalt/src/Video.js
cmswalker: If we want a constant listener, maybe we should call this `currentTime` instead of `startTime` ?

dangerismycat: > If we want a constant listener, maybe we should call this `currentTime` instead of `startTime` ?

I was initially thinking that keeping closer to the native API would be better, but since `currentTime` is read-write, we'd have to keep it in sync, which sounds like a bit of a headache. (And using the native attribute name without replicating the native behavior would be confusing.)

> It's probably the best way to avoid outside components updating the `currentTime` if they need to do it once video play is underway. Otherwise we'd need to expose `seek` API for the wrapper

Ok, if that's something we want to avoid, that's fine. From the [fastSeek docs](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/fastSeek) it sounds like this would be the prop to use to programmatically to a future time, but I'm also fine with leaving out that functionality for now if it's not needed:
> Note: If you need to seek with precision, you should set [HTMLMediaElement.currentTime](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/currentTime) instead.

packages/gestalt/src/Video.test.js
dangerismycat: I might be misunderstanding the timing of the logic, but I was thinking of something like passing a `startTime` and a `ref` and then reading the `currentTime` off the ref to confirm that it matches. I think that should be possible?

cmswalker: The attrs and properties of all the nodes are super mocked out under test in headless environments things like `disableremoteplayback` are there but stuff it actually tests against the video asset like times are proxied out

packages/gestalt/src/Video.js
cmswalker: Yeah fastSeek is cool but no chrome support 😢 . I'll update the lifecycle methods for startTime and hopefully that's clear enough for it to stick. Having a hard time envisioning a scenario where someone needs to programatically set `currentTime` multiple times without user interaction so hopefully viewing it as a startTime only is good enough

packages/gestalt/src/Video.test.js
dangerismycat: Ah cool, no worries. We _just_ switched to all headless (like this morning) and I haven't had a chance to play around with it much, so I'm not familiar with it yet. If some sort of logic test like that seems possible, let's do it; if it looks like it'd just be a snapshot, don't worry about adding the test.

cmswalker: I think snapshot is all we can do here since `node.currenttime` is mocked out totally in headless. My guess is any of the attributes that reflect video state vs the broad config ones aren't available in headless since its probably mocking them out

PR #2138, opened 2022-05-31
Text + Heading: Deprecate old size options

EduhCosta:
### Summary

Remove deprecated sizes from Heading and Text (`sm`, `md` and `lg`).

#### Why?

Align with design tokens and avoid arbitrary size usage.

To make any conversions using our codemods, run
`yarn codemod modifyPropValue [path-to-files] --component=Heading --previousProp=size --previousValue=[value to replace] --nextValue=[new value]`
and
`yarn codemod modifyPropValue [path-to-files] --component=Text --previousProp=size --previousValue=[value to replace] --nextValue=[new value]`

### Links

- [Jira](https://jira.pinadmin.com/browse/GESTALT-4162)

### Checklist

### DIFF
--- a/packages/gestalt/src/Heading.js
+++ b/packages/gestalt/src/Heading.js
@@ -18,16 +18,10 @@ const defaultHeadingLevels = {
   '400': 3,
   '500': 2,
   '600': 1,
-  'sm': 3,
-  'md': 2,
-  'lg': 1,
 };
 
 // Corresponds to the font-size design tokens
 const SIZE_SCALE = {
-  sm: 400,
-  md: 500,
-  lg: 600,
   '100': 100,
   '200': 200,
   '300': 300,
@@ -38,7 +32,7 @@ const SIZE_SCALE = {
 
 type AccessibilityLevel = 1 | 2 | 3 | 4 | 5 | 6 | 'none';
 type Overflow = 'normal' | 'breakWord';
-type Size = '100' | '200' | '300' | '400' | '500' | '600' | 'sm' | 'md' | 'lg';
+type Size = '100' | '200' | '300' | '400' | '500' | '600';
 
 type Props = {|
   /**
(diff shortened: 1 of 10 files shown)


packages/gestalt/src/Heading.js
```diff
@@ -18,16 +18,10 @@ const defaultHeadingLevels = {
   '400': 3,
   '500': 2,
   '600': 1,
-  'sm': 3,
-  'md': 2,
-  'lg': 1,
 };
 
 // Corresponds to the font-size design tokens
 const SIZE_SCALE = {
```
ayeshakmaz, line 24: Can this const now be removed since we no longer need to map from the old values?

packages/gestalt/src/Text.js
```diff
@@ -12,9 +12,6 @@ function isNotNullish(val): boolean {
 }
 
 const SIZE_SCALE = {
```
ayeshakmaz, line 14: Same question here, can this be removed?

ayeshakmaz:
@EduhCosta Can you also remove the lines in both prop descriptions (Heading and Text) that say "The "sm", "md", and "lg" values will soon be deprecated."?

packages/gestalt/src/Heading.js
AlbertCarreras: @ayeshakmaz It's mapping from string to number... do we need to keep that? Not super familiar

ayeshakmaz: I haven't tried - I would see if the string concatenation is smart enough without the conversion, otherwise using something like .Number would work. we're making a string out of this string, so hopefully that's not necessary

EduhCosta:
> @EduhCosta Can you also remove the lines in both prop descriptions (Heading and Text) that say "The "sm", "md", and "lg" values will soon be deprecated."?

Done

ayeshakmaz (approved):
Thank you! Looks great!

PR #2139, opened 2022-06-01
Internal: Add Pinterest metadata tag to connect Gestalt Pinterest account w/ Gestalt site

ponori:
## What
Adding the Pinterest HTML tag to the 

## Why

Adding Pinterest HTML tag to connect the Gestalt Pinterest account  on Pinterest to the https://gestalt.pinterest.systems/ so we can officially claim it.
https://user-images.githubusercontent.com/50343812/171476588-5a994f80-9e90-4b22-9050-c46e15b2051c.png

### DIFF
--- a/docs/pages/_document.js
+++ b/docs/pages/_document.js
@@ -26,6 +26,7 @@ class GestaltDocument extends Document {
     return (
       <Html lang="en" dir={dir}>
         <Head>
+          <meta name="p:domain_verify" content="752e3976762ef39258186e60a40bbe5a" />
           {/* eslint-disable-next-line @next/next/no-sync-scripts */}
           <script
             type="text/javascript"


docs/pages/_document.js
```diff
@@ -26,6 +26,7 @@ class GestaltDocument extends Document {
     return (
       <Html lang="en" dir={dir}>
         <Head>
+          <meta name="p:domain_verify" content="752e3976762ef39258186e60a40bbe5a" />
```
AlbertCarreras, line 29: @ponori Can you give more context for this? Why are we adding this?

ponori: Yeah, absolutely. I'm adding this tag to connect our Gestalt Pinterest account to the site so we can officially claim it.
Screen Shot 2022-06-01 at 11 28 32 AM

PR #2148, opened 2022-06-03
Data Visualization: Add usage guidelines with examples

ayeshakmaz:
#### What changed?

Add page for examples and guidelines around the new data visualization color palette

#### Why?

Provide examples of color pairings, and pairings to avoid 

### Links

- [Figma](https://www.figma.com/file/UOCLenizykiIi9yxs8k992/Working-file%3A-Data-viz-palette?node-id=834%3A27996)

### Checklist


docs/pages/data_visualization_guidelines.js
```diff
@@ -0,0 +1,264 @@
...
+type ColorCardProps = {|
+  count: number,
+|};
+function PaletteGenerator({ count }: ColorCardProps): Node {
+  return [...Array(count)].map((step, idx) => {
+    const tokenStep = idx + 1;
+    return (
+      <Box marginBottom={1} key={`color-${idx}`}>
+        <ColorTile
+          key={`${idx}`}
```
dangerismycat, line 25: this key isn't doing anything

docs/pages/data_visualization_guidelines.js
```diff
@@ -0,0 +1,264 @@
...
+                fullTokenName="color-data-visualization-05"
+              />
+            </Flex>
+            <Flex direction="column" gap={1}>
+              <ColorTile
+                textColor="inverse"
+                description="Data Visualization 08"
+                fullTokenName="color-data-visualization-08"
+              />
+              <Box marginBottom={6}>
```
dangerismycat, line 201: is this marginBottom necessary?

dangerismycat (approved):

docs/pages/data_visualization_guidelines.js
ayeshakmaz: yeah it's for when it wraps into one column on mobile. Not the cleanest solution, but I'll add a comment

docs/pages/data_visualization_guidelines.js
ayeshakmaz: good catch!

docs/pages/data_visualization_guidelines.js
dangerismycat: 👍 This is why I need to re-implement `gap` to include both column and row values 😬

ayeshakmaz: Yes please :P

PR #2153, opened 2022-06-13
Docs: Heading component design guidelines

ponori:
Adding design guidelines for the Heading component

### DIFF
--- a/docs/pages/heading.js
+++ b/docs/pages/heading.js
@@ -329,29 +329,35 @@ export default function DocsPage({ generatedDocGen }: {| generatedDocGen: DocGen
             cardSize="lg"
             defaultCode={`
 <Flex maxWidth={240} direction="column" gap={8}>
-  <Box color="secondary" padding={2} rounding={2}>
+  <Flex direction="column" gap={2}>
     <Text>breakWord (default):</Text>
-    <Heading size="400" overflow="breakWord" >
-      This is a long and Supercalifragilisticexpialidocious sentence.
-      次の単語グレートブリテンおよび北アイルランド連合王国で本当に大きな言葉
-    </Heading>
-  </Box>
+    <Box color="secondary" padding={2} rounding={2}>
+      <Heading size="400" overflow="breakWord">
+        This is a long and Supercalifragilisticexpialidocious sentence.
+        次の単語グレートブリテンおよび北アイルランド連合王国で本当に大きな言葉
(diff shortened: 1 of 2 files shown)


docs/pages/heading.js
```diff
@@ -99,64 +335,78 @@ export default function DocsPage({ generatedDocGen }: {| generatedDocGen: DocGen
...
-  "
-        name="Example: Levels"
-        defaultCode={`
-<Box>
+          />
+        </MainSection.Subsection>
+        <MainSection.Subsection
+          title="Accessibility level"
+          description="For accessibility purposes, we allow you to override the heading level.
+
```
AlbertCarreras, line 403: we allow you to override the heading level. Any guidance on why Heading with no level vs just Text weight="bold"?

I think the none value is legacy, maybe we could not mentioned it. Or maybe mention is legacy and Text is preferred

docs/pages/heading.js
```diff
@@ -1,23 +1,258 @@
...
+            href: '/typography',
+            onClick: () => {},
+          }}
+        />
+      </MainSection>
+      <MainSection name="Accessibility">
+        <MainSection.Subsection title="Logical order">
+          <Markdown text="Users will find a logical heading order and structure very helpful, especially if they have difficulty with reading and language, or if they use assistive devices such as a screen reader. A clear structure will help a screen reader user navigate the app without getting confused. Our headings default to a heading level based on size. For example:" />
+
+          <Table accessibilityLabel="Heading order">
```
AlbertCarreras, line 174: This table needs some love. Maybe we can improve spacing in a follow up PR

docs/pages/heading.js
```diff
@@ -99,64 +335,78 @@ export default function DocsPage({ generatedDocGen }: {| generatedDocGen: DocGen
...
 `}
-      />
-      <Example
-        id="align"
-        name="Example: Alignment"
-        defaultCode={`
-<Box>
+          />
+        </MainSection.Subsection>
+        <MainSection.Subsection title="Alignment">
```
AlbertCarreras, line 380: Can we keep text aligned with Text? It's the same section, so we could just opy paste from one another. This one is empty but we could have the same description as in Text.

AlbertCarreras (approved):
LGTM, have the same comments as in https://github.com/pinterest/gestalt/pull/2154

PR #2154, opened 2022-06-14
Docs: Text design guidelines

ponori:
Added design guidelines to the Text component

### DIFF
--- a/docs/pages/text.js
+++ b/docs/pages/text.js
@@ -40,7 +40,7 @@ export default function TextPage({ generatedDocGen }: {| generatedDocGen: DocGen
             type="do"
             description="Emphasize text inside of paragraphs by using a bold weight."
             defaultCode={`
-
+<Text>For campaign optimization and delivery, <Text weight="bold" inline>set optimization and delivery at the campaign level</Text> so all ad groups have the same values.</Text>
 `}
           />
           <MainSection.Card
@@ -48,7 +48,7 @@ export default function TextPage({ generatedDocGen }: {| generatedDocGen: DocGen
             type="don't"
             description="Emphasize text inside of paragraphs by underlining it; this can be confused with a [link](/link)."
             defaultCode={`
-
+ <Text>For campaign optimization and delivery, <Text underline inline>set optimization and delivery at the campaign level</Text> so all ad groups have the same values.</Text>
 `}
           />
         </MainSection.Subsection>
@@ -58,7 +58,10 @@ export default function TextPage({ generatedDocGen }: {| generatedDocGen: DocGen
             type="do"
             description="Use size to emphasize things like numbers that don’t define a page structure."
             defaultCode={`
-
+<Flex direction="column">
+  <Text size={200}>Impressions</Text>
+  <Text size={500} weight="bold">1.25M</Text>
+</Flex>
 `}
           />
           <MainSection.Card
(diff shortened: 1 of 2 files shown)


AlbertCarreras:
@ponori Are we splitting this upgrade in 2: design guidelines + engineering part? As they are, I would expect another round of changes to get them finished. See prop table for instance.

docs/pages/text.js
```diff
@@ -10,23 +10,164 @@ import MainSection from '../components/MainSection.js';
...
+            type="do"
+            description="Emphasize text inside of paragraphs by using a bold weight."
+            defaultCode={`
+<Text>For campaign optimization and delivery, <Text weight="bold" inline>set optimization and delivery at the campaign level</Text> so all ad groups have the same values.</Text>
+`}
+          />
+          <MainSection.Card
+            cardSize="md"
+            type="don't"
+            description="Emphasize text inside of paragraphs by underlining it; this can be confused with a [link](/link)."
```
AlbertCarreras, line 49: with Link instead of with a link

docs/pages/text.js
```diff
@@ -10,23 +10,164 @@ import MainSection from '../components/MainSection.js';
...
+          title="Accessible color"
+          description="For low-vision users, text color contrast is very important. To insure accessible contrast, stick to our [standard text colors](/color_usage#Standard-text-colors). See our [accessibility](/accessibility) page for design considerations and handy accessibility tools for checking color contrast."
+        />
+      </MainSection>
+      <MainSection
+        name="Localization"
+        description="Keep text simple and short to avoid truncation or line wrapping in UI controls like buttons when translating languages that require more characters. Avoid overriding our line-height settings, as this can result in text clipping for scripts, like Hindi, that have taller ascenders and descenders."
+      >
+        <MainSection.Subsection
+          title="Text-wrapping and hyphenation"
```
AlbertCarreras, line 149: We don't mention iOS in any other cmp page. Should we remove this?

docs/pages/text.js
```diff
@@ -10,23 +10,164 @@ import MainSection from '../components/MainSection.js';
...
+          description="Hyphenation on iOS is turned off by default to avoid incorrect word breaks when strings of text wrap to the next line. This is especially helpful for international languages where an incorrect word break can greatly change the meaning of a word or sentence."
+        />
+      </MainSection>
       <MainSection name="Variants">
         <MainSection.Subsection
           title="Alignment"
           description="Use this to adjust the positioning of text within wrapper elements."
```
AlbertCarreras, line 156: I'll add a comment here that to see forceRight/forceLeft we should toggle the RTL switch in Settings. I'd be great to add an example of when we want that, which I think it was related to numbers in tables????

docs/pages/text.js
```diff
@@ -262,6 +405,54 @@ export default function TextPage({ generatedDocGen }: {| generatedDocGen: DocGen
...
+          <MainSection.Subsection
+            description={`
+      **[Design tokens](/design_tokens)**
+      Values for text sizes, weights, families and colors.
+    `}
+          />
+          <MainSection.Subsection
+            description={`
+      **[Link](/link)**
+      Used as a text-only navigational element. Links usually appear within or directly following a paragraph or sentence. .
```
AlbertCarreras, line 452: . . !

AlbertCarreras (approved):
LGTM

docs/pages/text.js
ponori: Honestly, I have no context for this addition. I think this is definitely something we should evaluate in a future pass.

docs/pages/text.js
ponori: Good question. @hectoid what's your take on guidance for forceRight and forceLeft? I'm uncertain what the use-cases would be for this. 

Alberto brings up a great case re: numbers in tables, but besides that, I'm uncertain. One other thought could be when we display LTR words in RTL experiences...

docs/pages/text.js
ponori: Good catch - fixed!

PR #2155, opened 2022-06-14
Docs: Heading improvements

ponori:
A few small tweaks to the Heading docs to clean up visuals and add some clarifications.

### DIFF
--- a/docs/pages/heading.js
+++ b/docs/pages/heading.js
@@ -381,7 +381,7 @@ export default function DocsPage({ generatedDocGen }: {| generatedDocGen: DocGen
         </MainSection.Subsection>
         <MainSection.Subsection
           title="Alignment"
-          description={`Use \`align\` to adjust the positioning of text within wrapper elements`}
+          description={`Use \`align\` to adjust the positioning of text within container elements`}
         >
           <MainSection.Card
             cardSize="lg"
--- a/docs/pages/heading.js
+++ b/docs/pages/heading.js
@@ -381,7 +381,7 @@ export default function DocsPage({ generatedDocGen }: {| generatedDocGen: DocGen
         </MainSection.Subsection>
         <MainSection.Subsection
           title="Alignment"
-          description="Use `align` to adjust the positioning of text within wrapper elements."
+          description={`Use \`align\` to adjust the positioning of text within wrapper elements`}
         >
           <MainSection.Card
             cardSize="lg"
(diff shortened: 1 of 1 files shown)


docs/pages/heading.js
```diff
@@ -377,7 +379,10 @@ export default function DocsPage({ generatedDocGen }: {| generatedDocGen: DocGen
 `}
           />
         </MainSection.Subsection>
-        <MainSection.Subsection title="Alignment">
+        <MainSection.Subsection
+          title="Alignment"
+          description="Use `align` to adjust the positioning of text within wrapper elements."
```
dangerismycat, line 384: s/"wrapper"/"container"

docs/pages/heading.js
```diff
@@ -399,9 +404,10 @@ export default function DocsPage({ generatedDocGen }: {| generatedDocGen: DocGen
         </MainSection.Subsection>
         <MainSection.Subsection
           title="Accessibility level"
-          description="For accessibility purposes, we allow you to override the heading level.
+          description={`
+For accessibility purposes, we allow you to override the heading level.
 
-        We should have one level 1 per page &amp; levels should be appropriately nested. E.g. level 1 followed by level 2 &amp; level 2 followed by level 2 or level 3. We also allow headings without an accessibility level."
+We should have one level 1 per page &amp; levels should be appropriately nested. E.g. level 1 followed by level 2 &amp; level 2 followed by level 2 or level 3. We also allow headings without an accessibility level. We recommend against using \`accessibilityLevel="none"\`.`}
```
dangerismycat, line 410: These two sentences read a bit awkward together. Consider combining to something like "For certain specific situations, it is possible to use Heading with an accessibility level; however, we recommend against using this if possible."

dangerismycat (approved):

docs/pages/heading.js
ponori: Done!

docs/pages/heading.js
ponori: Not sure I follow what this means.

dangerismycat: Sorry, [`sed` syntax](https://en.wikipedia.org/wiki/Sed#Substitution_command): I'm suggesting that we replace "wrapper" with "container", which better describes the parent element.

ponori: Oh, nice! TIL! Thanks and fixed!

PR #2161, opened 2022-06-22
Card: fix `active` logic to support `false`

dangerismycat:
Card's `active` prop is intended to force the active state of the component. However, it currently can only be used to force the active state "on"; because of the way the logic is rewritten, attempting to force "off" falls through to continue using the hover state.

This PR fixes that logic to always respect the `active` value if passed.

### DIFF
--- a/packages/gestalt/src/Card.js
+++ b/packages/gestalt/src/Card.js
@@ -4,6 +4,10 @@ import classnames from 'classnames';
 import Box from './Box.js';
 import styles from './Card.css';
 
+function isNil(val) {
+  return val === undefined || val === null;
+}
+
 type Props = {|
   /**
    * Used to force the "active" hover state visually.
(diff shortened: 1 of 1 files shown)


packages/gestalt/src/Card.js
```diff
@@ -44,8 +48,7 @@ export default function Card({ active, children, image, onMouseEnter, onMouseLea
   };
 
   const classes = classnames(styles.card, {
-    // JS equality rules == null checks for `null` or `undefined` and leaves out `false`.
-    [styles.hover]: active || (active == null && hovered),
+    [styles.hover]: isNil(active) ? hovered : active,
```
dangerismycat, line 51: Boolean logic can be confusing, especially when boolean comparators are used with boolean prop values. This simplifies things: if `active` has a value,  use that. Otherwise we'll use the hover state.

PR #2163, opened 2022-06-23
Docs: Add new footers

ayeshakmaz:
#### What changed?

Add Resources footer for the home page, and update main footer

#### Why?

As part of our site redesign, update the resources and footers available on the site to make way-finding more prominent

Screen Shot 2022-06-23 at 11 48 39 AM

### Links

- [Jira](https://jira.pinadmin.com/browse/GESTALT-4387)
- [Figma](https://www.figma.com/file/emykswARRS1tQwpB7ThIPN/Gestalt-2022-Site?node-id=879%3A17966)

### Checklist


docs/components/App.js
```diff
@@ -28,6 +29,7 @@ export default function App({ children }: Props): Node {
       event.preventDefault();
       router.push(href);
     };
+    setIsHomePage(window.location.pathname === '/home');
```
ayeshakmaz, line 32: @dangerismycat What's the best way to do this? Window comes back as undefined in a flaky way, even if changing to window?.location?.pathname

docs/components/App.js
```diff
@@ -41,7 +42,9 @@ export default function App({ children }: Props): Node {
       window.gtag('config', 'UA-12967896-44', {
         page_path: window.location.pathname + window.location.search + window.location.hash,
       });
+      setIsHomePage(window?.location?.pathname === '/home');
```
dangerismycat, line 45: If we're only showing the resources footer on the home page, does it make sense to move it into that page specifically? Rather than include it in AppLayout?

docs/components/Footer.js
```diff
@@ -4,40 +4,80 @@ import { Box, Flex, Icon, Link, Text } from 'gestalt';
...
-            <Text color="default" weight="bold">
+            <Text size="100">
               <Link
                 accessibilityLabel="Visit Pinterest.com"
                 href="https://www.pinterest.com"
-                onClick={() => trackButtonClick('Pinterest logo')}
+                onClick={() => trackButtonClick('Pinterest copyright')}
               >
-                Pinterest
+                &copy; {new Date().getFullYear()} Pinterest
```
dangerismycat, line 37: ahh thank you so much for doing this the smart way 💯 😄 

(I wish all websites did)

docs/components/Footer.js
```diff
@@ -4,40 +4,80 @@ import { Box, Flex, Icon, Link, Text } from 'gestalt';
 import trackButtonClick from './buttons/trackButtonClick.js';
 
 export default function Footer(): Node {
+  const links = [
```
dangerismycat, line 7: nit: static consts like this can be defined outside the component, simplifying and making the component more readable (and a perf micro-optimization, but that doesn't matter)

docs/components/Footer.js
```diff
@@ -4,40 +4,80 @@ import { Box, Flex, Icon, Link, Text } from 'gestalt';
...
+    { title: "What's new", url: 'https://gestalt.pinterest.systems/home' },
+    { title: 'About us', url: 'https://gestalt.pinterest.systems/about_us' },
+    { title: 'Careers', url: 'https://www.pinterestcareers.com/', external: true },
```
dangerismycat, line 10: Consider pre-filling the search: https://www.pinterestcareers.com/job-search-results/?keyword=gestalt

docs/components/Footer.js
```diff
@@ -4,40 +4,80 @@ import { Box, Flex, Icon, Link, Text } from 'gestalt';
...
+    { title: 'Code sandbox', url: 'https://codesandbox.io/s/k5plvp9v8v', external: true },
+    { title: 'Github', url: 'https://github.com/pinterest/gestalt', external: true },
```
dangerismycat, line 12: super nit: they stylize their name as `GitHub`

docs/components/Footer.js
```diff
@@ -4,40 +4,80 @@ import { Box, Flex, Icon, Link, Text } from 'gestalt';
...
-    <Box color="lightGray" padding={4} mdPadding={6} lgPadding={8} role="contentinfo">
-      <Box paddingX={2} paddingY={1} display="flex" direction="column" mdDirection="row">
+    <Box
+      dangerouslySetInlineStyle={{
+        __style: {
+          backgroundColor: 'var(--color-orange-firetini-0)',
+        },
+      }}
+      padding={4}
+      role="contentinfo"
```
dangerismycat, line 23: 💯

docs/components/Footer.js
```diff
@@ -4,40 +4,80 @@ import { Box, Flex, Icon, Link, Text } from 'gestalt';
...
+          mdColumn={6}
+          padding={2}
+          display="inlineBlock"
+          mdDisplay="flex"
+          alignItems="center"
+          justifyContent="center"
+          wrap
+        >
+          {links.map((link, idx) => (
+            <Box marginBottom={2} mdMarginBottom={0} mdMarginEnd={4} key={idx}>
```
dangerismycat, line 53: Let's use either the link title or URL as the key to avoid using the index

docs/components/App.js
ayeshakmaz, line 45: @dangerismycat Is there a better way to do this?

docs/components/ResourcesFooter.js
```diff
@@ -0,0 +1,128 @@
...
+  heading: string,
+  isInternal?: boolean,
+  icon: 'figma' | 'slack' | 'eng',
+|};
+function LinkList({ items, heading, icon, isInternal = true }: LinkListProps): Node {
+  return (
+    <Flex alignItems="start" gap={2}>
+      <Box width="32px" display="flex">
+        {
+          {
```
dangerismycat, line 22: nit: defining an object literal inline like this isn't the most readable — consider defining outside the function body as a map

dangerismycat (approved):
It looks like the vertical spacing of the list items gets a bit wonky at smaller screen sizes:
Screen Shot 2022-06-23 at 12 39 07 PM
Screen Shot 2022-06-23 at 12 39 18 PM
Screen Shot 2022-06-23 at 12 39 55 PM
Screen Shot 2022-06-23 at 12 40 08 PM

This is a shortcoming of Flex, but can be fixed by adding top/bottom margins to the list items

docs/components/App.js
ayeshakmaz: Because fo the way our pages are styled, it involves a lot of extra styling to get the footer to display properly if I put it within the Home page content (it won't be full width or flush onto the other footer), so it felt like this was a better fit for how we approach laying out all these pages

docs/components/Footer.js
```diff
@@ -4,40 +4,80 @@ import { Box, Flex, Icon, Link, Text } from 'gestalt';
 import trackButtonClick from './buttons/trackButtonClick.js';
 
 export default function Footer(): Node {
+  const links = [
+    { title: "What's new", url: 'https://gestalt.pinterest.systems/home' },
```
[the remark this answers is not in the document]
ayeshakmaz: ahh good catch

docs/components/Footer.js
ayeshakmaz: hah same!

docs/components/App.js
```diff
@@ -12,6 +12,7 @@ type Props = {|
 
 export default function App({ children }: Props): Node {
   const router = useRouter();
+  const [isHomePage, setIsHomePage] = useState(false);
```
rlingineni: since `useRouter` is a hook, and it might cause a re-render on change,  maybe you could just directly use it as a direct variable, instead of setting `isHomePage` as state variable

```jsx
 const router = useRouter();
 const isHomePage = router.pathName === 'home';
```

rlingineni (commented):
Looks good

docs/components/App.js
dangerismycat: I'm not sure. This works for now at least

PR #2171, opened 2022-06-28
Font weight: change font weight from bold to semibold

AlbertCarreras:
#### What changed?

Follow up to https://github.com/pinterest/gestalt/pull/2047 to remove experiment and implement 600 font bold weight.

#### Why?

https://helium.pinadmin.com/mweb_gestalt_semibold_weight/
https://helium.pinadmin.com/web_gestalt_semibold_weight/

Neither experiment showed statistically significant  (positive/negative) trends.

Needs follow-up PR to update references to font weight in Docs. https://gestalt.pinterest.systems/typography#Font-weight
Screen Shot 2022-06-28 at 2 53 04 PM


packages/gestalt-design-tokens/tokens/font/base.json
```diff
@@ -27,10 +27,6 @@
       },
       "semibold": {
```
ayeshakmaz, line 28: Can we just call this "bold" since we no longer have another option? It keeps it backward compatible, esp with the tokens. So now the bold token just points to 600 instead of 700

ayeshakmaz (changes requested):

packages/gestalt-design-tokens/tokens/font/base.json
AlbertCarreras: @ayeshakmaz That was my first intention.

The Typography  page mentions semibold so I thought we wanted to keep that difference. Still think I should move it to bold?
 


https://gestalt.pinterest.systems/typography#Font-weight
Screen Shot 2022-06-28 at 7 47 32 PM

ayeshakmaz: Oh I think in iOS the phrasing they use is semibold, same way in android they use "medium"

ayeshakmaz: So it's not because of us, but the platform

dangerismycat: [MDN defines "bold" as 700](https://developer.mozilla.org/en-US/docs/Web/CSS/font-weight#values). Not sure if we want to break with that convention? (I could very easily be talked into it though)

ponori: Semibold is the appropriate term to use here. We're moving away from bold and to semibold.

ayeshakmaz: ahh ok, we need to label this as a breaking change then because we're removing a design token

ayeshakmaz: Will the prop on Text also be changing?

somerandomdude: It may be worth chatting with Hector before we remove the bold weight token. I think there's consideration of exploring Text using bold down the road.

AlbertCarreras: @ponori @ayeshakmaz I'll keep the token, we can deprecate down the road if there's more discussion needed around that

ayeshakmaz (approved):

PR #2175, opened 2022-06-29
Masonry: remove boolean flexible prop (Breaking change)

christianvuerings:
#### What changed?

Removes the `flexible` boolean prop from Masonry.

#### Why?
Consistency: `layout="flexible"` and `flexible` props do the same thing, so it's confusing what prop to use when

### FAQ

<details>
  <summary>Why not use <code>packages/gestalt-codemods/generic-codemods/modifyPropValue.js</code> ?</summary>
  We need to throw when the `flexible` prop is set on a layout other than the default, `flexible` or `serverRenderedFlexible`.
</details>

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -111,6 +111,12 @@
       "globals": {
         "page": true,
         "browser": true
+      },
+      "rules": {
+        "jest/expect-expect": [
+          "error",
+          { "assertFunctionNames": ["expect", "runInlineTest"] }
+        ]
       }
     },
     {
--- a/docs/pages/masonry.js
+++ b/docs/pages/masonry.js
@@ -1,5 +1,5 @@
 // @flow strict
-import { Component, type Node } from 'react';
+import { Component, type Node, type ElementProps } from 'react';
 import { Box, Masonry, Image, Label, Text } from 'gestalt';
 import GeneratedPropTable from '../components/GeneratedPropTable.js';
 import PageHeader from '../components/PageHeader.js';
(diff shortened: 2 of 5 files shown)


christianvuerings:
@dangerismycat Could you r+ https://phabricator.pinadmin.com/D906454? It converts `flexible` to `layout="flexible"` in webapp.

That way
* we can land the change in pinboard before landing the Gestalt change
* the gestalt upgrade diff will be cleaner

```bash
yarn codemod --parser=flow -t=packages/gestalt-codemods/59.0.0/masonry-update-flexible-prop.js ~/code/pinboard/webapp/
```

dangerismycat:
Do we want to remove the boolean prop or the layout option? Would something like `<Masonry flexible layout="uniformRow" />` be something we'd want to support?

christianvuerings:
> Do we want to remove the boolean prop or the layout option? Would something like `<Masonry flexible layout="uniformRow" />` be something we'd want to support?

We want to remove the boolean prop in this case. For more context also see https://github.com/pinterest/gestalt/pull/782:
> One other future goal: I'd like to remove the flexible bool prop. When true, Masonry currently uses a completely different layout engine. For instance, if the user asks for uniform row layout and specifies flexible={true}, the uniform layout flag will be ignored. Once this lands, I intend to convert all instances of flexible={true} to layout="flexible".

christianvuerings:
@dangerismycat fyi the related pinboard diff just landed so this should be good to merge / r+
https://phabricator.pinadmin.com/D906454

dangerismycat (approved):
Sounds good, thanks for the context Christian!

PR #2183, opened 2022-06-30
Table: update sortable header cell behavior

dangerismycat:
Current behavior isn't great when de-selecting a sortable header, with the arrow icon lingering due to the header cell having focus after being clicked:
https://user-images.githubusercontent.com/12059539/176942403-ead225fa-e627-4aad-81d3-5c8f5a539792.mov

This PR doesn't adjust the actual focus of the cell, but removes the focus UI state just after click.

### DIFF
--- a/docs/components/COMPONENT_DATA.js
+++ b/docs/components/COMPONENT_DATA.js
@@ -972,11 +972,11 @@ const GENERAL_COMPONENT_LIST: ListItemType = [
     category: 'Data',
     status: {
       accessible: {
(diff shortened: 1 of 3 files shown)


ayeshakmaz:
Can we update the A11Y scorecard to be partially accessible (similar to Popover) since the sorting piece is not accessible? Thanks!

ayeshakmaz (approved):

rlingineni (approved):

PR #2184, opened 2022-07-01
Masonry: remove deprecated layouts

christianvuerings:
#### What changed?

1 way to specify masonry layouts: string enums

#### Why?

Masonry has 3 different ways to specify layouts:

1. As a string enum: `'basic' | 'basicCentered' | 'flexible' | 'serverRenderedFlexible' | 'uniformRow'`
2. As a class: `LegacyMasonryLayout | LegacyUniformRowLayout`
3. As a symbol: `typeof DefaultLayoutSymbol | typeof UniformRowLayoutSymbol`

These 3 different ways are confusing and we should stick to 1

#### Context

In https://github.com/pinterest/gestalt/pull/782 we introduced the option to specify masonry layouts as strings but never completed the conversion:

> Because extra exports are annoying and we already split on what to use, I'd like to update to one true system: strings.

### Related diffs

- https://phabricator.pinadmin.com/D906449 - Replaces all deprecated layouts in pinboard/webapp
- #2175 - Replaces `flexible` prop with `layout="flexible"`

### DIFF
--- a/packages/gestalt/src/index.js
+++ b/packages/gestalt/src/index.js
@@ -34,8 +34,6 @@ import Letterbox from './Letterbox.js';
 import Link from './Link.js';
 import Mask from './Mask.js';
 import Masonry from './Masonry.js';
-import MasonryDefaultLayout from './layouts/MasonryLayout.js';
-import MasonryUniformRowLayout from './layouts/UniformRowLayout.js';
 import Modal from './Modal.js';
 import Module from './Module.js';
 import NumberField from './NumberField.js';
@@ -109,8 +107,6 @@ export {
   Link,
   Mask,
   Masonry,
-  MasonryDefaultLayout,
-  MasonryUniformRowLayout,
   Modal,
   Module,
   NumberField,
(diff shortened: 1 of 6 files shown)


packages/gestalt/src/index.js
```diff
@@ -34,8 +34,6 @@ import Letterbox from './Letterbox.js';
 import Link from './Link.js';
 import Mask from './Mask.js';
 import Masonry from './Masonry.js';
-import MasonryDefaultLayout from './layouts/MasonryLayout.js';
```
christianvuerings, line 37: Please see https://phabricator.pinadmin.com/D906449 where we remove these imports from pinboard/webapp

dangerismycat:
This seems fine to me but @liuyenwei I'd be interested in your review as well

christianvuerings:
@dangerismycat fyi the related pinboard diff just landed so this should be good to merge / r+ /cc @liuyenwei 
https://phabricator.pinadmin.com/D906449

dangerismycat (approved):

PR #2197, opened 2022-07-12
Masonry: Fix internal only flow issues

christianvuerings:
Fixes 3 flow fixme's and 2 ESLint disables related to `Position` in `Masonry`. These changes are internal only


AlbertCarreras (approved):
Thanks!

PR #2203, opened 2022-07-15
SideNavigation: small fixes

AlbertCarreras:
#### What changed?

SideNavigation: small fixes

### DIFF
--- a/packages/gestalt/src/SideNavigationGroup.js
+++ b/packages/gestalt/src/SideNavigationGroup.js
@@ -199,6 +199,8 @@ export default function SideNavigationGroup({
       <NestingProvider>
         {display === 'expandable' ? (
           <TapArea
+            accessibilityControls={display === 'expandable' ? itemId : undefined}
+            accessibilityExpanded={display === 'expandable' ? expanded : undefined}
             onMouseEnter={() => setHovered(true)}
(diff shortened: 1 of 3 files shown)


packages/gestalt/src/SideNavigationGroup.js
```diff
@@ -241,9 +243,11 @@ export default function SideNavigationGroup({
...
           />
         )}
-        <ul className={classnames(styles.ulItem)}>
-          {expanded || display === 'static' ? navigationChildren : null}
-        </ul>
+        {expanded || display === 'static' ? (
+          <ul id={itemId} className={classnames(styles.ulItem)}>
+            {navigationChildren}
+          </ul>
+        ) : null}
```
ayeshakmaz, line 250: good catch

ayeshakmaz (approved):

packages/gestalt/src/__snapshots__/SideNavigation.test.js.snap
```diff
@@ -912,7 +912,9 @@ exports[`SideNavigation renders nested directory 1`] = `
                 className="liItem"
               >
                 <div
+                  aria-controls=":r8:"
```
dangerismycat, line 915: What is this? Can we leave a comment in the code explaining its purpose?

PR #2207, opened 2022-07-17
Internal: remove unused greenkeeper references

christianvuerings:
We no longer use Greenkeeper - instead we use dependabot which is now part of GitHub: https://github.com/dependabot

From a cached [greenkeeper.io](https://greenkeeper.io/) version: 

> We’ll stop Greenkeeper’s operation as an independent service on June 3rd, 2020. For your dependency update needs, we have a brilliant alternative for you: [Snyk](https://snyk.io/). The team behind Greenkeeper have spent the past months working with them on next-generation dependency updates as a part of their open-source security product. We invite all of our users to migrate over: Snyk not only has a generous free plan for Open Source repos and small organisations, but also offers dev-friendly security features that Greenkeeper didn’t have. They’re good people, and your repos will be in competent and caring hands.


AlbertCarreras (approved):
Thanks!

PR #2210, opened 2022-07-20
Docs: Iconography guidelines

ponori:
Added Iconography guidelines to the docs


docs/pages/iconography.js
```diff
@@ -0,0 +1,402 @@
...
+                  nationality or language.
+                </Text>
+              }
+            />
+          </Flex.Item>
+        </Flex>
+      </MainSection>
+
+      <MainSection
+        name="Typography components"
```
ayeshakmaz, line 106: Should this be "Iconography components" or "Components with Icons" maybe?

docs/pages/iconography.js
```diff
@@ -0,0 +1,402 @@
...
+
+- [IconButton](/icon): Uses icons instead of text to convey available actions on a screen.
+- [Status](/status): A graphic indicator of an element’s state.
+- [Link](/link): External links use the visit icon as an external domain indicator.
+`}
+      />
+
+      <MainSection name="Characteristics">
+        <MainSection.Subsection title="Appearance">
+          <Markdown
```
ayeshakmaz, line 118: I don't remember so apologies if we've discussed this, but can this content go in the "description" prop of Subsection instead? It should still be rendered in markdown

docs/pages/iconography.js
ponori: Thanks! Fixed!

docs/pages/iconography.js
ponori: I added it this way as the content is intermixed w/ an image so it felt strange to have one block in a `description` prop and then another as a `Markdown` component. Thoughts?

ayeshakmaz (approved):

PR #2218, opened 2022-07-23
Flex: modify `gap` to support row/column values

dangerismycat:
This PR modifies Flex's `gap` prop from `0..12` to `0..12 | {| row: 0..12, column: 0..12 |}`. This allows the user to specify spacing for the both the row and column directions, which is not currently possible.

#### Why?

When I first implemented Flex, I frankly misunderstood [the spec](https://developer.mozilla.org/en-US/docs/Web/CSS/gap) and implemented it incorrectly. I didn't realize that providing a single value for `gap` is actually shorthand for `row-gap` and `column-gap`, not just a single value. So that implementation only worked with a single row or single column of items. If the items wrapped, there was no way to specify the spacing between rows, leading to squished layouts and forcing the user to use Box instead (and implement the children spacing manually 😞).

The original implementation is fine when wrapping isn't an issue:
Screen Shot 2022-07-22 at 4 54 36 PM

…but is no bueno when the items wrap:
Screen Shot 2022-07-22 at 4 54 47 PM

The new implementation allows for the vertical spacing to let this layout breathe:
Screen Shot 2022-07-22 at 4 54 57 PM

_A NOTE ABOUT THE CODEMOD:_
From now on, `gap={3}` will be equivalent to `gap={{ row: 3, column: 3 }}`. This matches [the spec](https://developer.mozilla.org/en-US/docs/Web/CSS/gap), so it's the best choice going forward. But that's not how the original implementation worked, so we actually to codemod something like `<Flex direction="row" gap={3}>` to `<Flex direction="row" gap={{ row: 3, column: 0 }}>`. This is ugly and unfortunate, but we can't assume that adding the missing spacing won't break layouts. In the future, most Flex usages will _probably_ just use the shorthand, but for now we need to replicate the current behavior. Blame 2020 Ryan for that. 😬


packages/gestalt-codemods/_next/flex-gap-expansion.js
```diff
@@ -0,0 +1,124 @@
+/*
+ * Converts
+ *  <Flex gap={3} /> to <Flex gap={{ row: 3, column: 0 }} />
+ */
+
+// yarn codemod --parser=flow -t=packages/gestalt-codemods/33.0.0/heading-replace-truncate-lineClamp.js relative/path/to/your/code
```
dangerismycat, line 6: I'll update this when I'm preparing to land and have an expected version number

packages/gestalt/src/Flex.js
```diff
@@ -21,7 +22,7 @@ type Props = {|
    */
   alignSelf?: 'auto' | 'start' | 'end' | 'center' | 'baseline' | 'stretch',
   /**
-   *
+   * Note that each child will be automatically wrapped in [Flex.Item](https://gestalt.pinterest.systems/flex#Flex.Item) to apply various styles. If specific flex styles are needed on a child, you can manually wrap it in Flex.Item to apply those styles. See [the Applying flex properties to children example](https://gestalt.pinterest.systems/flex#Applying-flex-properties-to-children) to learn more.
```
ayeshakmaz, line 25: The URLs here are now out of date 😅

packages/gestalt-codemods/_next/__testfixtures__/flex-gap-expansion.input.js
```diff
@@ -0,0 +1,48 @@
...
+        <Text>Item 2</Text>
+        <Text>Item 3</Text>
+      </Flex>
+
+      <Flex gap={condition ? gapValue : alternateGapValue}>
+        <Text>Item 1</Text>
+        <Text>Item 2</Text>
+        <Text>Item 3</Text>
+      </Flex>
+    </Box>
```
ayeshakmaz, line 46: To be extra careful, is it worth repeating these with direction="column"?

ayeshakmaz (approved):
So excited for this update

packages/gestalt-codemods/_next/__testfixtures__/flex-gap-expansion.input.js
dangerismycat: It's not really necessary…if you check out the codemod logic, the parts that determine what the new value is and where that new value goes are pretty independent. So it should be enough to test "does the new value go where expected given the `direction` value?" and "does the logic handle ternaries properly?" without needing to explicitly test every possible permutation.

packages/gestalt/src/Flex.js
dangerismycat: Good catch, updated all internal links in #2264

christianvuerings:
@dangerismycat curious, did we also consider using `rowGap` and `columnGap` as props instead of passing in an object to `gap`? Seems like it would be more consistent with what we do with e.g. `paddingX` and `marginTop`

dangerismycat:
> @dangerismycat curious, did we also consider using `rowGap` and `columnGap` as props instead of passing in an object to `gap`? Seems like it would be more consistent with what we do with e.g. `paddingX` and `marginTop`

Not really — that would allow for the weird state of `<Flex gap={2} rowGap={3} columnGap={4}>`, which isn't great. I'm not a fan of the padding/margin props either but also not eager to go through the pain of refactoring them 😅  For `gap` at least, I wanted to guide users to a [Pit of Success](https://blog.codinghorror.com/falling-into-the-pit-of-success/) rather than allow nonsensical states.

PR #2219, opened 2022-07-23
SelectList: Add support for `<optgroup>`, replace `items` prop with subcomponents

dangerismycat:
A recent blogpost mentioned [`<optgroup>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/optgroup), which I'd never heard of before. [This element allows for grouping/categorization within `<select>` items](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/select#advanced_select_with_multiple_features). This is a great fit for SelectList, as it could allow more users to use this simpler component rather than needing to use Dropdown. This PR adds support for `<optgroup>`…

…and also completely overhauls how users specify items in the list. The current `items` prop wouldn't scale well to support groups, leading to a tedious and confusing nested structure. Even without the introduction of groups, the array of objects isn't very readable, particularly if the user needs to reorder the items. Therefore, this PR creates SelectList.Option and SelectList.Group subcomponents. Users can compose these subcomponents as needed to create their list of items in a much more HTML-like format. And of course, an array of objects can be mapped to SelectList.Option when a more JS-esque approach is desired.

_NON-GOALS OF THIS PR_
- Addressing the Flow types of SelectList's `children` to ensure that only the subcomponents are used. This is the same problem we have with Dropdown and SideNav, and really everywhere we use subcomponents. I'll work on a solution for all these issues in the future.
- Using a more realistic example in the new Groups example. We can update this in the future. Anyone with a good idea for that example is welcome to fast-follow this PR with their fixes.

_Codemod_
```bash
yarn codemod --parser=flow -t=packages/gestalt-codemods/71.0.0/select-list-options-to-children.js relative/path/to/your/code
```


docs/pages/selectlist.js
```diff
@@ -1,39 +1,69 @@
...
-    {label:'Denmark', value: 'denmark'},
-    {label:'Egypt', value: 'egypt'},
-    {label:'France', value: 'france'},
-  ]}
-  size='lg'
-  label='Country'
-/>
+  size="lg"
+>
+  {[
```
dangerismycat, line 28: In the future we can update some/all of these examples to use the more HTML-ish approach, if desired. This is the quick+easy version for now

_HTML-ish approach_
```jsx
<SelectList>
  <SelectList.Option label="Algeria" value="algeria" />
  <SelectList.Option label="Belgium" value="belgium" />
  ...
</SelectList>
```

packages/gestalt-codemods/_next/select-list-options-to-children.js
```diff
@@ -0,0 +1,165 @@
+/*
+ * Converts
+ *  <SelectList options={foo} /> to <SelectList>{foo.map((item) => <SelectList.Option {...item} />)}</SelectList>
+ */
+
+// yarn codemod --parser=flow -t=packages/gestalt-codemods/_next/select-list-options-to-children.js relative/path/to/your/code
```
dangerismycat, line 6: I'll update this before merging when I have an expected version number

packages/gestalt-codemods/_next/select-list-options-to-children.js
```diff
@@ -0,0 +1,165 @@
...
+        // Reassign the new attributes, which no longer contain `options`
+        node.openingElement.attributes = newAttrs;
+        // We're adding children, so this is no longer self-closing and we need a closing element
+        node.openingElement.selfClosing = false;
+        node.closingElement = j.jsxClosingElement(j.jsxIdentifier(jsxElementName));
+
+        const mapTarget = options ? j.arrayExpression(options) : j.identifier(optionsVariableName);
+
+        // Add the existing options array as mapped children
+        node.children = [
```
dangerismycat, line 109: hoo boy 😅

packages/gestalt/src/SelectList.js
```diff
@@ -91,7 +92,7 @@ export default function SelectList({
     }
   };
 
-  const handleBlur = (event: SyntheticInputEvent<HTMLSelectElement>) => {
+  const handleBlur: (event: SyntheticInputEvent<HTMLSelectElement>) => void = (event) => {
```
dangerismycat, line 95: No real change, just matching the type format with the other handler

docs/pages/web/selectlist.js
```diff
@@ -194,6 +253,7 @@ export default function SelectListPage({ generatedDocGen }: {| generatedDocGen:
     };
 
     return (
+<<<<<<< HEAD:docs/pages/web/selectlist.js
```
AlbertCarreras, line 256: Need to clean this @dangerismycat

packages/gestalt-codemods/_next/select-list-options-to-children.js
AlbertCarreras: wow, thanks for writing this. Good reference for the future.  I would have just thrown an error on detection ha

AlbertCarreras (approved):
LGTM except for the merging conflict in the Docs! Thanks for adding this.

rlingineni:
This looks good. I'm still not sure when to use this vs dropdown groups. I think this [guidance](https://deploy-preview-2219--gestalt.netlify.app/web/selectlist#Usage-guidelines) might have to be updated? 

> The current items prop wouldn't scale well to support groups, leading to a tedious and confusing nested structure. 

Also, just wondering, is this an example of something that is considered a breaking change because all the select lists in Pinboard have to be updated?

There do seems to be some werid HEAD tags in places causing some of the sandboxes to break:

image

docs/pages/web/selectlist.js
dangerismycat: Ah thanks, good catch! More encouragement to keep converting pages to the new code example style 😬

dangerismycat:
> This looks good. I'm still not sure when to use this vs dropdown groups. I think this [guidance](https://deploy-preview-2219--gestalt.netlify.app/web/selectlist#Usage-guidelines) might have to be updated?

Yep, definitely.

> Also, just wondering, is this an example of something that is considered a breaking change because all the select lists in Pinboard have to be updated?

Absolutely, hence the `major` version bump and the codemod.

> There do seems to be some werid HEAD tags in places causing some of the sandboxes to break:

Those are from merge conflicts — but not anything an IDE would recognize as such because the code is "just a string". This is the biggest reason I'm excited about the new code editor! A functional UI is nice, but using real code for all these examples will catch so many errors like this.

packages/gestalt-codemods/_next/select-list-options-to-children.js
dangerismycat: I finally learned to read the "documentation" (haha) to figure this stuff out. Let me know the next time you're writing something like this and I'm happy to pair with you to teach you. It's pretty simple once you figure it out, just kinda tedious.

PR #2220, opened 2022-07-25
Table.Header: add 'display' prop

christianvuerings:
#### What changed?

* Add `display` prop to `Table.Header`
* Fix hydration issues on every page which used the `QualityChecklist` component
* Fix `Warning: Expected server HTML to contain a matching <table> in <div>.`

#### Why?

The `QualityChecklist` component renders a `<div />` as a direct descendant of `<thead />` which is not allowed:
Screen Shot 2022-07-25 at 1 25 57 PM

It results in the follow hydration error:
Screen Shot 2022-07-25 at 7 16 47 AM

#### Repro steps

* Clone the repro
* Go to `/modal` (or any other page with a `QualityChecklist` component
* Reload the page

#### Notes

* Bug was introduced in #2114 /cc @AlbertCarreras 
* Instead we could wrap `<Text />` in `<Box display="visuallyHidden" />` but that results in unnecessary space and an extra border

### DIFF
--- a/docs/components/QualityChecklist.js
+++ b/docs/components/QualityChecklist.js
@@ -1,6 +1,6 @@
 // @flow strict
 import { type Node } from 'react';
-import { Box, Text, Table, SlimBanner } from 'gestalt';
+import { Text, Table, SlimBanner } from 'gestalt';
 import COMPONENT_DATA from './COMPONENT_DATA.js';
 import MainSection from './MainSection.js';
 import StatusData from './StatusData.js';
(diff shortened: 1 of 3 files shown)


docs/components/QualityChecklist.js
```diff
@@ -32,17 +32,15 @@ export default function QualityChecklist({ component }: Props): Node {
             <col style={{ width: '20%' }} />
             <col style={{ width: '60%' }} />
           </colgroup>
-          <Box display="visuallyHidden">
```
christianvuerings, line 35: @AlbertCarreras this is the root of the issue - could you take a look at this diff?

AlbertCarreras: If this fixes Hydration issues... gonna make you a Saint in Spain, Saint Vuerings

AlbertCarreras (approved):

PR #2221, opened 2022-07-25
Docs: Introduce Sandpack for code examples

christianvuerings:
### What changed?

Convert code examples for `Modal` and `Sheet` to [Sandpack](https://sandpack.codesandbox.io/) ([repo](https://github.com/codesandbox/sandpack)).

> Sandpack is a component toolkit for creating live-running code editing experiences, using the power of CodeSandbox

### Why?

Benefits:
* **Example code is flow typed**
* **Example code is linted** (ESLint). We found quite a few violations of [no-unstable-nested-components](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/no-unstable-nested-components.md).
* **Example code uses prettier**. Avoids indentation issues we currently see with markdown
* **Example code is easy to codemod**. Since each example now lives in their own JS file, it's easy to codemod.
* **Built-in error overlay**. Uses CodeSandbox built-in error overlay which makes it easy to discover mistakes in our example code
* **No more hard to update external code** (we need a new fork each time). Issues found in current examples:
  * Legacy React `render` usage
  * Incorrect text sizes: `lg` instead of the new number sizes: `300`
  * Modal alignment: `align="left"` instead of `align="start"`
  * Button: `inline` boolean prop instead of `fullWidth`
* **No more broken cursor location**. This was an issue with the `react-live` editor
* No longer need `Edits made below will not be reflected in the example above, open the sandbox instead.`
* Fix broken `Do` examples which had `defaultCode` set but no `iframeContent` (e.g. first do example on modal page)
* Fix discrepancy between local code & CodeSandbox example. E.g. role example on modal page

### FAQ

> Do you think we should use `SandpackExample` for all of our code examples?

Yes, that way we have the same benefits for every code example & consistency across our docs.

> How long have you been dreaming about implementing this?

Since the beginning of 2022, that's when I first saw it being use on the [ReactJS beta docs](https://beta.reactjs.org/learn).

### Potential follow ups

* **Create/Use custom theme for code preview**
Right now we're using a built-in theme, we can create our own / use a different one if we want. 
* **Use `Monaco` editor instead of the CodeSandbox editor**
The [Monaco](https://microsoft.github.io/monaco-editor/) editor is what VSCode uses & would enable us to run e.g. ESLint in the browser / Go to definition on right click. ([example](https://codesandbox.io/s/vud7z?file=/src/Editor.tsx:666-742) / [docs](https://github.com/codesandbox/sandpack/issues/327))
* **Allow for more configurability of `SandpackExample`**. E.g. setting an example preview height
* **Use Gestalt master/current branch JS & CSS in dev mode**. We currently use the latest published Gestalt JS & CSS, this works well after a release but not as good during development - e.g. adding a new prop.


christianvuerings:
@dangerismycat @AlbertCarreras @ayeshakmaz @EduhCosta @ponori The PR is now ready for review.

Feel free to check out
* Benefits of using Sandpack (see [PR description](https://github.com/pinterest/gestalt/pull/2221#issue-1317056640))
* [Updated modal](https://deploy-preview-2221--gestalt.netlify.app/modal) & [updated sheet](https://deploy-preview-2221--gestalt.netlify.app/modal) code examples

AlbertCarreras (approved):
LGTM! I wonder if the idea is to migrate all components to this new system. I'd be great but a lot of work!

christianvuerings:
> LGTM! I wonder if the idea is to migrate all components to this new system. I'd be great but a lot of work!

@AlbertCarreras Yes, that would be the suggestion. That way we have consistency across all of our docs & have all the benefits everywhere (flow typed / ESLint / etc)

ayeshakmaz:
@christianvuerings Could you convert the PageHeader doc (or at least a few examples) by chance? Does it work well for components that need the full width of the page? Upsell is another example

christianvuerings:
> @christianvuerings Could you convert the PageHeader doc (or at least a few examples) by chance? Does it work well for components that need the full width of the page? Upsell is another example

@ayeshakmaz Definitely, I converted the `PageHeader` page to show that it works. Couple of notes:
* I had to slightly tweak the page container `width` to make it work.
* If you test it, make sure your window is maximized
* With `Upsell` & `PageHeader` I hope we eventually (once browser support is acceptable) switch to using [CSS container queries](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Container_Queries). That way, the components can be responsive to their container & not just the window width.

Screen Shot 2022-07-28 at 3 53 19 PM
Screen Shot 2022-07-28 at 3 53 27 PM
Screen Shot 2022-07-28 at 3 53 39 PM

And some more issues found:
* `key` prop not set on the example code
* Gestalt components with incorrect values
    * `Text` color set to `darkGray` instead of `default`
    * `Box` color set to `white` instead of `light`
* Invalid `disabled` prop on `BaseRow`
* `onSelect` set on `Dropdown.Link` instead of `href`
* `helperIconButton` missed `accessibilityControls` and `accessibilityExpanded`

Hoping to land this soon so we can avoid more merge conflicts & share these benefits across other doc pages

ayeshakmaz:
@christianvuerings Is there a way to configure this view to be top and bottom instead of left and right? I think this could potentially get frustrating
Screen Shot 2022-07-28 at 11 22 25 AM
, especially in a case like PageHeader, where the UI is very different depending on the width

christianvuerings:
> @christianvuerings Is there a way to configure this view to be top and bottom instead of left and right?

@ayeshakmaz Yes, that is possible but not something I would recommend:

* We loose the ability to see the changes side by side. That's also the reason why the row layout is the default for `Sandpack`.
* `PageHeader` / `CallOut` seem to be the odd components which rely on page width. Most other components do not
* If we use a column layout, we would have to scroll a lot to see the changes. That's an issue we have today.
  `Popover` example:
2022-07-28_vertical-layout-issue

In order to move the diff forward, I added a `layout` prop where the default is `row` but can be set to `column` for components like `PageHeader`.

Screen Shot 2022-07-28 at 10 44 08 PM

PR #2227, opened 2022-07-28
Docs: Move Development Page to Markdown

rlingineni:
#### What changed?

Removing the old development page with a new Markdown based one.

#### Why?

This is part of Shipping MDX based pages for the Gestalt Docs site

### Links

- [TDD](link to Paper doc)
- [Figma](link to Figma file)

### Checklist


docs/markdown/development.md
```diff
@@ -10,12 +10,11 @@ component: false
 - **Install node**
```
AlbertCarreras, line 10: missing links here

also can we remove bold style from links, and only show the underline?

docs/docs.css
```diff
@@ -70,9 +70,10 @@ li {
   margin-top: 24px;
 }
 
+/*
```
AlbertCarreras, line 73: should we remove this comment?

docs/components/DocsSideNavigation.js
```diff
@@ -56,9 +56,13 @@ export default function DocsSideNavigation({ border }: {| border?: boolean |}):
...
               {section.pages.map((componentName, i) => {
                 const href = `/${componentName.replace(/ /g, '_').replace(/'/g, '').toLowerCase()}`;
+
+                // in nextjs, if it's a dynamic route, the dynamic route id will be passed as part of the query obj
+                const { id: pathId } = router.query;
+
                 return (
                   <SideNavigation.TopItem
-                    active={router.pathname === href ? 'page' : undefined}
+                    active={router.pathname === href || `/${pathId}` === href ? 'page' : undefined}
```
AlbertCarreras, line 65: let's make sure this is still valid after Ayesha's navigation implementation.

docs/pages/development.js
```diff
@@ -1,500 +0,0 @@
-// @flow strict
```
christianvuerings, line 1: @rlingineni could we ensure that our `validateAccessibilitySpecs` check still works as expected? Basically we want to ensure every page on our docs site has an accessibility check. In the past we relied on those pages living in `docs/pages`
https://github.com/pinterest/gestalt/blob/master/scripts/validateAccessibilitySpecs.js

I'm also curious as to why we created a specific `markdown` directory, could/should we use `pages` instead? https://nextjs.org/docs/advanced-features/using-mdx

image

It might then also be possible to of the dynamic `docs/pages/[id].js` route and use `MDXProvider` instead.

docs/components/DocsSideNavigation.js
rlingineni: Good catch. I see some merge conflicts now, so I'll take a look!

docs/pages/development.js
rlingineni: Hi Christian!

Yeah sure! I assume it's `npm run a11y:validate`. I think they might not exist, so let me see what I can do to update this script and make the changes.

For the second thing:

The way pages is structured now in master is a bit different btw from when the PR was made.

Previously, all the files were just flat inside the directory, so it made spotting the `.md` files from the non-markdown files much harder. One of things we wanted is just finding the page we need and editing it quickly from Github.  

But with the new changes, I think it totally makes sense to put them in the same directory structure. And the previous approach could actually end up being more disorganized. I'll look into it and seeing if we can put it in the same directory

docs/markdown/development.md
rlingineni: Done!

docs/pages/development.js
rlingineni: Okay, so quick update on this. I think the `validateAccessibilitySpecs` check should be passing now. The markdown page is tied to the site index.

For the second part, we're actually using the markdown folder, and not an `.mdx` file alongside the rest of our pages. And this is because we're using`next-mdx-remote` instead of `@next/mdx` as our library. 

MDX Remote allows us to use a wrapper component and distinguish between the js pages and markdown ones. With true MDX, with the regular MDX library, we'd have to write code like:

```
export default ({ children }) => <MyLayoutComponent meta={meta}>{children}</MyLayoutComponent>
```

At the bottom of every MDX file to add an export with a wrapper. One of the goals was to minimize the overhead for writing markdown and going from Dropbox. And there was no generic way to do that.

docs/markdown/get_started/developers/development_process.md
```diff
@@ -166,14 +168,14 @@ When pushing new changes to GitHub, your PR title should be aligned with the sco
 
 Our versioning guidelines follow those outlined at **[semver.org](https://semver.org)**:
 
-- Patch: internal fixes, documentation changes, or package upgrades (anything that consumers of Gestalt don't need to worry about)
```
AlbertCarreras, line 169: Check the get_started/developers/development_process#Guidelines site, I updated the info re Codemods. Maybe try rebasing?

docs/docs-components/useGetSideNavItems.js
```diff
@@ -10,11 +10,18 @@ function convertNamesForURL(name) {
...
+  const { pathname, query } = useRouter();
   const { setIsSidebarOpen } = useNavigationContext();
 
   let nestingLevel = 0;
   const getNavItems = (navItem: siteIndexType, previousSectionName: string) => {
+    // in nextjs, if it's a dynamic route, the dynamic route id will be passed as part of the query obj
+    const { id: pathId } = query;
+    const urlPath = pathId ? pathId.join('/') : '';
+
+    const isActiveTab = (href: string) =>
```
AlbertCarreras, line 22: why do we need this ``/${urlPath}` === href`, in what cases wasnt working?

AlbertCarreras (approved):
LGTM, make the changes as per my comments and I can merge

docs/markdown/get_started/developers/development_process.md
rlingineni: Looks like the section is removed? I went ahead and tried to match them.

rlingineni:
@AlbertCarreras this looks all green, can I get a merge?

docs/markdown/get_started/developers/development_process.md
AlbertCarreras: All that info was removed because it moved to Releases section

PR #2236, opened 2022-07-31
Docs: Fix /_error renders on collage page

christianvuerings:
#### What changed?

Fix an issue where the collage page was making requests to `/undefined`

#### Why?

It results in multiple calls to `GestaltApp.getInitialProps` which render the `/_error` page

#### Root cause

#1662 /cc @dangerismycat
In that PR we fall back to `{}` for `image`, however that is truthy while before it was `undefined` which is falsy

```js
const image = images[index];
// Works as expected
if (image) {
```

```js
// Fails
const image = images[index] || {};
if (image) {
```


dangerismycat (approved):
Good catch! Thanks!

PR #2255, opened 2022-08-03
Docs: adjustments to Sandpack code containers

AlbertCarreras:
#### What changed?

Docs: adjustments to Sandpack code containers

Set default props for Sandpack containers

Variants and all examples except Best practices are open in SideNavigation, PageHeader, Sheet, and Modal

### DIFF
--- a/docs/docs-components/SandpackExample.js
+++ b/docs/docs-components/SandpackExample.js
@@ -14,6 +14,8 @@ import ShowHideEditorButton from './buttons/ShowHideEditorButton.js';
 import OpenInCodeSandboxButton from './buttons/OpenInCodeSandboxButton.js';
 import { useLocalFiles } from './contexts/LocalFilesProvider.js';
 
+const MIN_HEIGHT = 350;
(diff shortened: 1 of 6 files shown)


docs/docs-components/SandpackExample.js
```diff
@@ -57,9 +59,9 @@ function SandpackContainer({
             <SandpackCodeEditor
               wrapContent
               style={{
-                height,
+                height: height > MIN_HEIGHT ? height : MIN_HEIGHT,
```
dangerismycat, line 62: I believe these are just pass-through CSS properties, so this can probably use `minHeight` and not need this logic

dangerismycat (approved):

docs/docs-components/SandpackExample.js
AlbertCarreras: height can be specified for taller examples. I set some sidenavigation examples to 500. If parallel, we want the same height for the sandpack

PR #2259, opened 2022-08-04
Docs: convert multiple "foundation" page examples to Sandpack

christianvuerings:
#### What changed?

* Converts multiple "foundation" page examples to Sandpack
* Convert `hideControls` prop to `showControls` => makes it consistent with the `showEditor` prop on `SandpackExample`
* Remove t-shirt sizes `sm` and `md` from `SandpackExample`

#### Why?

For more context see #2221 

Flow issues found in this diff:

```
* Cannot create `Table` element because property `accessibilityLabel` is missing in  props [1] but exists in  `Props` [2].Flow(prop-missing)
* Cannot create `Badge` element because property `inline` is missing in  `Props` [1] but exists in  props [2].Flow(prop-missing)
* Cannot create `Image` element because  string [1] is incompatible with  number [2] in property `naturalWidth`.
* Cannot create `TapArea` element because: Either property `color` is missing in  `TapAreaType` [1] but exists in  props [2]. Or property `color` is missing in  `LinkTapAreaType` [3] but exists in  props [2].Flow(incompatible-type)
```

### DIFF
--- a/docs/docs-components/SandpackExample.js
+++ b/docs/docs-components/SandpackExample.js
@@ -14,8 +14,8 @@ import ShowHideEditorButton from './buttons/ShowHideEditorButton.js';
 import OpenInCodeSandboxButton from './buttons/OpenInCodeSandboxButton.js';
 import { useLocalFiles } from './contexts/LocalFilesProvider.js';
 
-const MIN_HEIGHT = 350;
-const MAX_WIDTH = 390;
+const MIN_EDITOR_HEIGHT = 350;
+const MAX_EDITOR_MOBILE_WIDTH = 390;
 
 async function copyCode({ code }: {| code: ?string |}) {
   try {
@@ -27,34 +27,20 @@ async function copyCode({ code }: {| code: ?string |}) {
 }
 
 function SandpackContainer({
-  name,
   layout,
-  previewHeight = 'md',
-  showEditor,
-  hideControls,
-  mobileView,
+  name,
+  previewHeight,
+  hideControls = false,
+  hideEditor = false,
 }: {|
+  layout: 'row' | 'column' | 'mobileRow',
   name: string,
-  layout: 'row' | 'column',
-  previewHeight?: 'sm' | 'md' | number,
-  showEditor: boolean,
+  previewHeight?: number,
   hideControls?: boolean,
-  mobileView?: boolean,
+  hideEditor?: boolean,
 |}) {
-  const [editorShown, setEditorShown] = React.useState(showEditor);
+  const [editorShown, setEditorShown] = React.useState(!hideEditor);
   const { sandpack } = useSandpack();
-
-  const CARD_SIZE_NAME_TO_PIXEL = {
-    sm: 236,
-    md: MIN_HEIGHT,
-  };
-
-  const height =
-    previewHeight !== 'sm' && previewHeight !== 'md'
-      ? previewHeight
-      : CARD_SIZE_NAME_TO_PIXEL[previewHeight];
-
-  const maxWidth = mobileView ? MAX_WIDTH : undefined;
 
   return (
     <React.Fragment>
@@ -62,8 +48,8 @@ function SandpackContainer({
         <SandpackLayout>
           <SandpackPreview
             style={{
-              maxWidth,
-              height,
+              maxWidth: layout === 'mobileRow' ? MAX_EDITOR_MOBILE_WIDTH : undefined,
+              height: previewHeight,
             }}
             showRefreshButton={false}
             showOpenInCodeSandbox={false}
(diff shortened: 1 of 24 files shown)


docs/docs-components/SandpackExample.js
```diff
@@ -59,9 +51,9 @@ function SandpackContainer({
             <SandpackCodeEditor
               wrapContent
               style={{
-                height: height > MIN_HEIGHT ? height : MIN_HEIGHT,
+                height: height != null && height > MIN_EDITOR_HEIGHT ? height : MIN_EDITOR_HEIGHT,
                 flex: layout === 'column' ? 'none' : null,
-                order: 1,
+                order: layout === 'column' ? 1 : null,
```
AlbertCarreras, line 56: why are we undoing this? 


I was moving the code editor below de example for consistency
Screen Shot 2022-08-04 at 9 22 32 PM

christianvuerings: @AlbertCarreras I did this since CodeSandbox also always has the editor on the left. So it makes it more consistent when users click `Open in CodeSandbox`.

Otherwise the code "jumps" from 1 side to the next:
image

image

For now I'll revert this change though, to get this diff landed

docs/docs-components/SandpackExample.js
```diff
@@ -112,17 +97,15 @@ export default function SandpackExample({
...
   code: ?string | (() => Node),
-  layout?: 'row' | 'column',
+  layout?: 'row' | 'column' | 'mobileRow',
   name: string,
-  previewHeight?: 'sm' | 'md' | number,
+  previewHeight?: number,
+  showControls?: boolean,
   showEditor?: boolean,
-  hideControls?: boolean,
-  mobileView?: boolean,
```
christianvuerings, line 125: @AlbertCarreras fyi I merged `mobileView` into the `layout` prop with `mobileRow`. It would be great if we could limit the amount of props available on `SandpackExample`

docs/docs-components/SandpackExample.js
```diff
@@ -72,37 +57,37 @@ function SandpackContainer({
             <SandpackCodeEditor
               wrapContent
               style={{
-                height: height > MIN_HEIGHT ? height : MIN_HEIGHT,
                 flex: layout === 'column' ? 'none' : null,
-                order: 1,
```
christianvuerings, line 77: @AlbertCarreras fyi, we no longer need to set `order` here - we only needed it when it was conditional.

christianvuerings:
@AlbertCarreras or @dangerismycat could either of you take a look? Would love to merge this asap to avoid further merge conflicts (e.g. `gap` changes)

dangerismycat:
@christianvuerings Sorry about that — Alberto and Ayesha are out this week, so it's just me. 😅  Taking a look now

docs/docs-components/SandpackExample.js
```diff
@@ -27,43 +27,29 @@ async function copyCode({ code }: {| code: ?string |}) {
...
   showEditor,
-  hideControls,
-  mobileView,
 }: {|
+  layout: 'row' | 'column' | 'mobileRow',
   name: string,
-  layout: 'row' | 'column',
-  previewHeight?: 'sm' | 'md' | number,
+  previewHeight?: number,
+  showControls: boolean,
```
dangerismycat, line 39: It seems like this prop should be optional; however, that would imply (to me) that the default value would be `false` and we'd only add the prop when needed. But I think our desired default is to show the controls (I believe this is why Alberto initially created this as `hideControls`).

What do you think?

docs/examples/layouts/layoutsExample.js
```diff
@@ -0,0 +1,91 @@
...
+
+        <TextField
+          label="TextField 1"
+          id="textfield1"
+          onChange={() => {}}
+          placeholder="Placeholder"
+        />
+
+        <Box
+          // Using Box instead of Flex + 'gap' for proper vertical spacing when text fields wrap
```
dangerismycat, line 30: now we no longer need this 😄

dangerismycat (approved):
We can update that SandpackExample prop in a future PR, so feel free to merge this if you're ok with it

docs/examples/layouts/layoutsExample.js
christianvuerings: @dangerismycat nice one, updated the example to use the general `gap` now.

docs/docs-components/SandpackExample.js
christianvuerings: @dangerismycat you both convinced me:
* changed `showControls` back to `hideControls`
* changed `showEditor` to `hideEditor` since the default is also to show the editor

PR #2265, opened 2022-08-07
Internal: Upgrade React Testing Library and update flow types

christianvuerings:
#### What changed?

* Upgrade React Testing Library and peer dependencies
* Update flow types for React testing library

#### Why?

Ultimately we want to add https://www.npmjs.com/package/jest-fail-on-console - however we saw a lot of errors because the previous version of React Testing library still uses React.render

#### Notes

Example issues found:
* Sheet: `Not wrapped in act`
* Sheet: `Function components cannot be given refs`


dangerismycat (approved):
Thanks for this Christian — especially the Flow types!

PR #2268, opened 2022-08-08
Internal: fix Flow issues with positioning utils

dangerismycat:
@christianvuerings I'd love your 👀 on this. Removing one `*` Flow type led to a lot of downstream type adjustments and I'd love a second set of eyes to make sure I didn't flub anything 😅

### DIFF
--- a/packages/gestalt/src/behaviors/TrapFocusBehavior.js
+++ b/packages/gestalt/src/behaviors/TrapFocusBehavior.js
@@ -29,7 +29,7 @@ function queryFocusableAll(el: HTMLDivElement) {
 
 const focusElement = (el: HTMLElement) => {
   // https://github.com/facebook/flow/issues/8705
-  // $FlowFixMe[method-unbinding]
+  // $FlowIssue[method-unbinding]
   if (typeof el.focus === 'function') {
     el.focus();
   }
--- a/packages/gestalt/src/utils/positioningUtils.js
+++ b/packages/gestalt/src/utils/positioningUtils.js
@@ -77,8 +77,10 @@ export const getTriggerRect = ({
   anchor: HTMLElement,
   positionRelativeToAnchor: boolean,
   scrollBoundaryContainerRef: ?HTMLElement,
-  // $FlowFixMe[unclear-type]
-|}): * => {
+|}): {|
+  relativeOffset: ?Coordinates,
+  triggerBoundingRect: ?ClientRect,
+|} => {
   let triggerBoundingRect;
   let relativeOffset;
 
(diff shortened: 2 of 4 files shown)


packages/gestalt/src/behaviors/TrapFocusBehavior.js
```diff
@@ -29,7 +29,7 @@ function queryFocusableAll(el: HTMLDivElement) {
 
 const focusElement = (el: HTMLElement) => {
   // https://github.com/facebook/flow/issues/8705
-  // $FlowFixMe[method-unbinding]
+  // $FlowIssue[method-unbinding]
```
dangerismycat, line 32: Unrelated to the rest of the changes — using a more precise suppression here to indicate that it's an issue on Flow's side

packages/gestalt/src/utils/positioningUtils.js
```diff
@@ -77,8 +77,10 @@ export const getTriggerRect = ({
   anchor: HTMLElement,
   positionRelativeToAnchor: boolean,
   scrollBoundaryContainerRef: ?HTMLElement,
-  // $FlowFixMe[unclear-type]
-|}): * => {
```
dangerismycat, line 81: This is the starting point of the changes — everything else is downstream of this change

christianvuerings: Love these changes, thank you Ryan.
Would be nice if we had no more `$FlowFixMe`'s in Gestalt's codebase, especially in exported code.

packages/gestalt/src/utils/positioningUtils.js
```diff
@@ -135,24 +137,24 @@ export function getPopoverDir({
...
   windowSize: Window,
   isScrollBoundaryContainer?: boolean,
 |}): MainDirections {
   // Calculates the available space if we were to place the popover in the 4 main directions
   // to determine which 'quadrant' to position the popover inside of
-  let up = triggerRect.top - popoverSize.height - CARET_HEIGHT;
-  let right = windowSize.width - popoverSize.width - CARET_HEIGHT - triggerRect.right;
-  let down = windowSize.height - popoverSize.height - CARET_HEIGHT - triggerRect.bottom;
-  let left = triggerRect.left - popoverSize.width - CARET_HEIGHT;
+  let up = (triggerRect?.top ?? 0) - popoverSize.height - CARET_HEIGHT;
```
christianvuerings, line 146: Should / could we do the fallback for `top` / `right` / `bottom` / `left` to `0` whenever we initialize `triggerRect`? If we do that though, make sure we have good tests in place so we don't introduce any bugs.

packages/gestalt/src/utils/positioningUtils.js
dangerismycat: Agreed. Not sure if it's something we'll be able to prioritize, but we can chip away at it with lil cleanup PRs.

packages/gestalt/src/utils/positioningUtils.js
dangerismycat: Possibly — I put minimal thought into this as it's all related to Popover, which there's a pretty good chance we replace with a third party library sometime soon-ish. It'll be really nice to offload that responsibility and delete all this code

PR #2269, opened 2022-08-09
Docs: fix Flex issues

dangerismycat:
😓

### DIFF
--- a/docs/pages/foundations/color/examples.js
+++ b/docs/pages/foundations/color/examples.js
@@ -201,7 +201,7 @@ export default function ColorExamplesPage(): Node {
             defaultCode={`
             <Flex>
               <Flex direction="column" gap={{ column: 2, row: 0 }}>
-                <Flex gap={{ column: 0, row: 2 }}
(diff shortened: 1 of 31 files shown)


docs/pages/foundations/color/examples.js
```diff
@@ -201,7 +201,7 @@ export default function ColorExamplesPage(): Node {
             defaultCode={`
             <Flex>
               <Flex direction="column" gap={{ column: 2, row: 0 }}>
-                <Flex gap={{ column: 0, row: 2 }}
+                <Flex gap={{ column: 0, row: 2 }}>
```
christianvuerings, line 204: @dangerismycat thanks for fixing these - one more reason as to why we want these examples to be flow typed. We can achieve that by using #2221 for every example /cc @AlbertCarreras

dangerismycat: Agreed — we're going to prioritize that migration and start converting pages soon (I hope to convert at least one this week just to get a feel for it). The lack of errors/Flow/linting on our code examples has bugged me for such a long time

PR #2271, opened 2022-08-10
Internal: disallow console.* calls in tests

christianvuerings:
#### What changed?

We disallow `console.*` calls such as `console.log` / `console.error` and `console.warn` in our jest tests.

### Why?

* Clean output when you run tests. No random console message when running the tests
* Find issues: deprecated or incorrect React testing usage / invalid examples
* Make `console.*` calls explicit, so it's intentional & clear when we use them (E.g. codemod runs)

#### Before
Screen Shot 2022-08-10 at 3 05 29 PM

#### After
Screen Shot 2022-08-10 at 3 06 06 PM

### Example issues found

* `Error: Uncaught [Error: Foo translations not available — please add translations to I18nProvider] `
* `Warning: toHaveDescription has been deprecated and will be removed in future updates. Please use toHaveAccessibleDescription.`
* `Warning: An update to InnerPopper inside a test was not wrapped in act(…).`

### Notes

* Now, if we do find a `console.*` which we can't get rid off, we should mock the `console.*` call. There are a few examples in this diff
* Set this to a `patch` version since no runtime code has changed (only tests)

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -117,7 +117,7 @@
       "rules": {
         "jest/expect-expect": [
           "error",
-          { "assertFunctionNames": ["expect", "runInlineTest"] }
+          { "assertFunctionNames": ["expect", "runInlineTest", "runTest"] }
         ]
       }
     },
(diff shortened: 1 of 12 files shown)


dangerismycat (approved):
Good idea — thanks Christian!

PR #2273, opened 2022-08-11
Internal: remove deepCloneReplacingUndefined

christianvuerings:
#### What changed?

Removes `deepCloneReplacingUndefined` from Gestalt's codebase

#### Why?

The output with or without the function doesn't change. The docs still seem to work.

### Links

Introduced in https://github.com/pinterest/gestalt/pull/1912 /cc @dangerismycat 
More context: https://github.com/pinterest/gestalt/pull/1912#discussion_r799760855


christianvuerings:
@dangerismycat could you validate that we no longer need this?

dangerismycat (approved):
I'm not seeing any errors either in the build or the preview deploy console…I wonder what changed?

PR #2275, opened 2022-08-11
Sheet: fix flicker when closing Sheet with React 18

### LINKED ISSUE #2156: React 18 causing Sheet to flash
React 18 flashes sheets

Go to this example in Sheet
https://gestalt.netlify.app/sheet#Preventing-close-on-outside-click

Click on View Sheet 
Then close the Sheet.

Now you can see after closing the sheet a flash of old Sheet, after the sheet gets closed.

This is mostly because of react 18 I presume.

christianvuerings:
# Summary

## What changed?

Fixes #2156: a flicker when you close `Sheet` in React 18

## Why?

So our users don't see this annoying flicker

## Notes

* `flushSync` should be used very sparingly. In this case, we only use it after an animation ends to ensure the callback gets fired at the correct time.

### Before
https://user-images.githubusercontent.com/127199/184133136-b027b350-6705-4ee2-82bb-d984bdd78681.mov

### After
https://user-images.githubusercontent.com/127199/184133154-b3dc6a23-6f83-4db0-a240-2c63b04c8f2b.mov

## Links

* https://reactjs.org/docs/react-dom.html
* https://github.com/radix-ui/primitives/pull/1344/files
* https://cs.github.com/firefox-devtools/react-contextmenu/blob/cac2555b8e0cda95258633aeed3cb9469cfd7e0d/src/ContextMenu.js?q=flushSync+react#L14


dangerismycat (approved):
Thanks for the fix, Christian! I'll merge+bump once Eduardo is finished with bumping his Heading PR

PR #2276, opened 2022-08-12
Masonry: add support for RTL text direction

dangerismycat:
Masonry currently does nothing in response to the user's right-to-left (RTL) settings. This doesn't really matter in full-width, infinite grids, like homefeed; the user isn't likely to even be aware if the first item is in the top left or the top right. However, this _does_ matter when the grid width is constrained, particularly if there are a finite number of items or the user is likely to be aware of the order of the items, e.g. the Homefeed Tuner (pins and topics). 
Page_Broken_HW

This PR adds logic to support laying out Masonry items right-to-left according to the computed text direction for the Masonry container. This will typically match the style set on the entire page (typically on `<html>` or `<body>`), but will also respect more "local" settings, since [the `dir` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/dir) can be set on any HTML element.

Note that when testing this out on our docs site, you'll need to refresh after toggling the site RTL setting. We could listen for changes to `dir` and force a reflow, but that seems quite unlikely given that users typically aren't toggling that setting.

### DIFF
--- a/packages/gestalt/src/Masonry.js
+++ b/packages/gestalt/src/Masonry.js
@@ -408,8 +408,8 @@ export default class Masonry<T: { ... }> extends ReactComponent<Props<T>, State<
         style={{
           top: 0,
           ...(isRtl ? { right: 0 } : { left: 0 }),
-          transform: `translateX(${isRtl ? '-' : ''}${left}px) translateY(${top}px)`,
-          WebkitTransform: `translateX(${isRtl ? '-' : ''}${left}px) translateY(${top}px)`,
+          transform: `translateX(${isRtl ? left * -1 : left}px) translateY(${top}px)`,
+          WebkitTransform: `translateX(${isRtl ? left * -1 : left}px) translateY(${top}px)`,
           width: layoutNumberToCssDimension(width),
           height: layoutNumberToCssDimension(height),
         }}
--- a/packages/gestalt/src/Masonry.js
+++ b/packages/gestalt/src/Masonry.js
@@ -395,6 +395,8 @@ export default class Masonry<T: { ... }> extends ReactComponent<Props<T>, State<
       isVisible = true;
     }
 
+    const isRtl = this.gridWrapper ? getComputedStyle(this.gridWrapper).direction === 'rtl' : false;
+
     const itemComponent = (
       <div
         className={[styles.Masonry__Item, styles.Masonry__Item__Mounted].join(' ')}
@@ -403,9 +405,8 @@ export default class Masonry<T: { ... }> extends ReactComponent<Props<T>, State<
         role="listitem"
         style={{
           top: 0,
-          left: 0,
-          transform: `translateX(${left}px) translateY(${top}px)`,
-          WebkitTransform: `translateX(${left}px) translateY(${top}px)`,
+          transform: `translateX(${isRtl ? '-' : ''}${left}px) translateY(${top}px)`,
+          WebkitTransform: `translateX(${isRtl ? '-' : ''}${left}px) translateY(${top}px)`,
           width: layoutNumberToCssDimension(width),
           height: layoutNumberToCssDimension(height),
         }}
--- a/packages/gestalt/src/Masonry.js
+++ b/packages/gestalt/src/Masonry.js
@@ -405,6 +405,7 @@ export default class Masonry<T: { ... }> extends ReactComponent<Props<T>, State<
         role="listitem"
         style={{
           top: 0,
+          ...(isRtl ? { right: 0 } : { left: 0 }),
           transform: `translateX(${isRtl ? '-' : ''}${left}px) translateY(${top}px)`,
           WebkitTransform: `translateX(${isRtl ? '-' : ''}${left}px) translateY(${top}px)`,
           width: layoutNumberToCssDimension(width),
--- a/packages/gestalt/src/Masonry.js
+++ b/packages/gestalt/src/Masonry.js
@@ -395,7 +395,7 @@ export default class Masonry<T: { ... }> extends ReactComponent<Props<T>, State<
       isVisible = true;
     }
 
-    const isRtl = this.gridWrapper ? getComputedStyle(this.gridWrapper).direction === 'rtl' : false;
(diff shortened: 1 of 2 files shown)


dangerismycat:
@christianvuerings I'd love your 👀  on this whenever you get a chance, since @liuyenwei is currently OOO (I know, I know, you both are 😅 )

packages/gestalt/src/Masonry.js
```diff
@@ -409,9 +411,9 @@ export default class Masonry<T: { ... }> extends ReactComponent<Props<T>, State<
         role="listitem"
         style={{
           top: 0,
-          left: 0,
-          transform: `translateX(${left}px) translateY(${top}px)`,
-          WebkitTransform: `translateX(${left}px) translateY(${top}px)`,
+          ...(isRtl ? { right: 0 } : { left: 0 }),
```
christianvuerings, line 414: @dangerismycat should we add a visual integration test? That way we can avoid regressions.

packages/gestalt/src/Masonry.js
```diff
@@ -401,6 +401,8 @@ export default class Masonry<T: { ... }> extends ReactComponent<Props<T>, State<
       isVisible = true;
     }
 
+    const isRtl = this.gridWrapper ? getComputedStyle(this.gridWrapper).direction === 'rtl' : false;
```
christianvuerings, line 404: Curious what the render/paint performance is of calling `getComputedStyle` on every Masonry item render. https://gist.github.com/paulirish/5d52fb081b3570c81e3a#calling-getcomputedstyle

* Could/should we use `document.dir`? It's basically what we also use for RLT in our CSS: `html[dir="rtl"]` - so we don't support setting `rtl` on part for the document.
* If we need to still use `getComputedStyle`, could we do it outside of the `map` which renders every Masonry item?

packages/gestalt/src/Masonry.js
dangerismycat: Eventually yes — we currently have visual diff testing disabled as there have been numerous issues since the Playwright switch and none of us have had the time to look into them (I'll have to address at least some of them in the near future as I get the old Masonry integration tests up and running again before other work on that component).

packages/gestalt/src/Masonry.js
dangerismycat: Good point. I had no idea `getComputedStyle` was so expensive — thanks for the heads-up! I'll switch to the simple `document.dir` check

dangerismycat:
@christianvuerings Any other thoughts on this current approach?

packages/gestalt/src/Masonry.js
```diff
@@ -401,6 +401,8 @@ export default class Masonry<T: { ... }> extends ReactComponent<Props<T>, State<
       isVisible = true;
     }
 
+    const isRtl = document?.dir === 'rtl';
```
christianvuerings, line 404: @dangerismycat major thing I would test for is SSR when we do the bump in pinboard since `document` is not available on the server.

christianvuerings (approved):

packages/gestalt/src/Masonry.js
dangerismycat: Given the optional chaining, I wouldn't think we'd see an error there. Would the main concern be layout thrash?

christianvuerings: @dangerismycat yes, when server rendering the items, `document` will be `undefined` which means `isRtl` will be `false` even if `html dir="rtl"` is set.

However, we also have these 3 server render style files in pinboard which are used during server side render. So those might need to be adjusted to make SSR working correctly without the layout shifting from LTR to RTL.

- `webapp/app/packages/gestaltExtensions/Masonry/styles/getFlexibleGridServerStyles.js`
- `webapp/app/packages/gestaltExtensions/Masonry/styles/getDuploServerSideStyles.js`
- `webapp/app/packages/gestaltExtensions/Masonry/styles/getClassicGridServerStyles.js`

# Notes

* It would be great if we could move those "ServerStyles" into Gestalt so we don't have this weird interconnection between the 2
* These "ServerStyle" files work by relying on the `static` class name we give the rendered items during SSR: https://github.com/pinterest/gestalt/blob/c8193d9b33816f07ffd11605793b75ed92aa9211/packages/gestalt/src/Masonry.js#L481 . That's a different render output than what we do on the client with `renderMasonryComponent`: https://github.com/pinterest/gestalt/blob/c8193d9b33816f07ffd11605793b75ed92aa9211/packages/gestalt/src/Masonry.js#L404

liuyenwei (changes requested):
requesting changes.   as-is, this change isn't safe to land since it would cause layout thrash during SSR for RTL layouts

liuyenwei (approved):
updating to approved.  thinking about it more, the changes are fine to land and we have the option to update the pinboard css styles in the same PR as the update so nothing technically blocking this from landing

packages/gestalt/src/Masonry.js
```diff
@@ -409,9 +411,9 @@ export default class Masonry<T: { ... }> extends ReactComponent<Props<T>, State<
         role="listitem"
         style={{
           top: 0,
-          left: 0,
-          transform: `translateX(${left}px) translateY(${top}px)`,
-          WebkitTransform: `translateX(${left}px) translateY(${top}px)`,
+          ...(isRtl ? { right: 0 } : { left: 0 }),
+          transform: `translateX(${isRtl ? '-' : ''}${left}px) translateY(${top}px)`,
```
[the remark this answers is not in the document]
dangerismycat: ah that's much better, good idea

packages/gestalt/src/Masonry.js
dangerismycat: Ah I wasn't aware of those files, thanks! I just added a ticket for myself to migrate those into Gestalt.

liuyenwei: fwiw - when we migrate those into Gestalt, we should consolidate as much as possible vs maintaining the multiple forks

dangerismycat: agreed!

PR #2279, opened 2022-08-12
Internal: remove unused scripts

christianvuerings:
#### What changed?

Remove unused scripts from the Gestalt repo

#### Why?

Unused code results in more maintenance and confusion as to why we still have these scripts around.


dangerismycat (approved):
Thanks! I'll merge this once I'm done with the current bump

PR #2281, opened 2022-08-13
Internal: lint shell files

christianvuerings:
#### What changed?

* Lint shell scripts in Gestalt's codebase
* Add `shellcheck` to recommended VSCode extensions

#### Why?

Guard against common bash mistakes

### Links

- https://www.shellcheck.net/


dangerismycat (approved):
Great idea, thanks for this!

PR #2282, opened 2022-08-16
SideNavigation: fixes in height and mobile implementation in sm breakpoint

AlbertCarreras:
#### What changed?

SideNavigation: fixes in height and mobile implementation in sm breakpoint in the Docs

All modes of SideNavigation now have height="100%" 

The Docs have a custom implementation of mobile view under the sm breakpoint. We force the mobile view using the DeviceTypeProvider and setting the value to 'phone'.  Sidenav under sm breakpoint closes on scroll and when clicking outside the sidenav. 

### BEFORE
Kapture 2022-08-17 at 21 17 45

### AFTER
Kapture 2022-08-17 at 21 10 42


rlingineni:
Something seems wonky on the homepage for me. Using 13inch mbp:

image



Mobile looks great!

docs/docs-components/AppLayout.js
```diff
@@ -18,11 +19,18 @@ type Props = {|
...
+  const { isSidebarOpen, setIsSidebarOpen } = useNavigationContext();
 
   const footerColor =
     colorScheme === 'dark' ? 'var(--color-gray-roboflow-700)' : 'var(--color-orange-firetini-0)';
 
+  useEffect(() => {
+    if (typeof window !== 'undefined')
+      window.addEventListener('scroll', () => setIsSidebarOpen(false));
+
+    return () => window.removeEventListener('scroll', () => setIsSidebarOpen(false));
```
dangerismycat, line 31: This isn't actually removing the event listener — these are independent anonymous functions, with different references. You'll need to name the function to be able to remove it:
```js
useEffect(() => {
  const handleScroll = () => setIsSidebarOpen(false);
  window.addEventListener('scroll', handleScroll);
  return () => {
    window.removeEventListener('scroll', handleScroll);
  };
});
```

AlbertCarreras: Thanks! good catch

AlbertCarreras: DONE

docs/docs-components/AppLayout.js
```diff
@@ -18,11 +19,19 @@ type Props = {|
 
 export default function AppLayout({ children, colorScheme, isHomePage }: Props): Node {
   const { isMobile } = useDocsDeviceType();
-  const { isSidebarOpen } = useNavigationContext();
+  const { isSidebarOpen, setIsSidebarOpen } = useNavigationContext();
 
   const footerColor =
     colorScheme === 'dark' ? 'var(--color-gray-roboflow-700)' : 'var(--color-orange-firetini-0)';
```
AlbertCarreras, line 26: @dangerismycat I did update it

dangerismycat (approved):

PR #2284, opened 2022-08-16
Video: fix hydration issue

christianvuerings:
# Summary

## What changed?

* Fix a hydration issue for the `Video` component.
* Update visual test for `Video`

## Why?

I noticed that the visual test of the `Video` component had a hydration error overlay on top of it. The root cause was a React hydration issue.

Please take a look at https://nextjs.org/docs/messages/react-hydration-error which explains the fix.

# FAQ

## Why don't I see the error overlay when going to the production website? https://gestalt.pinterest.systems/visual-test/Video

In production, we use the Next.js and React.js production version which don't show the error overlay. You do however see an error in the console which links to this URL: https://reactjs.org/docs/error-decoder.html/?invariant=423

> There was an error while hydrating. Because the error happened outside of a Suspense boundary, the entire root will switch to client rendering.

image

You don't see this console error in the deploy preview of this PR: https://deploy-preview-2284--gestalt.netlify.app/visual-test/Video


dangerismycat (approved):
Thanks Christian!

PR #2200, opened 2022-07-12
SideNavigation: V2

AlbertCarreras:
#### What changed?

SideNavigation: V2 release including nested directories

#### Why?

V2 in alpha release. More changes coming to the component including improved keyboard navigation 

Breaking changes: 

- SideNavigation.Item API changes with supporting codemods for manual update

 - `onSelected` to `onClick` 

```
yarn codemod detectManualReplacement ~/path/to/your/code
--component=SideNavigation
--subcomponent=Item
--prop=onSelected
```

 - `item` to `label`
 
```
yarn codemod detectManualReplacement ~/path/to/your/code
--component=SideNavigation
--subcomponent=Item
--prop=item
```

 - `SideNavigation.Item` to `SideNavigation.TopItem`
 
```
yarn codemod detectManualReplacement ~/path/to/your/code
--component=SideNavigation
--subcomponent=Item
--prop=href
```

### Links

- [Figma](https://www.figma.com/file/kK9TSF41utPLF2CIRoGLhX/Gestalt-SideNav-Component?node-id=285%3A4844)
- [TDD](https://paper.dropbox.com/doc/TDD-SideNavigation--BkRWqAZLEiLaYaoK1dA32uFvAg-4IWmy1gXM95tF7SNNKh2A)

### Checklist

- [x] Added unit and Flow Tests
- [x] Added documentation + accessibility tests
- [x] Verified accessibility: keyboard & screen reader interaction
- [x] Checked dark mode, responsiveness, and right-to-left support
- [x] Checked stakeholder feedback (e.g. Gestalt designers)

PR #2286, opened 2022-08-17
Internal: re-enable playwright visual tests

christianvuerings:
# Summary

## What changed?

Re-enable visual integration tests with these changes:

* **Easier to run**: Remove Docker dependency & linux screenshots
* **More consistent**: Only run playwright tests on macOS
* **Up-to-date**: Update `playwright`, `eslint-plugin-playwright` and `@axe-core/playwright`
* **Less flaky**:
  * set `maxDiffPixelRatio` and `maxDiffPixels` options
  * set `retries` to `3`
* **Faster**: We now [shard](https://playwright.dev/docs/test-parallel#shard-tests-between-multiple-machines) the runs in CI 

## Why?

To avoid regressions moving forward - we also saw it guards against hydration issues - e.g. #2284 

## Notes

* Visual integration tests were disabled in #2200 
* If we see any issues after this lands, change `npx playwright test playwright/` to `npx playwright test playwright/accessibility/` in `package.json`'s `playwright:test"`. That way, we only run the accessibility integration tests.

###  Ways to make tests less flaky

- Await for focus ring:
    - `await page.focus('button', { timeout: 2000 });`
- Await for tooltip to be visible:
    - `await page.locator('div[role="tooltip"]');`
- Await for all images to be loaded (e.g. Collage)
    - `await page.goto('/visual-test/Collage', {waitUntil: 'networkidle',});`


Dockerfile
```diff
@@ -1,12 +1,12 @@
-FROM mcr.microsoft.com/playwright:v1.22.0-focal
+FROM mcr.microsoft.com/playwright:v1.25.0-focal
```
AlbertCarreras, line 1: @christianvuerings What is this for?

AlbertCarreras:
@christianvuerings I wouldn't merge this until you're back. We disable them cos they were causing many problems. Here a list.


    - the error messing is not intuitive, very opaque
    - vis diffs fail… God knows why, the errors seem inconsistent
    - Docker requires syncing with local updates otherwise the linux test fail
            How to Configure Your Container to Sync Your Code Automatically
            1. Find the folder in your Docker container that has your code. ...
            2. Find the path to the folder on your laptop that has the same code.
            3. Add a host volume to your Docker-compose file. ...
            4. Switch from using node.js to nodemon. ...
            5. Run Docker Compose or Blimp.

    - There’s no interface, says it couldn’t render the image
        - Datepicker not picking date value in linux
        - cannot run individual test in linux, but all and they all get updated
        - Playwright images don’t display well on Github
        - Another component snapshot shows an error message in  the snapshot
        
        - Fonts differ
        - ComboBox positioning is different between linux/build
    - Different  behavior between local/CI

it wasn't easy and caused a lot of frustration. I'd be careful merging again

dangerismycat (changes requested):
I'd _really_ prefer to avoid re-enabling these tests until someone can dedicate time to figuring out our processes around them. It was very, very painful and disruptive to have them enabled, which is why we disabled them. I don't have time right now to dive deep into Playwright and why the tests were so flaky and slow. I'm fine with the trade-off of possible regressions in the short term.

Dockerfile
christianvuerings: @AlbertCarreras to upgrade playwright to the latest version. In this case, we use https://hub.docker.com/_/microsoft-playwright as a base image

christianvuerings:
@AlbertCarreras @dangerismycat thank you for the feedback and list of issues you encountered. It was totally fine to disable them since they blocked the team.

Couple of notes below of how we could potentially address them:

> Docker requires syncing with local updates otherwise the linux test fail

Lots of issue are due to Linux / Docker - instead, I wonder if we should use `macos` as a container everywhere. That way we wouldn't have to run docker anywhere.

> Another component snapshot shows an error message in  the snapshot

Was this #2284 ? If so, it means we now guard against hydration issues

christianvuerings:
@AlbertCarreras @dangerismycat want to take another look at this diff?

I updated the PR description with all the changes to make the playwright tests faster, less flaky & easier to update.

christianvuerings:
@dangerismycat @AlbertCarreras heads up that I also added a section to the docs about how to update the visual snapshots for a specific test:

image

Are there any other processes / things we should change?

christianvuerings:
@AlbertCarreras @dangerismycat What do you recommend as next steps for this PR? Please let me know if something is missing.

It's totally fine if we would have to disable the visual tests again if an issue would pop up. For now though, the tests should be way faster, less flaky and easier to maintain.

dangerismycat:
Hey @christianvuerings - Sorry for the radio silence on this one. We've been pretty busy / OOO, but I'll take a look at this PR again next week and possibly merge it. Thanks for this work!

dangerismycat (approved):

PR #2288, opened 2022-08-17
Component overview: updated per redesign

AlbertCarreras:
Component overview: updated as per redesign

- [Figma](https://www.figma.com/file/a2ar8cqq2nhoz9IssecW8y/Components-overview-updates?node-id=5%3A298file)

### DIFF
--- a/docs/pages/web/overview.js
+++ b/docs/pages/web/overview.js
@@ -11,18 +11,20 @@ import Accessibility from '../../graphics/foundations/accessibility.svg';
 
 const getIllustrationCardColor = (category: string, hasDarkBackground?: boolean) => {
   const tealBackgrounds = ['Foundations'];
-  const grayBackgrounds = ['Utilities', 'Building Blocks'];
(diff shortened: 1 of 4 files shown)


docs/pages/web/overview.js
```diff
@@ -113,15 +122,17 @@ function List({ array, title = '' }: {| array: ListItemType, title?: string |}):
 }
 
 export default function ComponentOverview(): Node {
-  const [order, setOrder] = useState('category');
+  const [order, setOrder] = useState('alphabetical');
```
dangerismycat, line 125: We don't want to start with categorical organization here? We already default to alphabetical organization in the sidenav (well, it's the only option we offer 😅) — would it not make sense to keep the default here by category? I think that also makes the page more browsable, which is the goal of the page.

dangerismycat (approved):

docs/pages/web/overview.js
AlbertCarreras: @hectoid  Thoughts?

PR #2292, opened 2022-08-17
Foundations: new Overview page

AlbertCarreras:
Foundations: new Overview page

- [Figma](https://www.figma.com/file/zHdav5Odp1YEl0jnaBgyOs/Foundations-overview?node-id=2%3A125)

Pending: some descriptions are pending but general implementation is done


docs/pages/foundations/overview.js
```diff
@@ -0,0 +1,38 @@
...
+            description="Here you will find the foundational elements for designing UI components, surfaces and patterns. These include graphic elements like iconography, color and typography to structural. They also cover elevation, layout and accessibility."
+            type="guidelines"
+          />
+        </IllustrationContainer>
+        <IllustrationSection title="" grid="auto-fill" min={312}>
+          {COMPONENT_DATA.foundations.map((element, idx) => (
+            <IllustrationCard
+              headingLevel={2}
+              key={idx}
+              href={element?.path ?? `/web/${element.name.replace(/\s/g, '_').toLowerCase()}`}
```
rlingineni, line 27: I feel like this that `convertToUrl` function we keep using everywhere? 

Should we redefine it here again, so it's all aligned, and then in a future update, we can refactor the code and store that function in a util file somewhere.

rlingineni:
LGTM!

rlingineni (approved):

docs/pages/foundations/overview.js
AlbertCarreras: Yeah, we should cleanup some of this if we reuse them.We can do  a cleanup follow up PR

PR #2297, opened 2022-08-18
Internal: run Algolia crawler on every PR merge

dangerismycat:
In order to keep the docs search up-to-date, we should re-crawl the site after every PR merge. [There's a GitHub Action to do  that.](https://github.com/algolia/algoliasearch-crawler-github-actions) This PR enables that Action. I've also added the necessary secrets to our repo settings. Hopefully this works? 🤞 (Not sure if there's any way to test aside from merging the PR, so once CI passes I'll merge this and then log on to Algolia to see if a new crawl was kicked off.)

### DIFF
--- a/.github/workflows/algolia.yml
+++ b/.github/workflows/algolia.yml
@@ -1,0 +1,9 @@
+name: Algolia crawler creation and crawl
+uses: algolia/algoliasearch-crawler-github-actions@v1
+id: algolia_crawler
+with:
+  crawler-user-id: ${{ secrets.CRAWLER_USER_ID }}
+  crawler-api-key: ${{ secrets.CRAWLER_API_KEY }}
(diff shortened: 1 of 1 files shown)


.github/workflows/algolia.yml
```diff
@@ -0,0 +1,9 @@
+name: Algolia crawler creation and crawl
```
christianvuerings, line 1: @dangerismycat is this a correct GitHub workflow format? E.g. when is this action supposed to execute?

I think we might need something along these lines:
https://github.com/algolia/algoliasearch-crawler-github-actions/blob/main/examples/basic.yml
(perhaps only re-crawl when we push to `master`?)

dangerismycat: Probably not! I have no idea what I'm doing here. 😅

Thanks for passing along that example. I'll pattern-match from that!

PR #2298, opened 2022-08-18
Internal: run Prettier on CSS files

christianvuerings:
# Summary

## What changed?

Use [Prettier](https://prettier.io/) on our CSS files

## Why?

* Consistent formatting across all of our codebase (JS & md already uses prettier

## Notes

* We already had [stylelint-config-prettier](https://www.npmjs.com/package/stylelint-config-prettier) installed but never used it.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,6 +1,5 @@
 __fixtures__
 __testfixtures__
-*.css
 build
 coverage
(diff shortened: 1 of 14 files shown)


AlbertCarreras (approved):
LGTM

PR #2304, opened 2022-08-19
IconButton: change default `iconColor` prop value from "gray" to "darkGray"

AlbertCarreras:
# Summary

#### What changed?

IconButton: change default iconColor prop value from "gray" to "darkGray"

#### Why?

As an engineer, I want the IconButtton's color and size props to default to Gestalt design recommendations so that implementation of typical usage requires less code and work.

Acceptance criteria

The default value of the iconColor prop should be changed from gray to darkGray.

####  Breaking changes codemod

Run 

`yarn codemod --parser=flow -t=packages/gestalt-codemods/66.0.0/iconButton-defaut-iconColor-change.js relative/path/to/your/code
`
- [Jira](https://jira.pinadmin.com/browse/GESTALT-2501)


packages/gestalt/src/IconButton.js
```diff
@@ -87,7 +87,7 @@ const IconButtonWithForwardRef: React$AbstractComponent<unionProps, unionRefs> =
     dangerouslySetSvgPath,
     disabled,
     icon,
-    iconColor,
+    iconColor = 'darkGray',
```
dangerismycat, line 90: Huh…we weren't setting the default color before? That makes me suspicious that it's still being defined somewhere…have you looked into that?

dangerismycat (approved):

packages/gestalt/src/IconButton.js
AlbertCarreras: Pog sets "gray" based on the default bgColor which is transparent.

PR #2316, opened 2022-08-22
Docs: Fix Overlapping Accessibility Checks

rlingineni:
It seems that our `validateAccessibilitySpecs.js` check didn't account for overlapping pages.

e.g. `iOS/Avatar` and `Web/Avatar`  just look for `Avatar.spec.mjs`

I's super easy to miss an accessibility check or add one. This updates the page finding algo to handle collisions with a prefix. 

So naming convention is `iOS_Avatar` or `web_Avatar`. It only looks for more specific names when a collision exists. The check will spit out the names that don't exist, so no need to worry about the naming schema.

For the majority of cases, we can keep things as is


dangerismycat:
I think I understand the issue, but can you provide more detail about the problem and your proposed solution? More comments (especially expected input/output types, as if we were using Flow in the script file) would be helpful.

rlingineni:
> I think I understand the issue, but can you provide more detail about the problem and your proposed solution? More comments (especially expected input/output types, as if we were using Flow in the script file) would be helpful.

I think I fixed the comments to be a bit better.

Initially we had paths like:

```
web/avatar
ios/avatar
ios/overview
web/overview
get_started/development_process
```

The previous script would flatten everything, and only return `avatar, overview, development_process` to be required. However, there are five distinct pages to be validated.

The new algorithm, looks at all the sitepaths, and handles overlapping page matches, by going one level deeper to distinguish the pages

```
web/avatar -> web_avatar
ios/avatar -> ios_avatar
ios/overview -> ios_overview
web/overview -> web_overview
get_started/development_process -> development_process
```

Hopefully that clarifies

rlingineni:
Also heads up #2319 has to be merged first

dangerismycat (approved):

PR #2325, opened 2022-08-24
Masonry: rename `comp` to `Item`

dangerismycat:
The current prop name `comp` on Masonry doesn't make it clear how the prop will be used: is it a function? A function component? A component of any sort? It's not clear.

Renaming this prop to `Item` makes it more clear that the prop will be used as a component. The ultimate goal is to ensure that users know this always needs to be a stable reference, which this barely addresses. However, future work will add a lint rule to ensure that `Item` is always stable, and will possibly rethink how we treat the item component to completely preclude this possibility.

More context in [this Slack thread](https://pinterest.slack.com/archives/C13KLG5P0/p1661293861162029)

## Codemod
```zsh
yarn codemod modifyProp ~/path/to/your/code --component=Masonry --previousProp=comp --nextProp=Item
```

_Notes_
@ponori This isn't one for the weekly digest, just the monthly eng newsletter.

### DIFF
--- a/packages/gestalt/src/Masonry.js
+++ b/packages/gestalt/src/Masonry.js
@@ -23,19 +23,19 @@ type Props<T> = {|
    */
   columnWidth?: number,
   /**
+   * The amount of vertical and horizontal space between each item, specified in pixels.
+   */
+  gutterWidth?: number,
+  /**
    * A React component (or stateless functional component) that renders the item you would like displayed in the grid. This component is passed three props: the item's data, the item's index in the grid, and a flag indicating if Masonry is currently measuring the item. *Note that this [must be a stable reference!](https://www.developerway.com/posts/react-re-renders-guide#part3.1)* If using a component declared within a parent function component, you must use [`useCallback`](https://reactjs.org/docs/hooks-reference.html#usecallback) to ensure a stable reference.
    */
-  comp: ComponentType<{|
+  Item: ComponentType<{|
     data: T,
     itemIdx: number,
     isMeasuring: boolean,
   |}>,
   /**
-   * The amount of vertical and horizontal space between each item, specified in pixels.
-   */
-  gutterWidth?: number,
-  /**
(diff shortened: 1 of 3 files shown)


jackhsu978 (approved):
Nice!

packages/gestalt/src/Masonry.js
```diff
@@ -22,20 +22,20 @@ type Props<T> = {|
...
   columnWidth?: number,
+  /**
+   * The amount of vertical and horizontal space between each item, specified in pixels.
+   */
+  gutterWidth?: number,
   /**
    * A React component (or stateless functional component) that renders the item you would like displayed in the grid. This component is passed three props: the item's data, the item's index in the grid, and a flag indicating if Masonry is currently measuring the item. *Note that this [must be a stable reference!](https://www.developerway.com/posts/react-re-renders-guide#part3.1)* If using a component declared within a parent function component, you must use [`useCallback`](https://reactjs.org/docs/hooks-reference.html#usecallback) to ensure a stable reference.
    */
-  comp: ComponentType<{|
+  Item: ComponentType<{|
```
AlbertCarreras, line 32: @dangerismycat Why are we capitalizing the prop name? should it be `item` rather than `Item`?

jackhsu978: @AlbertCarreras this is to indicate that it's not a regular JavaScript function, but a React component.

A variable that is a React component should be capitalized. For example:

```js
const Nav = type === 'top' ? TopNav : BottomNav;
return <Nav />
```

AlbertCarreras: Yeah, I knew that, wasn't sure about capitalized props

dangerismycat: @AlbertCarreras I added more context to the PR summary. Check out this Slack thread: https://pinterest.slack.com/archives/C13KLG5P0/p1661293861162029

PR #2329, opened 2022-08-24
Docs: Added iOS icon page

khalvy:
#### What changed?

Added the iOS icon page along with importing the Icon Button and Button graphics


docs/markdown/ios/icon.md
```diff
@@ -0,0 +1,136 @@
...
+fullwidth: true
+---
+
+<ImgContainer src="https://i.pinimg.com/originals/1d/52/31/1d5231c99699e72628135de682fede30.jpg" />
+
+## Usage guidelines
+
+<TwoCol>
+<Group>
+<Do title="Do" />
```
ayeshakmaz, line 13: I think this title should actually read "When to use" instead of "Do"

docs/markdown/ios/icon.md
```diff
@@ -0,0 +1,136 @@
...
+- Use as symbolic communication for elements that do not have room for text. In this case, ensure the icon choice is easily recognizable and makes sense to international users.
+- To help with quick scanning by adding rhythm and hierarchy to the design.
+
+</Group>
+<Group>
+<Dont title="Don't" />
```
ayeshakmaz, line 19: And this would be "When not to use"

docs/markdown/ios/icon.md
```diff
@@ -0,0 +1,136 @@
...
+</Group>
+</TwoCol>
+
+## Related
+
+<TwoCol>
+
+<IllustrationCard
+              title="Button"
+              href="#related"
```
ayeshakmaz, line 122: Can you remove this href line? it's not needed

docs/markdown/ios/icon.md
```diff
@@ -0,0 +1,136 @@
...
+              description="Button allows users to take actions, and make choices using text labels to express what action will occur when the user interacts with it."
+              color="green-matchacado-50"
+              image="button"
+            />
+
+<IllustrationCard
+              title="IconButton"
+              href="#related"
```
ayeshakmaz, line 130: Same here, this can be removed

ayeshakmaz (commented):
You'll also need to add an accessibility test for this page

See this PR as an example: https://github.com/pinterest/gestalt/pull/2331/files

It would called Icon_iOS.spec.mjs , within the playwright/accessibility folder

PR #2330, opened 2022-08-24
Popover: Fix accessibility bug, add option for accessibilityLabel

ayeshakmaz:
#### What changed?
Update Popover (and its inner components) to always include a role and aria-label - a requirement for accessibility. Also allow users to specify an accessibilityLabel for their Popover

#### Why?

We were disabling an a11y check, which caused issues in the main repo. All non-modal dialogs (Popovers) need a role. All items with role = dialog or listbox need a label. So this PR adds defaults for both to ensure every Popover is accessible by default (but also allows users to override the label to be more specific, which is preferred).

### Links

- [Jira](https://jira.pinadmin.com/browse/GESTALT-3554)

### Checklist

### DIFF
--- a/docs/pages/web/popover.js
+++ b/docs/pages/web/popover.js
@@ -532,7 +532,7 @@ To further assist screen readers, we recommend passing the following ARIA attrib
 - \`accessibilityExpanded\`: informs the screen reader whether Popover is currently open or closed. It populates [aria-expanded](https://www.w3.org/TR/wai-aria-practices/examples/menu-button/menu-button-links.html).
 - \`accessibilityControls\`: match with the \`id\` of the associated Popover whose contents or visibility are controlled by the interactive component so that screen reader users can identify the relationship between elements. It populates [aria-controls](https://www.w3.org/TR/wai-aria-practices/examples/menu-button/menu-button-links.html).
 
-For role attribute, use:
+For the \`role\` prop, use:
 - 'dialog' if the Popover is a dialog that prompts the user to enter information or requires a response.
 - 'menu' if the Popover presents a list of choices to the user.
 - 'listbox' if the Popover is a widget that allows the user to select one or more items (whose role is option) from a list. May also include a search option.
--- a/docs/pages/web/popover.js
+++ b/docs/pages/web/popover.js
@@ -97,6 +97,7 @@ function PopoverExample() {
       {open && (
         <Layer>
           <Popover
+            accessibilityLabel="Save to board"
             anchor={anchorRef.current}
             id="main-example"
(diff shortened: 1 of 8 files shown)


packages/gestalt/src/Popover.js
```diff
@@ -72,7 +77,7 @@ export default function Popover({
   onDismiss,
   positionRelativeToAnchor = true,
   color = 'white',
-  role,
+  role = 'dialog',
```
ayeshakmaz, line 80: Every Popover needs a role, and the safest default is dialog

packages/gestalt/src/Popover.js
```diff
@@ -64,6 +68,7 @@ type Props = {|
  * Popover is most appropriate for desktop screens and can contain a variety of elements, such as [Button](/button) and [Images](/image). Popover is also the container used to construct more complex elements like [Dropdown](/dropdown) and the board picker, pictured below, which allow people to choose the board to save a Pin to.
  */
 export default function Popover({
+  accessibilityLabel = 'Popover',
```
ayeshakmaz, line 71: A safe default label is always "Popover"

playwright/accessibility/Popover.spec.mjs
```diff
@@ -6,8 +6,5 @@ test('Popover Accessibility check', async ({ page }) => {
   await page.goto('/web/popover');
   await expectAccessiblePage({
     page,
-    rules: {
```
ayeshakmaz, line 9: Yay!

docs/pages/web/popover.js
```diff
@@ -617,11 +622,13 @@ function PopoverExample() {
       {open && (
         <Layer>
           <Popover
+            accessibilityLabel="Save to board"
             anchor={anchorRef.current}
             id="example-a11y"
             idealDirection="down"
             onDismiss={() => setOpen(false)}
             positionRelativeToAnchor={false}
+            role="listbox"
```
dangerismycat, line 631: nit: above, "For role attribute, use…"? (I can't comment on that line because GH 🙄)
- backticks around "role"
- "the" before "role"
- change "attribute" to "prop"

packages/gestalt/src/Popover.js
dangerismycat: not translated though… 😕

dangerismycat (approved):

packages/gestalt/src/Popover.js
ayeshakmaz: Yeah, once we get the i10n going, we can definitely fix this

PR #2340, opened 2022-08-29
Docs: Navigate to overview page with platform switcher

ayeshakmaz:
#### What changed?

When switching b/w platforms, navigate to the Overview page


dangerismycat (approved):
Possible next improvement: store the previous component the user had been on for a given platform, restore to that

PR #2354, opened 2022-09-01
Dropdown: support conditional children wrapped in Fragment

AlbertCarreras:
#### What changed?

Dropdown: support conditional children wrapped in Fragment #2354

#### Why?

 Dropdown is a composable / modular component which requires using its own subcomponents ( Dropdown.Item, Dropdown.Link, Dropdown.Section ) to build an instance of Dropdown. Subcomponents are passed as children.

The logic to render Dropdown's children can be complex and based on different conditions. Dropdown relies internally in recognizing each Dropdown subcomponent's display name. When rendering conditionally, Dropdown children must be wrapped in React.Fragment if it contains multiple items. In order to support conditional rendering, Dropdown needs to recognize React.Fragment and set its children in the children array, otherwise, it fails to render them.

Furthermore, Dropdown now supports consecutive arrays of arrays of Dropdown subcomponents. Previously, dropdown would not assign an index correctly because it couldn't access nested arrays. Now, Dropdown flattens arrays to allow more complex logic to render children. 

Added example (Subcomponent composability) to exemplify it. https://deploy-preview-2354--gestalt.netlify.app//web/dropdown#Subcomponent-composability

SEE EXAMPLE BELOW TO SEE NOW SUPPORTED CHILDREN COMPOSITIONS

```
      <Dropdown anchor={element} id="ex-1" onDismiss={mockOnDismiss}>
        {renderOptions && (
          <Fragment> >>> FRAGMENT (Dropdown now recognizes them)
            {[1, 2, 3, 4, 5, 6].map((x) => ( >>> ARRAYS OF ARRAYS (Dropdown now flattens them)
              <Dropdown.Item
                key={x}
                onSelect={onSelectMock}
                option={{ value: x.toString(), label: x.toString() }}
              />
            ))}
            {[7, 8, 9, 10, 11, 12].map((x) => (
              <Dropdown.Item
                key={x}
                onSelect={onSelectMock}
                option={{ value: x.toString(), label: x.toString() }}
              />
            ))}
          </Fragment>
        )}
      </Dropdown>,
```

### DIFF
--- a/packages/gestalt/src/Dropdown.js
+++ b/packages/gestalt/src/Dropdown.js
@@ -1,5 +1,5 @@
 // @flow strict
-import { Children, cloneElement, type Node, useState } from 'react';
+import { Fragment, Children, cloneElement, type Node, useState } from 'react';
 import Box from './Box.js';
 import Popover from './Popover.js';
 import Layer from './Layer.js';
@@ -23,13 +23,16 @@ function getChildrenOptions(childrenArray) {
   return childrenArray.reduce((accumulatedChildren, currentChild) => {
     const {
       props: { children: currentItemChildren },
+      type,
       type: { displayName },
     } = currentChild;
 
-    if (currentItemChildren && displayName === 'Dropdown.Section') {
+    if ((currentItemChildren && displayName === 'Dropdown.Section') || type === Fragment) {
       return [
         ...accumulatedChildren,
-        ...(Array.isArray(currentItemChildren) ? currentItemChildren : [currentItemChildren]),
+        ...(Array.isArray(currentItemChildren)
+          ? currentItemChildren.flat()
+          : [currentItemChildren]),
       ];
     }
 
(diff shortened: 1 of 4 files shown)


packages/gestalt/src/Dropdown.js
```diff
@@ -23,13 +23,16 @@ function getChildrenOptions(childrenArray) {
   return childrenArray.reduce((accumulatedChildren, currentChild) => {
     const {
       props: { children: currentItemChildren },
+      type,
       type: { displayName },
     } = currentChild;
 
-    if (currentItemChildren && displayName === 'Dropdown.Section') {
+    if ((currentItemChildren && displayName === 'Dropdown.Section') || typeof type === 'symbol') {
```
dangerismycat, line 30: This isn't really readable. Consider `child.type === Fragment`, or even better creating a helper function `isFragment` that encapsulates that logic

dangerismycat (approved):

packages/gestalt/src/Dropdown.js
AlbertCarreras: DONE. Thanks!

PR #2356, opened 2022-09-01
Internal: suggest index.js pattern for exporting Sandpack examples

AlbertCarreras:
Internal: suggest index.js pattern for exporting Sandpack examples

As per @rlingineni 's suggestion. To discuss. Argument for this change: it cleans up the component documentation file reducing the amount of imports. Some arguments against can be the difficulty to identify dead code if some examples are unused in the Docs.


rlingineni (approved):
Looks good.

Little naming nit: I know the term `Example` is repeated. So it's `Example.BlahExample` maybe we don't need the "example" suffix at the end.  

I know renaming can be tedious, so we can also just do it for the future ones probably.

AlbertCarreras:
I like the idea.

PR #2365, opened 2022-09-06
ComboBox: implement Sandpack code editors

AlbertCarreras:
ComboBox: implement Sandpack code editors

### DIFF
--- a/docs/examples/combobox/controlled.js
+++ b/docs/examples/combobox/controlled.js
@@ -1,0 +1,137 @@
+// @flow strict
+import { useState, type Node } from 'react';
+import { Box, ComboBox, Flex, Text } from 'gestalt';
+
+const US_STATES = [
+  'AK - Alaska',
+  'AL - Alabama',
+  'AR - Arkansas',
+  'AS - American Samoa',
+  'AZ - Arizona',
+  'CA - California',
+  'CO - Colorado',
+  'CT - Connecticut',
+  'DC - District of Columbia',
+  'DE - Delaware',
+  'FL - Florida',
+  'GA - Georgia',
+  'GU - Guam',
+  'HI - Hawaii',
+  'IA - Iowa',
+  'ID - Idaho',
+  'IL - Illinois',
+  'IN - Indiana',
+  'KS - Kansas',
+  'KY - Kentucky',
+  'LA - Louisiana',
+  'MA - Massachusetts',
+  'MD - Maryland',
+  'ME - Maine',
+  'MI - Michigan',
+  'MN - Minnesota',
+  'MO - Missouri',
+  'MS - Mississippi',
+  'MT - Montana',
+  'NC - North Carolina',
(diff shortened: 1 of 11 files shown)


docs/examples/combobox/controlled.js
```diff
@@ -0,0 +1,137 @@
+// @flow strict
+import { useState, type Node } from 'react';
+import { Box, ComboBox, Flex, Text } from 'gestalt';
+
+export default function Example(): Node {
+  const US_STATES = [
```
dangerismycat, line 6: nit: these sort of things shouldn't be defined in components, as they'll be recreated on each render. not important here but a good thing to keep in mind

docs/pages/web/combobox.js
```diff
@@ -6,59 +6,26 @@ import docgen, { type DocGen } from '../../docs-components/docgen.js';
...
-    've / ver',
-    'xe / xem',
-    'xie / xem',
-    'zie / zem',
-  ];
-
-  const options = PRONOUNS.map((pronoun, index) => ({ label: pronoun, value: 'value'+index }));
+    <Page title={generatedDocGen?.displayName}>
+      <PageHeader name={generatedDocGen?.displayName} description={generatedDocGen?.description}>
+        <SandpackExample code={main} name="No image source" hideEditor previewHeight={320} />
```
dangerismycat, line 26: nit: I don't understand how this name relates to the code example?

docs/pages/web/combobox.js
```diff
@@ -246,273 +110,23 @@ function ComboBoxExample(props) {
...
-        onClear={resetErrorMessage}
-        options={options}
-        placeholder="Add your pronouns"
-      />
-    </Box>
-  );
-}
-`}
+            sandpackExample={
+              <SandpackExample code={uncontrolled} name="uncontrolled" previewHeight={320} />
```
dangerismycat, line 114: nit: `name` is user-facing (it populates the accessibility label on an icon button) so we should use Sentence case

docs/pages/web/combobox.js
dangerismycat, line 26: nit: if we're using the same height for all the code examples, that should be a top-of-file const

dangerismycat (approved):

docs/examples/combobox/controlled.js
AlbertCarreras: DONE

docs/pages/web/combobox.js
AlbertCarreras: Didn't know they were a11y labels.

FIXED DONE

docs/pages/web/combobox.js
AlbertCarreras: DONE

docs/pages/web/combobox.js
AlbertCarreras: DONE

PR #2366, opened 2022-09-07
[TBDiscussed] I18nProvider, Video, SideNavigation: refactor I18nProvider to allow multiple components and implement default values to Video

AlbertCarreras:
#### What changed?

I18nProvider, Video: refactor I18nProvider to allow multiple components and implement default values to Video

#### Why?

I18nProvider had a highly complex logic to guarantee all components always had values set on the set I18nProvider in the app. I18nProvider had default key-values and enforced via Flow type and errors the correct and complete set of key-values in the app provider. 

The previous Flow type would not allow multiple component as it was getting confused about the missing props as it was expecting the same key-values in each component. To avoid this, we've simplified the Flow type to a generic type. This could be improve in the future with the implementation of a better Flow type implementation. I couldn't figure out a way to resolve the Flow error so implemented the basic Flow type that enables the logic to work properly.

The  useI18nContext also implemented two points of type enforcement throwing errors. This was happening under two circumstances: no initialContext and missing key-values for a particular component. These errors can be useful to support production code when translations are required, however, for all other apps/cases where translations aren't need it, it was enforcing a duplication of code.

With these changes, we get I18nProvider to work with multiple components guaranteeing a backup default value for those key-values translations not set in the app.

We don't use [/DocsI18nProvider.js](https://github.com/pinterest/gestalt/pull/2366/files#diff-932209fd686d1f3b61f3686bd6ef7264adb8ced9abbe4911f3cec6a84d886d03) anymore as we rely on the default values in the Docs. This approach would also be valid for other platforms such as internal tools

### Links

- [TDD](link to Paper doc)
- [Figma](link to Figma file)

### Checklist


packages/gestalt/src/contexts/I18nProvider.js
```diff
@@ -3,85 +3,80 @@ import { type Context, createContext, useContext } from 'react';
...
 
+  // For those nullish and missing key-values, complete them with the default values
   if (missingAndNullTranslations.length > 0) {
-    const multipleMissing = missingAndNullTranslations.length > 1;
-    const isNotProd = process.env.NODE_ENV !== 'production';
-
-    if (isNotProd) {
-      // Ideally we would throw an actual error here (at least in dev), but this wasn't working with the mocks in Pinboard
-      // eslint-disable-next-line no-console
-      console.error(
```
dangerismycat, line 75: The problem with this is that we're not communicating to the user that translations are missing, and our default strings aren't translated. So we would be showing untranslated strings in the app with no indication that anything is wrong. I never wanted to provide default translations in the first place, but at least the previous logic would alert the user to the issue.

dangerismycat (changes requested):
We need to keep iterating on this solution — we can't use untranslated defaults and also not communicate to users that nothing is wrong.

PR #2370, opened 2022-09-08
Docs: Add summary option for What's New page

ayeshakmaz:
#### What changed?

Add ability to show a summary for each What's New section

Screen Shot 2022-09-06 at 11 15 44 AM

#### Why?

Give readers a high-level overview of the content

### DIFF
--- a/docs/pages/BlogPosts.json
+++ b/docs/pages/BlogPosts.json
@@ -3,6 +3,7 @@
   "digests": [
     {
       "week": "September 02, 2022",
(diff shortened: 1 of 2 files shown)


dangerismycat (approved):
Feel free to merge this as-is, but what if we also showed the summary in the right side nav? Currently those links aren't all that useful, since users aren't likely to be looking by date, but by topic. Another idea: instead of a summary, we could show each post `title` in the right side nav?

PR #2371, opened 2022-09-08
Internal: implement ua-parser-js to detect device type in docs site

AlbertCarreras:
Docs: implement ua-parser-js to detect device in our docs site

To test changes, use https://deploy-preview-2371--gestalt.netlify.app/ in your mobile browser


dangerismycat (approved):
Thanks for including the Flow types! 🙂

PR #2372, opened 2022-09-08
SideNavigation: conditional rendering improvements

AlbertCarreras:
#### What changed?

SideNavigation: logic cleanup and conditional rendering improvements

#### Why?

Cleanup some logic to explicitly support React.Fragment as well as other conditional rendering of subcomponents.

### DIFF
--- a/packages/gestalt/src/SideNavigationSection.js
+++ b/packages/gestalt/src/SideNavigationSection.js
@@ -2,9 +2,9 @@
 import { type Node } from 'react';
 import classnames from 'classnames';
 import styles from './SideNavigation.css';
-import boxStyles from './boxWhitespace.css';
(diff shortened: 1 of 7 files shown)


packages/gestalt/src/SideNavigationSection.js
```diff
@@ -15,24 +15,24 @@ type Props = {|
    * Label for the section. See the [Sections](https://gestalt.pinterest.systems/web/sidenavigation#Sections) variant for more info.
    */
   label: string,
-  /**
-   * Private prop.
-   */
-  _hasMarginTop?: string,
```
dangerismycat, line 21: 💯

dangerismycat (approved):

PR #2373, opened 2022-09-09
Tabs: clarify indicator usage in docs

dangerismycat:
I heard feedback recently that some folks aren't clear about the `indicator` field on the `tabs` prop — including some who thought the current tab would have the red dot! This PR removes the indicator from most examples, and adds an example to specifically call out its usage.

### DIFF
--- a/docs/pages/web/tabs.js
+++ b/docs/pages/web/tabs.js
@@ -12,9 +12,9 @@ import AccessibilitySection from '../../docs-components/AccessibilitySection.js'
 
 export default function DocsPage({ generatedDocGen }: {| generatedDocGen: DocGen |}): Node {
   return (
-    <Page title="Tabs">
+    <Page title={generatedDocGen?.displayName}>
       <PageHeader
-        name="Tabs"
+        name={generatedDocGen?.displayName}
         description={generatedDocGen?.description}
         defaultCode={`
 function DefaultExample() {
@@ -25,8 +25,8 @@ function DefaultExample() {
       activeTabIndex={activeIndex}
       onChange={({ activeTabIndex }) => { setActiveIndex(activeTabIndex); }}
       tabs={[
-        { href: "#", text: "Explore", indicator: "dot" },
-        { href: "#", text: "Shop" },
+        { href: "#", text: "Explore" },
+        { href: "#", text: "Shop", indicator: "dot" },
         { href: "#", text: "Profiles" },
       ]}
       wrap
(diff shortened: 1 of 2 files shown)


docs/pages/web/tabs.js
```diff
@@ -25,8 +25,8 @@ function DefaultExample() {
       activeTabIndex={activeIndex}
       onChange={({ activeTabIndex }) => { setActiveIndex(activeTabIndex); }}
       tabs={[
-        { href: "#", text: "Explore", indicator: "dot" },
-        { href: "#", text: "Shop" },
+        { href: "#", text: "Explore" },
+        { href: "#", text: "Shop", indicator: "dot" },
```
dangerismycat, line 29: We still want to show what the component is capable of in this top example, but I want to make it more clear that the dot and underline are not indicating the same thing

docs/pages/web/tabs.js
```diff
@@ -343,15 +376,15 @@ function TabExample() {
           defaultCode={`
 function TabExample() {
   const [activeIndex, setActiveIndex] = React.useState(0);
-  const [isTransparent, setIsTransparent] = React.useState(true);
+  const [isTransparent, setIsTransparent] = React.useState(false);
```
dangerismycat, line 379: Somewhat unrelated — I think it's more effective to show the problem first, then allow the user to "solve" the problem by interacting with the switch. This is how we've handled the Wrapping example and it seems better to me.

packages/gestalt/src/Tabs.js
```diff
@@ -204,7 +204,7 @@ type Props = {|
    */
   onChange: OnChangeHandler,
   /**
-   *
+   * The array of tabs to be displayed. The active tab (as indicated by `activeTabIndex`) will be underlined. Use the [optional `indicator` field](https://gestalt.pinterest.systems/web/tabs#Indicator) to show a notification of new items on the tab. Though `text` currently accepts a React.Node, this is deprecated and will be replaced by a simple `string` type soon.
```
dangerismycat, line 207: > Though `text` currently accepts a React.Node, this is deprecated and will be replaced by a simple `string` type soon

I'm looking into this at some point — just trying to prevent additional usages for now

AlbertCarreras: @dangerismycat The underline looks funny here. Could we do a regular See variant X to learn more type of reference?

Screen Shot 2022-09-09 at 2 29 17 PM

AlbertCarreras (approved):

packages/gestalt/src/Tabs.js
dangerismycat: I was trying to avoid that to make the text flow better, but you're right, the underline+code is odd.

PR #2376, opened 2022-09-09
Docs: fix broken links from recent Algolia crawl

dangerismycat:
I noticed a ton of broken/old links from our recent crawl (301/302 and 500 responses). This PR updates those links.


docs/markdown/testing/redbox.md
```diff
@@ -210,18 +210,18 @@ Some interactive example to get the best out of redbox
 
 ## Resources
 
-**[RedboxGroup](/buttongroup)**
+**[RedboxGroup](/web/buttongroup)**
```
ayeshakmaz, line 213: Should we keep these testing files around? Seems like more work to keep them updated

dangerismycat: I'd be fine with deleting them. I think @rlingineni wanted to keep them around for reference but that's what git history is for 🙂

I'll put up another PR to remove them

dangerismycat: here's that PR: https://github.com/pinterest/gestalt/pull/2378

playwright/accessibility/Typography.spec.mjs
```diff
@@ -3,6 +3,6 @@ import { test } from '@playwright/test';
 import expectAccessiblePage from './expectAccessiblePage.mjs';
 
 test('Typography Accessibility check', async ({ page }) => {
-  await page.goto('/web/typography');
```
ayeshakmaz, line 6: Shouldn't this fail? surprised it didn't complain

dangerismycat: Gooooood question

docs/markdown/testing/redbox.md
ayeshakmaz: Thank you!

PR #2383, opened 2022-09-11
Internal: use multiline comments to enable flow docs

christianvuerings:
#### What changed?

Use multiline comments to enable flow docs

#### Why?

Makes the comments visible to where the props are being used, not just where they are defined.

Screen Shot 2022-09-11 at 4 57 00 AM
Screen Shot 2022-09-11 at 4 56 49 AM

### DIFF
--- a/docs/docs-components/PageHeader.js
+++ b/docs/docs-components/PageHeader.js
@@ -18,11 +18,19 @@ const buildSourceLinkUrl = (componentName) =>
 
 type Props = {|
   badge?: 'pilot' | 'deprecated',
-  // DEPRECATED: Use `children` instead of `defaultCode`
+  /**
+   * @deprecated : Use `children` instead of `defaultCode`
+   */
   defaultCode?: string,
   description?: string,
-  fileName?: string, // only use if name !== file name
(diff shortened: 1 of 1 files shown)


docs/docs-components/PageHeader.js
```diff
@@ -18,11 +18,19 @@ const buildSourceLinkUrl = (componentName) =>
 
 type Props = {|
   badge?: 'pilot' | 'deprecated',
-  // DEPRECATED: Use `children` instead of `defaultCode`
+  /**
+   * @deprecated : Use `children` instead of `defaultCode`
```
christianvuerings, line 22: Since flow out of the box doesn't yet support deprecated types, we use JSDoc comments instead:
* https://github.com/facebook/flow/issues/4646
* https://jsdoc.app/tags-deprecated.html

Note that in TypeScript, the JSDoc format is also recommended: https://github.com/microsoft/TypeScript/issues/390

dangerismycat: Neat! 😍

dangerismycat (approved):

PR #2384, opened 2022-09-12
Popover: Add support to handle content that overflows the viewport height

EduhCosta:
#### What changed?

When the popover content is bigger the 90% the viewport it will be repositioned on screen top and use a default padding to render itself.

#### Why?

On our current Dropdown component, when we have a list bigger than screen size our component behaves incorrectly. It overflows the screen and does not show options completely.

### Links

- [Jira](https://jira.pinadmin.com/browse/GESTALT-4787)
- [Developing tracked](https://docs.google.com/document/d/1DYc2xPPWNhhUpYRgbYDOA-qoS5fMUkizPE2nDNIt_5k/edit#)

### Checklist

## Screenshot testing

*Before:*
Screenshot 2022-09-13 at 10 33 05

*After:*
Screenshot 2022-09-13 at 10 40 52


packages/gestalt/src/Contents.js
```diff
@@ -202,6 +204,37 @@ class Contents extends Component<Props, State> {
...
 
+  /**
+   * Identify when is required to repositioning the popover, i.e., the windows popover has 90% of
+   * viewport (available screen height), and set the controller variable as `true`;
+   * When the height of popover is less than 90% the controller variable is `false`;
+   */
+  balancePopoverPosition(): PopoverOverride {
+    // Required because of SSR
+    if (!window || !document) {
+      return {};
```
dangerismycat, line 215: Consider instead returning `null` here, and either using nullish coalescing to an empty object on line 249, or conditionally spreading `overridePropsToMaxPopoverSize` on line 267

packages/gestalt/src/Contents.js
```diff
@@ -202,6 +204,37 @@ class Contents extends Component<Props, State> {
...
+    }
+
+    const viewportAvailable = window.innerHeight;
+    const popoverHeight = document.getElementById(this.props?.id || '')?.clientHeight;
```
dangerismycat, line 219: nit: destructure `id` from `this.props` at the top of the method, and using nullish coalescing (`??`) instead of `||` for default values (here and elsewhere)

packages/gestalt/src/Dropdown.js
```diff
@@ -221,7 +221,14 @@ export default function Dropdown({
       shouldFocus
       size="xl"
     >
-      <Box alignItems="center" direction="column" display="flex" flex="grow" margin={2}>
+      <Box
+        maxHeight="90vh"
```
dangerismycat, line 225: nit: alphabetize these props, please 🙂

dangerismycat (approved):
This looks great. Could you also add before/after screenshots to the PR summary so Future Us know exactly what these changes are addressing?

dangerismycat (approved):

PR #2385, opened 2022-09-13
SlimBanner: Adjust layout in small viewports

ayeshakmaz:
#### What changed?

Move SlimBanner action down to a new line in mobile viewports

#### Why?

Ensure the text is readable in small viewports. Note: This does not yet address the horizontal alignment of the icon and dismiss button in small viewports

Before:

Screen Shot 2022-09-12 at 5 57 55 PM

After:
Screen Shot 2022-09-12 at 5 58 25 PM

### Links

- [Jira](https://jira.pinadmin.com/browse/GESTALT-4605)
- [Figma](https://www.figma.com/file/MInxIhLJSbGDebM0939W5G/Gestalt-SlimBanner-Hand-off?node-id=213%3A1766)

### Checklist

### DIFF
--- a/packages/gestalt/src/SlimBanner.js
+++ b/packages/gestalt/src/SlimBanner.js
@@ -181,6 +181,8 @@ export default function SlimBanner({
       alignItems="center"
       color={isBare ? 'transparent' : backgroundColor}
       display="flex"
+      direction="column"
+      mdDirection="row"
       padding={isBare ? 0 : 4}
       paddingY={isBare ? 1 : 0}
       rounding={4}
@@ -216,9 +218,9 @@ export default function SlimBanner({
           <Flex.Item flex="none">
             <Flex alignItems="center" gap={{ row: 4, column: 0 }}>
               {primaryAction && (
-                <Flex.Item flex="none">
+                <Box display="none" mdDisplay="flex" flex="none">
                   <PrimaryAction {...primaryAction} />
-                </Flex.Item>
+                </Box>
               )}
 
               {dismissButton && <DismissButton {...dismissButton} />}
@@ -226,6 +228,11 @@ export default function SlimBanner({
           </Flex.Item>
         )}
       </Flex>
+      {shouldShowButtons && primaryAction && (
+        <Box display="flex" mdDisplay="none" flex="none" alignSelf="end" marginTop={4}>
+          <PrimaryAction {...primaryAction} />
+        </Box>
+      )}
     </Box>
   );
 }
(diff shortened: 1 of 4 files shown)


packages/gestalt/src/SlimBanner.js
```diff
@@ -216,16 +218,21 @@ export default function SlimBanner({
           <Flex.Item flex="none">
             <Flex alignItems="center" gap={{ row: 4, column: 0 }}>
```
ayeshakmaz, line 219: Do we have anything that can conditionally set alignItems? In the small viewports it needs to be `start`, otherwise `center`

AlbertCarreras:
Can we add a similar variant section like Pageheader? https://gestalt.pinterest.systems/web/pageheader#Responsive

packages/gestalt/src/SlimBanner.js
```diff
@@ -216,16 +218,21 @@ export default function SlimBanner({
...
-                </Flex.Item>
+                </Box>
               )}
 
               {dismissButton && <DismissButton {...dismissButton} />}
             </Flex>
           </Flex.Item>
         )}
       </Flex>
+      {shouldShowButtons && primaryAction && (
```
AlbertCarreras, line 231: isn't "shouldShowButtons" redundant?

isn't just `!isBare` what we need? we double check for && primaryAction 
&& dismissButton before each component

const shouldShowButtons = !isBare && (primaryAction || dismissButton);

packages/gestalt/src/SlimBanner.js
AlbertCarreras: not yet, matchMedia ain't recommended. and we don't have responsive props for aligment in Box.

AlbertCarreras (approved):

packages/gestalt/src/SlimBanner.js
dangerismycat: Not directly. We could look into adding responsive versions of the flex props if that would be useful. For now we could check viewport width in the component, but that would involve multiple sources of truth so not the best 😬

packages/gestalt/src/SlimBanner.js
ayeshakmaz: switched it to !isBare && primaryAction since flow needs the primaryAction check still

packages/gestalt/src/SlimBanner.js
ayeshakmaz: Ok yeah, I'll leave as is for now, and we can look into adding the responsive alignment

PR #2388, opened 2022-09-14
Flex: implement Sandpack code editor examples

AlbertCarreras:
Flex: implement Sandpack code editor examples

### DIFF
--- a/docs/examples/flex/flexItem.js
+++ b/docs/examples/flex/flexItem.js
@@ -1,0 +1,19 @@
+// @flow strict
+import { type Node } from 'react';
+import { Flex, Box, Button } from 'gestalt';
+
+export default function Example(): Node {
+  return (
+    <Flex width="100%" height="100%" justifyContent="center" alignItems="center">
+      <Box paddingX={2} paddingY={2} width="100%">
+        <Flex alignItems="center" gap={4}>
+          <Button text="Button 1" />
+          <Flex.Item flex="grow">
+            <Button text="Button 2" />
+          </Flex.Item>
+          <Button text="Button 3" />
+        </Flex>
+      </Box>
+    </Flex>
+  );
+}
--- a/docs/examples/flex/overflowing.js
+++ b/docs/examples/flex/overflowing.js
@@ -1,0 +1,23 @@
+// @flow strict
+import { type Node } from 'react';
+import { Flex, Box, Text, Icon, Badge } from 'gestalt';
+
+export default function Example(): Node {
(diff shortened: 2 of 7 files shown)


docs/examples/flex/flexItem.js
```diff
@@ -0,0 +1,19 @@
+// @flow strict
+import { type Node } from 'react';
+import { Flex, Box, Button } from 'gestalt';
+
+export default function Example(): Node {
+  return (
+    <Flex width="100%" height="100%" justifyContent="center" alignItems="center">
+      <Box borderStyle="sm" paddingX={2} paddingY={3} rounding={3} width="100%">
```
ayeshakmaz, line 8: The layout on this one is a bit funky now, maybe remove the border on this Box

docs/examples/flex/overflowing.js
```diff
@@ -0,0 +1,25 @@
+// @flow strict
+import { type Node } from 'react';
+import { Flex, Box, Text, Icon, Badge } from 'gestalt';
+
+export default function Example(): Node {
+  return (
+    <Flex width="100%" height="100%" justifyContent="center" alignItems="center">
+      <Box borderStyle="sm" padding={3} rounding={3}>
```
ayeshakmaz, line 8: Same here, can remove the border

docs/examples/flex/overflowing.js
```diff
@@ -0,0 +1,25 @@
...
+        <Flex alignItems="center" gap={3}>
+          <Icon accessibilityLabel="Private" icon="lock" />
+
+          <Flex.Item minWidth={0}>
+            <Text>
```
ayeshakmaz, line 13: We mention truncation in the description of this section. Should the Text have lineClamp={1} to show that? I don't see any difference in this example when the minWidth={0} is removed. 

Doesn't have to be fixed in this PR, just thought I'd bring it up while we're here

ayeshakmaz (approved):
Some comments, but looks good otherwise!

docs/examples/flex/flexItem.js
AlbertCarreras: DONE

docs/examples/flex/overflowing.js
AlbertCarreras: DONE

docs/examples/flex/overflowing.js
AlbertCarreras: DONE

PR #2403, opened 2022-09-19
Checkbox, RadioButton, RadioGroupButton: start align input

AlbertCarreras:
#### What changed?

Checkbox, RadioButton, RadioGroupButton: start align input

##### BEFORE
Screen Shot 2022-09-20 at 1 06 12 AM
Screen Shot 2022-09-20 at 1 06 34 AM

##### AFTER
Screen Shot 2022-09-20 at 1 05 45 AM
Screen Shot 2022-09-20 at 1 07 06 AM

#### Why?

Checkbox and RadioGroup controls currently are vertically center-aligned with the label (see attached image)
The Checkbox and RadioGroup controls (e.g., the actual checkbox / radio inputs) should be top aligned with the label. So that if/when the label wraps, the position of the checkbox / radio should not change relative to the first line of the label.

### Links

- [Jira](https://jira.pinadmin.com/browse/GESTALT-4272)

### DIFF
--- a/packages/gestalt/src/Checkbox.js
+++ b/packages/gestalt/src/Checkbox.js
@@ -156,13 +156,7 @@ const CheckboxWithForwardRef: React$AbstractComponent<Props, HTMLInputElement> =
 
   return (
     <Box>
-      <Box
-        alignItems={helperText || errorMessage || image ? 'start' : 'center'}
-        display="flex"
-        justifyContent="start"
-        marginStart={-1}
-        marginEnd={-1}
-      >
+      <Box alignItems="start" display="flex" justifyContent="start" marginStart={-1} marginEnd={-1}>
         <Label htmlFor={id}>
           <Box paddingX={1} position="relative">
             <input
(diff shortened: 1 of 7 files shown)


packages/gestalt/src/Checkbox.js
```diff
@@ -207,7 +201,10 @@ const CheckboxWithForwardRef: React$AbstractComponent<Props, HTMLInputElement> =
         </Label>
         {Boolean(image) && <Box paddingX={1}>{image}</Box>}
         {label && (
-          <Box display={labelDisplay === 'hidden' ? 'visuallyHidden' : 'block'}>
+          <Box
+            display={labelDisplay === 'hidden' ? 'visuallyHidden' : 'block'}
+            dangerouslySetInlineStyle={{ __style: { marginTop: '2px' } }}
```
ayeshakmaz, line 206: why is the 2px needed?

ayeshakmaz (commented):
If the 2px are needed from design, can you add a comment as to why?

packages/gestalt/src/Checkbox.js
AlbertCarreras: @ayeshakmaz 

Without the 2 pixels looks unbalanced. @ponori Thoughts on these 2 pxs?

NO pixels
Screen Shot 2022-09-20 at 1 07 06 AM

WITH 2 MARGIN TOP PIXELS
image

ponori: Yeah, I see what you're saying Alberto... Ayesha, my guess is that this is problematic b/c the spacing doesn't translate to a spacing token. Would be interesting to hear if there was a way to have our cake (proper alignment) and eat it too (utilize tokens).

dangerismycat: Does this need to be a token? We use little tweaks like this all over the place where mixing elements of different types (icons + text, checkbox + text, etc). I'm personally fine with these nudges as long as we make note of why the additional margin is included in a code comment

AlbertCarreras: COMMENT ADDED.
DONE

ayeshakmaz: Yeah agreed, don't think it needs to be a token, but definitely needs the comment. Thanks Alberto!

ayeshakmaz (approved):

PR #2392, opened 2022-09-16
RadioGroupButton: standardize helper message style with other form/control fields, renamed subtext

AlbertCarreras:
#### What changed?

RadioGroupButton: standardize helper message style with other form/control fields, renamed subtext

#### BEFORE

Screen Shot 2022-09-16 at 5 49 16 PM

##### AFTER

Screen Shot 2022-09-16 at 5 49 59 PM

#### Why?

Form controls: Textfield, TextArea, have the same helperText / errorMessage prop and style. Refactoring RadioGroup.Button to match style and API of all form fields.

####  Breaking changes codemod

1st Run 

`yarn codemod modifyProp ~/code/pinboard/webapp
--component=RadioGroup
--subcomponent=RadioButton
--previousProp=subtext
--nextProp=helperText
 `

### Links

- [Jira](https://jira.pinadmin.com/browse/GESTALT-4843)

PR #2391, opened 2022-09-16
Checkbox: standardize helper/error message style with other form/control fields, renamed subtext, deprecated hasError

AlbertCarreras:
#### What changed?

Checkbox: standardize helper/error message style with other form/control fields, renamed subtext, deprecated hasError

##### BEFORE
Screen Shot 2022-09-16 at 4 22 42 PM
Screen Shot 2022-09-16 at 4 22 36 PM

##### AFTER
Screen Shot 2022-09-16 at 4 22 49 PM
Screen Shot 2022-09-16 at 4 22 22 PM

#### Why?

Form controls: Textfield, TextArea, have the same helperText / errorMessage prop and style. Refactoring Checkbox to match style and API of all form fields.

####  Breaking changes codemod

1st Run 

`yarn codemod modifyProp ~/code/pinboard/webapp
--component=Checkbox
--previousProp=subtext
--nextProp=helperText`

2nd Run 

#### IMPORTANT: On this second codemod run, review each change to make sure the hasError ain't a critical prop to the proper behavior of that Checkbox instance. Make adjustments so not to break the code.

`yarn codemod modifyProp ~/code/pinboard/webapp --component=Checkbox --previousProp=hasError
 `

### Links

- [Jira](https://jira.pinadmin.com/browse/GESTALT-4843)

PR #2406, opened 2022-09-20
Checkbox, RadioButton: Implementing a11y pause

AlbertCarreras:
### What

Checkbox, RadioButton: Implementing a11y pause

### Why

It was removed in previous PR  https://github.com/pinterest/gestalt/pull/2392 & https://github.com/pinterest/gestalt/pull/2391

Reimplementing + adding pause in ErrorMessage for Checkbox as well.

Screen Shot 2022-09-20 at 2 22 28 AM
Screen Shot 2022-09-20 at 2 36 44 AM

### DIFF
--- a/packages/gestalt/src/FormHelperText.js
+++ b/packages/gestalt/src/FormHelperText.js
@@ -4,10 +4,16 @@ import { type Node } from 'react';
 import Box from './Box.js';
 import Text from './Text.js';
 
-export default function FormHelperText({ text }: {| text: string |}): Node {
+type Props = {|
+  text: string,
+  addA11yPause?: boolean,
+|};
+
+export default function FormHelperText({ text, addA11yPause = false }: Props): Node {
   return (
     <Box marginTop={2}>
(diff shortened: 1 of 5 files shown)


packages/gestalt/src/FormHelperText.js
```diff
@@ -4,10 +4,16 @@ import { type Node } from 'react';
 import Box from './Box.js';
 import Text from './Text.js';
 
-export default function FormHelperText({ text }: {| text: string |}): Node {
+type Props = {|
+  text: string,
+  addA11yPause?: boolean,
+|};
+
+export default function FormHelperText({ text, addA11yPause = false }: Props): Node {
```
ayeshakmaz, line 12: For the future: maybe we should combine these two, to just FormHelperText, with either a default or error variant

ayeshakmaz: Since they seem so repetitive, and we only show one at a time

ayeshakmaz (approved):

PR #2407, opened 2022-09-20
DefaultLabelProvider: replace I18nProvider with more flexible component

dangerismycat:
## The Actual Changes

From conversations with the team, I realized that I18nProvider was solving the wrong problem: the issue isn't providing _translated_ strings for common use cases, but providing any strings for those usages. Translation is an implementation detail, which some apps support and some don't. From further conversation with PJ, I realized that these strings won't always be for accessibility labels, but could potentially be for visual labels as well. Thus a rename of this component to `DefaultLabelProvider`, since it's used to provide default labels. 😅 

Since translations are no longer a requirement for this solution, that allows for more lax Flow typing of the Provider value. Namely it means that the value can be omitted completely, if the app owners are fine with using our default English strings for the supported labels. If the value is passed, we are still fairly strict with the Flow type, actually improving the DX by alerting implementors to any missing strings (which would otherwise only be noticed by using the app in a different language and using tools to see the accessibility labels).

Screen Shot 2022-10-06 at 5 43 13 PM

***

I'll leave the original summary of a previous version of this PR below as notes about the Flow exploration I did:

### Flow exploration (only somewhat relevant now)

We need to use I18nProvider for more components, but Flow complains about the union types created when adding additional components. This is ultimately a limit of static analysis: while we provide a generic type `C` for the component name provided to `useI18nContext`, Flow cannot know what the runtime component will be statically. Therefore any references to `propsTranslations[componentName]` must be a union of all possible `I18nContextType` key fields, and iterating through them (e.g. the `.filter` call) while attempting to access `componentTranslations` properties must accommodate all possible keys, not just those for the current component.

Consider this code:
```js
const translationPropNames = Object.keys(componentTranslations);
const nullTranslations = translationPropNames.filter(
  (propName) => typeof componentTranslations[propName] !== 'string',
);
```
Flow complains because `propName` is inferred to be `"accessibilityClearButtonLabel" | "accessibilityHidePasswordLabel" | "accessibilityShowPasswordLabel"`, and while Flow doesn't know which component object we'll be looking at, it knows that no object will have all of those fields.

One possible solution is to make the data structure a bit weaker:
```js
type I18nContextType = {|
  ComboBox: {|
    accessibilityClearButtonLabel: ?string,
    [string]: ?string,
  |},
  TextField: {|
    accessibilityHidePasswordLabel: ?string,
    accessibilityShowPasswordLabel: ?string,
    [string]: ?string,
  |},
|};
```
This isn't great. While any extra fields will be ignored within components, DX suffers as we're no longer able to let the user know if they have a typo or are providing a field that isn't yet supported.

Another solution is to cast the type of `propName`:
```js
const translationPropNames = Object.keys(componentTranslations);
const nullTranslations = translationPropNames.filter(
  (propName: string) => typeof componentTranslations[propName] !== 'string',
);
```
This is fine. We're telling Flow not to try too hard as `propName` can be any string. This seems a bit unsafe, and theoretically it is — but [this is inherent in how Flow treats objects as maps](https://flow.org/en/docs/types/objects/#toc-objects-as-maps):
> When an object type has an indexer property, property accesses are assumed to have the annotated type, even if the object does not have a value in that slot at runtime. It is the programmer’s responsibility to ensure the access is safe, as with arrays.

However, these checks were a bit belt-and-suspenders anyway — Flow already provides the safety we need here, in that attempting to use I18nProvider without providing translations will already throw a Flow error. As with the previous approach, we can improve this DX in dev by adding a more explicit error message (also checking for empty strings) and letting the dev know which component is missing translations. (We can no longer say which exact translations are missing, but this isn't important.)

Of note is that we're no longer providing default (English) strings. I never wanted to include those in the first place, but couldn't figure out the proper typing with my initial Pinboard implementation, so had to tack them on. Using English default strings too easily covers up missing translations and buries those errors.

[useful Flow sandbox link](https://flow.org/try/#0PQKgBAAgZgNg9gdzCYAoAlgWwA5wE4AuYA3mAQJ7YCmYAwnAHYFUAeBANGAMZ5UCGzek1YcwAVwDOVIczZgAvmCh44mMAHJefLgXUBuVKgrUwASQCMADgYyRAFUo0AvCQA+qMHVUAjOACE4FgAuNw9PMG0uKgkJdG90GHQKWhh+PD8xAgJGABk+byoYEIB+CQI8dAYAc3Yw13lazzsRADF0QoATEOJ3cIiuKJi4hKTyAAl0DqoABT4YhHwOvIKisFLyypqwz0jo2PjEigBlAAtEWfnF5cKSsorqxrB62vqDVC5GMrBKpPQ+GFschcxDC9EwvgCwRI236g32I2SqT46Uy2QY11WDDEMBgjwaYWabDanW6MN2QwOowmUwuEgWeCW+RuYCxOMeOwGe2Gh3Ip3OczpVyZmOxuLC+Pkbw+DC+UygfGxBDseD4MpgAnQnxCFmsgKVjjAwNBPn8gVJfXJ8J5KTSGSyuWFIXUNuR3wY2Ey6jxj0JBGJMC60ItnIpCPGkxmAvpjJWTupNGwUcWXrJIatoz5CFp0YxTszYETlwZKc8EqlnyIOpsjFkBBCeoAPFW9Q5qAA+Q3cLSCGsiJtWavCNitqhtgAUPwIfwBvbYAEoDNKvlXpioAG4RvCd5uzggAOlXcA3UzwBlYuEIYDlCpglYHh+PVFPhmMNAAav9JmDcAwqExOwAJAA0lQ5ASP2uq7iObZvFAYgMDomoMOIUg7kOBANrQIQfokHTfowf4EOOYQfDgBFMAAcnwmBUPWtRziEAEAKKpDRTAjhBg61iOnC0B2IKeEuRDYCo2ASMqqoSOqU6fJ2kjSLuY5obWC4kRW3CqD+hESWqGqyS4IlwGJOlSXpMoANqkVplHUVQAC6byCep5SSdJSGHtgVE0RInYAPLeAAVlQOh7gA1qBEhjlZ5FKiqukyTKqmeMAwBgCZbmyQQJwCBEvBgGUCQwCyoroBIJxgGOqQxGAiThTA5BzmpMpEKyMDpWZPkuC58XuaJXnRHuUAJMweBjjCY6GZ5tkhHcmxzoaHavnAUAaWRv7sXFpkJRI5mTf1dlgAAhE4LjqLN1QlmAqlhLwBBiHgyHRetsWuR1BjyIYQlkJtGUynJqEDnqY7qL6-odOoc5AA)


packages/gestalt/src/contexts/I18nProvider.js
```diff
@@ -9,27 +9,28 @@ import { type Context, createContext, useContext } from 'react';
  * - Add default translations to the mock file (./__mocks__/I18nProvider.js) and docs/components/contexts/DocsExperimentProvider.js
  */
 
+type nullOrString = null | string;
```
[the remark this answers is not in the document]
jackhsu978: nit: I think flow type alias should be PascalCase. Can you change this to `NullOrString`?

docs/examples/defaultlabelprovider/no-translations.js
```diff
@@ -0,0 +1,60 @@
+// @flow strict
+import { type Node, useState } from 'react';
```
dangerismycat, line 2: Note that we no longer need to import `React` — we can import only what's needed in the file, same as in all our other files

docs/pages/web/utilities/defaultlabelprovider.js
```diff
@@ -0,0 +1,50 @@
...
...estalt components have many required accessibility labels. This allows you to pass translated strings to ensure that users of all abilities and languages can use your app.\n\nHowever, certain labels are unlikely to change based on the specific implementation, e.g. the accessibility label for the close button on a modal. For supported labels on certain components, you can specify the label once, at the top level of your app, for use in all instances of that component throughout your app. This allows for certain accessibility label props to now be optional; the default label will be used unless overridden at the callsite.\n\nIf your app supports internationalization, pass translated strings to the Provider. If your app does not support internationalization, you may pass untranslated strings, or you may omit the \`labels\` prop or pass \`null\` to use our default labels. See [the source code](https://github.com/pinterest/gestalt/blob/master/packages/gestalt/src/contexts/DefaultLabelProvider.js) for the default English labels we provide.\n\nIn the examples below, type a character to make the "clear input" icon button appear, then inspect that element and note the accessibility label.`}
```
dangerismycat, line 21: For some reason using actual line breaks to break up the text was causing everything after the first paragraph to render as a code block, hence the `\n` instead 🤔

packages/gestalt/src/ComboBox.js
```diff
@@ -40,7 +41,7 @@ type Props = {|
   /**
    * Label to describe the clear button's purpose.
    */
-  accessibilityClearButtonLabel: string,
+  accessibilityClearButtonLabel?: string,
```
dangerismycat, line 44: Now optional since it's part of DefaultLabelProvider

packages/gestalt/src/contexts/DefaultLabelProvider.js
```diff
@@ -0,0 +1,66 @@
...
+type Props = {|
+  children: Node,
+  /**
+   * An object describing the default strings to be used by supported component labels throughout your app. If your app supports i18n, don't forget to translate your strings!
+   *
+   * Note that all supported labels for all supported components must be provided if using this Provider. Omit this prop to use default (English) strings for supported labels.
+   *
+   * See [the source code](https://github.com/pinterest/gestalt/blob/master/packages/gestalt/src/contexts/DefaultLabelProvider.js) for the specific shape of this object.
+   */
+  labels?: null | DefaultLabelContextType,
```
dangerismycat, line 43: Strangely, when using `?DefaultLabelContextType` here, the type is shown in full on the docs page. Using `null | DefaultLabelContextType` just shows that literal union. No idea why. 🤔 

This is actually an advantage here, though. This type is already complicated and will be very gnarly in the future, and the lack of formatting makes it look like crap on the docs pages. Given that this Provider is more of an "admin" feature, I'm fine with pointing people to the source code to see the actual object shape. It also means we maintain a single source of truth, which is very important with a living object like this.

packages/gestalt/src/contexts/I18nProvider.jsdom.test.js
```diff
@@ -1,89 +0,0 @@
...
-        <TestComponent />
-      </I18nProvider>,
-    );
-
-    // This is a bit roundabout — we don't really care that these strings are in the document, but that they were returned from the Hook correctly
-    expect(screen.getByText(/Hide password/)).toBeInTheDocument();
-    expect(screen.getByText(/Show password/)).toBeInTheDocument();
-  });
-
-  it('throws on unsupported component', () => {
```
dangerismycat, line 34: We no longer throw on unsupported components — Flow takes care of this check for us

docs/pages/web/utilities/defaultlabelprovider.js
```diff
@@ -0,0 +1,50 @@
...
+      >
+        <MainSection.Card
+          sandpackExample={
+            <SandpackExample code={translations} name="Translations example" previewHeight={320} />
+          }
+        />
+        <MainSection.Card
```
ayeshakmaz, line 28: May be helpful to add a title to these cards to reiterate "With translations" and "Without translations" for folks who don't read the summary

docs/pages/web/utilities/defaultlabelprovider.js
dangerismycat: ahhh thank you 🙏

docs/examples/defaultlabelprovider/translations.js
```diff
@@ -0,0 +1,71 @@
+// @flow strict
+import { type Node, useState } from 'react';
+import { Box, ComboBox, DefaultLabelProvider, Flex, Heading } from 'gestalt';
+
+const myI18nTranslator = (val) => val.toUpperCase();
+const labels = {
+  ComboBox: {
+    accessibilityClearButtonLabel: myI18nTranslator('Clear input'),
+  },
+  TextField: {
```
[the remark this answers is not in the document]
dangerismycat: We can't remove it — the object is all or nothing, either you provide strings for every label or you provide none. That's not great if your app only uses a subset of our components, but Pinboard is our primary focus and all labels are needed there. If we type this more loosely, then there isn't feedback that labels for a given component are missing.

I suspect our users will fall primarily into two buckets, either needing every label (Pinboard) or not needing any (internal tools). 

As to adding TextField: this object will eventually grow to include most (all?) components, so we don't need to show them all here. (Also, for TextField specifically: these labels are hidden behind an experiment for now.) ComboBox isn't the best example, so we can swap this out for an example that's easier to see when we implement such a label.

packages/gestalt/src/ComboBox.js
```diff
@@ -433,7 +436,9 @@ const ComboBoxWithForwardRef: React$AbstractComponent<Props, HTMLInputElement> =
           iconButton={
             controlledInputValue || textfieldInput || (tags && tags.length > 0) ? (
               <InternalTextFieldIconButton
-                accessibilityLabel={accessibilityClearButtonLabel}
+                accessibilityLabel={
+                  accessibilityClearButtonLabelProp ?? accessibilityClearButtonLabel
```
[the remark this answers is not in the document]
dangerismycat: Sure, renaming the default is probably more clear

docs/pages/web/utilities/defaultlabelprovider.js
dangerismycat: There are labels in the examples themselves 🙂

ayeshakmaz (approved):

PR #2408, opened 2022-09-21
Textfield, TextArea: add maxLength prop

AlbertCarreras:
#### What changed?

Textfield, TextArea: add maxLength prop

#### Why?
Screen Shot 2022-09-22 at 7 24 39 PM

Screen Shot 2022-09-22 at 7 23 21 PM

Screen Shot 2022-09-22 at 7 24 07 PM

Kapture 2022-09-21 at 23 28 13

As a user, I want to have a visual affordance for how many characters I can add to a TextArea so I can effectively write for the character limit.

Acceptance criteria

- The TextArea and TextField components should have a prop to enforce a maximum character limit.
maximum character limits should not be enforced unless the maxchars prop (or whatever name we give it) is explicitly set.
- The prop should only accept a positive integer.
When the prop is set with a valid number, a counter affordance should be displayed on the right side of TextField/TextArea below the input element (see Figma design link above).
- The counter should update whenever the number of characters within the input changes.

When helperText is displayed:
- The counter's width should be set by calculating the text set to the max character count and then adding 4 additional pixels to account for flex between variable number widths.
- Space should always be set aside for the Status error icon displayed in the error state so that helperText doesn't shift when the icon is made visible.

The following logic should control the counter's text:
- By default, the counter's text color should be set to "subtle"
- When the character count equal to 100% of the maximum characters, the text color should be set to "warning" and the warning Status icon displayed.
- When the maximum character limit has been hit, the input should not allow any more characters to be added.

### Links

- [Figma](https://www.figma.com/file/qi2yCevlp9ZZPYndRUl6bz/Input-maxchars-treatment?node-id=1543%3A3223)
- [Jira](https://jira.pinadmin.com/browse/GESTALT-4766)

### Checklist

### DIFF
--- a/docs/pages/web/textarea.js
+++ b/docs/pages/web/textarea.js
@@ -462,19 +462,18 @@ function Example(props) {
           title="Maximum length"
           description={`TextArea supports the native [maxlength](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/maxlength) input attribute. \`maxLength\` sets the maximum number of characters allowed to be entered by the user in TextArea. \`maxLength\` must be an integer value 0 or higher.
 
-When \`maxLength\` is passed to TextArea, the component displays a character counter as well as an [error Status](/status) when the user reaches or the controlled value exceeds the maximum length of characters.
+The user cannot exceed the maximum number of characters interacting with the component. Whenever possible, avoid setting initial values from the parent component's state that already exceed the \`maxLength\`.
 
-For controlled TextAreas, if the initial value in the external state exceeds the \`maxLength\` value, TextArea will display the full value; however, the user won't be able to add additional characters and they will only be allowed to delete characters. This case should be prevented.
+When \`maxLength\` is passed to TextArea, the component displays a character counter as well as a [warning or problem Status](/status) when the user reaches or the prepopulated controlled value exceeds the maximum length of characters.
 
-
-The first example shows an empty TextArea with \`maxLength\` set to 200 characters. The second example shows the error Status.`}
+The first example shows an empty TextArea with \`maxLength\` set to 200 characters. The second example shows the warning and problem Status.`}
         >
           <MainSection.Card
             cardSize="sm"
             defaultCode={`
 function TextAreaExample() {
   const [value, setValue] = React.useState('');
-  const maxLength = 200;
+  const characterCount = 200;
 
   return (
     <TextArea
@@ -487,7 +486,7 @@ function TextAreaExample() {
         onBlur={() => {}}
         onFocus={() => {}}
         rows={4}
-        maxLength={{ maxLengthChar: maxLength, errorAccessibilityLabel: 'Limit reached. You can only use 20 characters in this field.' }}
+        maxLength={{ characterCount, errorAccessibilityLabel: 'Limit reached. You can only use 200 characters in this field.' }}
       />
   );
 }
@@ -500,7 +499,7 @@ function TextAreaExample() {
   const [valueA, setValueA] = React.useState('Before and after via side by side display of pale woman with auburn hair using concealer. The image shows her using a brush with concealer under her eyes.  The second image shows her with full makeup.');
   const [valueB, setValueB] = React.useState('Before and after via side by side display of pale woman with auburn hair using concealer. The image shows her using a brush with concealer under her eyes, second image shows her with full makeup and says new.');
 
-  const maxLength = 200;
+  const characterCount = 200;
   const errorAccessibilityLabel = "Limit reached. You can only use 20 characters in this field.";
 
   return (
@@ -515,7 +514,7 @@ function TextAreaExample() {
         onBlur={() => {}}
         onFocus={() => {}}
         rows={4}
-        maxLength={{ maxLengthChar: maxLength, errorAccessibilityLabel }}
+        maxLength={{ characterCount, errorAccessibilityLabel }}
       />
       <TextArea
         id="maxLengthExceeded"
@@ -527,7 +526,7 @@ function TextAreaExample() {
         onBlur={() => {}}
         onFocus={() => {}}
         rows={4}
-        maxLength={{ maxLengthChar: maxLength, errorAccessibilityLabel }}
+        maxLength={{ characterCount, errorAccessibilityLabel }}
       />
     </Flex>
   );
--- a/docs/pages/web/textarea.js
+++ b/docs/pages/web/textarea.js
@@ -460,16 +460,17 @@ function Example(props) {
         </MainSection.Subsection>
         <MainSection.Subsection
           title="Maximum length"
-          columns={2}
-          description={`TextArea supports the native [maxlength](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/maxlength) input attribute. \`maxLength\` sets the maximum number of characters allowed to be entered by the user in TextArea.
-          
-For controlled TextAreas, if the initial value in the external state exceeds the \`maxLength\` value, TextArea will display the long value; however, the user won't be able to add additional characters and they will only be allowed to delete characters. 
-          
-When \`maxLength\` is passed to TextArea, the component displays a character counter as well as an [error Status](/status) when the user reaches or the controlled value exceeds the maximum length of characters. 
+          description={`TextArea supports the native [maxlength](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/maxlength) input attribute. \`maxLength\` sets the maximum number of characters allowed to be entered by the user in TextArea. \`maxLength\` must be an integer value 0 or higher.
 
-The first example shows an empty TextArea with \`maxLength\` set to 200 characters. The second example shows the error Status. The example below should be prevented as it shows TextArea displaying the error Status when the value lenght is exceeded.`}
+When \`maxLength\` is passed to TextArea, the component displays a character counter as well as an [error Status](/status) when the user reaches or the controlled value exceeds the maximum length of characters.
+
+For controlled TextAreas, if the initial value in the external state exceeds the \`maxLength\` value, TextArea will display the full value; however, the user won't be able to add additional characters and they will only be allowed to delete characters. This case should be prevented.
+
+
+The first example shows an empty TextArea with \`maxLength\` set to 200 characters. The second example shows the error Status.`}
         >
           <MainSection.Card
+            cardSize="sm"
             defaultCode={`
 function TextAreaExample() {
   const [value, setValue] = React.useState('');
@@ -485,7 +486,7 @@ function TextAreaExample() {
         value={value}
         onBlur={() => {}}
         onFocus={() => {}}
-        rows={2}
+        rows={4}
         maxLength={{ maxLengthChar: maxLength, errorAccessibilityLabel: 'Limit reached. You can only use 20 characters in this field.' }}
       />
   );
@@ -493,6 +494,7 @@ function TextAreaExample() {
 `}
           />
           <MainSection.Card
+            cardSize="lg"
             defaultCode={`
             function TextAreaExample() {
   const [valueA, setValueA] = React.useState('Before and after via side by side display of pale woman with auburn hair using concealer. The image shows her using a brush with concealer under her eyes.  The second image shows her with full makeup.');
@@ -502,7 +504,7 @@ function TextAreaExample() {
   const errorAccessibilityLabel = "Limit reached. You can only use 20 characters in this field.";
 
   return (
-    <Flex direction="column" gap={4}>
+    <Flex direction="column" gap={12}>
       <TextArea
         id="maxLengthReached"
         label="Alt text"
@@ -512,7 +514,7 @@ function TextAreaExample() {
         value={valueA}
         onBlur={() => {}}
         onFocus={() => {}}
-        rows={2}
+        rows={4}
         maxLength={{ maxLengthChar: maxLength, errorAccessibilityLabel }}
       />
       <TextArea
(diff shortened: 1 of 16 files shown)


docs/pages/web/textfield.js
```diff
@@ -700,6 +698,78 @@ function TextFieldExample() {
...
 }
+`}
+          />
+        </MainSection.Subsection>
+        <MainSection.Subsection
+          title="Maximum length"
+          columns={2}
+          description={`Textfield supports the native [maxlength](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/maxlength) input attribute. \`maxLength\` sets the maximum number of characters allowed to be entered by the user in Textfield.
+          
+For controlled TextFields, if the initial value in the external state exceeds the \`maxLength\` value, Textfield will display the long value; however, the user won't be able to add additional characters and they will only be allowed to delete characters.
```
ayeshakmaz, line 709: Not sure what "Textfield will display the long value" means... 

maybe "Textfield will display the full value"

docs/pages/web/textfield.js
```diff
@@ -700,6 +698,78 @@ function TextFieldExample() {
...
+        <MainSection.Subsection
+          title="Maximum length"
+          columns={2}
+          description={`Textfield supports the native [maxlength](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/maxlength) input attribute. \`maxLength\` sets the maximum number of characters allowed to be entered by the user in Textfield.
+          
+For controlled TextFields, if the initial value in the external state exceeds the \`maxLength\` value, Textfield will display the long value; however, the user won't be able to add additional characters and they will only be allowed to delete characters. 
+          
+When \`maxLength\` is passed to TextField, the component displays a character counter as well as an [error Status](/status) when the user reaches or the controlled value exceeds the maximum length of characters. 
+
+The first example shows an empty Textfield with \`maxLength\` set to 20 characters. The second example shows the error Status. The example below should be prevented as it shows Textfield displaying the error Status when the value lenght is exceeded.`}
```
ayeshakmaz, line 713: Typo: length

Not clear what "The example below should be prevented as it shows Textfield displaying the error Status when the value lenght is exceeded." is hoping to convey. 

Is it saying that you should prevent the user from being able to type after the limit is reached?

packages/gestalt/src/TextField.js
```diff
@@ -46,6 +51,10 @@ type Props = {|
    * Whether the label should be visible or not. If `hidden`, the label is still available for screen reader users, but does not appear visually. See the [label visibility variant](https://gestalt.pinterest.systems#Label-visibility) for more info.
    */
   labelDisplay?: 'visible' | 'hidden',
+  /**
+   * The maximum number of characters allowed in Textfield. See the [maximum lenght variant](https://gestalt.pinterest.systems/web/textfield##Maximum-length) for more details.
```
ayeshakmaz, line 55: typo: length

docs/pages/web/textarea.js
```diff
@@ -458,6 +458,81 @@ function Example(props) {
...
+        <MainSection.Subsection
+          title="Maximum length"
+          columns={2}
+          description={`TextArea supports the native [maxlength](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/maxlength) input attribute. \`maxLength\` sets the maximum number of characters allowed to be entered by the user in TextArea.
+          
+For controlled TextAreas, if the initial value in the external state exceeds the \`maxLength\` value, TextArea will display the long value; however, the user won't be able to add additional characters and they will only be allowed to delete characters. 
+          
+When \`maxLength\` is passed to TextArea, the component displays a character counter as well as an [error Status](/status) when the user reaches or the controlled value exceeds the maximum length of characters. 
+
+The first example shows an empty TextArea with \`maxLength\` set to 200 characters. The second example shows the error Status. The example below should be prevented as it shows TextArea displaying the error Status when the value lenght is exceeded.`}
```
ayeshakmaz, line 470: same comments as TextField

packages/gestalt/src/TextArea.js
```diff
@@ -42,6 +42,13 @@ type Props = {|
    * Whether the label should be visible or not. If `hidden`, the label is still available for screen reader users, but does not appear visually. See the [label visibility variant](https://gestalt.pinterest.systems/web/textarea#Label-visibility) for more info.
    */
   labelDisplay?: 'visible' | 'hidden',
+  /**
+   * The maximum number of characters allowed in TextArea. See the [maximum lenght variant](https://gestalt.pinterest.systems/web/textarea##Maximum-length) for more details.
```
ayeshakmaz, line 46: Typo: length

packages/gestalt/src/TextField.js
ayeshakmaz: typo in the URL too

packages/gestalt/src/TextArea.js
ayeshakmaz: typo in the URL

docs/pages/web/textarea.js
```diff
@@ -458,6 +458,81 @@ function Example(props) {
...
+        maxLength={{ maxLengthChar: maxLength, errorAccessibilityLabel: 'Limit reached. You can only use 20 characters in this field.' }}
+      />
+  );
+}
+`}
+          />
+          <MainSection.Card
+            defaultCode={`
+            function TextAreaExample() {
+  const [valueA, setValueA] = React.useState('Before and after via side by side display of pale woman with auburn hair using concealer. The image shows her using a brush with concealer under her eyes.  The second image shows her with full makeup.');
```
ayeshakmaz, line 498: Can we increase the lines visible in these examples to make it more clear?

docs/pages/web/textfield.js
AlbertCarreras: DONE

docs/pages/web/textfield.js
AlbertCarreras: We shouldn't set initial values from the parent component state that already exceed the maxLength. I'm not sure if this could be possible if we pre-populate fields...

packages/gestalt/src/TextField.js
AlbertCarreras: DONE

docs/pages/web/textarea.js
AlbertCarreras: DONE (tried !)

packages/gestalt/src/TextArea.js
AlbertCarreras: DONE

docs/pages/web/textarea.js
AlbertCarreras: DONE

docs/pages/web/textfield.js
ayeshakmaz: Ahh ok, maybe then 

"Whenever possible, avoid setting initial values from the parent component's state that already exceed the maxLength"?

dangerismycat: Would we not want to show the error state if the initial value exceeds the max length?

ponori: This is on me and in retrospect, I think we should display a warning icon as opposed to an error icon. Based on the current functionality, we don't allow the input to go beyond the character limit.

packages/gestalt/src/TextField.js
```diff
@@ -9,6 +9,11 @@ import { useDeviceType } from './contexts/DeviceTypeProvider.js';
 
 type Type = 'date' | 'email' | 'password' | 'tel' | 'text' | 'url';
 
+export type MaxLength = {|
+  maxLengthChar: number,
```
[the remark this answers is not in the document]
AlbertCarreras: Yeah! I like mumbling while coding haha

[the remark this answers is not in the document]
AlbertCarreras: @dangerismycat what about just maxLength?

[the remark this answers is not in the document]
dangerismycat: `maxLength={{ maxLength: 20 }}` is a bit repetitive, no?

docs/pages/web/textfield.js
dangerismycat: Input from the user, yes. But it's possible for the developer to specify an initial state for the input that exceeds the limit, which we would want to communicate. (This initial value may be determined dynamically, so just throwing an actual error wouldn't be a good idea.)

packages/gestalt/src/TextField.js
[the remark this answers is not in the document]
AlbertCarreras: DONE 

`maxLengthCharacters`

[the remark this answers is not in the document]
dangerismycat: Hm…the more I think about it, `maxLength={{ maxLengthCharacters: 20 }}` also seems pretty repetitive. Something like `maxLength={{ characterCount: 20 }}` flows a bit more naturally?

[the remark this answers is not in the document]
AlbertCarreras: DONE

packages/gestalt/src/InternalTextField.js
```diff
@@ -145,6 +151,10 @@ const InternalTextFieldWithForwardRef: React$AbstractComponent<Props, HTMLInputE
 
     const unstyledClasses = classnames(styles.unstyledTextField);
 
+    if (maxLength && maxLength?.characterCount < 0) {
```
dangerismycat, line 154: optional chaining is unnecessary after the preceding truthy check

packages/gestalt/src/TextArea.js
```diff
@@ -173,13 +183,18 @@ const TextAreaWithForwardRef: React$AbstractComponent<Props, HTMLTextAreaElement
       : {},
   );
 
+  if (maxLength && maxLength?.characterCount < 0) {
```
dangerismycat, line 186: ditto here

packages/gestalt/src/FormHelperText.js
```diff
@@ -1,21 +1,92 @@
...
+              </Text>
+            </Box>
+            <Flex gap={1}>
+              {maxLengthReached ? (
+                <Fragment>
+                  {/* This visually hidden error message is accessible by screenreaders. It alerts the user right after the maximum length is reached. */}
+                  <Box display="visuallyHidden" aria-live="assertive" role="alert">
+                    {maxLength?.errorAccessibilityLabel}
+                  </Box>
+                  <Status type={status} accessibilityLabel={maxLength?.errorAccessibilityLabel} />
```
dangerismycat, line 76: Optional chaining is unnecessary here given the truthy check on line 57

packages/gestalt/src/FormHelperText.js
```diff
@@ -1,21 +1,92 @@
...
+      <Flex gap={4}>
+        <Flex.Item flex="grow">
+          {text ? (
+            <Text color="subtle" size="100">
+              {addA11yPause ? <Box display="visuallyHidden">:</Box> : null}
+              {text}
+            </Text>
+          ) : null}
+        </Flex.Item>
+        {maxLength ? (
```
dangerismycat, line 57: Consider pulling this out to a separate component, which will also encapsulate all the logic from lines 30-44

packages/gestalt/src/FormHelperText.js
```diff
@@ -1,21 +1,92 @@
...
+export default function FormHelperText({
+  currentLength,
+  text,
+  maxLength,
+  addA11yPause = false,
+}: Props): Node {
+  const ref = useRef();
+  const [width, setWidth] = useState(undefined);
+
+  useEffect(() => {
```
dangerismycat, line 25: Why is this logic needed? This is a lot of additional complexity — can this be solved with Flex instead?

packages/gestalt/src/InternalTextField.js
AlbertCarreras: DONE

packages/gestalt/src/TextArea.js
AlbertCarreras: DONE

packages/gestalt/src/FormHelperText.js
AlbertCarreras: Screen Shot 2022-09-22 at 8 20 43 PM

packages/gestalt/src/FormHelperText.js
AlbertCarreras: DONE

packages/gestalt/src/FormHelperText.js
AlbertCarreras: This is calculating the max with this counter can have. We need to know the width of 100/100 so the width of the counter can be set and keep static

packages/gestalt/src/FormHelperTextCounter.js
```diff
@@ -0,0 +1,71 @@
+// @flow strict
+import { Fragment, useRef, useEffect, useState, type Node } from 'react';
+import Box from './Box.js';
+import Flex from './Flex.js';
+import Status from './Status.js';
+import Text from './Text.js';
+import { type MaxLength } from './TextField.js';
+
+type Props = {|
+  maxLength?: ?MaxLength,
```
dangerismycat, line 10: Since this component is gated by a truthy check on `maxLength` in FormHelperText, this type can be required, which will allow you to remove the optional chaining within this component

dangerismycat (approved):

packages/gestalt/src/FormHelperTextCounter.js
AlbertCarreras: DONE

PR #2415, opened 2022-09-26
Textfield: implemented spellcheck=false for type=password or email

AlbertCarreras:
#### What changed?

Under the hood implementation in Textfield: implemented `spellcheck=false` for `type=password` and `type=email`

#### Why?

This config is required to prevent passwords and usernames could be sent to spell-checking servers during login processes. More info here: https://www.androidpolice.com/google-chrome-servers-get-passwords-enhanced-spell-check/


rlingineni (approved):
Looks good! I didn't know that. Good call out

PR #2418, opened 2022-09-27
DeviceTypeProvider: new utility and documentation page

AlbertCarreras:
#### What changed?

DeviceTypeProvider: new utility and documentation page

#### Why?

Gestalt components can enable their mobile user interfaces using DeviceTypeProvider.

WARNING: This PR introduces 2 main changes to the previously undocumented DeviceTypeProvider: removes 'tablet' from the Flow type as Gestalt doesn't support tablet UI (and this is not planned to change in the near future) and renames `phone` to `mobile` to follow the standard way of referring to mobile platform. 

Despite the fact that DeviceTypeProvider was an undocumented utility, we are releasing this change as a major version. Use the following codemod to find DeviceTypeProvider in your codebase and make adjustments to upgrade to this version.

`yarn codemod detectManualReplacement ~/path/to/your/code --component=DeviceTypeProvider
`

NOTE: To test the modified Sandpack examples, replace "mobile" with "phone" as Sandpack only works with the latest published Gestalt version.

### DIFF
--- a/docs/examples/devicetypeprovider/implementation.js
+++ b/docs/examples/devicetypeprovider/implementation.js
@@ -1,0 +1,112 @@
+// @flow strict
+import React, { type Node } from 'react';
+import { Box, SideNavigation, DeviceTypeProvider, Button, Flex } from 'gestalt';
+
+export default function Example(): Node {
+  const [deviceType, setDeviceType] = React.useState('desktop');
+
+  return (
+    <Flex>
+      <Box padding={2}>
+        <Button
+          accessibilityControls="sidenav"
+          accessibilityLabel={`Toggle to ${deviceType === 'desktop' ? 'mobile' : 'desktop'} view`}
+          color="red"
+          text={`Toggle to ${deviceType === 'desktop' ? 'mobile' : 'desktop'} view`}
+          size="lg"
+          onClick={() => setDeviceType((value) => (value === 'desktop' ? 'mobile' : 'desktop'))}
+        />
+      </Box>
+      <Flex.Item flex="grow">
+        <DeviceTypeProvider deviceType={deviceType}>
+          <Box
+            display="flex"
+            alignItems="center"
+            justifyContent="center"
+            color="secondary"
+            height={500}
+          >
+            <Box height={450} width={280} color="default" id="sidenav">
+              <SideNavigation
+                title="Advertisement"
+                accessibilityLabel="Mobile device example"
+                dismissButton={{
+                  onDismiss: () => {},
+                  accessibilityLabel: 'Close navigation',
+                  tooltip: {
+                    accessibilityLabel: 'Close side navigation',
+                    text: 'Close navigation',
+                  },
+                }}
+              >
+                <SideNavigation.TopItem
+                  href="#"
+                  onClick={({ event }) => event.preventDefault()}
+                  label="Reporting"
+                  icon="ads-stats"
+                />
+                <SideNavigation.TopItem
+                  href="#"
+                  onClick={({ event }) => event.preventDefault()}
+                  label="Conversions"
+                  icon="replace"
+                />
+                <SideNavigation.Section label="Audiences">
+                  <SideNavigation.TopItem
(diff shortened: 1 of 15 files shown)


docs/pages/web/utilities/devicetypeprovider.js
```diff
@@ -1,22 +1,50 @@
...
+        type="info"
+        iconAccessibilityLabel="Recommendation"
+        message="Gestalt components that require DeviceTypeProvider to enable their mobile user interfaces have a Mobile variant section in their documentation page. Check each example to learn more about their particular implementations."
+      />
 
       <GeneratedPropTable generatedDocGen={generatedDocGen} />
+
+      <MainSection
+        name="Implementation"
+        description={`Gestalt components can have different user interfaces based on the device platform they are rendered: desktop or mobile.
```
dangerismycat, line 27: Gestalt components can have different interfaces depending on the user's device. We currently support `'desktop'` and `'mobile'`. Components default to a responsive "desktop" UI. DeviceTypeProvider is required to enable mobile-specific variants where available. This provider should be implemented at the top level of your app. Any additional nested DeviceTypeProviders will override the top-level configuration.

While device detection can (for now) be performed using the [user-agent string](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/User-Agent), this is [not generally recommended](https://developer.mozilla.org/en-US/docs/Web/HTTP/Browser_detection_using_the_user_agent). Thankfully, there are [better solutions](https://developer.mozilla.org/en-US/docs/Web/HTTP/Browser_detection_using_the_user_agent#avoiding_user_agent_detection)!

This example shows a component with different desktop and mobile UIs.

dangerismycat: ⬆️ meant as a suggestion for this text block

packages/gestalt/src/contexts/DeviceTypeProvider.js
```diff
@@ -3,23 +3,19 @@ import { type Context, type Node, createContext, useContext } from 'react';
 
 const defaultDeviceType = 'desktop';
 
-type DeviceType = 'desktop' | 'phone' | 'tablet';
```
dangerismycat, line 6: nit: I would've saved this change for a follow-up PR just so the preview deploy would work without the reviewer needing to fiddle with anything, but definitely don't revert it now. Just something to keep in mind for the future 🙂

dangerismycat (approved):
This is awesome, thanks!

docs/examples/devicetypeprovider/implementation.js
```diff
@@ -0,0 +1,112 @@
+// @flow strict
+import React, { type Node } from 'react';
+import { Box, SideNavigation, DeviceTypeProvider, Button, Flex } from 'gestalt';
+
+export default function Example(): Node {
+  const [deviceType, setDeviceType] = React.useState('desktop');
+
+  return (
+    <Flex>
+      <Box padding={2}>
```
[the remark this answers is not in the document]
AlbertCarreras: I wanted to add spacing around the Button so it's not touching the edges of the container. With Flex it would add space between items, not with parent

docs/examples/sidenavigation/subcomponent.js
```diff
@@ -0,0 +1,88 @@
+// @flow strict
+import React, { type Node } from 'react';
+import { SideNavigation } from 'gestalt';
+
+export default function Example(): Node {
+  return (
+    <SideNavigation accessibilityLabel="Subcomponent composability example">
+      {true && (
```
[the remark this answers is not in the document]
AlbertCarreras: good point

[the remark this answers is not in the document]
AlbertCarreras: DONE

docs/pages/web/utilities/devicetypeprovider.js
AlbertCarreras: ### DONE

PR #2425, opened 2022-10-03
Datapoint: fix casing in docs examples

dangerismycat:
As noted in [this Slack thread](https://pinterest.slack.com/archives/C014X9LTRCN/p1664488014513369), we need to be more consistent with the casing of abbreviations. This PR updates our examples to be consistent. It specifically does _not_ address adding new guidance around this, as we should work with content design folks to develop that guidance.

### DIFF
--- a/docs/pages/web/datapoint.js
+++ b/docs/pages/web/datapoint.js
@@ -1,15 +1,14 @@
 // @flow strict
 import { type Node } from 'react';
-import GeneratedPropTable from '../../docs-components/GeneratedPropTable.js';
-import PageHeader from '../../docs-components/PageHeader.js';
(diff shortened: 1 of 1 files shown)


docs/pages/web/datapoint.js
```diff
@@ -137,14 +152,14 @@ Be cautious of using Datapoint for large volumes of data. In cases where the con
             cardSize="md"
             title="Medium"
             defaultCode={`
-        <Datapoint size="md" title="Spend" value="$1.23M" />
+        <Datapoint size="md" title="Spend" value="$1.23m" />
```
AlbertCarreras, line 155: actual change

AlbertCarreras (approved):

PR #2426, opened 2022-10-03
SideNavigation: fix dismissButton focus

AlbertCarreras:
#### What changed?

SideNavigation: fix dismissButton focus

#### Why?


packages/gestalt/src/InternalDismissButton.js
```diff
@@ -0,0 +1,108 @@
...
+      height: innerRef?.current?.clientHeight,
+      width: innerRef?.current?.clientWidth,
+    });
+
+    const [isFocused, setFocused] = useState(false);
+    const [isHovered, setHovered] = useState(false);
+
+    const { isFocusVisible } = useFocusVisible();
+
+    const renderPogComponent = (selected?: boolean): Node => (
```
dangerismycat, line 48: This should be inlined below, no need for a separate function

packages/gestalt/src/InternalDismissButton.js
```diff
@@ -0,0 +1,108 @@
...
+
+type Props = {|
+  accessibilityLabel: string,
+  accessibilityControls?: string,
+  onClick?: AbstractEventHandler<
+    SyntheticMouseEvent<HTMLButtonElement> | SyntheticKeyboardEvent<HTMLButtonElement>,
+  >,
+|};
+
+const InternalDismissIconButtonWithForwardRef: React$AbstractComponent<Props, HTMLButtonElement> =
```
dangerismycat, line 19: Can you add comments to this file why we're creating a custom button and not using IconButton?

packages/gestalt/src/InternalDismissButton.js
```diff
@@ -0,0 +1,108 @@
...
+  forwardRef<Props, HTMLButtonElement>(function IconButton(props: Props, ref): Node {
+    const { accessibilityLabel, accessibilityControls, onClick } = props;
```
dangerismycat, line 21: nit: destructure props in function signature

packages/gestalt/src/InternalDismissButton.js
```diff
@@ -0,0 +1,108 @@
...
+
+    const { isFocusVisible } = useFocusVisible();
+
+    const renderPogComponent = (selected?: boolean): Node => (
+      <Pog
+        focused={isFocusVisible && isFocused}
+        hovered={isHovered}
+        icon="cancel"
+        selected={selected}
+        size="lg"
```
dangerismycat, line 54: This component is always the same size? Do we need to support other sizes?

packages/gestalt/src/SideNavigationMobile.js
```diff
@@ -1,32 +1,50 @@
...
-import IconButton from './IconButton.js';
 import styles from './SideNavigation.css';
 import borderStyles from './Borders.css';
-import getChildrenToArray from './getChildrenToArray.js';
+import getChildrenToArray from './SideNavigation/getChildrenToArray.js';
+import InternalDismissButton from './InternalDismissButton.js';
 import { useSideNavigation } from './contexts/SideNavigationProvider.js';
 import { type Props } from './SideNavigation.js';
 
+type SideNavigationMobileProps = {| ...Props, id: string |};
```
dangerismycat, line 15: nit: this would be a bit more readable if you rename the imported Props type and use "Props" here:
```js
import { type Props as SideNavigationProps } from '';

type Props = {|
  ...SideNavigationProps,
  id: string,
|};
```

dangerismycat (approved):

packages/gestalt/src/InternalDismissButton.js
AlbertCarreras: DONE

packages/gestalt/src/InternalDismissButton.js
AlbertCarreras: DONE

packages/gestalt/src/InternalDismissButton.js
AlbertCarreras: I'd keep it lg for now, We can support other sizes if neede.d

packages/gestalt/src/SideNavigationMobile.js
AlbertCarreras: indeed

AlbertCarreras: DONE

PR #2432, opened 2022-10-05
Docs: add component aliases to page header

dangerismycat:
Naming is one of the more difficult things we struggle with, and similar components often have different names across the industry. While we need to settle on specific names for sanity's sake, our users may not be accustomed to those names. This PR adds an optional "aliases" section to the docs page header so we can add these other names components are commonly known by. This _should_ get picked up on the next Algolia crawl and improve search results for folks using these other names.

Screen Shot 2022-10-05 at 4 11 21 PM


ayeshakmaz (approved):
Looks great, thank you!

PR #2435, opened 2022-10-07
Tokens: Adjust dark mode error text and icon

ayeshakmaz:
#### What changed?

Adjust error token for Text and Icon to have more contrast in dark mode

#### Why?

In order to provide better color contrast in dark mode, $color-text-error and $color-text-icon-error went from $color-red-pushpin-400 to $color-red-pushpin-300

### DIFF
--- a/docs/pages/web/datepicker.js
+++ b/docs/pages/web/datepicker.js
@@ -128,10 +128,8 @@ function DatePickerExample() {
 
       <MainSection name="Variants">
         <Example
-          description="
-    Use DatePicker to select date inputs.
-  "
-          name="Example: Basic Date Picker"
+          description="Use DatePicker to select date inputs."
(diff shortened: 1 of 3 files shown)


docs/pages/web/datepicker.js
```diff
@@ -128,10 +128,8 @@ function DatePickerExample() {
 
       <MainSection name="Variants">
         <Example
-          description="
-    Use DatePicker to select date inputs.
-  "
-          name="Example: Basic Date Picker"
+          description="Use DatePicker to select date inputs."
+          name="Basic Date Picker"
```
ayeshakmaz, line 132: As I was checking dark mode error text, I noticed these example names and removed the repetitive Example: in front of all of them for cleaner docs

dangerismycat (approved):

PR #2436, opened 2022-10-08
ModalAlert: New component for confirmation and acknowledgment

ayeshakmaz:
#### What changed?

New component ModalAlert 

Screen Shot 2022-10-07 at 6 16 45 PM

#### Why?

New component to support confirmation and acknowledgment modals. It is a narrowed-down view of Modal, and has 3 types. 

A follow-up PR will be made to add examples to the doc page

### Links

- [Jira](https://jira.pinadmin.com/browse/GESTALT-4025)
- [Figma](https://www.figma.com/file/EWc4nY8bGmjYUQhET9hTvr/Gestalt-AlertModal-Hand-off?node-id=2%3A174)

### Checklist


packages/gestalt/src/Modal.css
```diff
@@ -7,7 +7,7 @@
 .wrapper {
   composes: relative overflowAuto flex from "./Layout.css";
   composes: whiteBgElevated from "./Colors.css";
-  composes: rounding8 from "./Borders.css";
+  composes: rounding4 from "./Borders.css";
```
ayeshakmaz, line 10: Discussed with Hector and we're dropping this to 16px

packages/gestalt/src/AlertModal.js
```diff
@@ -0,0 +1,202 @@
...
+          </Box>
+        )}
+        <Flex.Item flex="grow">
+          {' '}
+          <Heading size="400" accessibilityLevel={1}>
+            {heading}
+          </Heading>
+        </Flex.Item>
+        {type === 'default' && (
+          <Box marginStart={6}>
```
dangerismycat, line 117: It seems like the parent Flex could have a `gap={4}`, with this Box setting `marginStart={2}`? That would obviate the need for the other `marginEnd={4}`

packages/gestalt/src/AlertModal.js
```diff
@@ -0,0 +1,202 @@
...
+            <IconButton
+              accessibilityLabel="Close modal"
```
dangerismycat, line 119: This string needs to be translated. Perhaps include it in DefaultLabelProvider?

packages/gestalt/src/AlertModal.js
```diff
@@ -0,0 +1,202 @@
...
+      fullWidth
+      role="button"
+      size="lg"
+      text={label}
+    />
+  );
+}
+
+/**
+ * An AlertModal is a simple modal dialog used to alert a user of an issue, or to request confirmation after a user-generated action. AlertModal overlays and blocks Page content until it is dismissed by the user.
```
dangerismycat, line 166: similar nit here: "user-triggered action" or similar?

dangerismycat: also: "Page" doesn't need to be capitalized

packages/gestalt/src/AlertModal.js
```diff
@@ -0,0 +1,202 @@
...
+}: Props): Node {
+  return (
+    <Modal
+      accessibilityModalLabel={accessibilityModalLabel}
+      align="start"
+      closeOnOutsideClick={type === 'default'}
+      footer={
+        <Flex
+          justifyContent="end"
+          gap={{
```
dangerismycat, line 185: would `gap={4}` work here?

docs/pages/web/alertmodal.js
```diff
@@ -0,0 +1,235 @@
...
+        >
+          <MainSection.Card cardSize="lg" />
+        </MainSection.Subsection>
+      </AccessibilitySection>
+
+      <MainSection
+        name="Localization"
+        description={`Be sure to localize the \`heading\` and \`accessibilityModalLabel\`props, as well as any other text elements within AlertModal. Note that localization can lengthen text by 20 to 30 percent. `}
+      />
+      <MainSection name="Variants">
```
dangerismycat, line 155: I'm concerned that there's nothing in the code to enforce these variants, which don't seem to correspond to `type`? If we're recommending specific versions of this component, we should make those easy to build, and make it hard/impossible to build versions that don't match those variants/guidelines. I could currently build something like
```js
<AlertModal
  type="default"
  heading="Nope, that's wrong"
  primaryAction={{
    accessibilityLabel: 'Oops',
    label: "C'est la vie",
  }}
  onDismiss={handleDismiss}
>
Whatever you just did isn't correct. Are you sure you want to continue?
</AlertModal>
```
…which really doesn't seem like something we'd ever want someone to build. The wrong `type` for the message, no secondary "cancel" action required, an action that doesn't actually do anything (neither `href` nor `onClick` are required). Are we ok with this? If we have specific guidelines for variants, should we guide users to those variants via the component API?

packages/gestalt/src/AlertModal.js
```diff
@@ -0,0 +1,202 @@
...
+  /**
+   * Supply the element(s) that will be used as AlertModal's main content. See the [Best Practices](https://gestalt.pinterest.systems/web/alertmodal#Best-practices) for more info.
+   */
+  children: Node,
+  /**
+   * The text used for AlertModal's heading.
+   */
+  heading: string,
+  /**
+   * Callback fired when AlertModal is dismissed by clicking on the backdrop outside of the AlertModal (if `closeOnOutsideClick` is true) or when the dismiss IconButton is clicked (for default AlertModals).
```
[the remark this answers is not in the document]
dangerismycat: also: we don't offer `closeOnOutsideClick` for this component

hectoid:
It hasn't been vetted yet. Sorry, missed that.

docs/pages/web/alertmodal.js
ayeshakmaz: The guidelines around actions aren't strict enough that we could enforce through props, imo. Errors and warnings sometimes have one action, sometimes have two. I think the only thing we could safeguard for is ensuring either onClick or href exists

packages/gestalt/src/AlertModal.js
```diff
@@ -0,0 +1,202 @@
...
+import { type Node } from 'react';
+import Box from './Box.js';
+
+import Button from './Button.js';
+import Flex from './Flex.js';
+import Heading from './Heading.js';
+import Icon from './Icon.js';
+import IconButton from './IconButton.js';
+import Modal from './Modal.js';
+import { type AbstractEventHandler } from './AbstractEventHandler.js';
```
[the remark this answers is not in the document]
dangerismycat: Sorry, I should've been more clear. You'll have to inline the type itself, otherwise we see this:

Screen Shot 2022-10-12 at 4 55 36 PM

Check out other components using autogen props tables (like NumberField) to see examples

packages/gestalt/src/AlertModal.js
```diff
@@ -0,0 +1,249 @@
...
+  children,
+  onDismiss,
+  heading,
+  primaryAction,
+  secondaryAction,
+}: Props): Node {
+  const { accessibilityDismissButtonLabel: accessibilityDismissButtonLabelDefault } =
+    useDefaultLabelContext('AlertModal');
+
+  if (primaryAction && primaryAction.href === undefined && primaryAction.onClick === undefined) {
```
dangerismycat, line 205: nit: these checks could share a helper function to not repeat the logic

docs/pages/web/alertmodal.js
dangerismycat: Ok, the href/onClick check helps. There's still the question of if we want to offer a limited number of specific types of AlertModals, or if the user is free to build whatever they need (given the already constrained props available).

Assuming it's the latter, we may need to reconsider the docs structure here. Most other pages use variants to explain specific props, but these seem more like guidelines or patterns. I took the variants to mean "these are specific types we offer", and I wonder if others would be confused as well. What do you think?

ayeshakmaz: We could rename to be Error, Warning, Multiple Actions, Single Action to be more prop-based, or like Type - Error, Type- Warning, For acknowledgments, and For confirmations?

ayeshakmaz: And yeah, in this case it's the latter

dangerismycat: Maybe both? I think the guidance you're currently providing is really valuable, I just want it to be clear that it's _guidance_ and not something that can be turned on/off with a specific prop. We could perhaps add this current guidance as a new Patterns section or something? (that can be a working name that we update in the future)

Really, we probably need _more_ guidance like this on our component pages. I just want to make sure we're not presenting it in a confusing way

ayeshakmaz: The amazing guidance and docs are all @hectoid - just the implementer : ) I've updated the names of the variants, Hector let me know what you think

packages/gestalt/src/ModalAlert.js
```diff
@@ -0,0 +1,244 @@
...
+import Box from './Box.js';
+import Button from './Button.js';
+import Flex from './Flex.js';
+import Heading from './Heading.js';
+import Icon from './Icon.js';
+import IconButton from './IconButton.js';
+import Modal from './Modal.js';
+import { useDefaultLabelContext } from './contexts/DefaultLabelProvider.js';
+
+type AbstractEventHandler<T: SyntheticEvent<HTMLElement> | Event, U = {||}> = ({|
```
dangerismycat, line 12: This type will need to be inlined, otherwise we get:
Screen Shot 2022-10-12 at 4 55 36 PM

```js
  onClick?: ({|
    event:
    dangerouslyDisableOnNavigation: () => void,
  |}) => void,
```

packages/gestalt/src/ModalAlert.js
```diff
@@ -0,0 +1,244 @@
...
+  /**
+   * Determines the icon and dismiss pattern of the ModalAlert. See the [warning](https://gestalt.pinterest.systems/web/alertmodal#Warning) and [error](https://gestalt.pinterest.systems/web/alertmodal#Error)  variants for more info.
+   */
+  type?: 'default' | 'warning' | 'error',
+  /**
+   * Main action for users to take on ModalAlert. If `href` is supplied, the action will serve as a link. See [OnLinkNavigationProvider](https://gestalt.pinterest.systems/web/utilities/onlinknavigationprovider) to learn more about link navigation.
+   * If no `href` is supplied, the action will be a button.
+   * The `accessibilityLabel` should follow the [Accessibility guidelines](https://gestalt.pinterest.systems/web/alertmodal#Accessibility).
+   */
+  primaryAction: {|
```
dangerismycat, line 63: These are `ActionDataType`, no?

packages/gestalt/src/ModalAlert.js
```diff
@@ -0,0 +1,244 @@
...
+      fullWidth
+      role="button"
+      size="lg"
+      text={label}
+    />
+  );
+}
+
+/**
+ * An ModalAlert is a simple modal dialog used to alert a user of an issue, or to request confirmation after a user-triggered action. ModalAlert overlays and blocks page content until it is dismissed by the user.
```
dangerismycat, line 190: Don't forget to add a full link to the docs page on the component name

packages/gestalt/src/ModalAlert.js
```diff
@@ -0,0 +1,244 @@
...
+  children,
+  onDismiss,
+  heading,
+  primaryAction,
+  secondaryAction,
+}: Props): Node {
+  const { accessibilityDismissButtonLabel: accessibilityDismissButtonLabelDefault } =
+    useDefaultLabelContext('ModalAlert');
+
+  [
```
dangerismycat, line 205: You can simplify this data structure a bit:
```js
for (const [key, value] of Object.entries({ primaryAction, secondaryAction })) {
  if ([value?.href, value?.onClick].every((item) => item === undefined)) {
    throw new Error(
        `Either an \`href\` or an \`onClick\` handler must be provided to \`${key}\`.`,
      );
  }
}
```

playwright/accessibility/ModalAlert.spec.mjs
```diff
@@ -0,0 +1,8 @@
+// @flow strict
+import { test } from '@playwright/test';
+import expectAccessiblePage from './expectAccessiblePage.mjs';
+
+test('ModalAlert Accessibility check', async ({ page }) => {
+  await page.goto('/web/alertmodal');
```
dangerismycat, line 6: Don't forget to update the URL (here and in links throughout)

dangerismycat (approved):

packages/gestalt/src/ModalAlert.js
ayeshakmaz: Ooo good call, thanks!

PR #2437, opened 2022-10-08
Collection: Short-circuit rendering in Collection

chrislloyd:
## Motivation

For a long time the largest performance impact for Masonry has been the time to render items (such as a Pin). While this is typically fast, minor differences in timings is noticable when frame/interaction boundaries are ~16ms (60fps).

Prior this PR, the approach has generally been "make the Pin faster to render" and that's yielded a lot of benefits. However, I believe that overlooks and obvious improvement.

## Description

This PR introduces a memoized `<Viewport>` that short-circuits rendering child elements unless the set of visible items in the viewport changes.

Most scroll events _don't actually_ change the items being rendered. `<Collection>` is wasting CPU time re-rendering items that have no actual change. This means each render is consistent, however it can also mean the CPU gets saturated quickly. By only rendering when an item enters/leaves the viewport, this change trades timing consitency with more CPU availability when rendering new items.

```
 ┌┐ ┌┐ ┌┐ ┌┐ ┌┐ ┌┐ ┌┐ ┌┐ ┌┐
 ││ ││ ││ ││ ││ ││ ││ ││ ││
 ││ ││ ││ ││ ││ ││ ││ ││ ││
 └┘ └┘ └┘ └┘ └┘ └┘ └┘ └┘ └┘
```

vs.

```
 ┌┐
 ││          ┌┐       ┌┐
 ││ ┌┐ ┌┐ ┌┐ ││ ┌┐ ┌┐ ││ ┌┐
 └┘ └┘ └┘ └┘ └┘ └┘ └┘ └┘ └┘
```

In addition, I've added a few memoizations of expensive calulations (figuring out the layout bounds + the actual windowing).

## Followups

This PR doesn't introduce any changes to Collection's interface, but I believe it's worth experimenting with other changes such as:

1. Adding `contains: strict` (or similar) to all parts of Collection seeing as the layout is fixed. This would allow for removing `layoutStyles.relative` too.
2. Changing the layout type to be like: `{[id: string]: Rect}` instead of relying on array indexes as identity. This means it's easier to reuse prior layouts when computing new layouts and React will have more efficient diffs.
3. Having Masonry use Collection.

## Test plan

Well... you're going to have to trust me on this one ;) Obviously CI passes but any performance claims requires real world testing that lives outside of the Gestalt project. Also now that Masonry doesn't use Collection (it should) it's harder to test. YMMV and please run an experiment or something.

### DIFF
--- a/packages/gestalt/src/Collection.js
+++ b/packages/gestalt/src/Collection.js
@@ -39,7 +39,7 @@
   4. The viewport can be any size. Most windowing/recycling solutions implement some sort of overscanning, however Collection leaves this up the the parent.
 
 */
-import { type Node } from 'react';
+import { type ComponentType, Fragment, memo, type Node, useMemo, useCallback } from 'react';
 import layoutStyles from './Layout.css';
 
 type Props = {|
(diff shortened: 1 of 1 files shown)


dangerismycat (commented):
Hi Chris! 👋 Thanks for the PR — looks interesting. Are the snapshot failures on Collage expected? Are those items that are now outside of the viewport?

chrislloyd:
I think it’s just because the snapshots are shallow and now there’s an intermediate component there. It might be a little while before I can fix but should be straightforward if you have time.Also, 👋On Oct 11, 2022, at 17:42, Ryan James ***@***.***> wrote:﻿
@dangerismycat commented on this pull request.

Hi Chris! 👋 Thanks for the PR — looks interesting. Are the snapshot failures on Collage expected? Are those items that are now outside of the viewport?

PR #2440, opened 2022-10-12
Docs: convert Hacking Gestalt page to use Sandpack

dangerismycat:
Of note: the previous ref example was no longer functional (pretty good indicator of why this isn't a great technique! 😅), so I updated to use TextField instead:
Screen Shot 2022-10-12 at 2 57 19 PM


ayeshakmaz (approved):
Tiny nit: in two places (css selectors and wrapping) the "Alternative" heading is italicized instead of bolded (unrelated, by nice to fix)

PR #2442, opened 2022-10-13
Masonry: add virtualBufferFactor prop

christianvuerings:
#### What changed?

Adds `virtualBufferFactor` as a prop to Masonry.

#### Why?

The virtual buffer factor impacts when masonry shows items when scrolling. Adding it as a prop enables us to experiment with different values and measure performance / user impact.

#### Next steps

* Experiment with different `virtualBufferFactor` values & measure the performance impact
* Consolidate `virtualBoundsBottom`, `virtualBoundsTop` and `virtualBufferFactor` into 1 prop

### DIFF
--- a/packages/gestalt/src/Masonry.js
+++ b/packages/gestalt/src/Masonry.js
@@ -82,6 +82,10 @@ type Props<T> = {|
    */
   virtualBoundsTop?: number,
   /**
+   * If `virtualize` is enabled, Masonry will only render items that fit in the viewport, plus some buffer. `virtualBufferFactor` allows customization of the buffer size, specified as a multiplier of the container height. It specifies the amount of extra buffer space for populating visible items. For example, if `virtualBufferFactor` is 2, then Masonry will render items that fit in the viewport, plus 2x the viewport height.
+   */
+  virtualBufferFactor?: number,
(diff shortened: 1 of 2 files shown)


packages/gestalt/src/Masonry.js
```diff
@@ -375,12 +378,19 @@ export default class Masonry<T: { ... }> extends ReactComponent<Props<T>, State<
...
+      virtualBoundsTop,
+      virtualBoundsBottom,
+      virtualBufferFactor,
+    } = this.props;
     const { top, left, width, height } = position;
 
     let isVisible;
-    if (scrollContainer) {
-      const virtualBuffer = this.containerHeight * VIRTUAL_BUFFER_FACTOR;
+    if (scrollContainer && virtualBufferFactor) {
```
dangerismycat, line 392: Is this check necessary, given that we're providing a default value in defaultProps? Or is this just to appease Flow?

dangerismycat (approved):

packages/gestalt/src/Masonry.js
christianvuerings: @dangerismycat just to appease flow

PR #2449, opened 2022-10-20
ModalAlert: Fix validation bug

ayeshakmaz:
#### What changed?

Check for value before checking for href and onclick


packages/gestalt/src/ModalAlert.js
```diff
@@ -172,7 +172,7 @@ export default function ModalAlert({
 
   Object.entries({ primaryAction, secondaryAction }).forEach(([key, value]) => {
     // $FlowFixMe[incompatible-use]
-    if ([value?.href, value?.onClick].every((item) => item === undefined)) {
+    if (value && [value?.href, value?.onClick].every((item) => item === undefined)) {
```
dangerismycat, line 175: nit: with the truthy check for `value`, you no longer need optional chaining

dangerismycat (approved):

PR #2451, opened 2022-10-21
Flex: add dataTestId

dangerismycat:
This is a common request from folks, and often either results in people using Box instead, or wrapping Flex in an additional Box just to set the id (resulting in DOM bloat).

This could be implemented as a passthrough prop (similar to Box), but I think that's a leaky abstraction. Though it is a passthrough component under the hood, we don't treat Flex as such. And we shouldn't expect our users to know how Flex is implemented. Therefore, I implemented this as an "official" new prop, `dataTestId`. This matches our other non-Box test ids.

## Future work

I need to update the existing `prefer-flex` lint rule to accommodate this new prop.

I'll also look at ways of finding and refactoring Boxes that wrap Flexes only for the test id.

### DIFF
--- a/packages/gestalt/src/Flex.js
+++ b/packages/gestalt/src/Flex.js
@@ -26,6 +26,10 @@ type Props = {|
    */
   children?: Node,
   /**
+   * Used to identify the element for testing purposes.
+   */
+  dataTestId?: string,
+  /**
    * Establishes the main axis, thus defining the direction flex items are placed in the flex container.
    */
   direction?: 'row' | 'column',
@@ -105,6 +109,7 @@ const allowedProps = [
 export default function Flex({
   alignItems,
   children: childrenProp,
+  dataTestId,
   direction = 'row',
   gap = 0,
   justifyContent,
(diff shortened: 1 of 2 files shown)


AlbertCarreras:
I'm ok with the change. So, in the past we had proposed creating a Testable component. Wrapping Flex with <Box datatestid> is basically the same approach. Why are we concerned here about DOM bloat but not with Testable?

AlbertCarreras (approved):

dangerismycat:
> in the past we had proposed creating a Testable component. Wrapping Flex with is basically the same approach. Why are we concerned here about DOM bloat but not with Testable?
> 
> If we go this route, seems like we might need to start adding more datatestid to more components.

@AlbertCarreras  Ultimately I don't think a Testable component will scale well. In addition to the DOM bloat, that approach won't work for many of our sub-components, necessitating adding `dataTestId` props to them explicitly (as we've done with Dropdown.Item and Dropdown.Link already). I think we will need to add `dataTestId` props to many/most of our components at some point.

AlbertCarreras:
For consistency, I'd also add it to Box, although this is OOO for this PR.

dangerismycat:
> For consistency, I'd also add it to Box, although this is OOO for this PR.

Agreed — its availability as a passthrough prop is a leaky abstraction there as well, and is not intuitive. I'll throw together a quick follow-up PR

PR #2453, opened 2022-10-21
Box: make dataTestId an explicit prop

dangerismycat:
As a follow-up to #2451, this PR adds an explicit `dataTestId` prop to Box. Previously we've relied on the passthrough nature of Box to allow users to use `data-test-id`. However, this is a leaky abstraction, as the user has to be aware of the implementation of Box to know that they can pass through this undocumented "prop". By making this prop explicit, we ensure that users who are unfamiliar with Gestalt are able to easily add the test ids needed.

## CODEMOD

```bash
yarn codemod modifyProp ~/path/to/your/code --component=Box --previousProp='data-test-id' --nextProp=dataTestId
```

### DIFF
--- a/packages/gestalt/src/Box.js
+++ b/packages/gestalt/src/Box.js
@@ -183,6 +183,10 @@ type Props = {
    */
   lgColumn?: Column,
   /**
+   * Used to identify the element for testing purposes.
+   */
+  dataTestId?: string,
+  /**
    * Establishes the main-axis, thus defining the direction flex items are placed in the flex container.
    *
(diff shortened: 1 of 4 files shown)


AlbertCarreras:
I think this implementation overrides any existing data-test-id. I think it’s a breaking change anyway. Can you double check? 
image

AlbertCarreras (approved):

packages/gestalt/src/Box.js
```diff
@@ -517,19 +521,10 @@ type Props = {
 // to strip out a few naughty properties that break style encapsulation
 // (className, style) or accessibility (onClick).
 const disallowedProps = [
-  'onClick',
   'className',
+  'data-test-id', // in favor of the dataTestId prop
+  'onClick',
   'style',
-  // We're currently also explicitly disallowing the deprecated marginLeft/Right
```
dangerismycat, line 523: These should have been removed long ago

dangerismycat:
I think I'm going to try a different approach for this at some point. Closing this for now

PR #2457, opened 2022-10-25
Docs: Add top of page examples

ayeshakmaz:
#### What changed?

Add top-of-page example to Tag, Card, SegmentedControl, and ButtonGroup

#### Why?

Ensure folks can see the component at first glance

### DIFF
--- a/docs/examples/buttongroup/defaultExample.js
+++ b/docs/examples/buttongroup/defaultExample.js
@@ -1,0 +1,14 @@
+// @flow strict
+import { type Node } from 'react';
+import { Button, ButtonGroup, Flex } from 'gestalt';
+
+export default function Example(): Node {
+  return (
+    <Flex alignItems="center" height="100%" justifyContent="center" width="100%">
+      <ButtonGroup>
+        <Button text="Cancel" />
+        <Button color="red" text="Send" />
+      </ButtonGroup>
+    </Flex>
+  );
+}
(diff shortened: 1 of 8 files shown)


AlbertCarreras (approved):

docs/examples/buttongroup/defaultExample.js
```diff
@@ -0,0 +1,14 @@
+// @flow strict
+import { type Node } from 'react';
```
dangerismycat, line 2: Note that when the user views the code for this example, they'll see
```js
import {} from 'react';
```
You can avoid this by using a different type, which doesn't need to be imported:
```js
export default function Example(): React$Node {
```

docs/examples/segmentedcontrol/defaultExample.js
```diff
@@ -0,0 +1,27 @@
+// @flow strict
+import { type Node, useState } from 'react';
```
dangerismycat, line 2: This one's fine, as there are other non-type imports. So this will show up as
```js
```
when the user views the code

docs/examples/segmentedcontrol/defaultExample.js
```diff
@@ -0,0 +1,27 @@
...
+import { Box, Text, SegmentedControl, Icon } from 'gestalt';
+
+export default function Example(): Node {
+  const [itemIndex, setItemIndex] = useState(0);
+  const items = [
```
dangerismycat, line 7: nit: consts like these should be defined outside the component, so they're not recreated on every render (this doesn't matter here, but we want to make sure the examples we show are production-ready code)

PR #2467, opened 2022-11-01
Link, Button, TapArea, IconButton: implemented default label to target="blank"

AlbertCarreras:
### Breaking change for major release

Update `DefaultLabelProvider` with new label.

```
Link: {
    accessibilityNewTabLabel: myI18nTranslator('Opens a new tab'),
  },
```

#### What changed?

Link, Button, TapArea, IconButton: implemented default label to target="blank" #2467

Added SlimBanners across the different documentation pages linking components to DefaultLabelProvider.

#### Why?

Note from WebAim — The vast majority of screen readers don't inform users that target= "_blank" links open in new tabs; folks could have a hard time realizing why they can't use the back button to return to the previous page and that they might need to switch back to the last tab to do so. 

We actually can create a label informing users that the link is going to open in a new tab. 
Label suggestion to announce: Opens a new tab

Reference: We must inform your users about this opening in a new tab behavior. Failure to do so goes against [WCAG 2.0 Guideline 3.2: Predictability](https://www.w3.org/TR/UNDERSTANDING-WCAG20/consistent-behavior.html).

####  TESTING

| COMPONENT      |  no accessibilityLabel | accessibilityLabel
| Link             |  Screen Shot 2022-11-01 at 2 37 21 PM | Screen Shot 2022-11-01 at 2 38 21 PM |
| Button         | Screen Shot 2022-11-01 at 2 40 50 PM | Screen Shot 2022-11-01 at 2 40 36 PM  |
| IconButton         | NA |  Screen Shot 2022-11-02 at 10 27 14 AM |
| TapArea      |  Screen Shot 2022-11-01 at 2 44 09 PM | Screen Shot 2022-11-01 at 2 45 10 PM  |
| Dropdown  |    Screen Shot 2022-11-01 at 2 46 25 PM    | NA           |

### Checklist

- [NA] Checked dark mode, responsiveness, and right-to-left support

### DIFF
--- a/packages/gestalt/src/Button.js
+++ b/packages/gestalt/src/Button.js
@@ -1,5 +1,5 @@
 // @flow strict
-import { forwardRef, useImperativeHandle, useRef, type Node } from 'react';
+import { Fragment, forwardRef, useImperativeHandle, useRef, type Node } from 'react';
 import classnames from 'classnames';
 import Flex from './Flex.js';
 import focusStyles from './Focus.css';
@@ -13,6 +13,8 @@ import InternalLink from './InternalLink.js';
 import Icon, { type IconColor } from './Icon.js';
 import { useColorScheme } from './contexts/ColorSchemeProvider.js';
 import { type AbstractEventHandler } from './AbstractEventHandler.js';
+import NewTabAccessibilityLabel, { getAriaLabel } from './NewTabAccessibilityLabel.js';
+import { useDefaultLabelContext } from './contexts/DefaultLabelProvider.js';
 
 const DEFAULT_TEXT_COLORS = {
   blue: 'inverse',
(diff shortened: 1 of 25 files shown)


packages/gestalt/src/Button.js
```diff
@@ -84,22 +86,34 @@ type unionProps = ButtonType | SubmitButtonType | LinkButtonType;
...
+        {icon ? (
+          <Icon
+            accessibilityLabel=""
+            color={textColor}
+            icon={icon}
+            size={SIZE_NAME_TO_PIXEL[size]}
+          />
+        ) : null}
+      </Flex>
+      {target === 'blank' ? <NewTabAccessibilityLabel target={target} /> : null}
```
ayeshakmaz, line 115: isn't it a bit redundant to check for target blank both here and within NewTabAccessibilityLabel?

ayeshakmaz (approved):
Very thorough, thank you!

packages/gestalt/src/Button.js
AlbertCarreras: @ayeshakmaz It seems so! But no, this is expected. Flex doesn't know if the component is returning null so it adds gap spacing. I wanted to reuse NewTabAccessibilityLabel. The other components don't have the Flex wrapper, but button does, so I choose the redundancy so I could reuse NewTabAccessibilityLabel.

AlbertCarreras: Good catch. Initially it was all inside the Flex. Flex doesn't know if the component is returning null so it adds gap spacing if we don't keep the condition externally. I wanted to reuse NewTabAccessibilityLabel. The other components don't have the Flex wrapper, but button does, so I choose the redundancy so I could reuse NewTabAccessibilityLabel.

But then I moved it out, so then it's not needed it anymore.

ayeshakmaz: nice!

PR #2468, opened 2022-11-01
Internal: Retry playwright step

christianvuerings:
#### What changed?

Retry the playwright step 2 times instead of failing immediately.

#### Why?

Playwright steps are occasionally failing, instead of requiring manual re-runs, let's retry the step.

### Notes

* GitHub doesn't supply a retry job out of the box. There are GitHub actions which do have this functionality but adding those are a security concern
  * https://github.com/marketplace/actions/retry-step
  * https://github.com/marketplace/actions/retry-action

### DIFF
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -28,7 +28,11 @@ jobs:
       - name: Run build
         run: yarn build
       - name: Run your tests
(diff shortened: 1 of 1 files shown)


dangerismycat (approved):
🤞

PR #2472, opened 2022-11-03
Internal: turn on `react/no-array-index-key` lint rule

dangerismycat:
[Rule definition](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/no-array-index-key.md)

[React uses `key` to identify items within arrays and maintain stability as those items change](https://reactjs.org/docs/lists-and-keys.html#keys). But it's generally [a bad idea to use array indices as keys](https://robinpokorny.com/blog/index-as-a-key-is-an-anti-pattern/).

While there are occasionally valid reasons to use indices as keys, in general we want to avoid that practice, so it makes sense to lint against it.

### DIFF
--- a/docs/docs-components/AccessibilityChecklist.js
+++ b/docs/docs-components/AccessibilityChecklist.js
@@ -39,28 +39,26 @@ function QualityTable({ accessibilityData }: {| accessibilityData?: AccessibleSt
         </Table.Header>
       </Box>
       <Table.Body>
-        {['a11yVisual', 'a11yScreenreader', 'a11yNavigation', 'a11yComprehension'].map(
-          (item, index) => {
-            const componentStatus = accessibilityData?.[item] ?? 'notAvailable';
+        {['a11yVisual', 'a11yScreenreader', 'a11yNavigation', 'a11yComprehension'].map((item) => {
+          const componentStatus = accessibilityData?.[item] ?? 'notAvailable';
 
-            return (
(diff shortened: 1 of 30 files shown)


packages/gestalt/src/AvatarGroup.js
```diff
@@ -120,6 +120,7 @@ const AvatarGroupWithForwardRef: React$AbstractComponent<Props, UnionRefs> = for
     <CollaboratorAvatar
       hovered={hovered}
       index={index}
+      // eslint-disable-next-line react/no-array-index-key
```
dangerismycat, line 123: Avoiding making changes in exported components to keep the footprint of this PR small

docs/docs-components/AccessibilityChecklist.js
```diff
@@ -39,28 +39,26 @@ function QualityTable({ accessibilityData }: {| accessibilityData?: AccessibleSt
...
-                </Table.Cell>
-                <Table.Cell>
-                  <Text>{COMPONENT_A11Y_STATUS_MESSAGING[item][componentStatus]}</Text>
-                </Table.Cell>
-              </Table.Row>
-            );
-          },
-        )}
+          return (
+            <Table.Row key={item}>
```
dangerismycat, line 46: Being pretty cavalier about keys in the docs — if we notice errors in the console about duplicate keys, we can adjust these

PR #2012, opened 2022-03-22
Image: update importance prop to fetchpriority

mziemer21:
### Summary
Updating importance prop to fetchpriority.

#### What changed?
[Priority-hints](https://wicg.github.io/priority-hints/) is going to ship in the next version of chrome. Since the origin-trial, they've updated the name to fetchpriority.

I've also removed the default 'auto' for fetchpriority and loading. React correctly handles if the prop is undefined and does not set it. The browser will default to 'auto' if the attribute is not set.

#### Why?

Once priority-hints ships we want to start running experiments with it.

### Breaking Changes

Run the following codemod to locate Image components with `importance` prop and rename it to `fetchpriority`.

`yarn codemod modifyProp ~/path/to/your/code --component=Image  --previousProp=importance --nextProp=fetchpriority`

### Checklist

- [X] Added unit and Flow Tests
- [X] Added documentation + accessibility tests

PR #2474, opened 2022-11-03
Image: replace `importance` prop with `fetchPriority`

edarliu:
## Breaking Change

major release - running a generic codemod to replace instances of 'importance' prop and replacing them with 'fetchPriority' 

```bash
--component=Image
--previousProp=importance
--nextProp=fetchPriority
```

### Summary

Updating 'importance' prop to 'fetchPriority' . This diff removes the deprecated 'importance' prop and replaces with fetchpriority.

There's an old PR [here](https://github.com/pinterest/gestalt/pull/2012) that aimed to do the same thing - just opening up a new one instead of having to jump through permissions hoops + rebasing troubles.

#### Why?

Now that fetchPriority is stable (at least for chromium-based browsers), we want to start running experiments with it to see if we can optimize Largest Contentful Paint on certain surfaces in order to improve SEO rankings, as Google uses LCP and other Core Web Vitals to rank pages in their search engine algorithm.

Specifically, one experiment we had in mind was to optimize for LCP on mWeb unauth BLP by increasing the fetchPriority of images in the viewport. Currently, LCP is potentially slowed down by JS bundles that are called at around the same time in the network waterfall - Chrome defaults these bundles to 'high' priority as well, so we want to see what happens to LCP as well as other potential side effects when upping the image priority to 'highest'.

Priority hints are only in Chromium-based browsers (Chrome, edge, etc). There's been a push for adding them to Firefox as well, but this hasn't been accomplished yet, so we'll need to add a flag for Chromium browsers in the future experiments. However, Chromium browsers make up a large percentage of our traffic (forget the exact number), so any experiment findings will still have a large impact.

### Checklist

### DIFF
--- a/packages/gestalt/src/Image.js
+++ b/packages/gestalt/src/Image.js
@@ -31,15 +31,15 @@ type Props = {|
    */
   elementTiming?: string,
   /**
+   * Priority hints provide developers a way to indicate a resource's relative importance to the browser, allowing more control over the order resources are loaded (only available via Chrome Origin Trial). \`"high"\`: the developer considers the resource to be high priority. \`"low"\`: the developer considers the resource to be low priority. \`auto\` the developer does not indicate a preference.
+   * Note that this feature is currently experimental; please see the [attribute spec](https://wicg.github.io/priority-hints/) for more details.
+   */
+  fetchPriority?: 'high' | 'low' | 'auto',
+  /**
    * Sets how the image is resized to fit its container. See the [Fit example](https://gestalt.pinterest.systems/web/image#fit) for more details.
    * Note: this doesn't work with srcSet or sizes.
    */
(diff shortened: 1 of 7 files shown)


AlbertCarreras:
@edarliu "at least for chromium-based browsers" what do you mean with "at least". What's the tradeoff of not being stable for all those other browsers?

edarliu:
@AlbertCarreras right now, priority hints are only in Chromium-based browsers (Chrome, edge, etc). There's been a push for adding them to Firefox as well, but this hasn't been accomplished yet, so we'll need to add a flag for Chromium browsers in the future experiments. However, Chromium browsers make up a large percentage of our traffic (forget the exact number), so any experiment findings will still have a large impact.

AlbertCarreras (approved):

PR #2479, opened 2022-11-07
Docs: fix duplicate key issue

christianvuerings:
# Summary

## What changed?

Caused by #2472 - fixes a duplicate key issue on every docs page

## Why?

Resolves errors in the browser console (happens on every page):

Screen Shot 2022-11-07 at 7 09 04 AM

## Alternatives

* Require a unique `id` for every `MainSectionSubsection` component
* Use combination of `title` and `description` - I tried this and it looks like we have quite a few places were both are empty

### DIFF
--- a/docs/docs-components/MainSectionSubsection.js
+++ b/docs/docs-components/MainSectionSubsection.js
@@ -94,8 +94,9 @@ function MainSectionSubsection({
               column: 4,
             }}
           >
-            {arrayChildren.map((child) => (
-              <Flex.Item flex="grow" key={child}>
+            {arrayChildren.map((child, index) => (
+              // eslint-disable-next-line react/no-array-index-key
+              <Flex.Item flex="grow" key={index}>
                 {child}
               </Flex.Item>
             ))}


docs/docs-components/MainSectionSubsection.js
```diff
@@ -94,8 +94,9 @@ function MainSectionSubsection({
               column: 4,
             }}
           >
-            {arrayChildren.map((child) => (
-              <Flex.Item flex="grow" key={child}>
+            {arrayChildren.map((child, index) => (
+              // eslint-disable-next-line react/no-array-index-key
```
AlbertCarreras, line 98: Could we use useId+index for the key so we don't have to disable this?

christianvuerings: @AlbertCarreras https://reactjs.org/docs/hooks-reference.html#useid

image

AlbertCarreras (approved):

dangerismycat:
Thanks @christianvuerings ! I suspected there might be a few of these

PR #2481, opened 2022-11-07
Toast: refactor 'button' prop to 'primaryAction' to enable upcoming redesign changes

AlbertCarreras:
## Breaking change

API change. From `button` (Node) to a restricted object `{ accessibilityLabel, href, label, onClick, rel, size, target }

Added a new undocumented prop `_dangerouslySetPrimaryAction`. Use `_dangerouslySetPrimaryAction` as an exception, only for legacy custom buttons components. 

Run the following codemod to detect button props and replace them manually:

```
--component=Toast
--prop=button
```
 

#### What changed?

Toast: refactor 'button' prop to enable upcoming redesign changes

#### Why?

'button' prop doesn't allow to control button sizes. refactoring to 'primaryAction' to enable upcoming redesign changes

### Checklist

### DIFF
--- a/packages/gestalt/src/Toast.js
+++ b/packages/gestalt/src/Toast.js
@@ -1,11 +1,14 @@
 // @flow strict
-import { type Element, type Node } from 'react';
+import { isValidElement, type Element, type Node } from 'react';
 import Box from './Box.js';
 import Flex from './Flex.js';
+import Link from './Link.js';
 import Mask from './Mask.js';
+import Button from './Button.js';
 import Text from './Text.js';
 import styles from './Toast.css';
 import useResponsiveMinWidth from './useResponsiveMinWidth.js';
+import ToastPrimaryAction from './ToastPrimaryAction.js';
 import { useColorScheme } from './contexts/ColorSchemeProvider.js';
 
 const TOAST_MAX_WIDTH_PX = 500;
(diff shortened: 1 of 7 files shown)


packages/gestalt/src/Toast.js
```diff
@@ -107,16 +124,28 @@ export default function Toast({
...
           </Flex.Item>
-
-          {button ? (
+          {primaryAction || _dangerouslySetPrimaryAction ? (
             // Allow button text to wrap on mobile
-            <Flex.Item flex={isMobileWidth ? 'shrink' : 'none'}>{button}</Flex.Item>
+            <Flex.Item flex={isMobileWidth ? 'shrink' : 'none'}>
+              {isValidElement(_dangerouslySetPrimaryAction) ? _dangerouslySetPrimaryAction : null}
+              {!_dangerouslySetPrimaryAction &&
+              primaryAction?.accessibilityLabel &&
```
dangerismycat, line 137: I like that we're making these checks, but I think we should move them to ToastPrimaryAction, where we can throw an actual error. Flow will already be yelling at the developer about missing fields; throwing an error would accomplish the same thing you're doing here, but with more direct feedback about what's missing and how to fix it.

packages/gestalt/src/ToastPrimaryAction.js
```diff
@@ -0,0 +1,43 @@
+// @flow strict
+import { type Node } from 'react';
+import Button from './Button.js';
+import Link from './Link.js';
+
+type Props = {|
+  accessibilityLabel?: string,
```
dangerismycat, line 7: `accessibilityLabel` and `label` should be required, no?

dangerismycat (approved):

packages/gestalt/src/Toast.js
AlbertCarreras: I can set an error to double down on the user feedback, though Flow is already covering it

packages/gestalt/src/ToastPrimaryAction.js
AlbertCarreras: DONE

PR #2484, opened 2022-11-07
SlimBanner: extend "message" prop to accept Text component for rich-styled messages

AlbertCarreras:
#### What changed?

SlimBanner: extend "message" prop to accept Text component for rich-styled messages

#### Why?

Supporting Text and string in message prop will solve 2 problems: supporting rich-style messages such as inline bold text within Text and supporting links within Text allowing the correct localization of text with inlined links. helperText cannot be used with that purpose as localization not always follows the same word order in different locales.

- Left: Previously supported
- Center: Previously unsupported
- Right: Previously unsupported

Screen Shot 2022-11-07 at 5 54 28 PM

### Links

- [Figma]( https://www.figma.com/file/MInxIhLJSbGDebM0939W5G/Gestalt-SlimBanner-Hand-off?node-id=602%3A2520)

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,7 +38,7 @@
 
 ### Minor
 
-- Icon: convert Tabs Docs to use Sandpack editor ([#2488](https://github.com/pinterest/gestalt/pull/2488)) - [Preview link](https://deploy-preview-2488--gestalt.netlify.app)
+- Icon: convert Icon Docs to use Sandpack editor ([#2488](https://github.com/pinterest/gestalt/pull/2488)) - [Preview link](https://deploy-preview-2488--gestalt.netlify.app)
 
 ## 81.2.0 (Nov 9, 2022)
 
--- a/docs/pages/web/slimbanner.js
+++ b/docs/pages/web/slimbanner.js
@@ -444,49 +444,46 @@ Combine SlimBanners with other components like [Callouts](/web/callout) or [Upse
           />
         </MainSection.Subsection>
         <MainSection.Subsection
-          description={`\`message\` props accepts both strings and [Text](/Text). Use strings for simple messages without any visual style. SlimBanner will handle the message style and the adherence to design guidelines. If a message with more complex style is required, such as bold text or inline links, use Text to inline texts and links within the same Text component.
+          description={`The \`message\` prop accepts both strings and [Text](/Text). Use strings for simple messages without any visual style. SlimBanner will handle the message style and the adherence to design guidelines. If a message with more complex style is required, such as bold text or inline links, use Text to inline text and links within the same Text component.
 
-The SlimBanner message strings can be complemented with a helper Link. When passing rich Text components, \`helperLink\` isn't rendered to prevent unnecessary visual load.
+The SlimBanner \`message\` string can be complemented with a \`helperLink\`. When passing a Text component, \`helperLink\` isn't rendered to prevent unnecessary visual load.
 
-Due to localization constrains, \`message\` and \`helperLink\` prop texts cannot belong to the same statement. They must be independent sentences separated by a period.
+Due to localization constrains, the contents of \`message\` and \`helperLink\` cannot belong to the same sentence. They must be independent sentences separated by a period.
 `}
           title="Message"
         >
           <MainSection.Card
-            title="Simple message string with helperText"
             cardSize="md"
             defaultCode={`
-<SlimBanner
-  type="info"
-  message="This ad group is part of a campaign that is using campaign budget optimization. Changes to schedule or budget must be made at the campaign level."
-  iconAccessibilityLabel="Information"
-  helperLink={{
-      text: 'Learn more',
-      accessibilityLabel: 'Learn more about campaign budget optimization',
-      href: 'http://www.pinterest.com',
+<Flex direction="column" gap={6}>
+  <Text weight="bold">Simple message string with helperText</Text>
+  <SlimBanner
+    type="info"
+    message="This ad group is part of a campaign that is using campaign budget optimization. Changes to schedule or budget must be made at the campaign level."
+    iconAccessibilityLabel="Information"
+    helperLink={{
+        text: 'Learn more',
+        accessibilityLabel: 'Learn more about campaign budget optimization',
+        href: 'http://www.pinterest.com',
+        onClick: () => {},
+      }}
+  />
+  <Text weight="bold">Rich message with Text component</Text>
+  <SlimBanner
+    type="recommendation"
+    message={<Text inline>The campaign <Text inline weight="bold">Back to School</Text> is regularly hitting its <Link inline href="">daily cap</Link>. Consider raising daily caps to increase scale for a similar CPC and CTR.</Text>}
+    primaryAction={{
+      accessibilityLabel: 'Increase spend',
+      label: 'Increase spend',
       onClick: () => {},
     }}
-/>
-`}
-          />
-          <MainSection.Card
-            title="Rich message with Text component"
-            cardSize="md"
-            defaultCode={`
-<SlimBanner
-  type="recommendation"
-  message={<Text inline>The campaign <Text inline weight="bold">Back to School</Text> is regularly hitting its <Link inline href="">daily cap</Link>. Consider raising daily caps to increase scale for a similar CPC and CTR.</Text>}
-  primaryAction={{
-    accessibilityLabel: 'Increase spend',
-    label: 'Increase spend',
-    onClick: () => {},
-  }}
-  dismissButton={{
-    accessibilityLabel: 'Dismiss banner',
-    onDismiss: () => {},
-  }}
-  iconAccessibilityLabel="Recommendation"
-/>
+    dismissButton={{
+      accessibilityLabel: 'Dismiss banner',
+      onDismiss: () => {},
+    }}
+    iconAccessibilityLabel="Recommendation"
+  />
+</Flex>
 `}
           />
         </MainSection.Subsection>
--- a/docs/pages/web/slimbanner.js
+++ b/docs/pages/web/slimbanner.js
@@ -444,11 +444,12 @@ Combine SlimBanners with other components like [Callouts](/web/callout) or [Upse
           />
         </MainSection.Subsection>
         <MainSection.Subsection
-          description={`The \`message\` prop accepts both strings and [Text](/Text). Use strings for simple messages without any visual style. SlimBanner will handle the message style and the adherence to design guidelines. If a message with more complex style is required, such as bold text or inline links, use Text to inline text and links within the same Text component.
+          description={`
+The \`message\` prop accepts either a string or [Text](/Text). Use a string for simple messages without any visual style. SlimBanner will handle the message style and adherence to design guidelines. If a message with more complex style is required, such as bold text or inline links, use Text to wrap your message with any additional Text or Link usages contained within.
 
 The SlimBanner \`message\` string can be complemented with a \`helperLink\`. When passing a Text component, \`helperLink\` isn't rendered to prevent unnecessary visual load.
 
-Due to localization constrains, the contents of \`message\` and \`helperLink\` cannot belong to the same sentence. They must be independent sentences separated by a period.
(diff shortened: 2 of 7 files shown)


docs/pages/web/slimbanner.js
```diff
@@ -445,11 +444,17 @@ Combine SlimBanners with other components like [Callouts](/web/callout) or [Upse
...
         </MainSection.Subsection>
         <MainSection.Subsection
-          description="The SlimBanner message can be complemented with a helper Link."
-          title="helperLink"
+          description={`\`message\` props accepts both strings and [Text](/Text). Use strings for simple messages without any visual style. SlimBanner will handle the message style and the adherence to design guidelines. If a message with more complex style is required, such as bold text or inline links, use Text to inline texts and links within the same Text component.
+
+The SlimBanner message strings can be complemented with a helper Link. When passing rich Text components, \`helperLink\` isn't rendered to prevent unnecessary visual load.
+
+Due to localization constrains, \`message\` and \`helperLink\` prop texts cannot belong to the same statement. They must be independent sentences separated by a period.
+`}
```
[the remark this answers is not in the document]
AlbertCarreras: DONE thanks

docs/pages/web/slimbanner.js
```diff
@@ -462,6 +467,26 @@ Combine SlimBanners with other components like [Callouts](/web/callout) or [Upse
       onClick: () => {},
     }}
 />
+`}
+          />
+          <MainSection.Card
```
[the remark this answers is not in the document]
AlbertCarreras: DONE

packages/gestalt/src/SlimBanner.js
```diff
@@ -116,18 +116,18 @@ type Props = {|
...
    * Label to describe the status icon’s purpose. See the [Accessibility guidelines](https://gestalt.pinterest.systems/web/slimbanner#Accessibility) for details on proper usage.
    */
   iconAccessibilityLabel?: string,
   /**
-   * Main content of SlimBanner. Content should be [localized](https://gestalt.pinterest.systems/web/slimbanner#Localization).
+   * Main content of SlimBanner. Content should be [localized](https://gestalt.pinterest.systems/web/slimbanner#Localization). See the [Message variant](https://gestalt.pinterest.systems/web/slimbanner#Message) to learn more.
    *
    */
-  message: string,
+  message: string | Element<typeof Text>,
```
dangerismycat, line 130: Screen Shot 2022-11-09 at 5 06 45 PM

We need to make sure this shows `React.Element`. You'll need to adjust the logic in [GeneratedPropTable](https://github.com/pinterest/gestalt/blob/d30cd30ba74efd4b1e6b8ed02abf71f6293bbe9d/docs/docs-components/GeneratedPropTable.js#L101)…since the current regex is checking for "Element" at the beginning of a string (to avoid matching "HTMLDivElement"), you may need to add a new `.replace` call that looks for "Element" with preceding whitespace

packages/gestalt/src/SlimBanner.js
```diff
@@ -202,15 +202,18 @@ export default function SlimBanner({
...
+                {message}
+                {helperLink ? (
+                  <Fragment>
+                    {' '}
+                    <HelperLink {...helperLink} />
+                  </Fragment>
+                ) : null}
+              </Text>
+            ) : null}
+            {typeof message !== 'string' && isValidElement(message) ? message : null}
```
dangerismycat, line 216: Do we want to enforce the usage of Text here? Otherwise the user can pass any valid React element here and suppress the Flow error. I think we might be able to use `message.type.name === 'Text'` — the user would be able to create their own Text component and pass that, but that seems a lot less likely (and much more intentional).

packages/gestalt/src/TableRowExpandable.js
```diff
@@ -59,6 +59,7 @@ export default function TableRowExpandable({
 }: Props): Node {
   const [expanded, setExpanded] = useState(false);
   const cs = hoverStyle === 'gray' ? cx(styles.hoverShadeGray) : null;
+
```
dangerismycat, line 62: This change should be reverted as it's irrelevant to the PR.

docs/pages/web/slimbanner.js
```diff
@@ -445,23 +444,46 @@ Combine SlimBanners with other components like [Callouts](/web/callout) or [Upse
           />
         </MainSection.Subsection>
         <MainSection.Subsection
-          description="The SlimBanner message can be complemented with a helper Link."
-          title="helperLink"
+          description={`The \`message\` prop accepts both strings and [Text](/Text). Use strings for simple messages without any visual style. SlimBanner will handle the message style and the adherence to design guidelines. If a message with more complex style is required, such as bold text or inline links, use Text to inline text and links within the same Text component.
```
dangerismycat, line 447: "…prop accepts _either a string or_ Text." Current wording makes it sound like the user could pass an array of strings. You'll need to adjust to the singular form in the next sentence as well.

No "the" needed in "…will handle the message style _and adherence_ to…"

Last sentence is kinda awkward: "…such as bold text or inline links, use Text to wrap your message with any additional Text or Link usages contained within." (or something like that)

docs/pages/web/slimbanner.js
```diff
@@ -445,23 +444,46 @@ Combine SlimBanners with other components like [Callouts](/web/callout) or [Upse
...
+
+The SlimBanner \`message\` string can be complemented with a \`helperLink\`. When passing a Text component, \`helperLink\` isn't rendered to prevent unnecessary visual load.
+
+Due to localization constrains, the contents of \`message\` and \`helperLink\` cannot belong to the same sentence. They must be independent sentences separated by a period.
```
dangerismycat, line 451: typo: "Due to localization _constraints_…"

Maybe add one more sentence at the end? We want to be a bit forceful here. Perhaps something like "Do not attempt to construct a compound sentence using `message` and `helperLink`."

packages/gestalt/src/SlimBanner.js
AlbertCarreras: Im going to try using React.children to make that check. I think it can be used with any prop

packages/gestalt/src/TableRowExpandable.js
AlbertCarreras: DONE

docs/pages/web/slimbanner.js
AlbertCarreras: DONE

docs/pages/web/slimbanner.js
AlbertCarreras: Screen Shot 2022-11-10 at 11 57 58 AM
Screen Shot 2022-11-10 at 11 58 30 AM

packages/gestalt/src/SlimBanner.js
dangerismycat: I don't see any changes to GeneratedPropTable on this PR? Are you making that change in a separate PR?

packages/gestalt/src/SlimBanner.js
dangerismycat: Ah ok — I haven't actually tried this technique in production code, so I wasn't sure if it would work. Make the best attempt you can…while throwing an error here on bad input would be ideal, it's not a dealbreaker if we can't and this current technique is the best we can do.

docs/pages/web/slimbanner.js
AlbertCarreras: .name doesn't work. I have to add the following to get the displayName available

```
Text.displayName = 'Text';

export default Text;
```

AlbertCarreras: DONE

packages/gestalt/src/SlimBanner.js
AlbertCarreras: Misplaced the reply https://github.com/pinterest/gestalt/pull/2484#discussion_r1019395689

AlbertCarreras: Fixed it by adding .displayName to 'Text'

dangerismycat: Nice! 🔥

docs/pages/web/slimbanner.js
```diff
@@ -589,7 +611,11 @@ Tooltip provides helpful information regarding an interactive UI element, typica
 }
 
 export async function getStaticProps(): Promise<{| props: {| generatedDocGen: DocGen |} |}> {
+  const docGen = await docgen({ componentName: 'SlimBanner' });
+
+  docGen.props.primaryAction.flowType.raw = '{| string | React.Element<typeof Text>>> |}';
```
dangerismycat, line 616: Ah, I see now. Let's make this change in GeneratedPropTable, as we want this to apply everywhere we use React.Element

packages/gestalt/src/SideNavigation/getChildrenToArray.js
```diff
@@ -17,7 +17,6 @@ const getChildrenToArray = ({
   const navigationChildren = [];
 
   let recursionLevel = 0;
-
```
dangerismycat, line 20: this is irrelevant to the PR, should be reverted

docs/pages/web/slimbanner.js
dangerismycat: I'm not seeing the updated `description` here?

packages/gestalt/src/SideNavigation/getChildrenToArray.js
AlbertCarreras: DONE

docs/pages/web/slimbanner.js
AlbertCarreras: FIXED

docGen.props.**message**. flowType.raw

packages/gestalt/src/SlimBanner.js
AlbertCarreras: DONE! Sorry I changed the wrong prop name ahahah

docs/pages/web/slimbanner.js
AlbertCarreras: Mmmm... let me double check

CHANGELOG.md
```diff
@@ -38,7 +38,7 @@
 
 ### Minor
 
-- Icon: convert Tabs Docs to use Sandpack editor ([#2488](https://github.com/pinterest/gestalt/pull/2488)) - [Preview link](https://deploy-preview-2488--gestalt.netlify.app)
+- Icon: convert Icon Docs to use Sandpack editor ([#2488](https://github.com/pinterest/gestalt/pull/2488)) - [Preview link](https://deploy-preview-2488--gestalt.netlify.app)
```
AlbertCarreras, line 41: Fix from previous PR

docs/pages/web/slimbanner.js
dangerismycat: I'm still not seeing updates to the copy:

> "…prop accepts either a string or Text." Current wording makes it sound like the user could pass an array of strings. You'll need to adjust to the singular form in the next sentence as well.
>
> No "the" needed in "…will handle the message style and adherence to…"
>
> Last sentence is kinda awkward: "…such as bold text or inline links, use Text to wrap your message with any additional Text or Link usages contained within." (or something like that)

docs/pages/web/slimbanner.js
```diff
@@ -589,7 +615,11 @@ Tooltip provides helpful information regarding an interactive UI element, typica
 }
 
 export async function getStaticProps(): Promise<{| props: {| generatedDocGen: DocGen |} |}> {
+  const docGen = await docgen({ componentName: 'SlimBanner' });
+
+  docGen.props.message.flowType.raw = '{| string | React.Element<typeof Text>>> |}';
```
dangerismycat, line 620: This is the logic that should be in GeneratedPropTable. We want to handle this case universally, not as a one-off

AlbertCarreras: I'll fix that in a separate PR. OOS for this PR

docs/pages/web/slimbanner.js
AlbertCarreras: DONE

dangerismycat (approved):

PR #2485, opened 2022-11-07
Image: use img for scaled images

dangerismycat:
Currently, when using scaled images (`fit="contain"` or `fit="cover"`), Image uses a div with a background image instead of the `` element. This doesn't work well for SEO and has been a request from Logged-Out Product for a while.

I asked around and no one has context on why we use our current approach. Given that it was present in [the initial Gestalt commit](https://github.com/pinterest/gestalt/commit/4d69ecc5243ed8d1b2662a5af3bbbb3d388b9a6a#diff-1d010f8be067a2311d2b1f955d874fb3103d56386ce4616c8bc212e1d6f13b09), I suspect that @chrislloyd knows? From what I can tell, [this is a workaround](https://www.stevefenton.co.uk/blog/2019/09/fixing-css-object-fit-for-internet-explorer/) because `object-fit` [isn't supported by IE11](https://caniuse.com/object-fit). However, since we no longer support IE, we can ditch this workaround and use `object-fit` directly. As shown below, this does not affect output in modern browsers:

_Chrome/Brave_
Screen Shot 2022-11-07 at 2 55 49 PM

_Firefox_
Screen Shot 2022-11-07 at 3 25 56 PM

_Safari_
Screen Shot 2022-11-07 at 3 24 51 PM

_Edge_
Screen Shot 2022-11-07 at 3 30 27 PM


packages/gestalt/src/Image.js
```diff
@@ -152,50 +153,42 @@ export default class Image extends PureComponent<Props> {
...
-          style={{
-            backgroundColor: color,
-            backgroundImage: `url('${src}')`,
-          }}
-          role={role}
-        />
-        {childContent}
-      </Box>
-    ) : (
+    const isScaledImage = fit === 'cover' || fit === 'contain';
```
AlbertCarreras, line 162: should we kill line 7?

packages/gestalt/src/Image.js
```diff
@@ -152,50 +153,42 @@ export default class Image extends PureComponent<Props> {
...
+    const fitStyles = fit === 'cover' || fit === 'contain' ? styles[fit] : undefined;
```
AlbertCarreras, line 163: `isScaledImage ` to be used here?

packages/gestalt/src/Image.js
```diff
@@ -152,50 +153,42 @@ export default class Image extends PureComponent<Props> {
...
           elementtiming={elementTiming}
           importance={importance}
           loading={loading}
           onError={this.handleError}
           onLoad={this.handleLoad}
+          role={role === 'presentation' ? 'presentation' : undefined}
           sizes={sizes}
           src={src}
           srcSet={srcSet}
-          role={role === 'presentation' ? 'presentation' : undefined}
```
AlbertCarreras, line 198: moved up

packages/gestalt/src/Image.js
```diff
@@ -152,50 +153,42 @@ export default class Image extends PureComponent<Props> {
...
+          {...(isScaledImage ? { style: { objectFit: fit } } : {})}
```
AlbertCarreras, line 191: natively implemented into img, previously implemented via div to support IE

packages/gestalt/src/Image.js
AlbertCarreras: Screen Shot 2022-11-08 at 2 14 25 PM


Does `background-size` work in . it was being applied to div, not sure if it works with 

Is it redundant with `objectFit` /  dead code?

packages/gestalt/src/Image.js
dangerismycat: "line 7"? I'm not sure I follow

packages/gestalt/src/Image.js
dangerismycat: > `isScaledImage ` to be used here?

Flow doesn't realize that "none" won't be an option if we use that variable, unfortunately

dangerismycat: > Does background-size work in . it was being applied to div, not sure if it works with 
> Is it redundant with objectFit / dead code?

Good call — was able to simplify the styles a bit

packages/gestalt/src/Image.js
AlbertCarreras: Yeah, line 7
Screen Shot 2022-11-09 at 11 19 05 AM

AlbertCarreras (approved):
LGTM with some pending comments and failing snaps

packages/gestalt/src/Image.js
dangerismycat: got it, good call

PR #2487, opened 2022-11-08
Table: new Table.RowDrawer subcomponent

AlbertCarreras:
#### What changed?

Table: new Table.RowDrawer subcomponent

Screen Shot 2022-11-11 at 12 23 18 PM

Also, removed divider between Row and RowExpandable and RowDrawer

Screen Shot 2022-11-11 at 12 42 55 PM

#### Why?

New component to support m10n requests. 

New featured (Drawer with SlimBanner) was tested under experimentation presenting good results.

### Checklist

### DIFF
--- a/docs/pages/web/table.js
+++ b/docs/pages/web/table.js
@@ -1521,8 +1521,8 @@ function Example() {
         >
           <GeneratedPropTable
             generatedDocGen={generatedDocGen.TableHeader}
-            name="Table.Header"
-            id="Table.Header"
+            name={generatedDocGen?.TableHeader?.displayName}
+            id={generatedDocGen?.TableHeader?.displayName}
           />
         </MainSection.Subsection>
         <MainSection.Subsection
@@ -1531,8 +1531,8 @@ function Example() {
         >
           <GeneratedPropTable
             generatedDocGen={generatedDocGen.TableBody}
-            name="Table.Body"
-            id="Table.Body"
+            name={generatedDocGen?.TableBody?.displayName}
+            id={generatedDocGen?.TableBody?.displayName}
           />
         </MainSection.Subsection>
         <MainSection.Subsection
@@ -1541,8 +1541,8 @@ function Example() {
         >
           <GeneratedPropTable
             generatedDocGen={generatedDocGen.TableFooter}
-            name="Table.Footer"
-            id="Table.Footer"
+            name={generatedDocGen?.TableFooter?.displayName}
+            id={generatedDocGen?.TableFooter?.displayName}
           />
         </MainSection.Subsection>
         <MainSection.Subsection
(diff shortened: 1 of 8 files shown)


docs/pages/web/table.js
```diff
@@ -1599,6 +1599,16 @@ function Example() {
             id="Table.RowExpandable"
           />
         </MainSection.Subsection>
+        <MainSection.Subsection
+          title={generatedDocGen?.TableRowDrawer?.displayName}
+          description={generatedDocGen?.TableRowDrawer?.description}
+        >
+          <GeneratedPropTable
+            generatedDocGen={generatedDocGen.TableRowDrawer}
+            name="Table.RowDrawer"
```
dangerismycat, line 1608: `name` and `id` can both also use displayName

packages/gestalt/src/TableRowDrawer.js
```diff
@@ -0,0 +1,67 @@
...
+
+/**
+ * Use [Table.RowDrawer](https://gestalt.pinterest.systems/web/table#Table.RowDrawer) to define a row drawer to display additional content.
+ */
+export default function TableRowDrawer({ children, drawerContents, id }: Props): Node {
+  const { stickyColumns } = useTableContext();
+  const rowRef = useRef();
+  const [columnWidths, setColumnWidths] = useState([]);
+
+  const cs = cx(styles.drawer);
```
dangerismycat, line 31: This isn't necessary — you can pass `styles.drawer` directly where `cs` is used

packages/gestalt/src/TableRowExpandable.js
```diff
@@ -58,7 +58,9 @@ export default function TableRowExpandable({
   hoverStyle = 'gray',
 }: Props): Node {
   const [expanded, setExpanded] = useState(false);
-  const cs = hoverStyle === 'gray' ? cx(styles.hoverShadeGray) : null;
+  const cstr = hoverStyle === 'gray' ? cx(styles.hoverShadeGray) : null;
```
dangerismycat, line 61: `classnames` isn't doing anything here — just use the class names directly where needed

`classnames` is only useful when combining multiple classes, especially if there's conditional logic for which classes are included

dangerismycat (approved):
Can you add screenshots of the new/changed UI to the PR summary?

docs/pages/web/table.js
AlbertCarreras: DONE applied retroactively

packages/gestalt/src/TableRowDrawer.js
AlbertCarreras: Didnt know this.
DONE

packages/gestalt/src/TableRowExpandable.js
AlbertCarreras: DONE

PR #2491, opened 2022-11-10
Badge: new experimental Badge redesign

AlbertCarreras:
#### What changed?

Badge: new experimental Badge redesign

| Where | ### Before      | ### After |
| Standalone | Screen Shot 2022-11-14 at 11 12 46 AM | Screen Shot 2022-11-14 at 11 12 08 AM |
| Standalone | Screen Shot 2022-11-14 at 4 44 02 PM | Screen Shot 2022-11-14 at 4 42 08 PM |
| PageHeader | Screen Shot 2022-11-14 at 4 46 00 PM |  Screen Shot 2022-11-14 at 7 18 41 PM  |
| DataPoint | Screen Shot 2022-11-14 at 7 20 08 PM | Screen Shot 2022-11-14 at 7 19 47 PM |
| Module | Screen Shot 2022-11-14 at 7 21 26 PM | Screen Shot 2022-11-14 at 7 21 15 PM |
| Dropdown | Screen Shot 2022-11-14 at 7 22 45 PM | Screen Shot 2022-11-14 at 7 22 30 PM |
| Sidenavigation | Screen Shot 2022-11-14 at 7 23 59 PM | Screen Shot 2022-11-14 at 7 24 31 PM |

### Stats
Screen Shot 2022-11-14 at 11 15 45 AM
Screen Shot 2022-11-14 at 11 16 55 AM

Total info Badges: 49 instances

### Links

- [Jira](https://jira.pinadmin.com/browse/GESTALT-4090)
- [Figma](https://www.figma.com/file/t0WpdhtjU244qaAiiBqoQs/Component-%7C-Badge?node-id=304%3A421)

### Checklist


docs/docs-components/contexts/DocsExperimentProvider.js
```diff
@@ -8,7 +8,7 @@ import { ExperimentProvider } from 'gestalt';
  * - Unless you want the experimental behavior live on the docs for everyone, REMOVE YOUR EXPERIMENT HERE before merging your PR!
  * */
 
-const enabledExperiments = [];
+const enabledExperiments = ['web_gestalt_redesigned_badge', 'mweb_gestalt_redesigned_badge'];
```
dangerismycat, line 11: See the note above — we should remove these so that users aren't confused why they're seeing different UI

docs/pages/web/badge.js
```diff
@@ -18,6 +18,20 @@ export default function BadgePage({ generatedDocGen }: {| generatedDocGen: DocGe
         defaultCode={`
 <Text>Update your pronouns in your profile settings <Badge text="New" /></Text>
     `}
+        slimBanner={
+          <SlimBanner
+            iconAccessibilityLabel="Component under experiment"
+            message={`Badge is under an experiment to update its visual design: font size, spacing, and a new information Icon and Tooltip for type="info". Badge changes are active on the Gestalt Docs.`}
```
dangerismycat, line 24: I don't think we'll want to show this yet until we provide the user a way to also see the non-experimental treatment

packages/gestalt/src/Badge.js
```diff
@@ -25,6 +38,10 @@ type Props = {|
    * Text displayed inside of the Badge. Sentence case is preferred.
    */
   text: string,
+  /**
+   *  Experimental prop, do not use. Adds a [Tooltip](/web/tooltip) on hover/focus of the Badge. To convey the interaction, it also displays an information Icon. See the [type](https://gestalt.pinterest.systems/web/badge#Type) variant to learn more.
```
dangerismycat, line 42: If this prop is experimental, let's preface it with an underscore. That will also automatically remove it from the generated props table.

packages/gestalt/src/Badge.js
```diff
@@ -38,7 +55,7 @@ type Props = {|
  * ![Badge dark mode](https://raw.githubusercontent.com/pinterest/gestalt/master/playwright/visual-test/Badge-dark.spec.mjs-snapshots/Badge-dark-chromium-darwin.png)
  *
  */
-export default function Badge({ position = 'middle', text, type = 'info' }: Props): Node {
+export default function Badge({ position = 'middle', text, type = 'info', tooltip }: Props): Node {
   const TYPE_COLOR_MAP = {
```
dangerismycat, line 59: unrelated nit: this object should be defined outside of the function so we're not recreating it on every render

packages/gestalt/src/Badge.js
```diff
@@ -50,10 +67,60 @@ export default function Badge({ position = 'middle', text, type = 'info' }: Prop
...
+    [colors.recommendationBase]: type === 'recommendation',
+    [colors.washDark]: type === 'darkWash',
+    [colors.washLight]: type === 'lightWash',
+  });
+
+  if (inBadgeExp) {
+    const badgeComponent = (
+      <div className={csNewBadge}>
+        <Flex alignItems="center" gap={1}>
+          {type === 'info' && tooltip?.text ? (
```
dangerismycat, line 98: nit: we should be able to make this check once and store it as `shouldUseTooltip` or something so this logic isn't repeated

dangerismycat (approved):
Thanks for including the before/after table!

I'd argue that this is not a `major` change — it's not a breaking change, as nothing will be different unless the user activates the experiment. I would consider it to be `minor`.

packages/gestalt/src/Badge.css
```diff
@@ -11,6 +11,31 @@
...
+}
+
+.info {
+  background-color: var(--g-blue);
+}
+
+.interactiveInfo {
+  background-color: var(--g-blue);
+}
+
```
ayeshakmaz, line 33: these can be deleted with the token addition

packages/gestalt/src/Badge.css
```diff
@@ -11,6 +11,31 @@
...
+.interactiveInfo:hover,
+.interactiveInfo:focus {
+  background-color: var(--g-blueHovered);
```
ayeshakmaz, line 36: And this should use the new token

packages/gestalt/src/Badge.css
```diff
@@ -11,6 +11,31 @@
   white-space: nowrap;
 }
 
+.NewBadge {
+  composes: antialiased sansSerif from "./Typography.css";
+  composes: inlineBlock from "./Layout.css";
+  composes: rounding1 from "./Borders.css";
+  composes: fontWeightSemiBold from "./Typography.css";
+  color: var(--color-text-inverse);
+  font-size: 14px;
```
ayeshakmaz, line 20: This should use a token

AlbertCarreras:
> Thanks for including the before/after table!
> 
> I'd argue that this is not a `major` change — it's not a breaking change, as nothing will be different unless the user activates the experiment. I would consider it to be `minor`.

Yeah, agree. It was major because I had a efault label, but after discussing w Ayesha we removed it.

docs/docs-components/contexts/DocsExperimentProvider.js
AlbertCarreras: This will be replaced with logic from https://github.com/pinterest/gestalt/pull/2511

docs/pages/web/badge.js
AlbertCarreras: ditto

packages/gestalt/src/Badge.js
AlbertCarreras: DONE

packages/gestalt/src/Badge.js
AlbertCarreras: DONE

packages/gestalt/src/Badge.js
AlbertCarreras: DONE

keeping last one for Flow

packages/gestalt/src/Badge.js
AlbertCarreras: DONE

packages/gestalt/src/Badge.js
AlbertCarreras: DONE

packages/gestalt/src/Badge.css
AlbertCarreras: DONE

packages/gestalt/src/Badge.css
AlbertCarreras: DONE

PR #2493, opened 2022-11-14
Design tokens: add Flow declaration files to JSON files

christianvuerings:
#### What changed?

Add Flow definitions to the design tokens JSON files

#### Why?

Previously these files were untyped which resulted in us having to ignore these errors

### Links

- Flow declaration files: https://flow.org/en/docs/declarations/


AlbertCarreras (approved):

dangerismycat (commented):
Thank you Christian! ❤️

PR #2503, opened 2022-11-16
Docs: fix GeneratedPropTable regex

AlbertCarreras:
#### What changed?

Docs: fix GeneratedPropTable regex

### DIFF
--- a/docs/docs-components/GeneratedPropTable.js
+++ b/docs/docs-components/GeneratedPropTable.js
@@ -104,7 +104,7 @@ export default function GeneratedPropTable({
         .replace(/ComponentType/g, 'React.ComponentType')
         // Replace "Element" with "React.Element" to match docs convention
         // Includes `<` to avoid picking up `HTMLDivElement` and similar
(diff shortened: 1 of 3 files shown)


docs/docs-components/GeneratedPropTable.js
```diff
@@ -104,7 +104,7 @@ export default function GeneratedPropTable({
         .replace(/ComponentType/g, 'React.ComponentType')
         // Replace "Element" with "React.Element" to match docs convention
         // Includes `<` to avoid picking up `HTMLDivElement` and similar
-        .replace(/^Element</g, 'React.Element<')
+        .replace(/Element</g, 'React.Element<')
```
dangerismycat, line 107: Huh. Not only was I wrong about possibly needing more targeted logic here, this is actually catching some cases I'd missed! 🎉 
Screen Shot 2022-11-16 at 2 09 18 PM
Screen Shot 2022-11-16 at 2 10 32 PM

dangerismycat (approved):

PR #2504, opened 2022-11-16
Internal: cleanup/refactor how we handle autogen props

dangerismycat:
- Get the "responsive" part of the prop table working again (we don't actually use this currently, but at least it works again in case we decide to bring it back)
- Remove the "href" override and refactor the last few usages of it
- Refactor how we handle overrides for autogen metadata to be more scalable
- Clean up some cruft related to autogen metadata

### DIFF
--- a/docs/docs-components/GeneratedPropTable.js
+++ b/docs/docs-components/GeneratedPropTable.js
@@ -9,12 +9,12 @@ function getResponsive(description?: string): {|
   responsive?: boolean,
 |} {
   const input = description ?? '';
-  const match = input.match(/(?<main>Responsive: (?<typeOverride>.*))/);
+  const match = input.match(/(?<main>Responsive: (?<responsive>.*))/);
   const groups = match?.groups ?? {};
 
   return {
     description: groups.main ? input.replace(groups.main, '') : description,
-    responsive: Boolean(groups.typeOverride?.replace(/'/g, '')),
+    responsive: Boolean(groups.responsive?.replace(/'/g, '')),
   };
 }
 
--- a/docs/docs-components/GeneratedPropTable.js
+++ b/docs/docs-components/GeneratedPropTable.js
@@ -48,22 +48,6 @@ function getDefaultValue(description?: string): {|
   };
 }
 
-// Support the older-style links in props, where the prop name is a link
-// As of Nov '22, only used in Module and Text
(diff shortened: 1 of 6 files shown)


docs/docs-components/GeneratedPropTable.js
```diff
@@ -75,22 +96,21 @@ export default function GeneratedPropTable({
         return null;
       }
 
-      // Remove domain so local development (localhost) isn't broken
-      const descriptionWithoutDomain = description?.replace(
-        /https:\/\/gestalt\.pinterest\.systems/g,
-        '',
+      const { description: descriptionWithoutOverrides, ...overrides } = [
+        removeDomain,
```
dangerismycat, line 100: This will be a lot more scalable. Any new override functions simply need to return the adjusted description and the added field, then add that function here

packages/gestalt/src/Modal.js
```diff
@@ -49,16 +49,12 @@ type Props = {|
   onDismiss: () => void,
   /**
    * The underlying ARIA role for the Modal. See the [Accessibility Role section](https://gestalt.pinterest.systems/web/modal#Role) for more info.
-   *
-   * Default: 'dialog'
```
dangerismycat, line 53: These weren't necessary, default value comes in normally

docs/docs-components/GeneratedPropTable.js
```diff
@@ -32,19 +48,8 @@ function getDefaultValue(description?: string): {|
   };
 }
 
-// Support the older-style links in props, where the prop name is a link
-function getHref(description?: string): {|
```
dangerismycat, line 36: No longer needed after refactors in Module, ScrollBoundaryContainer, and Text

packages/gestalt/src/Module.js
```diff
@@ -13,53 +13,36 @@ type BadgeType = {|
 |};
 
 type Props = {|
-  //
   /**
-   * Add a badge displayed after the title. Will not be displayed if `title` is not provided. Not to be used with `icon` or `iconButton`. Be sure to localize the text.
-   *
-   * Link: https://gestalt.pinterest.systems/web/text#static-badge
```
dangerismycat, line 20: Removed all these old-style links, many of which weren't correct anyway 😬

PR #2507, opened 2022-11-17
Toast: restricting `text` prop from Node to only string and Text

AlbertCarreras:
### Breaking change

Run the following codemod to list and manually replace those Node elements in `text` prop with Text

--component=Toast
--prop=text

#### What changed?

Toast: restricting `text` prop from Node to only string and Text

### DIFF
--- a/docs/pages/web/toast.js
+++ b/docs/pages/web/toast.js
@@ -64,6 +64,11 @@ export default function DocsPage({ generatedDocGen }: {| generatedDocGen: DocGen
 
       <AccessibilitySection name={generatedDocGen?.displayName} />
 
+      <MainSection
+        name="Localization"
+        description={`Remember to localize \`text\` and any string within \`primaryAction\`.`}
+      />
+
       <MainSection name="Variants">
         <MainSection.Subsection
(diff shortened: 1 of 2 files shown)


docs/pages/web/toast.js
```diff
@@ -125,8 +130,12 @@ function ToastExample() {
         </MainSection.Subsection>
 
         <MainSection.Subsection
-          description="When passing in your own Text component for `text`, do not specify `color` on Text. Toast will automatically pick the correct text color for the given `variant`."
-          title="Complex Text"
+          description={`
+The \`text\` prop accepts either a string or [Text](/Text). Use a string for guide toasts without any visual style. Toast will handle the text style and adherence to design guidelines.
+
+If a confirmation toasts text with more complex style is required, such as bold text or inline links, use Text to wrap your message with any additional Text or Link usages contained within. When passing in your own Text component for \`text\`, do not specify \`color\` on Text. Toast will automatically pick the correct text color for the given \`variant\`.
```
dangerismycat, line 136: typo: "a confirmation _toast's_ text with more complex…"

Also: not wanting users to specify `color` on Text sounds like something we should enforce with a lint rule

dangerismycat (approved):

docs/pages/web/toast.js
AlbertCarreras: DONE

PR #2509, opened 2022-11-17
Playwright: updating config to run on macOS 12 to fix snapshot differences due to different macOS versions between local and CI, & disabling them in CI

AlbertCarreras:
Playwright: updating config to run on macOS 12 to fix snapshot differences due to different macOS versions between local and CI #2509

Previously Playwright was running on macOS latest but Github uses macOS 11 for latest rather than 12.
Screen Shot 2022-11-17 at 3 24 56 PM 
https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources 

Most of Gestalt engs work on macOS 12.

This version difference was affecting the line height in some components causing many visual diff tests to fail.

Also, moving _DocsDeviceTypeProvider_ from "_app" to "App" so experiments don't get reflected in visual diff tests.

Screen Shot 2022-11-18 at 11 32 10 AM


dangerismycat (approved):
Thanks for this! Good idea to peg to 12 specifically, since IT determines the macOS version we're on so it's good to match that specifically

package.json
```diff
@@ -115,7 +115,7 @@
     "lint:css": "stylelint --max-warnings 0 \"**/*.css\"",
     "lint:js": "eslint --max-warnings 0 .",
     "lint:sh": "shellcheck $(find . -type f -iname '*.sh' ! -path '*/node_modules/*' ! -path '*/.git/*')",
-    "playwright:test": "npx playwright test --config=playwright/playwright.config.js",
+    "playwright:test": "npx playwright test playwright/accessibility/ --config=playwright/playwright.config.js",
```
AlbertCarreras, line 118: here it's were the vis snap shots test dont run

PR #2510, opened 2022-11-17
Internal: update visual snapshots to macos-12

christianvuerings:
#### What changed?

All developers upgraded to `macos-12` but our visual snapshot tests are still for `macos-11`.

#### Why?

So it's easier to update our snapshots locally

### Links

https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources

image


dangerismycat:
Is this a duplicate of #2509?

PR #2511, opened 2022-11-18
Docs: added cookies to support activation/deactivation of experiments from ExperimentProvidern in Docs

AlbertCarreras:
#### What changed?

Docs: added cookies to support activation/deactivation of experiments from ExperimentProvidern in Docs

#### Why?
Be able to support SlimBanner's primaryAction to toggle between active/inactive experiments to visualize changes from experiments.

Implemented in Textfield. See https://deploy-preview-2511--gestalt.netlify.app/web/textfield

Next ones:
Screen Shot 2022-11-18 at 1 12 00 PM
Screen Shot 2022-11-18 at 1 56 21 PM

### Links

- [TDD](link to Paper doc)
- [Figma](link to Figma file)

### Checklist

### DIFF
--- a/docs/pages/web/badge.js
+++ b/docs/pages/web/badge.js
@@ -6,7 +6,6 @@ import docgen, { type DocGen } from '../../docs-components/docgen.js';
 import PageHeader from '../../docs-components/PageHeader.js';
 import MainSection from '../../docs-components/MainSection.js';
 import QualityChecklist from '../../docs-components/QualityChecklist.js';
-
 import AccessibilitySection from '../../docs-components/AccessibilitySection.js';
 
 export default function BadgePage({ generatedDocGen }: {| generatedDocGen: DocGen |}): Node {
--- a/docs/pages/web/textfield.js
+++ b/docs/pages/web/textfield.js
@@ -6,8 +6,11 @@ import Page from '../../docs-components/Page.js';
 import GeneratedPropTable from '../../docs-components/GeneratedPropTable.js';
 import docgen, { type DocGen } from '../../docs-components/docgen.js';
 import QualityChecklist from '../../docs-components/QualityChecklist.js';
-
 import AccessibilitySection from '../../docs-components/AccessibilitySection.js';
+import {
+  BareSlimBannerExperiment,
+  SlimBannerExperiment,
+} from '../../docs-components/SlimBannerExperiment.js';
 
 export default function TextFieldPage({ generatedDocGen }: {| generatedDocGen: DocGen |}): Node {
   return (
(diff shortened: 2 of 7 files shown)


docs/pages/web/badge.js
```diff
@@ -6,7 +6,6 @@ import docgen, { type DocGen } from '../../docs-components/docgen.js';
 import PageHeader from '../../docs-components/PageHeader.js';
 import MainSection from '../../docs-components/MainSection.js';
 import QualityChecklist from '../../docs-components/QualityChecklist.js';
-
```
dangerismycat, line 9: unrelated change, should be reverted for a cleaner PR

docs/pages/web/textfield.js
```diff
@@ -36,6 +39,35 @@ function Example(props) {
...
+                experiments === generatedDocGen?.displayName
+                  ? 'Deactivate component experiments in the Docs'
+                  : 'Activate component experiments in the Docs',
+              label:
+                experiments === generatedDocGen?.displayName
+                  ? 'Deactivate experiments'
+                  : 'Activate experiments',
+              onClick: () =>
+                setExperiments(
+                  experiments === generatedDocGen?.displayName ? '' : generatedDocGen?.displayName,
```
dangerismycat, line 65: It looks like this is all-or-nothing, yes? We should provide users more granularity here. Users should be able to toggle individual experiments.

docs/pages/web/textfield.js
```diff
@@ -543,6 +575,45 @@ function Example(props) {
           />
         </MainSection.Subsection>
 
+        <MainSection.Subsection title="Password">
+          <SlimBanner
+            iconAccessibilityLabel="Component under experiment"
+            message={
+              experiments === generatedDocGen?.displayName
+                ? 'The current Textfield example is displaying experimental changes.'
```
dangerismycat, line 583: nit: consider consolidating this text by using a template literal and moving the logic to where it's needed, e.g.
```jsx
message={`The current TextField example is ${experiments === generatedDocGen?.displayName ? '' : 'NOT '}displaying experimental changes.`}
```

docs/pages/web/textfield.js
AlbertCarreras: I tried stringifying an array of experiments but it couldn't get to work. It's not like we have >1 experiments at the time for each single component. This is a first iteration of this implementation. With cmp name I hit both desktop and web experiments in a clean way. 

If there's need for refactoring this to allow overlapping experiments we can rethink the implementation.

AlbertCarreras:
This PR could have been R+ed. @dangerismycat

docs/pages/web/textfield.js
AlbertCarreras: DONE

docs/pages/web/badge.js
dangerismycat: Right, it's cleanup that can be addressed the next time someone does work on this file. As it stands, anyone looking into the git history of this file will see this PR, adding noise to their search as this PR has no material changes to the file.

docs/pages/web/textfield.js
dangerismycat: What didn't work about stringifying an array or object of experiments?

PR #2513, opened 2022-11-18
Design Tokens: Add new info background color

ayeshakmaz:
#### What changed?

In order to support an upcoming interactive Badge, add a color token to use for hover situations

(also removes timestamps from Android files per their request)


AlbertCarreras (approved):

packages/gestalt-design-tokens/config.json
```diff
@@ -98,6 +98,7 @@
     },
     "android": {
       "transformGroup": "android-custom",
+      "options": { "fileHeader": "androidCustomHeader" },
```
AlbertCarreras, line 101: what's this?

AlbertCarreras (approved):

packages/gestalt-design-tokens/config.json
ayeshakmaz: Update to the file headers for android to remove the timestamp

PR #2514, opened 2022-11-18
Modal, Sheet: Update shadows to new tokens

ayeshakmaz:
#### What changed?

Utilize new raised shadow options for Sheet and Modal headers and footers


AlbertCarreras:
@ayeshakmaz Can we add a screenshot if there's a visual difference between the two? I'm just curious

AlbertCarreras (approved):

PR #2518, opened 2022-11-22
Toast: new Toast design adjustments

AlbertCarreras:
#### What changed?

Toast: new Toast design adjustments

All changes behind experiments
| change      |  BEFORE | AFTER |
| ----------- | ----------- |  ----------- |
| Width under sm breakpoint: Fix so that below smallest breakpoint, width gets set to "100%"        |   Screen Shot 2022-11-28 at 5 12 47 PM Kapture 2022-11-28 at 11 45 45     |    Kapture 2022-11-28 at 11 41 54    |
| lineClamp={2}       |    Screen Shot 2022-11-28 at 5 13 18 PM    |   Screen Shot 2022-11-28 at 5 11 44 PM    |
| marginStart/End to 8 boints   |        |    Screen Shot 2022-11-28 at 5 12 05 PM   |
| start alignment of Text   |   Screen Shot 2022-11-28 at 5 30 40 PM |    Screen Shot 2022-11-28 at 5 29 17 PM |
| error icon size to 24 px   |   |    |

### Links

- [Jira](https://jira.pinadmin.com/browse/GESTALT-4028)
- [TDD](https://paper.dropbox.com/doc/Gestalt-TDD-Toast--BqZRSpB7ziwSUFAwTIGjaPx2Ag-EC6LAr0uO808c6GrjH9FO)
- [Figma](http://localhost:8888/web/toast#Error)

### DIFF
--- a/packages/gestalt/src/Toast.js
+++ b/packages/gestalt/src/Toast.js
@@ -17,7 +17,7 @@ const TOAST_MAX_WIDTH_PX = 500;
 const TOAST_WIDTH_PX = 330;
 
 const EXP_SIZE_THUMBNAIL = 32;
-const EXP_TOAST_MAX_WIDTH = 716;
+const EXP_SIZE_ICON = 24;
 
 type Props = {|
   /**
@@ -96,8 +96,11 @@ export default function Toast({
     let textColorOverrideStyles = isDarkMode
       ? styles.textColorOverrideDark
       : styles.textColorOverrideLight;
+
     if (isErrorVariant) {
-      textColorOverrideStyles = styles.textColorOverrideLight;
+      textColorOverrideStyles = inToastExp
+        ? styles.textErrorColorOverrideLight
+        : styles.textColorOverrideLight;
     }
 
     textElement = <span className={textColorOverrideStyles}>{text}</span>;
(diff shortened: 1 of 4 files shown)


ponori:
This may be the best PR is the history of PRs...

packages/gestalt/src/Toast.js
```diff
@@ -112,69 +112,86 @@ export default function Toast({
...
     );
+    return (
+      <Fragment>
+        <Box
+          // Ensure that maxWidth isn't greater than viewport width (for small screens)
+          maxWidth={`min(${EXP_TOAST_MAX_WIDTH}px, 100vw)`}
+          role="status"
+          rounding={4}
+          smDisplay="none"
+          width="100%" // only full width below the smallest breakpoint
```
ayeshakmaz, line 179: Can you elaborate on what's being solved here, maybe with screenshots? I'm not sure I follow why this update is needed

packages/gestalt/src/Toast.js
```diff
@@ -112,69 +116,89 @@ export default function Toast({
...
+          marginStart={8}
+          maxWidth={`min(${EXP_TOAST_MAX_WIDTH}px, 100vw)`} // Ensure that maxWidth isn't greater than viewport width (for small screens)
+          role="status"
+          rounding={4}
+          smDisplay="none"
+          width="100%" // only full width below the smallest breakpoint
+        >
+          {toastContent}
+        </Box>
+        <Box
```
dangerismycat, line 189: Alberto's right, this is how the responsive `display` props are intended to be used. I've never particularly liked it. 😅 But it does seem like the best solution here.

To dedupe props: though we lint against it in Pinboard, I would create a `sharedProps` object and spread it over Boxes. That does make future codemods slightly less efficient (upgrades here will have to be made manually), but avoids drift between the two Boxes, which seems like the more likely source of bugs.

dangerismycat (approved):
Before/After table with gifs is 😍

packages/gestalt/src/Toast.js
AlbertCarreras: Just moved everything to a div. 

margins were simmetric so no concerns re RTL

PR #2519, opened 2022-11-22
Table.Drawer: fix subcomponent so no empty drawer shows no extra space

AlbertCarreras:
#### What changed?

Table.Drawer: fix subcomponent so no empty drawer shows no extra space

Screen Shot 2022-11-22 at 1 22 57 PM

After dismissing SlimBanner

### BEFORE

Screen Shot 2022-11-22 at 1 22 51 PM

### AFTER

Screen Shot 2022-11-22 at 1 20 07 PM

### DIFF
--- a/packages/gestalt/src/TableRowDrawer.js
+++ b/packages/gestalt/src/TableRowDrawer.js
@@ -49,14 +49,16 @@ export default function TableRowDrawer({ children, drawerContents, id }: Props):
 
   return (
     <Fragment>
-      <tr aria-details={id} ref={rowRef}>
+      <tr aria-details={drawerContents ? id : undefined} ref={rowRef}>
         {Number(stickyColumns) > 0 ? Children.map(children, renderCellWithAdjustedIndex) : children}
       </tr>
-      <tr id={id}>
-        <td className={styles.drawer} colSpan={Children.count(children) + 1}>
-          <Box padding={2}>{drawerContents}</Box>
-        </td>
-      </tr>
+      {drawerContents ? (
+        <tr id={id}>
+          <td className={styles.drawer} colSpan={Children.count(children) + 1}>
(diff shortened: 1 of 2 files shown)


packages/gestalt/src/TableRowDrawer.js
```diff
@@ -54,7 +54,7 @@ export default function TableRowDrawer({ children, drawerContents, id }: Props):
       </tr>
       <tr id={id}>
         <td className={styles.drawer} colSpan={Children.count(children) + 1}>
-          <Box padding={2}>{drawerContents}</Box>
+          <Box padding={drawerContents ? 2 : 0}>{drawerContents}</Box>
```
ayeshakmaz, line 57: If there is no drawer content, should this <tr> even be rendered?

ayeshakmaz, line 57: If there is no drawer content, should this even be rendered?

AlbertCarreras: @ayeshakmaz Indeed. Good observation.

DONE

packages/gestalt/src/TableRowDrawer.js
```diff
@@ -52,11 +52,13 @@ export default function TableRowDrawer({ children, drawerContents, id }: Props):
       <tr aria-details={id} ref={rowRef}>
```
ayeshakmaz, line 52: This id should be left off if there are no drawerContents anymore

AlbertCarreras: Good a11y catch

AlbertCarreras: DONE

ayeshakmaz (approved):

PR #2520, opened 2022-11-22
SlimBanner: implement below small breakpoint start alignment of elements

AlbertCarreras:
#### What changed?

SlimBanner: implement below small breakpoint start alignment of elements

NOTE: The implementation should be refactor taking advantage of Flex responsive props.

#### BEFORE

Screen Shot 2022-11-22 at 3 12 41 PM

#### AFTER

screen_shot_2022-11-22_at_3 03 27_pm

### Links

- [Jira](https://jira.pinadmin.com/browse/GESTALT-5068)

### DIFF
--- a/packages/gestalt/src/SlimBanner.js
+++ b/packages/gestalt/src/SlimBanner.js
@@ -176,68 +176,97 @@ export default function SlimBanner({
   // Buttons not allowed on compact SlimBanners
   const shouldShowButtons = !isBare && (primaryAction || dismissButton);
 
-  return (
-    <Box
-      alignItems="center"
-      color={isBare ? 'transparent' : backgroundColor}
-      display="flex"
-      direction="column"
-      mdDirection="row"
-      padding={isBare ? 0 : 4}
-      paddingY={isBare ? 1 : 0}
-      rounding={4}
-      width="100%"
-    >
-      <Flex alignItems="center" gap={{ row: isBare ? 2 : 4, column: 0 }} flex="grow" width="100%">
-        {!isDefault && iconAccessibilityLabel && (
-          <Flex.Item alignSelf={shouldShowButtons ? undefined : 'start'}>
-            <Icon
-              accessibilityLabel={iconAccessibilityLabel}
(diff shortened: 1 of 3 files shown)


packages/gestalt/src/SlimBanner.js
```diff
@@ -176,68 +176,95 @@ export default function SlimBanner({
...
-              size={16}
-            />
-          </Flex.Item>
-        )}
+  const startIcon =
+    !isDefault && iconAccessibilityLabel ? (
+      <Flex.Item alignSelf={shouldShowButtons ? undefined : 'start'}>
+        <Icon accessibilityLabel={iconAccessibilityLabel} color={iconColor} icon={icon} size={16} />
+      </Flex.Item>
+    ) : null;
```
AlbertCarreras, line 184: this null is needed, it was just added. prevent ghost spacing

ponori (approved):
This is looking really good from a design perspective Alberto  - nice work!

I think there's still some room to noodle on the spacing/padding, but I'd argue that's outside the scope of this work.

packages/gestalt/src/SlimBanner.js
```diff
@@ -176,68 +176,97 @@ export default function SlimBanner({
...
+  ) : null;
+
+  const actions = !isBare && primaryAction && (
+    <Box display="flex" mdDisplay="none" flex="none" alignSelf="end" marginTop={4}>
+      <PrimaryAction {...primaryAction} />
     </Box>
   );
+
+  return (
+    <Fragment>
```
ayeshakmaz, line 228: This method is going to create issues for testing by duplicating all the elements in the DOM. Can we do this instead with CSS? I'll reach out tomorrow, but let's not merge this yet

ayeshakmaz (changes requested):

packages/gestalt/src/SlimBanner.js
AlbertCarreras: Ok... le me exploreee

PR #2521, opened 2022-11-23
Tag: fix background color and dark mode

ayeshakmaz:
#### What changed?

Fix color disparity and update truncated view in dark mode

Before

Screen Shot 2022-11-22 at 5 06 03 PM

After

Screen Shot 2022-11-22 at 5 05 47 PM


AlbertCarreras (approved):
LGTM except the snapshots

PR #2525, opened 2022-11-29
Toast, Spinner, DefaultLabelProvider: new component features from new redesign, new `color` prop for Spinner, new labels to DefaultLabelProvider

AlbertCarreras:
# BREAKING CHANGES

##### TOAST API CHANGES

| BEFORE | AFTER | codemod
| text | text | |
| _NA_ | dismissButton | |
| _NA_ | helperLink | |
| primaryAction | primaryAction | |
| thumbnail | thumbnail | |
| thumbnailShape | _deprecated_ | *1 |
| variant | type | *2 |

*1
```
--component=Toast
--previousProp=thumbnailShape
```

*2
```
--component=Toast
--previousProp=variant
--nextProp=type
```

Also, updated [DefaultLabelProvider](https://gestalt.pinterest.systems/web/utilities/defaultlabelprovider) for Toast

# Summary

#### What changed?

Toast: new component features from new redesign
Removal experiments and consolidate changes

##### Visual testing
| New design / feature | Image |
| ----------- | ----------- |
| Regular text   |  Screen Shot 2022-11-30 at 9 55 23 PM   |
| Text truncation, respects helperLink   | Screen Shot 2022-11-30 at 10 02 11 PM|
| Success, error, progress types | Screen Shot 2022-11-30 at 9 53 42 PM |
| Dismissable |   Screen Shot 2022-11-30 at 9 53 37 PM |

### Links

- [TDD](link to Paper doc)
- [Figma](link to Figma file)

### Checklist

### DIFF
--- a/docs/pages/web/toast.js
+++ b/docs/pages/web/toast.js
@@ -133,16 +133,18 @@ function ToastExample() {
 
         <MainSection.Subsection
           description={`
-The \`text\` prop accepts either a string or [Text](/web/text). Use a string for guide toasts without any visual style. Toast will handle the text style and adherence to design guidelines.
+The \`text\` prop accepts either a string or [Text](/web/text). Use a string for guide toasts without any visual style. Toast will handle the text style and adherence to design guidelines. Regular strings are subject to two-line truncation.
 
-If  confirmation toast's text with more complex style is required, such as bold text or inline links, use Text to wrap your message with any additional Text or Link usages contained within. When passing in your own Text component for \`text\`, do not specify \`color\` on Text. Toast will automatically pick the correct text color for the given \`variant\`.
+If  confirmation toast's text with more complex style is required, such as bold text, inline links, or no truncation, use Text to wrap your message with any additional Text or Link usages contained within. When passing in your own Text component for \`text\`, do not specify \`color\` on Text. Toast will automatically pick the correct text color for the given \`type\`.
 `}
           title="Text"
+          columns={2}
         >
           <MainSection.Card
-            cardSize="lg"
+            cardSize="md"
             defaultCode={`
-<Flex justifyContent="center" width="100%">
+<Flex direction="column" gap={4} justifyContent="center">
+  <Toast text="Signed as Three Crafty Ladies" />
   <Toast
     text={
       <Text inline>
@@ -156,36 +158,13 @@ If  confirmation toast's text with more complex style is required, such as bold 
 </Flex>
 `}
           />
-        </MainSection.Subsection>
-        <MainSection.Subsection title="Error">
           <MainSection.Card
-            cardSize="lg"
+            cardSize="md"
             defaultCode={`
-<Flex justifyContent="center" width="100%">
+<Flex direction="column" justifyContent="center" width="100%" gap={4}>
   <Toast
-    text="You're offline."
-    variant="error"
-    primaryAction={{ accessibilityLabel: 'Reconnect', label: 'Reconnect' }}
-  />
-</Flex>
-`}
-          />
-        </MainSection.Subsection>
-
-        <MainSection.Subsection title="Thumbnail">
-          <MainSection.Card
-            cardSize="lg"
-            defaultCode={`
-<Flex justifyContent="center" width="100%">
-  <Toast
-    text={
-      <Text inline>
-        Saved to{' '}
-        <Link display="inline" target="blank" href="https://www.pinterest.com/search/pins/?q=home%20decor">
-          Home decor
-        </Link>
-      </Text>
-    }
+    text="Your account admin rights were successfully saved"
+    primaryAction={{ accessibilityLabel: 'Undo', label: 'Undo' }}
     thumbnail={
       <Image
         alt="Modern ceramic vase pin."
@@ -195,9 +174,19 @@ If  confirmation toast's text with more complex style is required, such as bold 
       />
     }
   />
+  <Toast
+    text="Your account admin rights were successfully saved."
+    dismissButton={{ onDismiss: () => {} }}
+    helperLink={{
+      text: "Go to settings",
+      accessibilityLabel: "Go to settings",
+      href: "#",
+    }}
+  />
 </Flex>
 `}
           />
+          helperLink
         </MainSection.Subsection>
 
         <MainSection.Subsection title="Primary action">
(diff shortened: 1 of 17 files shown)


packages/gestalt/src/Toast.js
```diff
@@ -44,15 +68,14 @@ type Props = {|
   /**
    * An optional thumbnail image to displayed next to the text.
    */
-  thumbnail?: Node,
-  /**
-   * The masked shape of the thumbnail.
-   */
-  thumbnailShape?: 'circle' | 'square',
+  thumbnail?:
```
dangerismycat, line 71: Using an object here seems a bit belt-and-suspenders given the `displayName` checks. I think we're fine typing this as a simple union of the acceptable components and using `displayName` where we're checking the object key. (Unless I'm missing some wacky usage — I don't think there's anything like "SterlingAvatar" that people use, is there?)

AlbertCarreras: @dangerismycat The only caveat us that we are cloning them to override sizes. I think this is fine, was just curious if we thought maybe objects could be more "neat".

packages/gestalt/src/Spinner.js
```diff
@@ -34,6 +38,7 @@ type Props = {|
  */
 export default function Spinner({
   accessibilityLabel,
+  color = 'subtle',
```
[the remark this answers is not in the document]
AlbertCarreras: DONE

packages/gestalt/src/Toast.js
```diff
@@ -4,26 +4,50 @@ import Box from './Box.js';
...
    * Allows to insert a custom button for user interaction. Do not use except for allowed cases where primaryAction doesn't support functionality required in it.
    */
   _dangerouslySetPrimaryAction?: Node,
+  /**
+   * Adds a dismiss button to Toast. See the [Dismissible variant](https://gestalt.pinterest.systems/web/toast#Dismissible) for more info.
+   * The `accessibilityLabel` should follow the [Accessibility guidelines](https://gestalt.pinterest.systems/web/toast#Accessibility).
+   *
+   */
+  dismissButton?: {|
+    accessibilityLabel: string,
```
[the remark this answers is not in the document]
AlbertCarreras: Yes. DONE

packages/gestalt/src/Toast.js
```diff
@@ -63,27 +86,38 @@ type Props = {|
...
-  });
-
   const { name: colorSchemeName } = useColorScheme();
   const isDarkMode = colorSchemeName === 'darkMode';
-  const isErrorVariant = variant === 'error';
+
+  const isErrorToast = type === 'error';
+  const isProgressToast = type === 'progress';
+  const isSuccessToast = type === 'success';
+  const isNeutralToast = !isErrorToast && !isProgressToast && !isSuccessToast;
```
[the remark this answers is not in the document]
AlbertCarreras: DONE

packages/gestalt/src/Toast.js
```diff
@@ -63,27 +86,38 @@ type Props = {|
...
 
   const responsiveMinWidth = useResponsiveMinWidth();
   const isMobileWidth = responsiveMinWidth === 'xs';
 
-  let containerColor = isDarkMode ? 'light' : 'dark';
-  let textColor = isDarkMode ? 'dark' : 'light';
+  let containerColor = 'inverse';
```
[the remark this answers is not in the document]
AlbertCarreras: DONE

packages/gestalt/src/Image.js
```diff
@@ -100,6 +100,8 @@ export default class Image extends PureComponent<Props> {
     loading: 'auto',
   };
 
+  static displayName: ?string = 'Image';
```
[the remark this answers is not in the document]
dangerismycat: ah, I think is because of how React types `static displayName`:
https://github.com/facebook/flow/blob/v0.111.3/lib/react.js#L87

packages/gestalt/src/ToastComponents.js
```diff
@@ -0,0 +1,176 @@
+// @flow strict
```
dangerismycat, line 1: We can use nested directories now, so let's break these out into individual files in `packages/gestalt/src/Toast/`

packages/gestalt/src/Toast.js
dangerismycat: Hm…I'm not sure I see the need for the object? I think I'm missing something though. Let's discuss in next Tuesday's eng meeting? This is an interesting topic

packages/gestalt/src/ToastComponents.js
AlbertCarreras: Lets chat about this, I dont think it works when the subfile is accessing components: Box etc. Just if it doesnt

packages/gestalt/src/Toast.js
AlbertCarreras: Keeping as is as per https://flow.org/en/docs/types/unions/#toc-disjoint-object-unions

AlbertCarreras: After chatting with Jack

AlbertCarreras: we could use `thumbnail instanceof "Image"` to narrow type and it works. 

With Flow better not do much polymorphism with objects, usually disjoint object unions are recommended

docs/pages/web/toast.js
```diff
@@ -156,36 +158,13 @@ If  confirmation toast's text with more complex style is required, such as bold
...
-      <Text inline>
-        Saved to{' '}
-        <Link display="inline" target="blank" href="https://www.pinterest.com/search/pins/?q=home%20decor">
-          Home decor
-        </Link>
-      </Text>
-    }
+    text="Your account admin rights were successfully saved"
+    primaryAction={{ accessibilityLabel: 'Undo', label: 'Undo' }}
     thumbnail={
```
dangerismycat, line 168: This needs to change, yes? You may want to move the current Toast examples out to separate files to catch errors like this one.

docs/pages/web/toast.js
```diff
@@ -220,6 +209,98 @@ If  confirmation toast's text with more complex style is required, such as bold
 </Flex>`}
           />
         </MainSection.Subsection>
+
+        <MainSection.Subsection title="Dismissable">
```
dangerismycat, line 213: Instead of digging the hole deeper, can we create these new examples using Sandpack?

packages/gestalt/src/Toast.js
```diff
@@ -59,82 +103,90 @@ type Props = {|
...
+  const isTextNode = typeof text !== 'string' && Children.only(text).type.displayName === 'Text';
+
+  if (isTextNode) {
     let textColorOverrideStyles = isDarkMode
       ? styles.textColorOverrideDark
       : styles.textColorOverrideLight;
 
-    if (isErrorVariant) {
+    if (type === 'error') {
       textColorOverrideStyles = styles.textErrorColorOverrideLight;
```
dangerismycat, line 135: Not quite related to these changes, but do we need a dark mode error override here as well? (apologies if I've asked this before — though that would indicate that an explanatory comment here would be helpful)

packages/gestalt/src/Toast.js
```diff
@@ -40,15 +85,14 @@ type Props = {|
...
   /**
-   * The masked shape of the thumbnail.
+   * See the [type variant](https://gestalt.pinterest.systems/web/toast#Type) to learn more.
    */
-  thumbnailShape?: 'circle' | 'square',
-  /**
-   * Use the `'error'` variant to indicate an error message. Generally not recommended given the ephemeral nature of Toasts.
-   */
-  variant?: 'default' | 'error',
+  type?: 'success' | 'error' | 'progress',
```
dangerismycat, line 95: I think this needs a `default` to indicate the default state rather than relying on the absence of the prop (or passing a nullish value)

dangerismycat (commented):
I'm a bit hesitant to make breaking changes to multiple components in the same PR unless absolutely necessary. Can we separate out the Spinner change into a separate PR and merge/bump that one first?

docs/pages/web/toast.js
AlbertCarreras: This is coming afterwards https://github.com/pinterest/gestalt/pull/2551 I'll do that in that one

docs/pages/web/toast.js
AlbertCarreras: DONE

packages/gestalt/src/Toast.js
AlbertCarreras: DONE

packages/gestalt/src/Spinner.js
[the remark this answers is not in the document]
AlbertCarreras: To not make it a breaking change here, I'll invert this in a follow up rather a pre PR.

AlbertCarreras:
Tries

1. Screenshot 2023-03-09 at 10 47 39 AM
2. Screenshot 2023-03-09 at 10 50 52 AM

This always prints FALSE
Screenshot 2023-03-09 at 10 52 22 AM

packages/gestalt/src/Toast.js
AlbertCarreras: See https://github.com/pinterest/gestalt/pull/2525#issuecomment-1462300583

dangerismycat: Well dang. At least we tried! Thank you for documenting how each approach didn't work, much appreciated both for now and for the future.

PR #2526, opened 2022-11-29
Text: add forwardRef to component

AlbertCarreras:
#### What changed?

Text: add forwardRef to component

#### Why?

Accessing the ref can be helpful to detect if the text is being truncated. This can be helpful when the Text component is using text nodes rather than strings. If we detect truncation, we can manually set the title prop.

We need this change to detect truncation inside Toast (which remove the helperLink) and therefore we can stack the helperLink again back at the toast so not to lose functionality.

### Links

- [TDD](link to Paper doc)
- [Figma](link to Figma file)

### Checklist

### DIFF
--- a/packages/gestalt/src/Text.js
+++ b/packages/gestalt/src/Text.js
@@ -1,5 +1,5 @@
 // @flow strict
-import { type Node } from 'react';
+import { forwardRef, type AbstractComponent, type Element, type Node } from 'react';
 import cx from 'classnames';
 import colors from './Colors.css';
 import styles from './Text.css';
@@ -10,6 +10,7 @@ function isNotNullish(val): boolean {
   return val !== null && val !== undefined;
 }
 
+type As = 'span' | 'div';
 type Overflow = 'normal' | 'breakWord' | 'noWrap';
 type Size = '100' | '200' | '300' | '400' | '500' | '600';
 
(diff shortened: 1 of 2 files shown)


packages/gestalt/src/Text.js
```diff
@@ -10,6 +10,7 @@ function isNotNullish(val): boolean {
   return val !== null && val !== undefined;
 }
 
+type As = 'span' | 'div';
```
dangerismycat, line 13: 💯

packages/gestalt/src/Text.js
```diff
@@ -52,6 +53,10 @@ type Props = {|
    * Indicates how the text content should handle overflowing its container. See the [overflow and truncation variant](https://gestalt.pinterest.systems/web/text#Overflow-and-truncation) for more details.
    */
   overflow?: Overflow,
+  /**
+   * Ref that is forwarded to the underlying element. See the [ref variant](https://gestalt.pinterest.systems/web/text#Refs) for more details.
+   */
+  ref?: HTMLDivElement | HTMLSpanElement,
```
dangerismycat, line 59: Huh…usually we have to disable `react/no-unused-prop-types` as Flow expects `ref` to be destructured with the rest of the props. I wonder what the difference is here?

dangerismycat (approved):

packages/gestalt/src/Text.js
AlbertCarreras: I noticed that too. There are other examples where it doesn't complain.

PR #2530, opened 2022-11-30
Docs: Add Icon descriptions to library page

ayeshakmaz:
#### What changed?

Add a tooltip for descriptions of Icons on the Icon Library page

Screen Shot 2022-11-30 at 3 49 28 PM

#### Why?

Give designers and developers an example of when to use each icon

Note: Layla will be adding the rest of the descriptions in a follow up PR 

### Links

- [Figma](https://www.figma.com/file/RBlfOYmL123whCtlglXIsG/Iconography-page-redesign?node-id=360%3A3389&t=Ndz6EXOsnEahAIjm-0)

### Checklist

### DIFF
--- a/docs/pages/foundations/iconography/ICON_DATA.json
+++ b/docs/pages/foundations/iconography/ICON_DATA.json
@@ -2,11 +2,13 @@
   "icons": [
     {
       "name": "add",
-      "categories": "Add"
+      "categories": "Add",
+      "description": "Use this icon to indicate the ability to create or add content"
     },
     {
       "name": "add-circle",
-      "categories": "Add"
+      "categories": "Add",
+      "description": "Use this icon to indicate the ability to display more content or additional information"
     },
     {
       "name": "add-layout",
@@ -22,11 +24,13 @@
     },
     {
       "name": "ad",
-      "categories": "Ads and measurements"
+      "categories": "Ads and measurements",
+      "description": "Use this icon to indicate an ad type"
     },
     {
       "name": "ad-group",
-      "categories": "Ads and measurements"
+      "categories": "Ads and measurements",
+      "description": "Use this icon to indicate a group of ads accounts"
     },
     {
       "name": "ads-overview",
@@ -438,7 +442,8 @@
     },
     {
       "name": "accessibility",
-      "categories": "Utility and tools"
+      "categories": "Utility and tools",
+      "description": "Use this icon to indicate accessibility features or adjustments"
     },
     {
       "name": "apps",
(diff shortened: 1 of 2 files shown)


docs/pages/foundations/iconography/library.js
```diff
@@ -53,72 +54,79 @@ type IconData = {|
...
-      tapStyle="compress"
-      onTap={onTap}
-      onMouseEnter={() => setHovered(true)}
-      onMouseLeave={() => setHovered(false)}
-      onFocus={() => setHovered(true)}
-      onBlur={() => setHovered(false)}
-    >
-      <Box
-        borderStyle="sm"
+    <Tooltip text={iconDescription} accessibilityLabel={iconDescription} idealDirection="down">
```
AlbertCarreras, line 72: Should the a11yLabel just be "" if the tooltip text is the same?

docs/pages/foundations/iconography/ICON_DATA.json
```diff
@@ -2,11 +2,13 @@
   "icons": [
     {
       "name": "add",
-      "categories": "Add"
+      "categories": "Add",
+      "description": "Use this icon to indicate the ability to create or add content"
```
dangerismycat, line 6: Maybe it's more apparent when seeing all the messages together, but "use this icon" seems a bit repetitive and wordy. Perhaps just "Indicates the ability to…"?

docs/pages/foundations/iconography/library.js
```diff
@@ -53,72 +54,79 @@ type IconData = {|
     | 'Time'
     | 'Toggle'
     | 'Utility and tools',
+  description: string,
 |};
 
-function IconTile({ iconName, onTap }: {| iconName: IconName, onTap: () => void |}) {
+function IconTile({
+  iconName,
+  iconDescription = 'Description coming soon',
```
dangerismycat, line 62: Another possibility: conditionally wrap the content with Tooltip based on the presence of iconDescription (might be less noisy in the short term)

dangerismycat (approved):

AlbertCarreras (approved):

docs/pages/foundations/iconography/library.js
ayeshakmaz: The tooltip text only gets read on focus with keyboard if it's in the a11y label (otherwise folks would have to quick nav it, so I think this is a better experience)

docs/pages/foundations/iconography/ICON_DATA.json
ayeshakmaz: Hmm I'll check with Cintia on this one and let Layla know

docs/pages/foundations/iconography/library.js
ayeshakmaz: I thought about that, but given description is required, and Layla's making the updates right after, I think it'll be ok for a day

PR #2535, opened 2022-12-02
Sheet: reduce internal padding to from 8 to 6 boints

AlbertCarreras:
#### What changed?

Sheet: reduce internal padding to from 8 to 6 boints

#### Why?
Style unification betweem Modal/Sheet

### Links

- [Jira](https://jira.pinadmin.com/browse/GESTALT-5087) 
- [Figma](https://www.figma.com/file/vjhfBsOtHw0wVg67vqwz1v/Gestalt-for-web?node-id=2508%3A17902&t=y3qTRUfavn0IBoXJ-1)


packages/gestalt/src/Sheet.js
```diff
@@ -85,7 +85,7 @@ Internal components <Header> and <DismissButton>
 */
 function Header({ heading }: {| heading: string |}) {
   return (
-    <Box display="flex" justifyContent="start" padding={8}>
+    <Box display="flex" justifyContent="start" padding={6}>
```
dangerismycat, line 88: nit: if we think these padding values will always match, consider using a const

dangerismycat (approved):

PR #2536, opened 2022-12-02
Modal: add default padding to main content and heading and padding prop

AlbertCarreras:
# Breaking change

We are changing internal padding. To prevent styling inside Modal breaking, we should set padding="default" into all Modals to keep the same style logic as before.

We also have to remove any padding in custom headings

Run the following codemod to locate Modals in the codebase and update them manually

`yarn codemod detectManualReplacement ~/path/to/your/code --component=Modal
`

**Codebase impact magnitude**
Screen Shot 2022-12-02 at 2 47 20 PM

app/packages/gestaltExtensions/Modal/DesktopModal.js
app/packages/gestaltExtensions/Modal/MobileModal.js

# Summary

#### What changed?

Modal: add default padding to main content and heading and padding prop

Adding padding = {6} (24 px) in header and main content (footer already had). padding="none" removes padding = {6}  on main content

#### Why?

Style unification between Modal/Sheet

### Links

- [Jira](https://jira.pinadmin.com/browse/GESTALT-5098) 
- [Figma](https://www.figma.com/file/vjhfBsOtHw0wVg67vqwz1v/Gestalt-for-web?node-id=2508%3A17900&t=y3qTRUfavn0IBoXJ-1)

### DIFF
--- a/packages/gestalt/src/Modal.js
+++ b/packages/gestalt/src/Modal.js
@@ -40,10 +40,6 @@ type Props = {|
    */
   footer?: Node,
   /**
-   * The main Modal content has a default padding. For those cases where full bleed is needed, set `fullBleed` to "false".
-   */
-  fullBleed?: boolean,
-  /**
    * The text used for Modal's heading. See the [Heading variant](https://gestalt.pinterest.systems/web/modal#Heading) for more info.
    */
   heading?: Node,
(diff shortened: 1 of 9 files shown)


packages/gestalt/src/Modal.js
```diff
@@ -39,6 +39,10 @@ type Props = {|
    * Supply the element(s) that will be used as Modal's custom footer. See the [Best Practices](https://gestalt.pinterest.systems/web/modal#Best-practices) for more info.
    */
   footer?: Node,
+  /**
+   * The main Modal content has a default padding. For those cases where full bleed is needed, set `fullBleed` to "false".
+   */
+  fullBleed?: boolean,
```
dangerismycat, line 45: How about something like `padding?: 'default' | 'none'`?

ayeshakmaz: +1

AlbertCarreras: DONE

dangerismycat (approved):

PR #2545, opened 2022-12-07
Spinner: convert docs examples to Sandpack

dangerismycat:


docs/examples/spinner/main.js
```diff
@@ -0,0 +1,21 @@
...
+import { Fragment, type Node, useState } from 'react';
+import { Box, Button, Spinner } from 'gestalt';
+
+export default function Example(): Node {
+  const [show, setShow] = useState(true);
+
+  return (
+    <Fragment>
+      <Box padding={2}>
+        <Button
```
AlbertCarreras, line 11: Not sure if we need the Button hide/show functionality in the main example

AlbertCarreras (approved):

docs/pages/web/spinner.js
```diff
@@ -56,35 +45,25 @@ function SpinnerExample() {
...
+      <AccessibilitySection
+        name={generatedDocGen?.displayName}
+        description={`
+      Be sure to include \`accessibilityLabel\`. Labels should relate to the specific part of the product where Spinner is being used (e.g. "Loading homefeed" when used on the homefeed surface). Don't forget to localize the label!
+      `}
+      />
 
       <MainSection name="Variants">
-        <Example
+        <MainSection.Subsection
```
AlbertCarreras, line 57: Maybe we should add a variant "displayed" like the "paused" in Pulsar where we show the implementation.

docs/examples/spinner/main.js
dangerismycat: As with the other: agreed that it's not a pattern we use much, but this is how the docs have been

docs/pages/web/spinner.js
dangerismycat: Yeah, this page needs a lot more love. I'm going to leave as-is for now (the goal was just to convert the examples to Sandpack), but we should revisit

PR #2546, opened 2022-12-07
Pulsar: migrate docs examples to Sandpack

dangerismycat:


docs/examples/pulsar/size.js
```diff
@@ -0,0 +1,23 @@
...
+export default function Example(): Node {
+  const [sizePX, setSizePx] = useState();
+
+  return (
+    <Box padding={2}>
+      <Flex alignItems="center" direction="column" gap={4}>
+        <NumberField
+          id="size variant pixel input"
+          label="Enter Pulsar size in pixels"
+          min={20}
```
AlbertCarreras, line 14: max={200} otherwise it gets cut

docs/examples/pulsar/main.js
```diff
@@ -0,0 +1,19 @@
...
+
+export default function Example(): Node {
+  const [isPulsing, setIsPulsing] = useState(true);
+
+  const text = isPulsing ? 'Click to hide' : 'Click to show';
+
+  return (
+    <Box padding={2}>
+      <Flex alignItems="center" direction="column" gap={4}>
+        <Button text={text} onClick={() => setIsPulsing((currVal) => !currVal)} size="md" />
```
AlbertCarreras, line 13: same though, not sure about adding buttons in main example

AlbertCarreras (approved):

docs/examples/pulsar/main.js
dangerismycat: In the sense that interactive top examples isn't a common pattern in our docs, agreed. But this does faithfully reproduce what we currently have. ¯\_(ツ)_/¯ I don't have strong feelings either way

docs/examples/pulsar/positioning.js
```diff
@@ -0,0 +1,17 @@
...
+import { type Node } from 'react';
+import { Box, Button, Pulsar } from 'gestalt';
+
+export default function Example(): Node {
+  return (
+    <Box padding={2}>
+      <Box margin={12}>
+        <Button text="New feature - click me!" />
+
+        <Box position="absolute" top marginStart={8} marginTop={3}>
```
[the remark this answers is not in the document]
dangerismycat: holy cow that's so much more complex 😂 given the low usage of this component (most appearances in the app use that Experiences component I believe), I'm going to stick with this simple example for now. I've been chatting with PJ about improving/replacing this component and it sounds like there's interest from Growth teams, so I think unless we get specific requests we can make minimal changes for now and wait till a more substantive overhaul

docs/examples/pulsar/size.js
dangerismycat: good catch!

PR #2548, opened 2022-12-07
Tag: migrate docs examples to Sandpack

dangerismycat:


docs/examples/tag/dismissable.js
```diff
@@ -0,0 +1,40 @@
...
+          onClick={() => {
+            setTags((currTags) => [...currTags, generateTag()]);
+          }}
+          text="Add tag"
+        />
+
+        <Flex alignItems="center" gap={2} height="100%" justifyContent="center" width="100%" wrap>
+          {tags.map((item, index) => (
+            // eslint-disable-next-line react/no-array-index-key
+            <Fragment key={index}>{item}</Fragment>
```
AlbertCarreras, line 34: Why not setting the key on the <Item> itself?

AlbertCarreras (approved):

docs/examples/tag/dismissable.js
dangerismycat: The array is comprised of "rendered" Tags, so there isn't JSX to set the tag on. It's akin to the difference between something like
```js
[() => renderTag(), () => renderTag()]
```
and
```js
[renderTag(), renderTag()]
```

PR #2549, opened 2022-12-07
Docs: strip eslint disable comments from Sandpack examples

dangerismycat:


docs/removeFlowTypesLoader.js
```diff
@@ -2,24 +2,37 @@
...
   this.cacheable();
 
   const options = {};
-  const removed = flowRemoveTypes(source).toString().replace('//       strict', '');
+  // Remove Flow declarations/types
+  const removed = flowRemoveTypes(source)
+    .toString()
+    // Also remove what's left of the initial declaration comment
+    .replace('//       strict', '')
+    // Remove eslint disable comments
```
dangerismycat, line 18: This doesn't cover block-level disables. We don't currently have the need, but this can be updated if we do in the future

PR #2558, opened 2022-12-10
Docs: Add 2022 Year in Review page

ayeshakmaz:
New page for the year in review Recap for 2022

Figma: https://www.figma.com/file/xv1ICdQ3v6sdNEcA0u66i5/Gestalt-YIR-2022?node-id=8%3A1218&t=jQEAZMw4tJjMSvi6-0


docs/graphics/home-page/HeroGraphic.js
```diff
@@ -1,7 +1,7 @@
 // @flow strict
 import { type Node } from 'react';
 
-export default function Hero(): Node {
+export default function Hero({ animate = true }: {| animate?: boolean |}): Node {
```
dangerismycat, line 4: nit: I would think animation wouldn't be the default state…like, here's a Hero, and here's a Hero that's ✨animated✨
and from a "reading the code" perspective, it's a bit unintuitive that the Hero is always animated unless you turn that off — like I would expect `<Hero />` and `<Hero animate />` vs `<Hero />` and `<Hero animate={false} />` (I feel like boolean props should always default to false, that `boolProp={false}` is a code smell)

also, super nit: do you mind renaming this `HeroGraphic` so it matches the file name?

playwright/accessibility/Year in Review 2022.spec.mjs
```diff
@@ -0,0 +1,10 @@
+// @flow strict
+import { test } from '@playwright/test';
+import expectAccessiblePage from './expectAccessiblePage.mjs';
+
+// Skip since the tests times out
+// eslint-disable-next-line jest/no-disabled-tests
```
dangerismycat, line 6: 😢

dangerismycat (approved):

docs/graphics/home-page/HeroGraphic.js
ayeshakmaz: This is the home page hero, which we would want to animate always, but we're only turning it off now, because there's a new animation on top. So once we remove the banner, it'd just go back to <Hero/>

PR #2563, opened 2022-12-12
Popover: migrate docs examples to Sandpack

dangerismycat:
Also added one new variant about Popover visibility on page load. Didn't modify content of existing examples, just migrated them as-is

### DIFF
--- a/docs/examples/popover/a11y.js
+++ b/docs/examples/popover/a11y.js
@@ -13,6 +13,24 @@ import {
   Image,
 } from 'gestalt';
 
+const images = [
+  {
+    url: 'https://i.ibb.co/s3PRJ8v/photo-1496747611176-843222e1e57c.webp',
+    title: 'Fashion',
+    alt: 'Thumbnail image: a white dress with red flowers',
+  },
+  {
+    url: 'https://i.ibb.co/swC1qpp/IMG-0494.jpg',
+    title: 'Food',
+    alt: 'Thumbnail image: a paella with shrimp, green peas, red peppers and yellow rice',
+  },
+  {
+    url: 'https://i.ibb.co/PFVF3JH/photo-1583847268964-b28dc8f51f92.webp',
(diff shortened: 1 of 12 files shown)


docs/examples/popover/dontCaret.js
```diff
@@ -0,0 +1,161 @@
...
+            anchor={anchorRef.current}
+            id="main-example"
+            idealDirection="down"
+            onDismiss={() => setOpen(false)}
+            positionRelativeToAnchor={false}
+            size="xl"
+            showDismissButton
+            showCaret
+          >
+            <Box width={360}>
```
AlbertCarreras, line 138: I'd to 250 here so it displays nicely

docs/pages/web/popover.js
```diff
@@ -162,360 +55,64 @@ function PopoverExample() {
...
-            </Box>
-          </Popover>
-        </Layer>
-      )}
-    </React.Fragment>
-  );
-}
-  `}
+            sandpackExample={
+              <SandpackExample code={main} hideEditor name="Do - Lightweight" previewHeight={400} />
```
AlbertCarreras, line 66: I'd do 250 in the Box width so it displays nicely.

docs/examples/popover/a11y.js
```diff
@@ -99,6 +99,7 @@ export default function DismissButton(): Node {
           <Button color="red" onClick={() => {}} size="lg" text="Save" />
         </Flex>
       </Box>
+
```
AlbertCarreras, line 102: I'd do 250 here so it displays nicely too.

AlbertCarreras (approved):

docs/examples/popover/dontCaret.js
dangerismycat: I split the difference and went with 300 (250 seemed a tad narrow)

PR #2564, opened 2022-12-13
Tag: major refactor, add warning state

dangerismycat:
### Summary

Screen Shot 2022-12-14 at 3 15 33 PM

#### What changed?

Breaking changes:
- `errorMessage` replaced with a new `type` prop
- Warning state added with new `type` prop
- `s/removeIconAccessibilityLabel/accessibilityRemoveIconLabel` to match our convention
- Consolidated union types: `accessibilityRemoveIconLabel` is always optional (provided by DefaultLabelProvider), `onRemove` is always optional but throws if not provided and the Tag is not disabled (this also allows us to use autogen props finally!)

##### CODEMODS
```bash
yarn codemod modifyProp ~/path/to/your/code --component=Tag --previousProp=removeIconAccessibilityLabel --nextProp=accessibilityRemoveIconLabel
```
```bash
yarn codemod modifyPropValue ~/path/to/your/code --component=Tag --previousProp=errorMessage --nextProp=type --nextValue=error
```

#### Why?

We've had existing tech debt: `errorMessage` was poorly named, as the message was only shown to the user as the accessibility label for the error icon. That label doesn't need to be specific, and presenting the prop as a "message" implies that it's used to communicate to the end user.

More recently, there is need for a warning state for Tag (see Figma link below).

### Links

- [Jira](https://jira.pinadmin.com/browse/GESTALT-5162)
- [Figma](https://www.figma.com/file/V8mHev9JXgurPoyi6zwJf9/Audience-Sharing?node-id=925%3A92038&t=sSSEI99yrkxnf0v1-1)

### Checklist


AlbertCarreras:
re codemod, you dont need quotes previousProp='removeIconAccessibilityLabel'  >> previousProp=removeIconAccessibilityLabel

packages/gestalt/src/Tag.js
```diff
@@ -1,34 +1,55 @@
...
-import useFocusVisible from './useFocusVisible.js';
 import focusStyles from './Focus.css';
-import touchableStyles from './Touchable.css';
+import Icon from './Icon.js';
 import styles from './Tag.css';
+import Text from './Text.js';
+import touchableStyles from './Touchable.css';
+import useFocusVisible from './useFocusVisible.js';
+
+const backgroundColorByType = {
```
AlbertCarreras, line 15: nit Object.freeze()?

packages/gestalt/src/Tag.js
```diff
@@ -67,44 +94,50 @@ export default function Tag(props: Props): Node {
     },
   );
 
-  const borderStyle =
-    disabled && !errorMessage
-      ? {
-          __style: { border: `solid 1px ${colorGray200}` },
-        }
-      : undefined;
+  const borderStyle = disabled && !hasIcon ? { border: `solid 1px ${colorGray200}` } : {};
```
AlbertCarreras, line 97: can we inline this if not reused?

packages/gestalt/src/Tag.js
```diff
@@ -67,44 +94,50 @@ export default function Tag(props: Props): Node {
...
+
+  if (!disabled && !onRemove) {
+    throw new Error('Non-disabled Tags must include an onRemove handler');
```
AlbertCarreras, line 100: Im ok,  but I feel this is an opaque error. If people set disable value dynamically, this error might get thrown after the engineer has landed the code. I think I would do it required and for disable eng would have to pass () => {}

AlbertCarreras (approved):

packages/gestalt/src/Tag.js
ayeshakmaz: +1, I'd treat it like Checkbox where the handler is required, even if it could become disabled at some point

ayeshakmaz:
Can you add screenshots since the examples won't register yet?

packages/gestalt/src/Tag.js
dangerismycat: That's a good point!

dangerismycat:
@ayeshakmaz Added screenshots. The "warning" colors on interaction might need tweaking, but that will be easier to QA once this PR is merged (and there are no existing usages in the codebase, so doesn't really matter if any needed tweaks are a fast-follow).

PR #2569, opened 2022-12-15
Tag: add Best Practices, Localization to docs

dangerismycat:


docs/examples/tag/dontInteractive.js
```diff
@@ -0,0 +1,11 @@
+// @flow strict
+import { type Node } from 'react';
+import { Flex, Tag } from 'gestalt';
+
+export default function Example(): Node {
+  return (
+    <Flex alignItems="center" justifyContent="center" height="100%" width="100%">
+      <Tag disabled onRemove={() => {}} text="Visit Pinterest" />
```
AlbertCarreras, line 8: This Tag is no interactive, we should add a TapArea around it?

AlbertCarreras (approved):

docs/examples/tag/dontInteractive.js
dangerismycat: sure, that's a good idea

PR #2581, opened 2023-01-04
Link: replace `inline` prop with `display`

dangerismycat:
#### What changed?

Currently Link's `inline` prop maps to `display: inline-block`, which doesn't wrap the contained text in an "inline" way — it wraps the entire block. This is not intuitive and interferes with common design patterns. This PR replaces `inline` with `display: 'inline' | 'inlineBlock' | 'block'`.

Screen Shot 2023-01-03 at 5 29 23 PM

#### Why?

Because inline-level elements ignore any height, width, or padding set on them, we must use `inline-block` for styles like our click/tap compress animation. However, we only set `tapStyle="compress` on 11 usages (out of 780 total Link usages), so it's not worth optimizing for this barely-used feature. Since we can't get rid of that feature completely, we have to accommodate by modifying `inline` to `display`. While the new prop is more complex, we can address this via docs updates.

### Codemod

Note that this codemod replaces all instances of `inline` with `display="inlineBlock"`, as this most faithfully reproduces the current behavior. However, most usages likely _should_ use `display="inline"`. I updated the usages on the Link docs page to use "inline", since they're easily verified and better convey the intended usage. We should revisit all other existing usages across our other docs pages and components and replace "inlineBlock" with "inline" where we can.

```bash
yarn codemod modifyPropValue ~/path/to/your/code \
  --component=Link \
  --previousProp=inline \
  --nextProp=display \
  --previousValue=true \
  --nextValue=inlineBlock
```

### Links

- [Jira](https://jira.pinadmin.com/browse/GESTALT-5172)

### DIFF
--- a/docs/docs-components/Page.js
+++ b/docs/docs-components/Page.js
@@ -63,7 +63,7 @@ export default function Page({
 
         {!hideEditLink && (
           <Box marginTop={12}>
-            <Link href={editPageUrl} target="blank" inline>
+            <Link href={editPageUrl} target="blank" display="inlineBlock">
               <Text underline>Edit page on GitHub</Text>
             </Link>
           </Box>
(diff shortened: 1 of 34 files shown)


AlbertCarreras (approved):
LGTM

PR #2584, opened 2023-01-04
Internal: actually fix autogen props default values

dangerismycat:
This _actually_ fixes the broken autogen props defaults, introduced in #2504. This also reverts the revert of a dependabot patch, which wasn't the culprit (#2583).

### DIFF
--- a/docs/docs-components/GeneratedPropTable.js
+++ b/docs/docs-components/GeneratedPropTable.js
@@ -4,47 +4,50 @@ import { type DocGen } from './docgen.js';
 import PropTable from './PropTable.js';
 
 // Note if the prop has responsive versions (e.g. margin, smMargin, mdMargin, lgMargin)
-function getResponsive(description?: string): {|
+function getResponsive(description?: string): {
   description?: string,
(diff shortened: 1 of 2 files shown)


docs/docs-components/GeneratedPropTable.js
```diff
@@ -4,47 +4,50 @@ import { type DocGen } from './docgen.js';
...
+  ...
+} {
   const input = description ?? '';
   const match = input.match(/(?<main>Default: (?<defaultValue>.*))/);
   const groups = match?.groups ?? {};
 
   return {
     description: groups.main ? input.replace(groups.main, '') : description,
-    defaultValue: groups.defaultValue?.replace(/'/g, ''),
+    ...(groups.defaultValue ? { defaultValue: groups.defaultValue?.replace(/'/g, '') } : {}),
```
dangerismycat, line 50: This pattern was the culprit. Because this previously returned an object with `defaultValue` every time, it would always overwrite the actual default value with this one, which was almost always undefined

AlbertCarreras (approved):

PR #2592, opened 2023-01-06
Internal: add flow declarations for @testing-library/user-event

dangerismycat:
`flow-typed` doesn't have a declarations file for our version of `@testing-library/user-event` (v14), but does have types for [the previous version](https://github.com/flow-typed/flow-typed/blob/e420967f1d88dc042409e9e70ac034ffb968a222/definitions/npm/%40testing-library/user-event_v13.x.x/flow_v0.104.x-/user-event_v13.x.x.js). This PR copy/pastes those v13 types into a flow-typed file for v14. This seems mostly correct so far, as it allows us to remove the "untyped-import" suppressions without surfacing any new errors. However, the API is likely slightly different for our version, so we should feel free to update these types in the future as needed if they don't seem to accurately reflect the API.

### DIFF
--- a/flow-typed/npm/@testing-library/user-event_v14.x.x.js
+++ b/flow-typed/npm/@testing-library/user-event_v14.x.x.js
@@ -1,0 +1,174 @@
+/*
+****** NOTE this is copied from the v13.x.x on flow-typed ******
+*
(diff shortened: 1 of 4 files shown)


AlbertCarreras (approved):
Thanks!

PR #2594, opened 2023-01-06
Icons: Add keywords to icon search page

ayeshakmaz:
#### What changed?

Add keywords to make a better search experience for the Icon library page 

#### Why?

Find icons by keywords instead of just exact icon name

Screen Shot 2023-01-06 at 4 45 24 PM

### Links

- [Jira](https://jira.pinadmin.com/browse/GESTALT-5001)
- [Figma](https://www.figma.com/file/RBlfOYmL123whCtlglXIsG/Iconography-page-redesign?node-id=442%3A3384&t=U2cj3eIDMKoIpsEg-0)

### DIFF
--- a/docs/pages/foundations/iconography/library.js
+++ b/docs/pages/foundations/iconography/library.js
@@ -116,6 +116,15 @@ function findIconByCategory(icon?: string, category: string) {
   );
 }
 
+function iconHasKeyword(iconName?: string, searchTerm: string) {
+  return (
+    iconCategoryData.icons.find(
+      ({ name, keywords }) =>
+        name === iconName && keywords?.find((word) => word.includes(searchTerm)),
+    ) !== undefined
+  );
+}
+
 export default function IconPage(): Node {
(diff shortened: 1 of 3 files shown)


docs/pages/foundations/iconography/library.js
```diff
@@ -116,6 +116,15 @@ function findIconByCategory(icon?: string, category: string) {
   );
 }
 
+function iconHasKeyword(iconName?: string, searchTerm: string) {
```
AlbertCarreras, line 119: this part here could be optimize probably  mapping into key values rather but it works and for Docs doesnt seem a big deal!

AlbertCarreras (approved):

docs/pages/foundations/iconography/library.js
ayeshakmaz: It's a bit more complicated logic so that it matches partial matches, so instead of keywords.includes, which would only match "create" but not "crea", it look through each keyword and then does include

PR #2611, opened 2023-01-13
Internal: Flow fixes

dangerismycat:
This addresses some of our FlowFixMe's. Some are fixed. Others are converted to FlowExpectedError, where there is not a workaround to make Flow happy. There shouldn't be any runtime changes — potential areas are called out in comments.

### DIFF
--- a/packages/gestalt/src/ModalAlert.js
+++ b/packages/gestalt/src/ModalAlert.js
@@ -173,8 +173,11 @@ export default function ModalAlert({
     useDefaultLabelContext('ModalAlert');
 
   Object.entries({ primaryAction, secondaryAction }).forEach(([key, value]) => {
-    // $FlowFixMe[incompatible-use]
-    if (value && [value.href, value.onClick].every((item) => item === undefined)) {
+    if (
+      value &&
+      typeof value === 'object' &&
+      [value.href, value.onClick].every((item) => item === undefined)
+    ) {
       throw new Error(
         `Either an \`href\` or an \`onClick\` handler must be provided to \`${key}\`.`,
       );
--- a/packages/gestalt/src/VideoControls.js
+++ b/packages/gestalt/src/VideoControls.js
@@ -101,11 +101,11 @@ function VideoControls({
     setShowFullscreenButton(
       typeof document !== 'undefined' &&
         (document.fullscreenEnabled ||
-          // $FlowFixMe[prop-missing]
+          // $FlowExpectedError[prop-missing]
           document.webkitFullscreenEnabled ||
-          // $FlowFixMe[prop-missing]
+          // $FlowExpectedError[prop-missing]
           document.mozFullScreenEnabled ||
-          // $FlowFixMe[prop-missing]
+          // $FlowExpectedError[prop-missing]
           document.msFullscreenEnabled),
     );
   }, []);
--- a/packages/gestalt/src/boxTransforms.js
+++ b/packages/gestalt/src/boxTransforms.js
@@ -470,7 +470,7 @@ export function buildStyles<T: Object>({
   // eslint-disable-next-line no-restricted-syntax
   for (const prop in props) {
     if (
-      // $FlowFixMe[method-unbinding]
+      // $FlowExpectedError[method-unbinding]
       Object.prototype.hasOwnProperty.call(propToFn, prop) &&
       !omitProps.includes(prop) &&
       (!allowlistProps || allowlistProps.includes(prop))
--- a/packages/gestalt/src/style.js
+++ b/packages/gestalt/src/style.js
@@ -75,6 +75,5 @@ export const toProps = ({ className, inlineStyle }: Style): ToPropsOutput => {
     props.style = inlineStyle;
   }
 
-  // $FlowFixMe[incompatible-exact]
-  return props;
+  return { ...props };
 };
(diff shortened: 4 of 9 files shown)


packages/gestalt/src/ModalAlert.js
```diff
@@ -173,8 +173,11 @@ export default function ModalAlert({
     useDefaultLabelContext('ModalAlert');
 
   Object.entries({ primaryAction, secondaryAction }).forEach(([key, value]) => {
-    // $FlowFixMe[incompatible-use]
-    if (value && [value.href, value.onClick].every((item) => item === undefined)) {
+    if (
+      value &&
+      typeof value === 'object' &&
```
dangerismycat, line 178: this is the addition to satisfy Flow

packages/gestalt/src/VideoControls.js
```diff
@@ -101,11 +101,11 @@ function VideoControls({
     setShowFullscreenButton(
       typeof document !== 'undefined' &&
         (document.fullscreenEnabled ||
-          // $FlowFixMe[prop-missing]
+          // $FlowExpectedError[prop-missing]
```
dangerismycat, line 104: These fields are non-standard, so aren't expected to be part of the official Flow type for `document`

packages/gestalt/src/boxTransforms.js
```diff
@@ -470,7 +470,7 @@ export function buildStyles<T: Object>({
   // eslint-disable-next-line no-restricted-syntax
   for (const prop in props) {
     if (
-      // $FlowFixMe[method-unbinding]
+      // $FlowExpectedError[method-unbinding]
```
dangerismycat, line 473: There isn't really a way around this. Flow doesn't like binding Object's `hasOwnProperty` to a different `this`. ESLint doesn't like calling `Object.hasOwnProperty` directly, since it might not exist or may have been overwritten.

packages/gestalt/src/style.js
```diff
@@ -75,6 +75,5 @@ export const toProps = ({ className, inlineStyle }: Style): ToPropsOutput => {
     props.style = inlineStyle;
   }
 
-  // $FlowFixMe[incompatible-exact]
-  return props;
+  return { ...props };
```
dangerismycat, line 78: This is the only possible runtime change. It _shouldn't_ matter, but there is the possibility

packages/gestalt/src/useFocusVisible.js
```diff
@@ -83,7 +83,7 @@ function setupGlobalFocusEvents() {
   // However, we need to detect other cases when a focus event occurs without
   // a preceding user event (e.g. screen reader focus). Overriding the focus
   // method on HTMLElement.prototype is a bit hacky, but works.
-  // $FlowFixMe[method-unbinding]
+  // $FlowExpectedError[method-unbinding]
```
dangerismycat, line 86: ditto here, no way I know of to appease Flow with this type of monkey-patching

packages/gestalt/src/utils/matchMedia.js
```diff
@@ -3,19 +3,19 @@
 export function addListener(mediaQuery: MediaQueryList, callback: MediaQueryListListener): void {
   // addEventListener on mediaQuery is not supported in all browsers (Edge / Safari)
   // https://developer.mozilla.org/en-US/docs/Web/API/MediaQueryList/addListener
-  // $FlowFixMe[method-unbinding]
+  // $FlowExpectedError[method-unbinding]
```
dangerismycat, line 6: ditto here with non-standard fields that aren't on the official Flow type

packages/gestalt/src/utils/positioningUtils.js
```diff
@@ -44,7 +44,7 @@ export const getContainerNode = ({
   // append the portal
   let containerNode = null;
   // currentNode references the DOM node used while traversing up nodes in the DOM tree
-  let currentNode = initialPositionRef;
+  let currentNode: ?HTMLElement | Node = initialPositionRef;
```
dangerismycat, line 47: On line 41 we're establishing that `initialPositionRef` is `?HTMLElement`, but on line 60 we're potentially re-assigning `currentNode` to the parent node of `currentNode`, which is of type `Node`

packages/gestalt/src/utils/wrapWithComponent.js
```diff
@@ -12,7 +12,7 @@ const isComponent = (AComponent: UnknownComponent, AnotherComponent: UnknownComp
 
 // Checks whether `element` is a React element of type `Component` (or one of
 // the passed components, if `Component` is an array of React components).
-function isElementOfType<P>(element: ?Node, Component: ChildrenArray<ComponentType<P>>): boolean {
+function isElementOfType<P>(element: ?Node, Component: ComponentType<P>): boolean {
```
dangerismycat, line 15: This was wrong and whoever wrote it wrong should feel bad

(it was me 😅 )

packages/gestalt/src/utils/wrapWithComponent.js
```diff
@@ -28,10 +28,8 @@ function isElementOfType<P>(element: ?Node, Component: ChildrenArray<ComponentTy
   const { type } = element;
   const Components = Array.isArray(Component) ? Component : [Component];
 
-  return Components.some(
-    // $FlowFixMe[incompatible-call]
-    (AComponent) => typeof type !== 'string' && isComponent(AComponent, type),
-  );
+  // $FlowExpectedError[incompatible-call]
```
dangerismycat, line 31: Flow is still unhappy about how we're handling `type` here, but that's expected: Flow wants us to know more than we possibly could

AlbertCarreras (approved):

PR #2616, opened 2023-01-20
Masonry: replace the `Item` prop with the `renderItem` prop

jackhsu978:
### Precursor to Breaking change

This PR adds a prop to the Masonry API. This prop is optional for now, but will eventually become required.

No changes are required now, but if you want to get a head start: run the following codemod to detect Masonry in your code base.

```
--component=Masonry
```

You can also automate the renaming: 

```
--component=Masonry
--previousProp=Items
--nextProp=renderItems
```

but still need to change the value manually following the steps described below:

Screenshot 2023-01-20 at 11 00 28 AM

#### What changed?

In `Masonry`, deprecate the `Item` prop (which is expected to be a React component) in favor of `renderItem`, which can be any function that returns a React Node.

#### Why?

When React renders the `<App />` component here:
```ts
const MyGrid = ({ Item }) => <Item />;

const App = ({ foo }) => <MyGrid Item={() => <GridItem foo={foo} />} />;
```

The virtual DOM is `<App><MyGrid><Item><GridItem /></Item></MyGrid></App>`

When React re-renders `<App />`, [React reconciliation](https://reactjs.org/docs/reconciliation.html) algorithm sees that the `<Item>` element is an instance of a brand new React component, since the anonymous function `() => <GridItem />` is a new function for each render call. The reconciliation algorithm would unmount the old `<Item(old)><GridItem /></Item(old)>` subtree and mount the new `<Item(new)><GridItem /></Item(new)>` subtree. The `<GridItem />` gets unmounted as a result, since its parent is unmounted. This is a problem because we do not want the grid items to be unmounted when the parent is re-rendered.

Instead, consider we do this:
```ts
const MyGrid = ({ renderItem }) => renderItem();

const App = ({ foo }) => <MyGrid renderItem={() => <GridItem foo={foo} />} />;
```

The virtual DOM is `<App><MyGrid><GridItem /></MyGrid></App>`

When React re-renders `<App />`, each React component in the virtual DOM is unchanged. React reconciliation can keep the components mounted and `<GridItem />` does not get unmounted as a result.

See this example https://codesandbox.io/s/sandpack-project-forked-qmh0tf?file=/App.js for a simple demonstration

### DIFF
--- a/packages/gestalt/src/Masonry.js
+++ b/packages/gestalt/src/Masonry.js
@@ -32,7 +32,7 @@ type Props<T> = {|
   gutterWidth?: number,
   /**
    * A React component (or stateless functional component) that renders the item you would like displayed in the grid. This component is passed three props: the item's data, the item's index in the grid, and a flag indicating if Masonry is currently measuring the item. *Note that this [must be a stable reference!](https://www.developerway.com/posts/react-re-renders-guide#part3.1)* If using a component declared within a parent function component, you must use [`useCallback`](https://reactjs.org/docs/hooks-reference.html#usecallback) to ensure a stable reference.
-   * This is deprecated in favor of the `renderItem` prop
+   * This is DEPRECATED and will be removed in the next major release. Please use `renderItem` prop instead.
    */
   Item?: ComponentType<{|
     data: T,
@@ -40,7 +40,7 @@ type Props<T> = {|
     isMeasuring: boolean,
   |}>,
   /**
-   * An array of items to display that contains the data to be rendered by `<Item />`.
+   * An array of items to display that contains the data to be rendered by `renderItem()` (fallback to the deprecated `<Item />` if `renderItem` is not passed).
    */
   items: $ReadOnlyArray<T>,
   /**
--- a/packages/gestalt/src/Masonry.js
+++ b/packages/gestalt/src/Masonry.js
@@ -1,5 +1,5 @@
 // @flow strict
-import { type Node, Component as ReactComponent } from 'react';
+import { type ComponentType, type Node, Component as ReactComponent } from 'react';
 import debounce, { type DebounceReturn } from './debounce.js';
 import FetchItems from './FetchItems.js';
 import styles from './Masonry.css';
@@ -31,7 +31,16 @@ type Props<T> = {|
    */
   gutterWidth?: number,
   /**
-   * An array of items to display that contains the data to be rendered by `renderItem()`.
+   * A React component (or stateless functional component) that renders the item you would like displayed in the grid. This component is passed three props: the item's data, the item's index in the grid, and a flag indicating if Masonry is currently measuring the item. *Note that this [must be a stable reference!](https://www.developerway.com/posts/react-re-renders-guide#part3.1)* If using a component declared within a parent function component, you must use [`useCallback`](https://reactjs.org/docs/hooks-reference.html#usecallback) to ensure a stable reference.
+   * This is deprecated in favor of the `renderItem` prop
+   */
+  Item?: ComponentType<{|
+    data: T,
+    itemIdx: number,
+    isMeasuring: boolean,
+  |}>,
+  /**
+   * An array of items to display that contains the data to be rendered by `<Item />`.
    */
   items: $ReadOnlyArray<T>,
   /**
@@ -67,7 +76,7 @@ type Props<T> = {|
   /**
    * A function that renders the item you would like displayed in the grid. This function is passed three props: the item's data, the item's index in the grid, and a flag indicating if Masonry is currently measuring the item.
    */
-  renderItem: ({|
+  renderItem?: ({|
     +data: T,
     +itemIdx: number,
     +isMeasuring: boolean,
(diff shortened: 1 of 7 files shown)


AlbertCarreras:
LGTM, thanks for this change and the description. 

I'll let Ryan approve it as he's been more involved in the conversation.

jackhsu978:
@AlbertCarreras Thanks for adding the nice upgrade instructions. In theory, we could write a codemod like this:

```js
Item={Item}
```
to
```js
renderItem={({ data, itemIdx, isMeasuring }) => <Item data={data} itemIdx={itemIdx} isMeasuring={isMeasuring} />
```

In our codebase, there are only 16 places where this codemod would help and I think writing codemod would actually be slower than just manually change them.

I chose minor release because I was thinking about adding the `renderItem` prop and deprecating (not removing) the `Item` prop in a minor release first, then remove the deprecated `Item` prop in a major release. Do you think that's better or overkill? If we want to just do a major release, then I should update the PR to remove the `Item` prop.

cc @dangerismycat

dangerismycat:
Taking a look at this PR now 👀 

> In Pinboard repo, there are only 16 places where this codemod would help and I think writing codemod would actually be slower than just manually change them.

Maybe, but it's also good practice 😅 Not a dealbreaker for me, but we do typically include codemods for relatively simple API changes like this one.

> I chose minor release because I was thinking about adding the `renderItem` prop and deprecating (not removing) the `Item` prop in a minor release first, then remove the deprecated `Item` prop in a major release. Do you think that's better or overkill? If we want to just do a major release, then I should update the PR to remove the `Item` prop.

Agreed this should be `minor` and should not remove `Item` yet. We'll need to add some simple logic to use `renderItem` if present, otherwise use `Item`. (We'll need to temporarily make both props optional — we can throw an error if neither prop is provided, and once the existing Pinboard usages are refactored, the `major` PR can remove `Item` and make `renderItem` required.)

AlbertCarreras:
@dangerismycat @jackhsu978 

The benefit of doing it minor is that the Pinboard refactor (major/breaking change) can come at a later time and we can refactor each Masonry instance one at a time. Then make breaking change and technically no refactor would be needed in the bump PR.

However, I'm not familiar with Masonry implementations. The change seems pretty straightforward to me. I'd argue that if we can codemod it... then it's safe to go major and be done in 2 PRs rather than 20. If the implementations are convoluted and codemoding is not possible as it requires manual review... then I'd would keep it minor.

Summary, if changes are this straightforward, I don't see the point of breaking down the refactor into multiple PRs and Gestalt PRs.

`Item={Item}`

to this:

`renderItem={({ data, itemIdx, isMeasuring }) => <Item data={data} itemIdx={itemIdx} isMeasuring={isMeasuring} />
`

It all depends on the implementations and how complex and risky refactoring this gets. 


Re codemods, I'd be nicer to write a custom one but I would not prioritize that for a component with such a few instances. Not worth the eng effort, IMO.

dangerismycat:
> The benefit of doing it minor is that the Pinboard refactor (major/breaking change) can come at a later time and we can refactor each Masonry instance one at a time. Then make the breaking change (prop removal) and technically no refactor would be needed in the bump PR.
> 
> However, I'm not familiar with Masonry implementations. The change seems pretty straightforward to me. I'd argue that if we can codemod it... then it's safe to go major and be done in 2 PRs rather than 20. If the implementations are convoluted and codemoding is not possible as it requires manual review... then I'd would keep it minor.
> 
> Summary, if changes are this straightforward, I don't see the point of breaking down the refactor into multiple PRs and Gestalt PRs.

Given how foundational Masonry is to Pinterest and how many critical metrics rely on it, I'd advocate for the more conservative approach here even though the changes seem straightforward. It does appear to be a simple prop replacement, but if something were to go wrong somehow, the fix would be a lot simpler if the old prop still exists and we only need to emergency land reverts in Pinboard.

I wouldn't recommend this approach with most (any?) other components, just Masonry.

jackhsu978:
@dangerismycat I am fine with either one release or two releases. But for the codemod, I am concerned that I may have to postpone this given the extra time it will take me to learn and implement the codemod. Will you have the bandwidth with writing the codemod? I will hate to postpone this impactful change for a codemod that, in theory, everyone can easily live without.

dangerismycat:
> @dangerismycat I am fine with either one release or two releases. But for the codemod, I am concerned that I may have to postpone this given the extra time it will take me to learn and implement the codemod. Will you have the bandwidth with writing the codemod? I will hate to postpone this impactful change for a codemod that, in theory, everyone can easily live without.

Codemod can wait till the `major` PR — no need for this one if we're just introducing an optional new prop. And yep, can definitely help write the codemod!

packages/gestalt/src/Masonry.js
```diff
@@ -377,13 +386,23 @@ export default class Masonry<T: { ... }> extends ReactComponent<Props<T>, State<
...
 
+  renderItem(item: {| +data: T, +itemIdx: number, +isMeasuring: boolean |}): Node {
+    const { Item, renderItem } = this.props;
+    if (renderItem) {
+      return renderItem(item);
+    }
+    if (Item) {
+      return <Item data={item.data} itemIdx={item.itemIdx} isMeasuring={item.isMeasuring} />;
+    }
+    return null;
```
dangerismycat, line 397: This should probably throw an error if we get here?

dangerismycat (approved):

packages/gestalt/src/Masonry.js
jackhsu978: addressed in https://github.com/pinterest/gestalt/pull/2616/commits/74257966b14a3574ea6e3c7e7073f10fff98f930

jackhsu978: that was bad. fixed again here: https://github.com/pinterest/gestalt/pull/2616/commits/88c61479059b0b13a29977170a0413cc45c773be

PR #2618, opened 2023-01-20
Modal, ModalAlert: implement adaptive Modal, ModalAlert for mobile

AlbertCarreras:
#### What changed?

Modal, ModalAlert: implement adaptive Modal, ModalAlert for mobile'

Screenshot 2023-01-20 at 6 49 24 PM

#### Why?

`window.body.style.overflow = 'hidden';`
Explanation: 
BEFORE https://www.loom.com/share/8812b70e200d48c2800b5a26fd3767c7
AFTER https://www.loom.com/share/9ccbeca20e1740cf8e97f0b45776d90d

### Links

- [Figma](https://www.figma.com/file/VespNbaZGqiY7M0PMSIipE/Mobile-Specs?node-id=403%3A4136&t=ars1hhsSCiJKXBE9-0)

### Checklist


docs/examples/modal/mobileExample.js
```diff
@@ -0,0 +1,40 @@
+// @flow strict
+import React, { type Node } from 'react';
+import { Flex, Layer, Modal, Box, Text, DeviceTypeProvider, Button } from 'gestalt';
+
+export default function Example(): Node {
+  const [showModal, setShowModal] = React.useState(false);
```
dangerismycat, line 6: nit: import `useState` directly (here and the other examples)

(I'm writing a lint rule for this, just not finished yet 😅)

docs/examples/modal/mobileExample.js
```diff
@@ -0,0 +1,40 @@
...
+
+  return showModal ? (
```
dangerismycat, line 8: Probably no need to hide the button when the modal is showing, right? It should overlay the button?

docs/pages/web/modal.js
```diff
@@ -244,6 +245,59 @@ export default function ModalPage({ generatedDocGen }: {| generatedDocGen: DocGe
             }
           />
         </MainSection.Subsection>
+        <MainSection
+          name="Mobile"
+          description={`Modal requires [DeviceTypeProvider](/web/utilities/devicetypeprovider) to enable its mobile user interface. The example below shows the mobile platform UI and its implementation.
+
+For mobile, all \`size\` are unified into a full mobile viewport Modal. Notice that \`subHeading\` gets moved from the header to the main content container.
```
dangerismycat, line 252: typo: "For mobile, all _sizes_ are unified…"

Not sure if "mobile" is necessary the second time. I would think this would be "…into a _full-viewport_ modal."

docs/pages/web/modalalert.js
```diff
@@ -298,6 +299,61 @@ export default function ModalAlertPage({ generatedDocGen }: {| generatedDocGen:
           />
         </MainSection.Subsection>
       </MainSection>
+
+      <MainSection
+        name="Mobile"
+        description={`ModalAlert requires [DeviceTypeProvider](/web/utilities/devicetypeprovider) to enable its mobile user interface. The example below shows the mobile platform UI and its implementation.
+
+For mobile, all \`size\` are unified into a full mobile viewport Modal. Notice that \`subHeading\` gets moved from the header to the main content container.
```
dangerismycat, line 307: Same nits here. (This is really making me want to build that modular docs thing! Imagine only writing this once 😍)

packages/gestalt/src/MobileModal.js
```diff
@@ -0,0 +1,211 @@
+// @flow strict
```
dangerismycat, line 1: Doesn't need to happen in this PR necessarily, but I'd put this in a `/modal` directory. I think we should only include exported components in the top-level directory

packages/gestalt/src/Modal.js
```diff
@@ -100,6 +102,7 @@ function Header({
  * A [Modal](https://gestalt.pinterest.systems/web/modal) displays content that requires user interaction. Modals appear on a layer above the page and therefore block the content underneath, preventing users from interacting with anything else besides the Modal. Modal should be used to gather short bits of information from the user. For confirmation of an action or acknowledgment, use [ModalAlert](https://gestalt.pinterest.systems/web/modalalert).
  *
  * ![Modal light mode](https://raw.githubusercontent.com/pinterest/gestalt/master/playwright/visual-test/Modal.spec.mjs-snapshots/Modal-chromium-darwin.png)
+ *  ![Modal mobile](https://raw.githubusercontent.com/pinterest/gestalt/master/playwright/visual-test/Modal-mobile.spec.mjs-snapshots/Modal-mobile-chromium-darwin.png)
```
dangerismycat, line 105: Should this have a dark mode version as well?

packages/gestalt/src/ModalAlert.js
```diff
@@ -156,6 +160,7 @@ function ModalAlertAction({ data, type }: {| data: ActionDataType, type: string
  * A [ModalAlert](https://gestalt.pinterest.systems/web/modalalert) is a simple modal dialog used to alert a user of an issue, or to request confirmation after a user-triggered action. ModalAlert overlays and blocks page content until it is dismissed by the user.
  *
  * ![ModalAlert light mode](https://raw.githubusercontent.com/pinterest/gestalt/master/playwright/visual-test/ModalAlert.spec.mjs-snapshots/ModalAlert-chromium-darwin.png)
+ * ![ModalAlert mobile](https://raw.githubusercontent.com/pinterest/gestalt/master/playwright/visual-test/ModalAlert-mobile.spec.mjs-snapshots/ModalAlert-mobile-chromium-darwin.png)
```
dangerismycat, line 163: ditto here

packages/gestalt/src/Modal.jsdom.test.js
```diff
@@ -1,20 +1,113 @@
 // @flow strict
```
dangerismycat, line 1: Thanks for adding/updating all these tests!

packages/gestalt/src/contexts/DefaultLabelProvider.js
```diff
@@ -20,7 +20,7 @@ export type DefaultLabelContextType = {|
   Link: {|
     accessibilityNewTabLabel: string,
   |},
-  ModalAlert: {|
+  Modal: {|
```
dangerismycat, line 23: I keep going back and forth on this one. Is it confusing to use the same label for two components? I want to say "yes", but also seems like extra boilerplate to have two basically identical fields. I don't know 😓

dangerismycat (approved):

docs/examples/modal/mobileExample.js
AlbertCarreras: DONE

docs/examples/modal/mobileExample.js
AlbertCarreras: Indeed, copied from sidenavigation, but yeah, good point

DONE

packages/gestalt/src/MobileModal.js
AlbertCarreras: As soon as you import from the upper directory, it throws an error. I've tried! But you can only do that with files that are consumed in the upper level

packages/gestalt/src/Modal.js
AlbertCarreras: I want to keep it simple. I think one sample od dark mode is enough. Then if mobileis available the same. Otherwise is too much  boilerplate

packages/gestalt/src/contexts/DefaultLabelProvider.js
AlbertCarreras: Yeah feel the same. I kind of feel it should be 1 component 1 key, but I already did this with Link, TapArea/Button/IconButton all consume from Link  "accessibilityNewTabLabel: 'Opens a new tab'".

But then... how do we detect where this are being consumed? I should be easy component <> key. 

We can always reconsider this and add more keys

packages/gestalt/src/MobileModal.js
dangerismycat: That lint rule should only care about importing from a parent directory. Check out the organization I did for Masonry — putting this file into `/Modal/MobileModal.js` and then importing that into `/Modal.js` should be fine

PR #2630, opened 2023-01-27
Sheet: implement dismiss confirmation modal and `dismissConfirmation` prop

AlbertCarreras:
#### What changed?

Sheet: implement dismiss confirmation modal and dismissConfirmation prop #2630

Screenshot 2023-01-30 at 1 08 21 PM

#### Why?

It's been a long-time request from eng. Some Sheets have forms in them and dismissing the Sheet leads to losing the information entered. 

There are 6 action point for dismissing Sheet:

1. ESC key >> component controlled
2. Dismiss button >> component controlled
3. Backdrop click >> component controlled
4. onDismissStart render prop (subheading) >> Eng has access and control
5. onDismissStart render prop (content) >> Eng has access and control
6. onDismissStart render prop (footer) >> Eng has access and control

`dismissConfirmation` forces confirmation from user if it tries to dismiss Sheet via 1,2,3 action points.

It renders a Popover that alerts the user and request user action.

### Links

- [TDD](link to Paper doc)
- [Figma](link to Figma file)

### Checklist

Kapture 2023-01-26 at 20 34 27

### DIFF
--- a/docs/pages/web/sheet.js
+++ b/docs/pages/web/sheet.js
@@ -142,7 +142,7 @@ When Sheet opens, focus should be placed on the first interactive element within
       </AccessibilitySection>
       <MainSection
         name="Localization"
-        description={`Be sure to localize the \`heading\`, \`accessibilityDismissButtonLabel\`, \`accessibilitySheetLabel\` props and well as any custom string in \`dismissConfirmation\`. Note that localization can lengthen text by 20 to 30 percent.`}
+        description={`Be sure to localize the \`heading\`, \`accessibilityDismissButtonLabel\`, \`accessibilitySheetLabel\` props and well as any custom strings in \`dismissConfirmation\`. Note that localization can lengthen text by 20 to 30 percent.`}
       />
       <SlimBanner
         iconAccessibilityLabel="Localize the default label"
@@ -280,15 +280,15 @@ Sheet comes in 3 sizes: small (\`sm\`), medium (\`md\`), and large (\`lg\`).
 The three internally-controlled or component-controlled dismiss actions are:
 - when the \`ESC\` key is pressed
 - when the backdrop is clicked
-- when the dismiss IconButtons are clicked
+- when the dismiss IconButton is clicked
 
-The externally-controlled dismiss actions require implementing the callback \`onDismissStart\`. See the [animation variant](#Animation) to learn more.
+The externally-controlled dismiss actions (\`subHeading\`, \`children\`, and \`footer\`) require implementing the callback \`onDismissStart\`. See the [animation variant](#Animation) to learn more.
 
 Sheets can contain forms or be part of flows where the user is required to submit infomation. If a Sheet is dismissed involuntarily, the data entered by the user could not be saved and lost. This can create a bad user experience.
 
 To prevent dismissing Sheet involuntary, we can use \`dismissConfirmation\`. When provided, it will open a confirmation modal each time component-controlled dismiss actions are triggered.
 
-The confirmation modal has a flexible API. When the \`dismissConfirmation\` prop is set to "true", Sheet uses defauld texts and labels.
+The confirmation modal has a flexible API. When the \`dismissConfirmation\` prop is set to an empty object "dismissConfirmation={{}}", Sheet uses default texts and labels. See the default content below:
 
 - Message: "Are you sure you want to dismiss?"
 - Subtext: "You will lose all of your changes. This cannot be undone."
@@ -297,7 +297,7 @@ The confirmation modal has a flexible API. When the \`dismissConfirmation\` prop
 - Secondary action text: "No, go back."
 - Secondary action label: "No, go back to the sheet."
 
-All texts and labels can be customized using the \`dismissConfirmation\` prop. We can pass an object with custom strings. For any missing strings, Sheet uses the default ones.
+All texts and labels can be customized using the \`dismissConfirmation\` prop. We can pass an object with custom strings. For any missing strings, Sheet uses the default ones. See the \`dismissConfirmation\` prop Flow type to learn more about the optional texts and labels than can be customized.
 `}
         >
           <SlimBanner
--- a/docs/pages/web/sheet.js
+++ b/docs/pages/web/sheet.js
@@ -326,7 +326,7 @@ function AccessibilityExample() {
       {shouldShow && (
         <Layer zIndex={sheetZIndex}>
           <Sheet
-            dismissConfirmation
+            dismissConfirmation={{}}
             accessibilityDismissButtonLabel="Close audience creation sheet"
             accessibilitySheetLabel="Audience list creation for new campaign"
             heading="Create a new audience list"
--- a/docs/pages/web/sheet.js
+++ b/docs/pages/web/sheet.js
@@ -1,5 +1,6 @@
 // @flow strict
 import { type Node } from 'react';
+import { SlimBanner } from 'gestalt';
 import GeneratedPropTable from '../../docs-components/GeneratedPropTable.js';
 import PageHeader from '../../docs-components/PageHeader.js';
 import MainSection from '../../docs-components/MainSection.js';
@@ -141,7 +142,18 @@ When Sheet opens, focus should be placed on the first interactive element within
       </AccessibilitySection>
       <MainSection
         name="Localization"
-        description={`Be sure to localize the \`heading\`, \`accessibilityDismissButtonLabel\` and \`accessibilitySheetLabel\` props. Note that localization can lengthen text by 20 to 30 percent.`}
+        description={`Be sure to localize the \`heading\`, \`accessibilityDismissButtonLabel\`, \`accessibilitySheetLabel\` props and well as any custom string in \`dismissConfirmation\`. Note that localization can lengthen text by 20 to 30 percent.`}
+      />
+      <SlimBanner
+        iconAccessibilityLabel="Localize the default label"
+        message="Sheet's dismiss IconButton and confirmation modal consume default text and labels from DefaultLabelProvider. Make sure to localize the default strings with DefaultLabelProvider."
+        type="recommendationBare"
+        helperLink={{
+          text: 'Learn more',
+          accessibilityLabel: 'Learn more about DefaultLabelProvider',
+          href: '/web/utilities/defaultlabelprovider',
+          onClick: () => {},
+        }}
       />
       <MainSection name="Variants">
         <MainSection.Subsection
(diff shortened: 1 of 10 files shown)


hectoid:
nice! Close/Cancel isn't the clearest language, as it becomes difficult to tell exactly what you are doing. Suggest:
Secondary button: No, go back
Primary button: Yes, close

docs/pages/web/sheet.js
```diff
@@ -141,7 +142,18 @@ When Sheet opens, focus should be placed on the first interactive element within
       </AccessibilitySection>
       <MainSection
         name="Localization"
-        description={`Be sure to localize the \`heading\`, \`accessibilityDismissButtonLabel\` and \`accessibilitySheetLabel\` props. Note that localization can lengthen text by 20 to 30 percent.`}
+        description={`Be sure to localize the \`heading\`, \`accessibilityDismissButtonLabel\`, \`accessibilitySheetLabel\` props and well as any custom string in \`dismissConfirmation\`. Note that localization can lengthen text by 20 to 30 percent.`}
```
dangerismycat, line 145: typo: "…custom _strings_ in `dismissConfirmation`."

docs/pages/web/sheet.js
```diff
@@ -261,6 +273,221 @@ Sheet comes in 3 sizes: small (\`sm\`), medium (\`md\`), and large (\`lg\`).
...
           />
         </MainSection.Subsection>
+        <MainSection.Subsection
+          title="Dismiss confirmation"
+          description={`There are two ways Sheet can be dismissed: internally-controlled and externally-controlled dismiss actions.
+
+The three internally-controlled or component-controlled dismiss actions are:
+- when the \`ESC\` key is pressed
+- when the backdrop is clicked
+- when the dismiss IconButtons are clicked
```
dangerismycat, line 283: There are multiple dismiss icon buttons?

docs/pages/web/sheet.js
```diff
@@ -261,6 +273,221 @@ Sheet comes in 3 sizes: small (\`sm\`), medium (\`md\`), and large (\`lg\`).
...
+
+The externally-controlled dismiss actions require implementing the callback \`onDismissStart\`. See the [animation variant](#Animation) to learn more.
```
dangerismycat, line 285: Should we list what the three externally-controlled dismiss methods are here?

docs/pages/web/sheet.js
```diff
@@ -261,6 +273,221 @@ Sheet comes in 3 sizes: small (\`sm\`), medium (\`md\`), and large (\`lg\`).
...
+
+Sheets can contain forms or be part of flows where the user is required to submit infomation. If a Sheet is dismissed involuntarily, the data entered by the user could not be saved and lost. This can create a bad user experience.
+
+To prevent dismissing Sheet involuntary, we can use \`dismissConfirmation\`. When provided, it will open a confirmation modal each time component-controlled dismiss actions are triggered.
+
+The confirmation modal has a flexible API. When the \`dismissConfirmation\` prop is set to "true", Sheet uses defauld texts and labels.
```
dangerismycat, line 291: typo: "…_default_ texts and labels"

dangerismycat: Also: I would end that last sentence with a `:`, since it's followed by a list of the various texts and labels

packages/gestalt/src/Sheet.js
```diff
@@ -8,33 +8,64 @@ type NodeOrRenderProp = Node | (({| onDismissStart: () => void |}) => Node);
 
 type Props = {|
   /**
-    Supply a short, descriptive label for screen-readers as a text alternative to the Dismiss button. See the [Accessibility section](#Accessibility) for more info.
+    Supply a short, descriptive label for screen-readers as a text alternative to the Dismiss button. See the [Accessibility section](https://gestalt.pinterest.systems/web/sheet#Accessibility) for more info.
```
dangerismycat, line 11: ah thank you 🙏

packages/gestalt/src/Sheet.js
```diff
@@ -8,33 +8,64 @@ type NodeOrRenderProp = Node | (({| onDismissStart: () => void |}) => Node);
...
   /**
-   * Callback fired when the Sheet in/out animations end. See the [animation](#Animation) variant to learn more.
+   * Callback fired when the Sheet in/out animations end. See the [animation](https://gestalt.pinterest.systems/web/sheet#Animation) variant to learn more.
    */
   onAnimationEnd?: ({| animationState: 'in' | 'out' |}) => void,
+  /**
+   * When supplied, it will disable component-controlled dismiss actions (ESC key press, backdrop click, or built-in dismiss IconButtons) and launch a confirmation Popover next to the dismiss IconButton requesting user confirmation before proceding. See the [dismiss confirmation](https://gestalt.pinterest.systems/web/sheet#Dismiss-confirmation) variant to learn more.
+   */
+  dismissConfirmation?:
+    | boolean
```
dangerismycat, line 42: I don't really like the polymorphism of this prop…but I also can't think of a better solution. 😬 Would passing an empty object literal instead of `true` be better? Would it be more annoying? Less intuitive?

packages/gestalt/src/Sheet.jsdom.test.js
```diff
@@ -99,8 +100,7 @@ describe('Sheet', () => {
       </Sheet>,
     );
 
-    // eslint-disable-next-line testing-library/no-node-access -- Please fix the next time this file is touched!
```
dangerismycat, line 102: ❤️

packages/gestalt/src/Sheet.jsdom.test.js
```diff
@@ -296,4 +296,233 @@ describe('Sheet', () => {
...
+        closeOnOutsideClick
+        onDismiss={mockOnDismiss}
+      >
+        <section />
+      </Sheet>,
+    );
+    // eslint-disable-next-line testing-library/no-node-access -- Please fix the next time this file is touched!
+    const backDrop = screen.getByRole('dialog').parentElement?.firstElementChild;
+    if (!(backDrop instanceof HTMLElement)) {
+      throw new Error('Backdrop should be an HTMLElement');
```
dangerismycat, line 317: Throwing in a test?

packages/gestalt/src/Sheet.jsdom.test.js
```diff
@@ -296,4 +296,233 @@ describe('Sheet', () => {
...
+        dismissConfirmation
+        accessibilityDismissButtonLabel="Dismiss"
+        accessibilitySheetLabel="Test Sheet"
+        closeOnOutsideClick
+        onDismiss={mockOnDismiss}
+      >
+        <section />
+      </Sheet>,
+    );
+    // eslint-disable-next-line testing-library/no-node-access -- Please fix the next time this file is touched!
```
dangerismycat, line 314: 😭

packages/gestalt/src/Sheet.jsdom.test.js
```diff
@@ -296,4 +296,233 @@ describe('Sheet', () => {
 
     expect(mockOnDismiss).toHaveBeenCalledTimes(0);
   });
+
+  it('should not dismiss from the backdrop click when closeOnOutsideClick and dismissConfirmation are true', () => {
```
dangerismycat, line 300: Thanks for adding tests!

packages/gestalt/src/SheetConfirmationPopover.js
```diff
@@ -0,0 +1,161 @@
...
+import Text from './Text.js';
+import TrapFocusBehavior from './behaviors/TrapFocusBehavior.js';
+import { useDefaultLabelContext } from './contexts/DefaultLabelProvider.js';
+import { ESCAPE } from './keyCodes.js';
+
+type Props = {|
+  onDismiss: () => void,
+  anchor: ?HTMLElement,
+  dismissConfirmation?:
+    | boolean
```
dangerismycat, line 18: Here I don't think we need the polymorphism. When this component is rendered, we know the parent Sheet's `dismissConfirmation` is either `true` or an object. There also isn't need for those fields to be nested here. So in InternalSheet, we can do something like
```js
{showPopover && (
  <SheetConfirmationPopover
    anchor={dismissButtonRef.current}
    onDismiss={() => {
      setShowPopover(false);
      dismissButtonRef?.current?.focus();
    }}
    {...(typeof dismissConfirmation === 'object' ? dismissConfirmation : {})}
  />
)}
```
allowing for a flatter props structure here:
```js
type Props = {|
  anchor: ?HTMLElement,
  onDismiss: () => void,
  message?: string,
  subtext?: string,
  primaryAction?: {|
    accessibilityLabel?: string,
    text?: string,
    onClick?: ({|
      event:
    |}) => void,
  |},
  secondaryAction?: {|
    accessibilityLabel?: string,
    text?: string,
    onClick?: ({|
      event:
    |}) => void,
  |},
|};
```

packages/gestalt/src/Sheet.js
dangerismycat: Actually the more I think about it I think an empty object is a better solution. It'll work anyways…and it's communicating the same thing ("I'd like a dismiss confirmation") while also communicating "I don't care to override any of the default labels"

dangerismycat: Like I think the polymorphism is also unintuitive, so if we're deciding against two unintuitive solutions, the one that makes more sense from a type perspective seems better?

packages/gestalt/src/SheetConfirmationPopover.js
```diff
@@ -0,0 +1,161 @@
...
+
+  useEffect(() => {
+    if (confirmationButtonRef.current) {
+      confirmationButtonRef.current.focus();
+    }
+  }, [confirmationButtonRef]);
+
+  // Handle onDismiss triggering from ESC keyup event
+  useEffect(() => {
+    function handleKeyDown(event: {| keyCode: number, stopPropagation: () => void |}) {
```
dangerismycat, line 73: Do we need to add the `event` type here? Shouldn't Flow be able to infer that?

packages/gestalt/src/SheetConfirmationPopover.js
```diff
@@ -0,0 +1,161 @@
...
+      if (event.keyCode === ESCAPE) {
+        event.stopPropagation();
+      }
+    }
+
+    window.addEventListener('key', handleKeyDown);
```
dangerismycat, line 79: This should be `'keydown'`, yes?

packages/gestalt/src/SheetConfirmationPopover.js
```diff
@@ -0,0 +1,161 @@
...
+      role="dialog"
+      size="md"
+    >
+      <TrapFocusBehavior>
+        <Box padding={3} width="100%">
+          <Flex direction="column" gap={4}>
+            <Box role="alert">
+              <Flex direction="column" gap={2} alignItems="center" width="100%">
+                <Text weight="bold">
+                  {typeof dismissConfirmation === 'boolean'
```
dangerismycat, line 100: The props changes mentioned above would simplify all this logic as well

docs/pages/web/sheet.js
AlbertCarreras: In my mind yes, for the user no. In the code, there are 2 haha

packages/gestalt/src/SheetConfirmationPopover.js
AlbertCarreras: DONE. thanks

packages/gestalt/src/SheetConfirmationPopover.js
AlbertCarreras: DONE

packages/gestalt/src/SheetConfirmationPopover.js
AlbertCarreras: DONE

packages/gestalt/src/SheetConfirmationPopover.js
AlbertCarreras: DONE

docs/pages/web/sheet.js
AlbertCarreras: DONE

docs/pages/web/sheet.js
AlbertCarreras: DONE

docs/pages/web/sheet.js
AlbertCarreras: DONE

docs/pages/web/sheet.js
AlbertCarreras: DONE

packages/gestalt/src/Sheet.jsdom.test.js
AlbertCarreras: DONE

packages/gestalt/src/Sheet.js
AlbertCarreras: DONE

packages/gestalt/src/Sheet.jsdom.test.js
dangerismycat: I'm still seeing the line about throwing an error. Not a blocker, just seems odd to need to throw in a test?

packages/gestalt/src/InternalSheet.js
```diff
@@ -28,8 +29,34 @@ type InternalSheetProps = {|
   children: NodeOrRenderProp,
   closeOnOutsideClick: boolean,
   footer: NodeOrRenderProp,
-  heading: NodeOrRenderProp,
+  heading?: string,
   onAnimationEnd: ?({| animationState: 'in' | 'out' |}) => void,
+  dismissConfirmation?: {|
```
dangerismycat, line 34: There's no need to keep this level of nesting — spread these props to SheetConfirmationPopover in InternalSheet, keep the props type here more flat

packages/gestalt/src/InternalSheet.js
```diff
@@ -74,17 +106,21 @@ export default function InternalSheet({
 
   // Handle onDismiss triggering from ESC keyup event
   useEffect(() => {
-    function handleKeyUp(event: {| keyCode: number |}) {
-      if (event.keyCode === ESCAPE) {
+    function handleKeyDown(event: {| keyCode: number |}) {
```
dangerismycat, line 109: I'm still seeing the `event` Flow type here — is it needed? Shouldn't Flow be able to infer it?

packages/gestalt/src/SheetConfirmationPopover.js
```diff
@@ -0,0 +1,143 @@
...
+                  dismissConfirmation?.secondaryAction?.accessibilityLabel ??
+                  dismissConfirmationSecondaryActionTextLabelDefault
+                }
+                color="gray"
+                text={
+                  dismissConfirmation?.secondaryAction?.text ??
+                  dismissConfirmationSecondaryActionTextDefault
+                }
+                onClick={({ event }) => {
+                  if (typeof dismissConfirmation !== 'boolean') {
```
dangerismycat, line 113: This logic is no longer relevant

packages/gestalt/src/SheetConfirmationPopover.js
```diff
@@ -0,0 +1,143 @@
...
+                accessibilityLabel={
+                  dismissConfirmation?.primaryAction?.accessibilityLabel ??
+                  dismissConfirmationPrimaryActionTextLabelDefault
+                }
+                text={
+                  dismissConfirmation?.primaryAction?.text ??
+                  dismissConfirmationPrimaryActionTextDefault
+                }
+                onClick={({ event }) => {
+                  if (typeof dismissConfirmation !== 'boolean') {
```
dangerismycat, line 130: Here too

dangerismycat (approved):

packages/gestalt/src/SheetConfirmationPopover.js
AlbertCarreras: DONE

packages/gestalt/src/InternalSheet.js
AlbertCarreras: DONE

packages/gestalt/src/InternalSheet.js
AlbertCarreras: DONE

packages/gestalt/src/SheetConfirmationPopover.js
AlbertCarreras: DONE

packages/gestalt/src/InternalSheet.js
```diff
@@ -28,10 +29,36 @@ type InternalSheetProps = {|
   children: NodeOrRenderProp,
   closeOnOutsideClick: boolean,
   footer: NodeOrRenderProp,
-  heading: NodeOrRenderProp,
+  enabledDismiss: boolean,
+  heading?: string,
+  message?: string,
+  message?: string,
```
dangerismycat, line 35: It looks like `message` got duplicated here.

Also: I wouldn't recommend spreading these props yet. To InternalSheet, the abstraction of `dismissConfirmation` as an object makes sense, as that's just a small part of what InternalSheet does (and really InternalSheet doesn't do anything with the prop aside from pass it along). But it does make sense to spread the `dismissConfirmation` fields to SheetConfirmationPopover, since that component does care about the individual fields. So something like
```js
// Sheet.js
type Props = {|
  dismissConfirmation: {| message, primaryAction, etc |}
|};
...
<InternalSheet dismissConfirmation={dismissConfirmation} />
...
```
```js
// InternalSheet.js
type Props = {|
  dismissConfirmation: {| message, primaryAction, etc |}
|};
...
<SheetConfirmationPopover {...dismissConfirmation} />
...

```
```js
// SheetConfirmationPopover.js
type Props = {|
  message: string,
  primaryAction: {| .. |},
  etc
|}
```
Does that make sense? Like continue to use the object abstraction as a way of keeping the related props together through Sheet and InternalSheet, then spread them where they're being used (SheetConfirmationPopover) and the object abstraction is no longer necessary

AlbertCarreras: DONE/UNDONE

PR #2631, opened 2023-01-30
Upsell: extend `message` prop API from string to string | Text

AlbertCarreras:
#### What changed?
Screenshot 2023-01-30 at 6 41 09 PM

Upsell: extend `message` prop API from string to string | Text

#### Why?

More flexibility

### Checklist


packages/gestalt/src/Upsell.css
```diff
@@ -9,3 +9,32 @@ html:not([dir="rtl"]) .rtlPos {
   right: 0;
   top: 0;
 }
+
+.textColorOverrideLight * {
```
dangerismycat, line 13: I believe you mentioned this in the Slack convo, but let's fast-follow with a PR consolidating the repeated styles/markup to an internal message component

dangerismycat (approved):

PR #2640, opened 2023-02-09
Internal: fix "open in codesandbox" dependencies

AlbertCarreras:
#### What changed?

Internal: fix "open in codesandbox" dependencies
Fixes a bug when opening codesandbox from example in Datepicker

Screenshot 2023-02-08 at 11 45 08 PM

### DIFF
--- a/docs/docs-components/handleCodeSandbox.js
+++ b/docs/docs-components/handleCodeSandbox.js
@@ -61,11 +61,8 @@ const handleCodeSandbox = async ({ code, title }: {| code: string, title: string
           dependencies: {
             react: 'latest',
             'react-dom': 'latest',
-            // $FlowIssue[exponential-spread]
-            ...(baseComponents.length > 0 ? { gestalt: 'latest' } : {}),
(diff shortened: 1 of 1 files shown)


docs/docs-components/handleCodeSandbox.js
```diff
@@ -61,11 +61,8 @@ const handleCodeSandbox = async ({ code, title }: {| code: string, title: string
           dependencies: {
             react: 'latest',
             'react-dom': 'latest',
-            // $FlowIssue[exponential-spread]
-            ...(baseComponents.length > 0 ? { gestalt: 'latest' } : {}),
-            ...(additionalComponents.includes('DatePicker')
-              ? { 'gestalt-datepicker': 'latest' }
-              : {}),
+            gestalt: 'latest',
+            'gestalt-datepicker': 'latest',
```
dangerismycat, line 65: Thanks for fixing this. 🙂 I'm curious if including `gestalt-datepicker` for all sandboxes will increase loading times for those not using DatePicker? We may need to re-add that logic to ensure we only load this package when necessary

PR #2647, opened 2023-02-16
Internal: add first integration test for Masonry

dangerismycat:
Once upon a time there existed a suite of integration tests for Masonry. It first lived in this repo, then was migrated to Pinboard, then was turned off, then was deleted. womp womp.

This PR is the first of many adding this test suite back. The approach is fairly different, using a regular (but unlinked) route on our docs site instead of setting up a bespoke mini-app, and using Playwright instead of Puppeteer.

I started with a test that I thought looked pretty simple, but really had no idea. The test file has two tests. The first test now passes completely. The second test no longer throws any errors, but also times out without passing. I'm still trying to debug what's going on there, but I wanted to get this code checked in to get other 👀 on it, and I'll keep chipping away at the skipped test.

### DIFF
--- a/playwright/masonry/fetch-more.spec.mjs
+++ b/playwright/masonry/fetch-more.spec.mjs
@@ -1,0 +1,64 @@
+// @flow strict
+import { expect, test } from '@playwright/test';
+import getServerURL from './utils/getServerURL.mjs';
+import waitForRenderedItems from './utils/waitForRenderedItems.mjs';
+import selectors from './utils/selectors.mjs';
+
+test.describe('Masonry: fetch more', () => {
+  test('trigger a call to "fetchMore" when container resizes', async ({
+    page,
+  }) => {
+    // Start with a small screen.
+    await page.setViewportSize({ width: 400, height: 400 });
+    await page.goto(getServerURL());
+    await waitForRenderedItems(page, { targetItemsGTE: 20 });
+
+    const initialFetchCount = await page.evaluate(
+      () => window.TEST_FETCH_COUNTS
+    );
+    // TODO[@rjames]: re-enable this eventually (possibly needs better SSR support?)
+    // expect(initialFetchCount).toBe(0);
+
+    // Fetches more if the viewport is big enough.
+    await page.setViewportSize({ width: 2000, height: 1000 });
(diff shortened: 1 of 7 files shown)


playwright/masonry/fetch-more.spec.mjs
```diff
@@ -0,0 +1,64 @@
+// @flow strict
```
dangerismycat, line 1: adapted from https://phabricator.pinadmin.com/rPdafa476383ce74f8a8425241049c9e4f4208059b#change-Kz9yKfWwwTFo

playwright/masonry/utils/getServerURL.mjs
```diff
@@ -0,0 +1,38 @@
...
+
+const normalizeValue = (val) => {
+  if (typeof val === 'boolean') {
+    return val ? '1' : '0';
+  }
+  return String(val);
+};
+
+/*::
+type Options = ?{|
```
dangerismycat, line 14: https://flow.org/en/docs/types/comments/

playwright/masonry/fetch-more.spec.mjs
```diff
@@ -0,0 +1,64 @@
...
+  }) => {
+    // Start with a small screen.
+    await page.setViewportSize({ width: 400, height: 400 });
+    await page.goto(getServerURL());
+    await waitForRenderedItems(page, { targetItemsGTE: 20 });
+
+    const initialFetchCount = await page.evaluate(
+      () => window.TEST_FETCH_COUNTS
+    );
+    // TODO[@rjames]: re-enable this eventually (possibly needs better SSR support?)
```
AlbertCarreras, line 19: is this TODO still valid? I dont see any disabled

playwright/masonry/utils/resizeWidth.mjs
```diff
@@ -0,0 +1,24 @@
+// @flow strict
```
AlbertCarreras, line 1: Are we using this file rn?

AlbertCarreras (approved):

playwright/masonry/utils/resizeWidth.mjs
dangerismycat: If not, we will be very soon (in the rest of the suite of tests)

PR #2656, opened 2023-02-22
Update package.json

moti-h2:
Please DO NOT APPROVE THIS


dangerismycat (changes requested):
Not sure what this is — if this change is needed, please re-open and give us more context on it

PR #2657, opened 2023-02-22
Docs: Remove Year in Review banner

ponori:
Given it's nearly March, it seems appropriate to remove the 2022 year in review banner.

### DIFF
--- a/docs/docs-components/AppLayout.js
+++ b/docs/docs-components/AppLayout.js
@@ -50,7 +50,6 @@ export default function AppLayout({ children, colorScheme }: Props): Node {
     <Box minHeight="100vh" color="default">
       <SkipToContent />
       <Header />
-      {isHomePage}
       {isSidebarOpen && (
         <Fragment>
           {/* The <div> element has a child <button> element that allows keyboard interaction */}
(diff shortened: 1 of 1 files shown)


docs/docs-components/AppLayout.js
```diff
@@ -51,7 +50,7 @@ export default function AppLayout({ children, colorScheme }: Props): Node {
     <Box minHeight="100vh" color="default">
       <SkipToContent />
       <Header />
-      {isHomePage && <YearInReviewBanner />}
+      {isHomePage}
```
dangerismycat, line 53: This line needs to be removed completely

ponori:
Thanks! Done.

AlbertCarreras:
@ponori @dangerismycat Should we do deeper cleanup? Maybe not reverting the full PR.

There are still a few lines that should be cleanup for example https://github.com/pinterest/gestalt/pull/2558/files#diff-90409f25a385b55363f71806274620855df572340cc7f936fdb08c583445e58dR18

https://github.com/pinterest/gestalt/pull/2558/files#diff-ca731702917dd5c437bb5549907d0dae2a2d512305a5531bc2e34b7a9354dcc7

dangerismycat (approved):
@AlbertCarreras I think the rest of those changes from Ayesha's PR we would keep, as we want the Year in Review page to still be available — we just don't need to link to it from the home page. Perhaps we could add a link to it on the Roadmap page or something, though that could happen in a later PR.

PR #2562, opened 2022-12-12
Popover: Controlled height based on size of contents

EduhCosta:
### Summary

This change handle and fix the following error, reported to us by [this thread](https://pinterest.slack.com/archives/C13KLG5P0/p1658953735281499):
"When we have more content inner dropdown children than the viewport height supports and the anchor is positioned on second half (bottom half) of screen, the initial elements is not accessible" 

The images below provide you the bug view:

Screen Shot 2022-07-27 at 1 28 10 PM

BUG's reproduction: https://codesandbox.io/s/dropdown-size-error-3jo09u

## Gestalt component involved

- Popover
- Dropdown
- Content

## Solution

We fix the max height of children's Popover content to 90% of the viewport's vertical size. If the viewport contains a `ScrollBoundaryContainer` component, we use the height of this component to set the max height of popover content. In other words, the Popover max size is `90%  screen height` or `90% of ScrollBoundaryContainer height`.

The second solution part is "top positioning" If we have a negative top value, which means if the `Controller` sets the top with a negative value (< 0), the popover commonly will not allow access to part of its internal content. In front of this case, we set the top's value as `5%` of the height available (following the height decisions explained above).

_[More about Controller and Popover structure here](https://paper.dropbox.com/doc/Popovers-History--BuQSFUfQbWwCgcd5HWMlpQ3VAg-e5Yrz9jiksN0ce8c893EJ)_

To understand better each concept, please read the comments attached on code.

## How to test it?

1. Access our current branch deploy: https://deploy-preview-2562--gestalt.netlify.app/
2. Go to Dropdown page: `/web/dropdown`
3. Resizes your screen to position the first example button to the second half of screen, like the image below:
Example of screen size
4. Expand the code example;
5. Use one the following codes to test the solution:

#### Example 01
```js
function Example() {
  const [open, setOpen] = React.useState(false);
  const [selected, setSelected] = React.useState(null);
  const anchorRef = React.useRef(null);
  const onSelect = ({ item }) => setSelected(item);

  const items = new Array(100).fill({ name: "item" }).map((item, index) => (
    <Dropdown.Item
      key={item.name + index}
      onSelect={onSelect}
      option={{ value: 'item ' + index, label: 'Item ' + index }}
      selected={selected}
    />
  ));

  return (
    <Flex height="100%" justifyContent="center">
      <Button
        accessibilityControls="demo-dropdown-example"
        accessibilityExpanded={open}
        accessibilityHaspopup
        iconEnd="arrow-down"
[...]

PR #2666, opened 2023-02-24
Popover, Dropdown: revert https://github.com/pinterest/gestalt/pull/2562 and add maxheight 30vh to Dropdown

AlbertCarreras:
#### What changed?

Popover, Dropdown: revert https://github.com/pinterest/gestalt/pull/2562 and add `maxheight="30vh"` to Dropdown

BEFORE
Screenshot 2023-02-24 at 3 52 49 PM

AFTER
Screenshot 2023-02-24 at 3 54 39 PM

#### Why?

The solution in https://github.com/pinterest/gestalt/pull/2562 does not work with small height ScrollBoundaryContainers. An overengineered solution that wasn't scalable to future problems.

Reducing the logic like we do in other components implementations, `maxHeight="30vh"` that is just so simple we can all understand how it works in most use cases and implementations.

More context here https://paper.dropbox.com/doc/Popovers-History--BzXlCWDyIuGRkiRJ0Nt_TEc5Ag-e5Yrz9jiksN0ce8c893EJ

### DIFF
--- a/packages/gestalt/src/Contents.js
+++ b/packages/gestalt/src/Contents.js
@@ -202,25 +202,6 @@ class Contents extends Component<Props, State> {
     }
   };
 
-  calcTopHeight() {
-    if (!window || !document) {
-      return { top: null, height: null };
-    }
-
-    const { scrollBoundaryContainerRef, positionRelativeToAnchor } = this.props;
-
-    // Define the height based the reference to render: ScrollBoundaryContainer or screen viewport
-    const height = scrollBoundaryContainerRef?.offsetHeight ?? window.innerHeight ?? 0;
-
-    // 5% of height available
-    const top = (height / 10) * 0.5;
-
-    // 90% of height available on container reference
-    const elementHeight = (height / 10) * 9;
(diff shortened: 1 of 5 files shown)


packages/gestalt/src/Contents.js
```diff
@@ -202,25 +202,6 @@ class Contents extends Component<Props, State> {
...
 
-  calcTopHeight() {
-    if (!window || !document) {
-      return { top: null, height: null };
-    }
-
-    const { scrollBoundaryContainerRef, positionRelativeToAnchor } = this.props;
-
-    // Define the height based the reference to render: ScrollBoundaryContainer or screen viewport
-    const height = scrollBoundaryContainerRef?.offsetHeight ?? window.innerHeight ?? 0;
```
EduhCosta, line 213: Did you try remove only this `scrollBoundaryContainerRef?.offsetHeight` of code?
You are right, when I created this code set the `scrollBoundaryContainerRef?.offsetHeight` as a controller of size.

PR #2667, opened 2023-02-28
Masonry: Add  `handle-item-updates` integration test

dangerismycat:
Another Masonry integration test working! This one tests Masonry's ability to reflow the layout after item sizes have updated. Also fixed other errors related to these integration tests as noted


docs/integration-test-helpers/masonry/MasonryContainer.js
```diff
@@ -238,9 +238,9 @@ export default class MasonryContainer extends Component<Props, State> {
   };
 
   // $FlowFixMe[unclear-type]
-  renderItem: ({| data: any |}) => Node = ({ data }) => {
+  renderItem: ({| data: any, itemIdx: number |}) => Node = ({ data, itemIdx }) => {
     const { expanded } = this.state;
-    return <ExampleGridItem expanded={expanded} {...data} />;
+    return <ExampleGridItem expanded={expanded} data={data} itemIdx={itemIdx} />;
```
dangerismycat, line 243: This wasn't passing through ExampleGridItem's props correctly before — perils of an unsealed Props type!

docs/pages/integration-test/masonry.js
```diff
@@ -7,6 +7,12 @@ import { generateExampleItems } from '../../integration-test-helpers/masonry/ite
 
 const measurementStore = Masonry.createMeasurementStore();
 
+// This is the counterpart to `normalizeValue` in `playwright/masonry/utils/getServerURL.mjs`
+// This will likely need to be updated in the near future
+function booleanize(value: '1' | '0'): boolean {
```
dangerismycat, line 12: `normalizeValue` can use more than just `'1' | '0'`for query param values, but I haven't come across that yet — I'll update this when I do

playwright/masonry/utils/selectors.mjs
```diff
@@ -1,10 +1,10 @@
 // @flow strict
 
 const selectors = {
-  afterGrid: '.afterGrid',
-  gridItem: '[data-grid-item], .Collection-Item',
```
dangerismycat, line 5: This class name isn't valid due to our hashing, but also isn't needed. There aren't other `data-grid-item`s outside of Masonry.

playwright/masonry/utils/waitForRenderedItems.mjs
```diff
@@ -82,7 +82,7 @@ export default async function waitForRenderedItems(
         const { documentElement } = document;
         if (
           typeof scrollHeight !== 'undefined' &&
-          documentElement?.scrollHeight !== scrollHeight
+          documentElement?.scrollHeight >= scrollHeight
```
dangerismycat, line 85: As previously written, this check is a bit odd — we shouldn't need to care that we're at an exact scrollHeight, just that we've surpassed a minimum

playwright/playwright.config.js
```diff
@@ -1,7 +1,7 @@
 // @flow strict
 const { devices } = require('@playwright/test');
 
-const reporter /*: string */ = process.env.CI ? 'github' : 'list';
+const reporter /*: 'github' | 'list' */ = process.env.CI ? 'github' : 'list';
```
dangerismycat, line 4: Does this matter? No. Is it more exact and technically more safe? Sure. ¯\_(ツ)_/¯

AlbertCarreras (approved):

PR #2672, opened 2023-02-28
PopoverEducational: move examples to Sandpack + added `shouldFocus` to props

AlbertCarreras:
#### What changed?

PopoverEducational: move examples to Sandpack + added `shouldFocus` to props

### DIFF
--- a/docs/examples/popovereducational/customContent.js
+++ b/docs/examples/popovereducational/customContent.js
@@ -3,11 +3,11 @@ import { type Node, useRef, useEffect, useState } from 'react';
 import { Box, Flex, IconButton, Button, Text, PopoverEducational } from 'gestalt';
 
 export default function Example(): Node {
-  const [openA, setOpenA] = useState(false);
+  const [open, setOpen] = useState(false);
   const anchorRefA = useRef();
 
   useEffect(() => {
-    setOpenA(true);
+    setOpen(true);
   }, []);
 
   return (
@@ -20,7 +20,7 @@ export default function Example(): Node {
         ref={anchorRefA}
         size="lg"
       />
-      {openA && (
+      {open && (
         <PopoverEducational
           accessibilityLabel={`Description of new "More ideas" feature`}
           anchor={anchorRefA.current}
--- a/docs/examples/popovereducational/customContent.js
+++ b/docs/examples/popovereducational/customContent.js
@@ -1,0 +1,44 @@
+// @flow strict
+import { type Node, useRef, useEffect, useState } from 'react';
+import { Box, Flex, IconButton, Button, Text, PopoverEducational } from 'gestalt';
+
+export default function Example(): Node {
+  const [openA, setOpenA] = useState(false);
+  const anchorRefA = useRef();
+
+  useEffect(() => {
+    setOpenA(true);
+  }, []);
+
+  return (
+    <Flex alignItems="center" justifyContent="center" height="100%" width="100%">
+      <IconButton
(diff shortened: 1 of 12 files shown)


packages/gestalt/src/PopoverEducational.js
```diff
@@ -166,10 +171,11 @@ export default function PopoverEducational({
         onDismiss={onDismiss}
         positionRelativeToAnchor
         showCaret
+        shouldFocus={shouldFocus ?? false}
         role={primaryAction && !children ? 'dialog' : role}
         size={size}
       >
-        {children ?? message ? (
+        {!children && message ? (
```
AlbertCarreras, line 178: This is the right logic, see snapshot change.

docs/examples/popovereducational/customContent.js
```diff
@@ -0,0 +1,44 @@
+// @flow strict
+import { type Node, useRef, useEffect, useState } from 'react';
+import { Box, Flex, IconButton, Button, Text, PopoverEducational } from 'gestalt';
+
+export default function Example(): Node {
+  const [openA, setOpenA] = useState(false);
```
dangerismycat, line 6: nit: the `A` suffix isn't necessary when there's nothing else to differentiate it from

docs/pages/web/popovereducational.js
```diff
@@ -368,9 +98,11 @@ function Example() {
...
+
+If PopoverEducational is already visible, we need its content to be keyboard accessible in sequential order. Don't use Layer to wrap PopoverEducational as it would move PopoverEducational outside the DOM hierarchy of the parent component and it will lose contextual sequencial order. The content will placed last in the keyboard navigations sequence, becoming unreachable in its content context. Moreover, make sure PopoverEducational is implemented right after the code of the anchor element so that it navigates the popover right after the anchor.
 
-If PopoverEducational is already visible, we need its content to be keyboard accessible in sequential order. Don't use Layer to wrap PopoverEducational as it would move PopoverEducational outside the DOM hierarchy of the parent component and it will lose contextual sequencial order. The content will placed last in the keyboard navigations sequence, becoming unreachable in its content context.
+If PopoverEducational
```
dangerismycat, line 105: this looks like a mistake?

packages/gestalt/src/PopoverEducational.js
dangerismycat: That's not what I was suggesting. You don't need to reference `children` in the markup twice. The logic is saying:
- if there's `children`, render that
- else if there isn't children:
  - if there's `message`, render that with some other markup
  - else render null

So in code, we can express that as
```js
{children ?? message ? {markup} : null}
```
There is no need for an additional `children && null` line. This makes the relationship between `children` and `message` very clear, and eliminates any possibility for bugs where a reference to `children` gets changed and we render both `message` _and_ `children`.

dangerismycat (approved):

packages/gestalt/src/PopoverEducational.js
AlbertCarreras: @dangerismycat We don't control the padding in children (as there are many existing instances) so if there's children, there can't be message+primaryAction. Children and message are exclusive.

AlbertCarreras: But it doesn't work 
Screenshot 2023-03-01 at 12 43 44 AM

AlbertCarreras: It needs parethneses
Screenshot 2023-03-01 at 12 44 58 AM

PR #2675, opened 2023-03-01
Dropdown, Popover: Add `maxHeight` prop

EduhCosta:
#### What changed?

We add the new property `maxHeight` to Dropdown component to control the size of Popover container and  removing `ScrollBoundaryContainer` calculation when this property is defined.

#### Why?

To understand better, please read the thread: https://pinterest.slack.com/archives/C13KLG5P0/p1676662974832019

### Explanation

https://capture.dropbox.com/tCVVMlf5LTJoLY7q

### DIFF
--- a/packages/gestalt/src/Contents.js
+++ b/packages/gestalt/src/Contents.js
@@ -45,6 +45,7 @@ type OwnProps = {|
   shouldFocus?: boolean,
   triggerRect: ?ClientRect,
   width: ?number,
+  __dangerouslyIgnoreScrollBoundaryContainerSize?: boolean,
 |};
 
 type HookProps = {|
@@ -207,10 +208,16 @@ class Contents extends Component<Props, State> {
       return { top: null, height: null };
     }
 
-    const { positionRelativeToAnchor } = this.props;
+    const {
+      __dangerouslyIgnoreScrollBoundaryContainerSize,
+      scrollBoundaryContainerRef,
+      positionRelativeToAnchor,
+    } = this.props;
 
     // Define the height based the reference to render: ScrollBoundaryContainer or screen viewport
-    const height = window.innerHeight ?? 0;
+    let height = window.innerHeight ?? 0;
+    if (__dangerouslyIgnoreScrollBoundaryContainerSize) height = window.innerHeight ?? 0;
+    else height = scrollBoundaryContainerRef?.offsetHeight ?? window.innerHeight ?? 0;
 
     // 5% of height available
     const top = (height / 10) * 0.5;
--- a/packages/gestalt/src/Contents.js
+++ b/packages/gestalt/src/Contents.js
@@ -216,8 +216,9 @@ class Contents extends Component<Props, State> {
 
     // Define the height based the reference to render: ScrollBoundaryContainer or screen viewport
     let height = window.innerHeight ?? 0;
-    if (__dangerouslyIgnoreScrollBoundaryContainerSize) height = window.innerHeight ?? 0;
-    else height = scrollBoundaryContainerRef?.offsetHeight ?? window.innerHeight ?? 0;
+    if (!__dangerouslyIgnoreScrollBoundaryContainerSize) {
+      height = scrollBoundaryContainerRef?.offsetHeight ?? window.innerHeight ?? 0;
+    }
 
     // 5% of height available
     const top = (height / 10) * 0.5;
--- a/packages/gestalt/src/Contents.js
+++ b/packages/gestalt/src/Contents.js
@@ -216,8 +216,8 @@ class Contents extends Component<Props, State> {
 
     // Define the height based the reference to render: ScrollBoundaryContainer or screen viewport
     let height = window.innerHeight ?? 0;
-    if (!__dangerouslyIgnoreScrollBoundaryContainerSize) {
-      height = scrollBoundaryContainerRef?.offsetHeight ?? window.innerHeight ?? 0;
+    if (!__dangerouslyIgnoreScrollBoundaryContainerSize && scrollBoundaryContainerRef) {
+      height = scrollBoundaryContainerRef.offsetHeight;
     }
 
     // 5% of height available
--- a/packages/gestalt/src/Contents.js
+++ b/packages/gestalt/src/Contents.js
@@ -207,10 +207,10 @@ class Contents extends Component<Props, State> {
       return { top: null, height: null };
     }
 
-    const { scrollBoundaryContainerRef, positionRelativeToAnchor } = this.props;
+    const { positionRelativeToAnchor } = this.props;
 
     // Define the height based the reference to render: ScrollBoundaryContainer or screen viewport
-    const height = scrollBoundaryContainerRef?.offsetHeight ?? window.innerHeight ?? 0;
+    const height = window.innerHeight ?? 0;
 
     // 5% of height available
     const top = (height / 10) * 0.5;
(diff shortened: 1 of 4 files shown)


packages/gestalt/src/Dropdown.js
```diff
@@ -220,8 +222,18 @@ export default function Dropdown({
       role="menu"
       shouldFocus
       size="xl"
+      __dangerouslyIgnoreScrollBoundaryContainerSize={
```
AlbertCarreras, line 225: let's do something more explicit.

1. this is a boolean, https://www.30secondsofcode.org/articles/s/javascript-boolean-trap
2. __dangerouslySetMaxWidth?: "30vh"

packages/gestalt/src/Dropdown.js
```diff
@@ -220,8 +222,18 @@ export default function Dropdown({
...
+      }
     >
-      <Box alignItems="center" direction="column" display="flex" flex="grow" margin={2}>
+      <Box
+        alignItems="center"
+        direction="column"
+        display="flex"
+        flex="grow"
+        margin={2}
+        maxHeight={__dangerouslyIgnoreScrollBoundaryContainerSize ? '30vh' : 'auto'}
```
AlbertCarreras, line 235: let's leave        ` maxHeight={__dangerouslyIgnoreScrollBoundaryContainerSize ? '30vh' : undefined} `

I'm not sure if auto equals undefined here, therefore, let's leave the equivalent.

Screenshot 2023-03-06 at 6 14 54 PM

AlbertCarreras: Or maybe somehting like         {...(lineClamp ? { style: { WebkitLineClamp: lineClamp } } : {})}

packages/gestalt/src/Contents.js
```diff
@@ -207,10 +208,17 @@ class Contents extends Component<Props, State> {
...
+      __dangerouslyIgnoreScrollBoundaryContainerSize,
+      scrollBoundaryContainerRef,
+      positionRelativeToAnchor,
+    } = this.props;
 
     // Define the height based the reference to render: ScrollBoundaryContainer or screen viewport
-    const height = scrollBoundaryContainerRef?.offsetHeight ?? window.innerHeight ?? 0;
+    let height = window.innerHeight ?? 0;
+    if (!__dangerouslyIgnoreScrollBoundaryContainerSize) {
+      height = scrollBoundaryContainerRef?.offsetHeight ?? window.innerHeight ?? 0;
```
AlbertCarreras, line 220: I'm curious, why do we use `scrollBoundaryContainerRef?.offsetHeight` instead of  `window.innerHeight`, can't we just use `window.innerHeight`?

packages/gestalt/src/Dropdown.js
AlbertCarreras: Also, add comment on why we are doing this here. It's undocumented but we should leave a description for internal knowldge

// __dangerouslySetMaxWidth can be set when `ScrollBoundaryContainer` is set to overflow="visible" but therefore limits the height of the Popover-based component.  Some cases require

packages/gestalt/src/Controller.js
```diff
@@ -150,6 +151,10 @@ class Controller extends Component<Props, State> {
           shouldFocus={shouldFocus}
           triggerRect={triggerBoundingRect}
           width={width}
+          __dangerouslyIgnoreScrollBoundaryContainerSize={
+            // eslint-disable-next-line no-underscore-dangle
+            this.props.__dangerouslyIgnoreScrollBoundaryContainerSize
```
AlbertCarreras, line 156: `          __dangerouslyIgnoreScrollBoundaryContainerSize={__dangerouslySetMaxWidth === '30vh"}`

packages/gestalt/src/Contents.js
AlbertCarreras:
```
    let height = window.innerHeight ?? 0;
    if (!__dangerouslyIgnoreScrollBoundaryContainerSize) {
      height = scrollBoundaryContainerRef?.offsetHeight ?? height; >> redundant
```

AlbertCarreras: maybe 

```
    let height = window.innerHeight ?? 0;
    if (!__dangerouslyIgnoreScrollBoundaryContainerSize && scrollBoundaryContainerRef?.offsetHeight ) {
      height = scrollBoundaryContainerRef?.offsetHeight
```

AlbertCarreras:
What happens if we do not set 30vh but we do remove the ScrollBoundaryContainer reference? Would the Dropdown go to the 90vh?

packages/gestalt/src/Contents.js
dangerismycat: Either of these works. I have a slight preference for the first one but not that strong.

a nit on the second logic is that optional chaining wouldn't be necessary after verifying the existence of scrollBoundaryContainerRef in the conditional

packages/gestalt/src/Dropdown.js
dangerismycat: I believe you mean `dangerouslySetMaxHeight`.

Does this need to be `dangerous` though? What if we just give users a `maxHeight` prop and recommend against using it in most situations?

packages/gestalt/src/Dropdown.js
dangerismycat: The failing snapshots below are an example of Alberto's concerns with using `'auto'` instead of `undefined`

packages/gestalt/src/Dropdown.js
AlbertCarreras: that works too

AlbertCarreras: maxHeight works. although I wanna know about what happens if first: https://github.com/pinterest/gestalt/pull/2675#issuecomment-1457221445

packages/gestalt/src/Dropdown.js
AlbertCarreras: If we follow Ryan's comment, then we should document the prop https://github.com/pinterest/gestalt/pull/2675#discussion_r1127246052

Which honestly makes sense not keeping undocumented.

Let's not add comment but document the prop "prop"erly  xd

EduhCosta:
> What happens if we do not set 30vh but we do remove the ScrollBoundaryContainer reference? Would the Dropdown go to the 90vh?

Yep, Dropdown will go.
But this solution will trigger a "dangerous behaviour", what happened on the previous case shared [on this thread](https://pinterest.slack.com/archives/G01GN7CE99S/p1677704953449709?thread_ts=1677703596.447549&cid=G01GN7CE99S), and based on @dangerismycat [concerns](https://pinterest.slack.com/archives/G01GN7CE99S/p1677704989625679?thread_ts=1677703596.447549&cid=G01GN7CE99S) we need booth props working together, to avoid the `height increasing` not expected behaviour.

packages/gestalt/src/Contents.js
EduhCosta: @AlbertCarreras 
We handled this prop based on `scrollBoundaryContainer` height because in some cases/screens on Pinboard when we define the size based on screen, the Popover renders under the TopBar. And we not have a case(tracked) where the scrollBoundaryContainer is under the TopBar; 

PS:. The zIndex is not fixing the Popover visibility to these cases. e.g. `/pin-builder`

packages/gestalt/src/Controller.js
EduhCosta: This was made on Popover component level and propagate to down, because the component structure documented [here](https://paper.dropbox.com/doc/Popovers-History--B0B6_pXAfccwGF~0PRE7Xi1MAg-e5Yrz9jiksN0ce8c893EJ).

AlbertCarreras (approved):
There are some cleanups needed for good to go

PR #2702, opened 2023-03-07
Internal: Add Troubleshooting section to Readme

rlingineni:
Adding a troubleshooting section here that links to our development FAQ

PR #2705, opened 2023-03-08
Github Workflow: Auto-Format and Commit

rlingineni:
### Summary
Run `yarn format` and commit the results if there are auto-fixes. 

This enables us to do simple changes from the browser, and not worry about linting. 

Since we already have a build preview in place, this should be a good addition to faster changes from the web.

The source for this workflow is derived from this blog post:
https://mskelton.medium.com/auto-formatting-code-using-prettier-and-github-actions-ed458f58b7df

#### What changed and Why?

For a PR as tiny as this one https://github.com/pinterest/gestalt/pull/2702. It's frustrating to pull the branch code, and do the changes on my machine, run it locally and blah, because of a tiny formatting issue. It's likely an extra space. 

Likewise, to make changes to a [file](https://github.com/pinterest/gestalt/blob/master/docs/pages/get_started/faq.js) like this (e.g. add a new section). It should be fairly easy to do by copy-pasting, and adding an answer from the web console, but formatting becomes a pain. 

Soo hoping this step makes these kind of changes faster


AlbertCarreras:
@rlingineni run prettier, maybe for the last time?

.github/workflows/main.yml
```diff
@@ -0,0 +1,20 @@
+name: Format Files
+on:
+  pull_request:
+    branches: [main]
```
dangerismycat, line 4: Our repo doesn't have a `main` — this will need to be `master`

.github/workflows/main.yml
```diff
@@ -0,0 +1,20 @@
...
+jobs:
+  format:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          ref: ${{ github.head_ref }}
+      - uses: actions/setup-node@v2
+        with:
+          node-version: '14.x'
```
dangerismycat, line 14: Node 14 is already a year past the end of active support and will soon cease security support: https://endoflife.date/nodejs

Let's use Node 18 here if possible, or at least 16

.github/workflows/main.yml
```diff
@@ -0,0 +1,20 @@
+name: Format Files
+on:
+  pull_request:
+    branches: [main]
+jobs:
+  format:
```
dangerismycat, line 6: Without the `permissions` field I'm not sure that this is going to be able to do anything

dangerismycat (changes requested):
I'm a bit hesitant to give write permissions to a third-party tool to solve something as minor as running prettier. Are we sure the need outweighs the potential security risk?

.github/workflows/format.yml
```diff
@@ -0,0 +1,31 @@
...
+      - uses: actions/setup-node@v2
+        with:
+          node-version: '18.x'
+      - name: Install Prettier
+        run: npm install -g prettier@2.8.0
+      - name: Run Formatter
+        run: yarn format
+      - name: Set Author Details
+        run: |
+            git config --global user.email "rlingineni@pinterest.com"
```
dangerismycat, line 25: You can use the same account we use for releases, [pinterest-gestalt](https://github.com/pinterest-gestalt)

I would set the email as `pinterest-gestalt@gmail.com` and name as "Pinterest-Gestalt"

rlingineni:
@dangerismycat this is looking green now

dangerismycat:
I'm still seeing this step as waiting?
Screen Shot 2023-03-14 at 4 39 16 PM

rlingineni:
> I'm still seeing this step as waiting? Screen Shot 2023-03-14 at 4 39 16 PM

Yeah, it went green and then un-green after another change. I'll take a look tmmrw

rlingineni:
Abandoning this PR because it's running a previous version of a workflow file that doesn't exist, and that's causing things to fail (I think). Will remake it

PR #2706, opened 2023-03-08
Internal: turn off `import/no-relative-parent-imports` lint rule

dangerismycat:
This rule is preventing us from keeping our main directory clean by moving child components into directories. Eventually we should add a webpack resolver to allow us to use absolute paths for all imports.


AlbertCarreras (approved):
fingers crosssed

PR #2724, opened 2023-03-13
Sticky: convert to autogen props and Sandpack examples

dangerismycat:


docs/pages/web/sticky.js
```diff
@@ -114,7 +31,21 @@ function Example() {
 }
 
 export async function getServerSideProps(): Promise<{| props: {| generatedDocGen: DocGen |} |}> {
+  const generatedDocGen = await docgen({ componentName: 'Sticky' });
+
+  ['bottom', 'left', 'right', 'top'].forEach((prop) => {
```
dangerismycat, line 36: The union typing with Threshold is pretty useful, so side-stepping that with these manual types

PR #2731, opened 2023-03-15
Github Workflow: Auto-Format and Commit

rlingineni:
See comments on https://github.com/pinterest/gestalt/pull/2705

### Summary
Run `yarn format` and commit the results if there are auto-fixes. 

This enables us to do simple changes from the browser, and not worry about linting. 

Since we already have a build preview in place, this should be a good addition to faster changes from the web.

The source for this workflow is derived from this blog post:
https://mskelton.medium.com/auto-formatting-code-using-prettier-and-github-actions-ed458f58b7df

#### What changed and Why?

For a PR as tiny as this one https://github.com/pinterest/gestalt/pull/2702. It's frustrating to pull the branch code, and do the changes on my machine, run it locally and blah, because of a tiny formatting issue. It's likely an extra space. 

Likewise, to make changes to a [file](https://github.com/pinterest/gestalt/blob/master/docs/pages/get_started/faq.js) like this (e.g. add a new section). It should be fairly easy to do by copy-pasting, and adding an answer from the web console, but formatting becomes a pain. 

Soo hoping this step makes these kind of changes faster

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,7 +16,9 @@ jobs:
     name: 'Format and Commit'
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
+        repository: ${{ github.event.pull_request.head.repo.full_name }}
+        ref: ${{ github.head_ref }
       - uses: actions/setup-node@v2
         with:
           node-version: '18.x'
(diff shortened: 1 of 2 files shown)


rlingineni:
@dangerismycat @AlbertCarreras , sadly killing this PR with auto-formatting. Learned more GH workflows now though.

Here's the issues:

1. If the origin branch of the PR is from a fork, then the action can't write to it. So in this PR, the origin is from rlingineni/gestalt. The workflow doesn't have access

2. I can add a conditional to run this job only on non-fork branches, but that was also making our CI flow look nasty. This might be useful, but the cases here are smaller.

A worthy attempt! One of these days, I'll have to try a setup from GH codespaces or some of the cloud IDEs to see how that improves my workflow.

PR #2734, opened 2023-03-16
OverlayPanel: internal fixes, cleanup

AlbertCarreras:
OverlayPanel: internal fixes

- Backdrop was not being animated
- Final microseconds flash of the OverlayPanel before disappearing: using useLayoutEffect rather useEffect for smooth transition

BEFORE: 
Brave Browser - Badge - Gestalt 2023-03-16 at 12 49 41 AM

AFTER:
Brave Browser - Badge - Gestalt 2023-03-16 at 12 46 30 AM


packages/gestalt/src/OverlayPanel.js
```diff
@@ -116,6 +117,18 @@ function OverlayPanel({
   );
 }
 
+const DismissingElement = ({
```
dangerismycat, line 120: Nothing about this seems specific to OverlayPanel, so I think this is ripe for more abstraction. Consider something like this:
```js
// ./animation/DismissElement.js

type Props = {| children: ({| onDismissStart: () => void |}) => Node |};

export default function DismissElement({ children }: Props): Node {
  const { onExternalDismiss } = useAnimation();
  return children({ onDismissStart: onExternalDismiss });
}
```
```js
// OverlayPanel.js
import DismissElement from './animation/DismissElement.js';

OverlayPanel.DismissElement = DismissElement;
```
```js
// MobileSheet.js
import DismissElement from './animation/DismissElement.js';

MobileSheet.DismissElement = DismissElement;
```
etc

docs/pages/web/overlaypanel.js
```diff
@@ -259,9 +260,24 @@ OverlayPanel comes in 3 sizes: small (\`sm\`), medium (\`md\`), and large (\`lg\
...
 
-      OverlayPanel also provides \`onAnimationEnd\`, a callback that gets triggered at the end of each animation. The callback has access to \`animationState\` to identify the end of each 'in' and 'out' animation for cases where the two events trigger different responses..
+      OverlayPanel also provides \`onAnimationEnd\`, a callback that gets triggered at the end of each animation. The callback has access to \`animationState\` to identify the end of each 'in' and 'out' animation for cases where the two events trigger different responses.
       `}
         >
+          <SlimBanner
+            iconAccessibilityLabel="Warning"
+            message={
+              <Text inline>
+                Don&apos;t use OverlayPanel&apos;s <code>onDismissStart</code> render prop available
```
dangerismycat, line 270: Can you add something about _why_ people shouldn't use those renderProps? Something about them being deprecated and removed soon

dangerismycat (approved):

PR #2737, opened 2023-03-17
Masonry: add lazy hydration for test route and remaining three integration tests

dangerismycat:
The remaining three integration tests all require being able to stop and run assertions on the SSR output before hydration. This was pretty simple previously, since we were running a custom mini-app that we had full control over. However, in Next.js Land, this is a bit more complicated. I explored and discarded a number of options:
- [A custom server](https://nextjs.org/docs/advanced-features/custom-server). This seemed way too heavy, total overkill. Not worth the time/effort.
- [A custom `Document`](https://nextjs.org/docs/advanced-features/custom-document). We do this currently, so not a lot of additional effort to implement…if I could figure out how, which I couldn't. This isn't a common feature, so not supported out of the box (or at least not mentioned in the documentation anywhere). I dug around in the source code for a while trying to figure out how to make this work, but kept looking elsewhere as it seemed overly complex (and like I'd probably have to reach into Next.js internals, which felt bad).
- [Using this experimental config](https://github.com/vercel/next.js/pull/11949). Unfortunately it only works on prod, and since I'm currently running the Masonry integration tests locally, that's a no-go.

I even found [a couple of years old tweet from Dan Abramov](https://twitter.com/dan_abramov/status/1456305021352652813) and asked him if he found a good solution, but didn't get any useful information from him.

However, I did find a tiny package from that Twitter thread that did the trick: [react-lazy-hydration](https://github.com/hadeeb/react-lazy-hydration). It only has ~350 stars on GH and hasn't been updated for a couple of years, so it's nothing I'd ever trust to use in prod. But it does what I need it to and it unblocked me, so that's our solution for now.

With that functionality in hand, this PR also adds those three remaining integration tests.


docs/integration-test-helpers/masonry/getFlexibleGridServerStyles.js
```diff
@@ -49,9 +49,11 @@ export default function getFlexibleGridServerStyles({
 `;
 
   for (let i = minColumns; i < maxColumns + 1; i += 1) {
-    const minWidth = i === minColumns ? 0 : (i - 1) * maxItemWidth;
+    // NOTE: THIS HAS CHANGED FROM PINBOARD PROD
```
dangerismycat, line 52: @liuyenwei This was necessary to get this test to pass. Since we only use this layout in prod on mweb with very limited screen size variation, I wonder if this is a change we want to make there as well — or if the test should be changed slightly.

liuyenwei: i suspect the test should change.  if this was an issue in production, we'd have issues with our CLS score

docs/pages/integration-test/masonry.js
```diff
@@ -18,10 +19,19 @@ function booleanize(value: string): boolean {
   return Boolean(value);
 }
 
+// LazyHydrate doesn't like to be used without any props, so we have to add it conditionally
+function MaybeLazyHydrate({ children, ssrOnly }: {| children: Node, ssrOnly: boolean |}) {
+  if (ssrOnly) {
+    return <LazyHydrate ssrOnly>{children}</LazyHydrate>;
```
liuyenwei, line 25: huh, TIL: https://github.com/facebook/react/issues/10923#issuecomment-338715787

dangerismycat: Interesting…how would you see us using that here? (that thread definitely would've been useful to me in a previous job where I was using D3 with React)

dangerismycat:
https://github.com/pinterest/gestalt/pull/2737#discussion_r1140470008

After chatting offline we think this _is_ something that needs to be updated in prod, so I'm going to make that change in Pinboard

docs/pages/integration-test/masonry.js
liuyenwei: just added that because its what LazyHydrate does under the hood (was curious about how it handles it and i found that)

liuyenwei (approved):

PR #2738, opened 2023-03-17
IconButtonFloating: Add `disabled` prop

EduhCosta:
#### What changed?

This PR is adding a property `disabled` to IconButtonFloating.

#### Why?

[This deprecation ticket](https://jira.pinadmin.com/browse/GESTALT-5478) is changing a Pinboard code based which has the follow behaviour:

1. While the user not select one of pins on organize screen, the IconButtonFloating, on footer of screen, should be disabled;
2. After one of pins is selected the IconButtonFloating should be enabled;

To get more context, please look at [this video](https://capture.dropbox.com/8RzrpABsVStH5XpS)

The video above was sent to @hectoid and @ponori and both of them understanding the `disabled` prop should be present on this component.

### Links

- [Jira](https://jira.pinadmin.com/browse/GESTALT-5478)

### Checklist

- [X] Checked stakeholder feedback (e.g. @hectoid e @ponori )

### DIFF
--- a/packages/gestalt/src/IconButtonFloating.js
+++ b/packages/gestalt/src/IconButtonFloating.js
@@ -22,6 +22,10 @@ type Props = {|
    */
   accessibilityLabel: string,
   /**
+   * When disabled, IconButtonFloating looks inactive and cannot be interacted with
+   */
+  disabled?: boolean,
+  /**
    * Icon displayed in IconButtonFloating to convey the behavior of the component. Refer to the [iconography](/foundations/iconography/library) guidelines regarding the available icon options.
    */
   icon: $Keys<typeof icons>,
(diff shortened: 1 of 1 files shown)


packages/gestalt/src/IconButtonFloating.js
```diff
@@ -21,6 +21,10 @@ type Props = {|
    * String that clients such as VoiceOver will read to describe the icon button. Always localize the label. See [Accessibility section](https://gestalt.pinterest.systems/web/iconbuttonfloating#Accessibility) for more info.
    */
   accessibilityLabel: string,
+  /**
+   * When disabled, IconButtonFloating looks inactive and cannot be interacted with
```
EduhCosta, line 25: This description on copied from `IconButton` disabled prop, as you can see [Here](https://gestalt.pinterest.systems/web/iconbutton#IconButtonProps).

PR #2773, opened 2023-03-29
Docs: final improvements to development view: cleanup/refactor

AlbertCarreras:
Docs: final improvements to development view: cleanup/refactor


dangerismycat (commented):
This seems fine, but any reason why it was merged without review?

PR #2774, opened 2023-03-29
Docs: Cleaned up our Figma plugins section

ponori:
Removed references to legacy plugins and added a promo image to highlight the Pinterest Design plugin.

### DIFF
--- a/docs/pages/get_started/designers.js
+++ b/docs/pages/get_started/designers.js
@@ -44,17 +44,13 @@ In your first couple of weeks at Pinterest, you will receive an invite from your
           description={`
 **Please note: The following plugins are only available for Pinterest employees.**
 
-[**Gestalt docs for Figma**](https://www.figma.com/community/plugin/977755389228415646/Gestalt-docs-for-Figma-(Beta))
-View the Gestalt documentation without leaving Figma!
-
-[**Pinsert**](https://www.figma.com/community/plugin/763812093925718603/Pinsert)
-Insert Pins, Boards, and Profiles from Pinterest.
+[![promo image for the pinterest design figma plugin](https://i.pinimg.com/originals/2b/4b/b4/2b4bb4f0c0e82ddc8f37917c82738b7d.png)](https://www.figma.com/community/plugin/1215463263194174399)
+[**The Pinterest Design plugin**](https://www.figma.com/community/plugin/1215463263194174399)
+The Pinterest Design Figma plugin is _the_   design tool to work faster and smarter. You can add Pinterest images directly to designs, export icons in production-ready formats, automatically convert UI to dark mode, reference design system documentation and more!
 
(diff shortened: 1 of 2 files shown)


docs/pages/get_started/designers.js
```diff
@@ -44,17 +44,13 @@ In your first couple of weeks at Pinterest, you will receive an invite from your
           description={`
 **Please note: The following plugins are only available for Pinterest employees.**
 
-[**Gestalt docs for Figma**](https://www.figma.com/community/plugin/977755389228415646/Gestalt-docs-for-Figma-(Beta))
-View the Gestalt documentation without leaving Figma!
-
-[**Pinsert**](https://www.figma.com/community/plugin/763812093925718603/Pinsert)
-Insert Pins, Boards, and Profiles from Pinterest.
+[![promo image for the pinterest design figma plugin](https://i.pinimg.com/originals/2b/4b/b4/2b4bb4f0c0e82ddc8f37917c82738b7d.png)](https://www.figma.com/community/plugin/1215463263194174399)
```
AlbertCarreras, line 47: can we use here **PrivateLink** instead?

Screenshot by Dropbox Capture

AlbertCarreras, line 47: Also can we place the image below the text?

docs/pages/get_started/designers.js
```diff
@@ -44,17 +44,13 @@ In your first couple of weeks at Pinterest, you will receive an invite from your
           description={`
 **Please note: The following plugins are only available for Pinterest employees.**
```
AlbertCarreras, line 45: We might not need this is we move the links to show the private icon as per comment below

AlbertCarreras (approved):

docs/pages/get_started/designers.js
ponori: Doesn't look like that's possible to use w/in `MainSection.Subsection`'s description prop. :(

ponori: Screen Shot 2023-03-29 at 1 20 00 PM
Hmm... From my eye, that muddies the water as to which plugin this image refers to. 

The intention by this lockup was to follow the same structure we use for component images:
Screen Shot 2023-03-29 at 1 21 24 PM

docs/pages/get_started/designers.js
ponori: Yeah, man, I'd _love_ to be able to use `PrivateLink` and thus get rid of this line. Maybe at some point we can update our `MainSection.Subsection` component to use MDX?

docs/pages/get_started/designers.js
AlbertCarreras: MMMm

rlingineni: I like below the text too! Because in the component image, we treat it as a visual array, and it's the main focus.

In this case, we're reading top to bottom, so I think images inline with the text, than starting the paragraph with the image works better

PR #2778, opened 2023-03-30
Internal: standardize format for a11y integration test naming

dangerismycat:
I noticed recently that our accessibility integration test file naming is chaos…different casing, whitespace, and names that have no bearing on the route they hit. This makes it difficult to see which routes have coverage, as demonstrated by the seven duplicate test files I deleted, as well as a test for a route we no longer support (without a redirect, this is basically testing a blank page). 

The new structure follows the route path, converting `/` to `__` but otherwise staying true to the URL (snake_case, not camelCase or PascalCase). Though this adds a bit more pain whenever we adjust paths (something we shouldn't be doing frequently anyway), this makes it very easy to find the relevant test file for a path, and to know what name a new test file should use.

This PR also reworks (simplifies) our validation script to take advantage of the consistent file format.

I tried something a bit fancier where the tests would use their own file name to determine the path in `page.goto`, but I don't think that's possible in Playwright tests.

It would also be really cool if we could obviate the test files altogether with a script that basically creates test files on the fly from everything in `docs/pages`, but that's an improvement for another day.

I also discovered a couple of lil other fixes, which I'll call out in comments.


docs/docs-components/siteIndex.js
```diff
@@ -185,7 +185,7 @@ const siteIndex: $ReadOnlyArray<siteIndexType> = [
           'Grammar',
           'Inclusive language',
           'UI elements',
-          'Syntax structure',
+          'Syntax and structure',
```
dangerismycat, line 188: This appeared to be a typo

docs/markdown/foundations/content_standards/syntax_and_structure.md
```diff
@@ -1,27 +1,27 @@
 ---
-title: Syntax and Structure
+title: Syntax and structure
```
dangerismycat, line 2: This was the only change I made in this file. The other changes were prettier

docs/redirects.js
```diff
@@ -138,6 +138,11 @@ const misc = [
     destination: '/get_started/developers/contributing/development_process',
     permanent: true,
   },
+  {
```
dangerismycat, line 141: This is the only new redirect. The others are just alphabetizing

playwright/accessibility/foundations__messaging__overview.spec.mjs
```diff
@@ -2,7 +2,7 @@
 import { test } from '@playwright/test';
 import expectAccessiblePage from './expectAccessiblePage.mjs';
 
-test('Data viz palette check', async ({ page }) => {
-  await page.goto('/foundations/data_visualization/palette');
+test('Messaging overview accessibility check', async ({ page }) => {
```
dangerismycat, line 5: git's being weird here. This wasn't converting a test to a different path — probably from a duplicate test I deleted

scripts/validateAccessibilitySpecs.js
```diff
@@ -31,50 +31,35 @@ const getAllSitePaths = (index) => {
...
+  const a11ySpecFiles = (await globby([specFileGlob])).map((file) =>
+    file
+      .toLowerCase()
+      // remove the playwright/accessibility/ prefix
+      .replace(/^playwright\/accessibility\//, '')
+      // remove the .spec.mjs suffix
+      .replace(/\.spec\.mjs$/, ''),
   );
 
+  const pagesWithoutA11ySpecFiles = routes.filter((page) => !a11ySpecFiles.includes(page));
```
dangerismycat, line 53: The consistent format makes this check much simpler — convert the route path from the index to match the file name format and do a simple comparison

playwright/accessibility/foundations__forms__structure_and_behavior.spec.mjs
```diff
@@ -3,6 +3,6 @@ import { test } from '@playwright/test';
 import expectAccessiblePage from './expectAccessiblePage.mjs';
 
 test('Structure and behavior Accessibility check', async ({ page }) => {
-  await page.goto('/foundations/forms/composition_+_behavior');
```
dangerismycat, line 6: this was just wrong 😬

AlbertCarreras (dismissed):

dangerismycat:
@rlingineni The thing to keep in mind is that these tests are not testing components, they're testing routes. We actually need to add accessibility tests for our components themselves as the current tests can't "see" any of the Sandpack examples, as they're rendered in iframes. So basing the test name on the route rather than the component makes the most sense to me.

PR #2787, opened 2023-03-30
Docs: refactored examples for Modal, ModalAlert, OverlayPanel examples for better display

AlbertCarreras:
Docs: refactored examples for Modal, ModalAlert, OverlayPanel examples for better display


docs/examples/modalalert/doClearCommunicate.js
```diff
@@ -1,33 +1,48 @@
...
-import { type Node } from 'react';
-import { ModalAlert, CompositeZIndex, FixedZIndex, Layer, Text } from 'gestalt';
-
-const HEADER_ZINDEX = new FixedZIndex(10);
-const zIndex = new CompositeZIndex([HEADER_ZINDEX]);
+import { Fragment, useState, type Node } from 'react';
+import { Box, Button, ModalAlert, CompositeZIndex, FixedZIndex, Layer, Text } from 'gestalt';
 
 export default function DoClearCommunicate(): Node {
+  const [showComponent, setShowComponent] = useState(false);
```
dangerismycat, line 6: These should be initialized as `true` so the modals are shown as the user scrolls through the page, yes?

docs/examples/modalalert/doClearCommunicate.js
```diff
@@ -1,33 +1,48 @@
...
+
+  const HEADER_ZINDEX = new FixedZIndex(10);
```
dangerismycat, line 8: Not sure why these were moved inside the components in the examples — consts like these that don't depend on component props should be defined outside the function component so they're not re-created on every render

dangerismycat (commented):
Thanks for pulling these changes out of the other PR!

PR #2788, opened 2023-03-31
ModalAlert: make sure docs examples are open without user interaction

dangerismycat:
#2787 introduced changes that caused ModalAlert examples to require user interaction to be visible, degrading the UX. This PR makes sure that the examples are visible as the user scrolls through the page.


rlingineni:
Ooh nice!

rlingineni (approved):

PR #2791, opened 2023-03-31
IconButtonFloating: add required `tooltip` prop

dangerismycat:
#### What changed?

This PR adds a required `tooltip` prop to IconButtonFloating. The prop value is identical to IconButton's.

#### Why?

For the same reason we added `tooltip` to IconButton: icon meaning is not always obvious to the user, and providing a tooltip gives the user more context so they know what to expect when interacting with the button.

IconButtonFloating usage is still very low (2 instances in Pinboard currently), so jumping straight to a required prop isn't a big lift.

### Links

- [Jira](https://jira.pinadmin.com/browse/GESTALT-6028)

### DIFF
--- a/docs/examples/iconbuttonfloating/doForScroll.js
+++ b/docs/examples/iconbuttonfloating/doForScroll.js
@@ -64,6 +64,9 @@ export default function Example(): Node {
           icon="question-mark"
           onClick={() => setOpen((prevVal) => !prevVal)}
           selected={open}
+          tooltip={{
+            text: 'Help & Resources Menu',
+          }}
         />
       </Box>
       {open && (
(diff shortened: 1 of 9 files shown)


docs/examples/iconbuttonfloating/doForScroll.js
```diff
@@ -64,6 +64,9 @@ export default function Example(): Node {
           icon="question-mark"
           onClick={() => setOpen((prevVal) => !prevVal)}
           selected={open}
+          tooltip={{
+            text: 'Help & Resources Menu',
```
AlbertCarreras, line 68: nit 

SHold we only upper case first letter? in all follwing examples too

AlbertCarreras (approved):

PR #2792, opened 2023-03-31
TextField: add 'inputMode' prop

AlbertCarreras:
#### What changed?

TextField: add 'inputMode' prop

FOLLOW UP. DatePicker: add 'inputMode' = none to internal TextField to only show calendar, and don't display virtual keyboard

image

#### Why?

Improve mobile support


packages/gestalt/src/TextField.js
```diff
@@ -48,6 +48,10 @@ type Props = {|
    * A unique identifier for the input.
    */
   id: string,
+  /**
+   * Optionally specify the type of data that might be entered by the user while editing the element or its contents. This allows a browser to display an appropriate virtual keyboard. See the [enterKeyHint variant](https://gestalt.pinterest.systems/web/textfield#EnterKeyHint) for more info.
```
dangerismycat, line 52: Probably more like "This allows _mobile browsers_ to display an appropriate…", yes? Virtual keyboards aren't ever a thing on desktop?

packages/gestalt/src/TextField.js
```diff
@@ -48,6 +48,10 @@ type Props = {|
...
+   */
+  inputMode?: 'none' | 'text' | 'decimal' | 'numeric',
```
dangerismycat, line 54: Can we also add `tel`, `email`, and `url`? Those would be some great UX wins

docs/pages/web/textfield.js
```diff
@@ -689,22 +693,93 @@ function Example(props) {
...
 - "send": delivering the text to its target
+
+The \`inputMode\` prop presents to the user a more accurate action label for the enter key on virtual keyboards. These are the values for each use case:
+
+- "none": No virtual keyboard. For when the page implements its own keyboard input control, for example DatePicker displays the calendar picker.
+- "text": Standard input keyboard for the user's current locale.
+- "decimal": Fractional numeric input keyboard containing the digits and decimal separator for the user's locale (typically "." or ",").
+- "numeric": Numeric input keyboard, but only requires the digits 0–9.
+
+Use \`type\` when TextField needs to capture phone numbers, emails or URLs.
```
dangerismycat, line 704: Would including those as `inputMode`s also be helpful? Or should we add those automatically for their respective `type`s?

packages/gestalt-datepicker/src/DatePickerTextField.js
```diff
@@ -67,6 +67,7 @@ function DatePickerTextField(props: Props) {
             }}
             errorMessage={errorMessage}
             helperText={helperText}
+            inputMode="none"
```
dangerismycat, line 70: This is a breaking change that we should pull out to a separate PR

dangerismycat (changes requested):
Only real concern is pulling out the DatePicker change. The other comments are niceties that I think we should add, but not blockers

docs/pages/web/textfield.js
AlbertCarreras: Screenshot by Dropbox Capture

packages/gestalt-datepicker/src/DatePickerTextField.js
AlbertCarreras: Breaking change or more like a bug fix?

packages/gestalt/src/TextField.js
AlbertCarreras: DONE

packages/gestalt-datepicker/src/DatePickerTextField.js
AlbertCarreras: DONE

docs/pages/web/textfield.js
dangerismycat: Got it, good call 👍

packages/gestalt-datepicker/src/DatePickerTextField.js
dangerismycat: a bug fix can be a breaking change 🙂

dangerismycat (approved):

PR #2794, opened 2023-03-31
DatePicker: added Year/Month dropdown options, removed virtual keyboard, upgraded react-datepicker to 4.0.0

AlbertCarreras:
#### BREAKING CHANGE

No codemod needed. We are removing virtual keyboard support for DatePicker so they don't conflict with each other. Also, upgrading react-datepicker to 4.0.0

#### What changed?

DatePicker: added Year/Month dropdown options + DatePicker:  remove virtual keyboard on mobile

https://user-images.githubusercontent.com/10593890/229939820-76c3a6c0-d38b-477d-96a8-0dfd9f75f889.MP4

Brave Browser - DatePicker - Gestalt 2023-04-04 at 3 08 21 PM

Also, 

Upgraded library to 4.0.0. to get latest Popper v2. utils.

#### Why?

DatePicker: add 'inputMode' = none to internal TextField to only show calendar, and don't display virtual keyboard

image

https://user-images.githubusercontent.com/10593890/229680720-c440d4cd-2546-4170-8861-b63deb1755da.MP4

### Links

- [TDD](link to Paper doc)
- [Figma](link to Figma file)

### Checklist

### DIFF
--- a/packages/gestalt-datepicker/package.json
+++ b/packages/gestalt-datepicker/package.json
@@ -14,7 +14,7 @@
   ],
   "dependencies": {
     "classnames": "^2.2.6",
-    "react-datepicker": "4.0.0"
+    "react-datepicker": "4.11.0"
   },
   "peerDependencies": {
     "gestalt": ">0.0.0",
--- a/packages/gestalt-datepicker/package.json
+++ b/packages/gestalt-datepicker/package.json
@@ -14,7 +14,7 @@
   ],
   "dependencies": {
     "classnames": "^2.2.6",
-    "react-datepicker": "4.11.0"
+    "react-datepicker": "4.0.0"
   },
   "peerDependencies": {
     "gestalt": ">0.0.0",
--- a/packages/gestalt-datepicker/package.json
+++ b/packages/gestalt-datepicker/package.json
@@ -14,7 +14,7 @@
   ],
   "dependencies": {
     "classnames": "^2.2.6",
-    "react-datepicker": "2.16.0"
+    "react-datepicker": "4.0.0"
   },
   "peerDependencies": {
     "gestalt": ">0.0.0",
--- a/packages/gestalt-datepicker/src/DatePicker.css
+++ b/packages/gestalt-datepicker/src/DatePicker.css
@@ -377,6 +377,43 @@ html:not([dir="rtl"]) :global ._gestalt .react-datepicker__navigation--next {
   margin-right: 8px;
 }
 
+/* Year / Month dropdown  */
+
+:global ._gestalt .react-datepicker__month-select,
+:global ._gestalt .react-datepicker__year-select {
+  background-color: var(--g-colorGray0);
+  border-color: var(--color-border-container);
+  border-radius: 16px;
+  border-style: solid;
+  border-width: 2px;
+  color: var(--color-text-subtle);
+  cursor: pointer;
+  font-family: var(--font-family-default-latin);
+  font-size: var(--font-size-200);
+  height: 40px;
+  margin-bottom: 10px;
+  margin-top: 10px;
+  padding: 5px 16px; /* right padding prevents it from running into arrow */
+  position: relative;
+  width: 100%;
+}
+
+:global ._gestalt .react-datepicker__month-select:focus,
+:global ._gestalt .react-datepicker__year-select:focus {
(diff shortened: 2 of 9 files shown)


packages/gestalt-datepicker/package.json
```diff
@@ -14,7 +14,7 @@
   ],
   "dependencies": {
     "classnames": "^2.2.6",
-    "react-datepicker": "2.16.0"
+    "react-datepicker": "4.0.0"
```
dangerismycat, line 17: It looks like the current version is `4.11.0`. Can we upgrade all the way to that?

packages/gestalt-datepicker/src/DatePicker.css
```diff
@@ -359,22 +359,41 @@ html:not([dir="rtl"]) :global ._gestalt .react-datepicker__navigation--next {
...
+  border-style: solid;
+  border-width: 2px;
+  color: var(--color-text-subtle);
+  cursor: pointer;
+  font-family: var(--font-family-default-latin);
+  font-size: var(--font-size-200);
+  height: 40px;
+  margin-bottom: 10px;
+  margin-top: 10px;
+  padding: 5px 16px; /* right padding prevents it from running into arrow */
```
dangerismycat, line 378: This is applying padding to both left and right sides…is that intentional? If so, you might want to update the comment to something like "horizontal padding prevents…"

Also, does this work fine with RTL?

packages/gestalt-datepicker/src/DatePicker.js
```diff
@@ -131,6 +131,10 @@ type Props = {|
    * Required for date range selection. Pass a ref object to DatePicker to autofocus on the unselected date range field. See the [date range picker example](https://gestalt.pinterest.systems/web/datepicker#rangePicker) to learn more.
    */
   ref?: Element<'input'>, // eslint-disable-line react/no-unused-prop-types
+  /**
+   * Show a select list for quick selection of year and month. See the [showMonthYearDropdown example](https://gestalt.pinterest.systems/web/datepicker#showMonthYearDropdown) to learn more.
+   */
+  showMonthYearDropdown?: boolean,
```
dangerismycat, line 137: Do we have a use case for only showing the month dropdown, or only showing the year? I wonder if this would be better as something like 
```js
dropdowns?: $ReadOnlyArray<'month' | 'year'>,

...

<ReactDatePicker
  dropdownMode={dropdowns.length > 0 ? 'select' : undefined}
  showMonthDropdown={dropdowns.includes('month')}
  showYearDropdown={dropdowns.includes('year')}
```

packages/gestalt-datepicker/src/DatePicker.jsdom.test.js
```diff
@@ -5,10 +5,17 @@ import DatePicker from './DatePicker.js';
 
 const initialDate = new Date(2018, 11, 14);
 
-function DatePickerWrap() {
+function DatePickerWrap(showMonthYearDropdown) {
```
dangerismycat, line 8: This shouldn't be working — you're passing this as a prop below, so you need to destructure it. As written, this is the equivalent of
```js
function DatePickerWrap(props) {
  return (
    <DatePicker
      showMonthYearDropdown={!!props}
    />
  );
}
```

dangerismycat (approved):

packages/gestalt-datepicker/package.json
AlbertCarreras: They jump from 4 to 11 in like 2 releases. I don't think the left 3 matter based on the comments. I can do that, we won't be any benefit though

dangerismycat: It looks like there are a lot of bugfixes and other improvements in there: https://github.com/Hacker0x01/react-datepicker/releases

I'm always a bit nervous about `x.0.0` releases, as they're typically followed by a slew of fixes. Unless there's a specific reason not to, we should always try to update to the most current version to get those fixes.

AlbertCarreras: DONE

AlbertCarreras: 4.11.0 Introduces a bug
Screenshot by Dropbox Capture

dangerismycat: No worries. We should upgrade in the near-ish future, but not worth worrying about for now

PR #2796, opened 2023-04-03
Table: implemented header/footer borders

AlbertCarreras:
#### What changed?

Table: implemented header/footer borders

Scrolling borders
Screenshot by Dropbox Capture

Fixed borders
Screenshot by Dropbox Capture

### Links

- [Jira](https://jira.pinadmin.com/browse/GESTALT-5934)
- [Figma](https://www.figma.com/file/HbBD5AgCNvupMVKBkTgdsH/Table-Enhancements?node-id=611-49291&t=rMa6Dw7RQUHf6B2q-0)

### DIFF
--- a/packages/gestalt/src/Table.css
+++ b/packages/gestalt/src/Table.css
@@ -7,40 +7,54 @@
 
 .th {
   composes: alignStart from "./Typography.css";
-  composes: paddingY3 from "./boxWhitespace.css";
-  composes: paddingX3 from "./boxWhitespace.css";
 }
 
+.th,
 .td {
   composes: paddingY3 from "./boxWhitespace.css";
   composes: paddingX3 from "./boxWhitespace.css";
 }
 
-.stickyHeader tr th {
-  background-color: var(--g-colorGray0);
-  position: sticky;
-  top: 0;
-  z-index: 2;
+.stickyHeader {
+  composes: raisedTop from "./Borders.css";
 }
 
+.stickyFooter {
+  composes: raisedBottom from "./Borders.css";
+}
+
+.stickyHeader,
+.stickyHeader tr th {
+  top: 0;
+}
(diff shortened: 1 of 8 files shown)


packages/gestalt/src/Table.css
```diff
@@ -7,40 +7,54 @@
...
 }
 
+.th,
 .td {
   composes: paddingY3 from "./boxWhitespace.css";
   composes: paddingX3 from "./boxWhitespace.css";
 }
 
+.stickyHeader {
+  box-shadow: var(--elevation-raised-top);
```
dangerismycat, line 19: We're already declaring these styles as `.raisedTop` and `.raisedBottom` in Borders.css. Let's compose these styles from those classes to keep bundle sizes down.

packages/gestalt/src/Table.css
```diff
@@ -7,40 +7,54 @@
...
+.stickyFooter tr td {
+  z-index: 2;
+}
+
+.columnSticky,
+.stickyHeader,
+.stickyFooter,
+.stickyHeader tr th,
+.stickyFooter tr td {
+  background-color: var(--g-colorGray0);
```
dangerismycat, line 53: Similar note here and throughout: where we're already declaring these styles in generic classes, let's compose those to create the styles needed here

dangerismycat (approved):

packages/gestalt/src/Table.css
AlbertCarreras: DONE

packages/gestalt/src/Table.css
AlbertCarreras: DONE

AlbertCarreras: composition only allowed with single

PR #2798, opened 2023-04-03
DatePicker: remove hard coded date formats and replaced to dynamically feed from date-fns

AlbertCarreras:
#### BREAKING CHANGE

We remove internal support of placeholder locales. They now depend on getting the format from date-fns.

#### What changed?

DatePicker: remove hard coded date formats and replaced to dynamically feed from date-fns

#### Why?

Manually hard coded formats was a bad idea. We added new locales and missed adding the formats. They are provided by date-fns , we should leverage that

https://deploy-preview-2798--gestalt.netlify.app/web/datepicker#localeData


dangerismycat (approved):

rlingineni:
Wait, how would we know the implications of this? Like do we know if all datepickers are providing a placeholder, and just using our default

AlbertCarreras:
> Wait, how would we know the implications of this? Like do we know if all datepickers are providing a placeholder, and just using our default

We inform our library works with 
Screenshot by Dropbox Capture


But I'm happy to make it a breaking change, as potentially could break the internals if someone is not using it.

PR #2799, opened 2023-04-04
DatePicker:  remove virtual keyboard on mobile

AlbertCarreras:
#### BREAKING CHANGE

No codemod needed. We are removing virtual keyboard support for DatePicker so they don't conflict with each other.

#### What changed?

DatePicker:  remove virtual keyboard on mobile

#### Why?

DatePicker: add 'inputMode' = none to internal TextField to only show calendar, and don't display virtual keyboard

image

https://user-images.githubusercontent.com/10593890/229680720-c440d4cd-2546-4170-8861-b63deb1755da.MP4

### DIFF
--- a/packages/gestalt-datepicker/src/DatePickerTextField.js
+++ b/packages/gestalt-datepicker/src/DatePickerTextField.js
@@ -67,6 +67,7 @@ function DatePickerTextField(props: Props) {
             }}
             errorMessage={errorMessage}
             helperText={helperText}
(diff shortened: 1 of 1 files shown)


AlbertCarreras:
I think I'm going to wait until we add month/year, so the change is not super restrictive in mobile.

Also, maybe on mobile we enable month/year dropdowns by default so the experience is not that bad,

PR #2802, opened 2023-04-04
Docs: Removing Paragraph BG Color for MD documents

rlingineni:
Thanks for creating a PR! Please follow this template and delete items/sections that are not relevant to your changes, including these instructions.

### Summary

Dark mode shows a paragraph highlight.

image

image

#### What changed?

Changing paragraphs to a transparent background. Didn't find instances of where a fixed bg color was needed. Turns out for most MD pages w/ paragraphs, this seems like an issue.

### DIFF
--- a/docs/docs.css
+++ b/docs/docs.css
@@ -83,7 +83,7 @@ pre {
 }
 
 .Markdown p {
-  background-color: var(--g-colorGray0);
+  background-color: transparent;
   margin: 0;
 }
 


dangerismycat (approved):

docs/docs.css
```diff
@@ -83,7 +83,7 @@ pre {
 }
 
 .Markdown p {
-  background-color: var(--g-colorGray0);
+  background-color: transparent;
```
AlbertCarreras, line 86: `--color-transparent`. xp

rlingineni: Where are these defined? I was trying to find them

dangerismycat: global CSS that comes from ColorSchemeProvider:
Screen Shot 2023-04-04 at 3 30 46 PM

if you grep for `gestaltTheme` you'll find where that class name is being set

rlingineni: I was looking at the transparent, I think regular transparent is probably okay for us here

PR #2807, opened 2023-04-05
Docs: Add white bg to darkmode images, remove padding

rlingineni:
Change image container so that the darkmode bg is white

Remove padding by default on all images. Can be added again with `addPadding` flag

image

image

### DIFF
--- a/docs/docs-components/MarkdownPage.js
+++ b/docs/docs-components/MarkdownPage.js
@@ -236,14 +236,14 @@ const components = {
             alignItems="center"
             justifyContent="center"
           >
-              <Image
-                src={src}
-                alt={alt}
-                width={layout === 'fill' ? undefined : width || '100%'}
-                height={layout === 'fill' ? undefined : height || '100%'}
-                layout={layout}
-                objectFit="contain"
-              />
+            <Image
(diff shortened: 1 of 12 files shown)


docs/docs-components/MarkdownPage.js
```diff
@@ -198,20 +198,26 @@ const components = {
...
-    noPadding,
+    addPadding,
   }: {|
     src: string,
     caption?: string,
     alt?: string,
     width?: number,
     height?: number,
-    noPadding?: boolean,
+    addPadding?: boolean,
```
dangerismycat, line 208: While I do appreciate you flipping this prop (good practice when dealing with boolean props), let's try a slightly different approach:
```js
padding?: 'standard' | 'none',
```
where "none" is the default value? That gives us flexibility to add other padding options in the future while avoiding a boolean prop for something that isn't inherently boolean ("padding" could have any number of values)

dangerismycat (approved):

docs/docs-components/MarkdownPage.js
rlingineni: Makes sense! Fixed

PR #2812, opened 2023-04-07
Internal: add TypeScript declaration files

AlbertCarreras:
### Breaking change

If your app was previously using https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/gestalt or https://www.npmjs.com/package/@types/gestalt, Gestalt now supports TypeScrip declaration files.

They are maintained manually. The types are based on the ones from https://www.npmjs.com/package/@types/gestalt but they were reviewed manually. Many errors were found and they all were corrected.

Please, report if you find any incorrect TypeScript type for any of the components 

#### What changed?

Internal:  add TypeScript declaration files

#### Why?

Support Gestalt in apps using TypeScript rather than Flow. 
Internalize ownership of https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/gestalt/index.d.ts for better maintenance

Test 
`yarn tsc` with no issues
Screenshot by Dropbox Capture

alpha release in bextools
Screenshot by Dropbox Capture
Screenshot by Dropbox Capture
Screenshot by Dropbox Capture

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -9,5 +9,5 @@
   "grunt.autoDetect": "off",
   "javascript.validate.enable": false,
   "npm.packageManager": "yarn",
-  "typescript.validate.enable": false
+  "typescript.validate.enable": true
 }
--- a/.gitignore
+++ b/.gitignore
@@ -16,9 +16,10 @@ packages/gestalt-datepicker/dist/*
 packages/eslint-plugin-gestalt/dist/*
 packages/gestalt-design-tokens/dist
 
-# the flow export file as it's maintained by hand
+# the following files as they are maintained by hand
 !packages/gestalt/dist/gestalt.js.flow
 !packages/gestalt/dist/gestalt-datepicker.js.flow
+!packages/gestalt/dist/index.d.ts
 
 playwright/test-results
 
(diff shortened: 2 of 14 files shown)


rlingineni:
try /alpha release

rlingineni:
test /alpha release

rlingineni:
@AlbertCarreras - when you get the chance, can you reset this back to your original commit

.vscode/settings.json
```diff
@@ -9,5 +9,5 @@
   "grunt.autoDetect": "off",
   "javascript.validate.enable": false,
```
jackhsu978, line 10: should `javascript.validate.enable` also be turned to true?

jackhsu978 (approved):
LGTM. Added a question.

docs/pages/web/iconbutton.js
```diff
@@ -63,7 +63,7 @@ export default function DocsPage({ generatedDocGen }: {| generatedDocGen: DocGen
           },
           {
             name: 'iconColor',
-            type: `"darkGray" | "gray" | "red" | "white"`,
+            type: `"darkGray" | "gray" | "red" | "white" | "brandPrimary"`,
```
dangerismycat, line 66: This fix should really be a separate PR in case this one needs to be reverted for any reason

Feel free to merge this one once it's ✅ 
https://github.com/pinterest/gestalt/pull/2899

.vscode/settings.json
AlbertCarreras: No, as all .js files have Flow and TS doesn't like that.

PR #2818, opened 2023-04-10
SheetMobile: fixes and docs additions

AlbertCarreras:
### Breaking change

We removed accessibilityLabel as the default is the recommended one as it informs about "Bottom sheet" which is a recommended practice and shouldn't be overiden.

 Run to remove prop

```
--component=SheetMobile
--previousProp=accessibilityLabel
```

#### What changed?

SheetMobile:  fixes and docs additions

Added transition example
Added SVG for SheetMobile

Pending, adding grabbing funcitonality.


packages/gestalt/src/SheetMobile.js
```diff
@@ -150,7 +145,6 @@ function SheetMobile({
   if (size === 'full')
     return (
       <FullPage
-        accessibilityLabel={accessibilityLabel}
```
dangerismycat, line 153: Should this prop be removed from the FullPage and PartialPage components as well?

dangerismycat (approved):

packages/gestalt/src/SheetMobile.js
AlbertCarreras: @dangerismycat Left it there for now, as I might try to reuse this for Modal, and modal do has an accessibility label.

PR #2823, opened 2023-04-12
Internal: add `realisticPinHeights` option for Masonry integration test route

dangerismycat:
The Masonry integration tests currently use a grid that makes no attempt to render realistic Pin items, as that's largely irrelevant for the features being tested. However, those unrealistic item heights make that grid useless for testing future 2-column module work. This PR introduces an option to use "realistic" pin heights taken from a one-day sample (shoutout to @juliacochran!) that includes pin heights by both the number of impressions and the number of pins that day. (I'm not sure which of these we should use to generate the most "realistic" set, so both are included.) We use one of those fields as a weight, then create a number line and randomly select from it. (TBH I only barely understand the stats work here — shoutout to my wife @megtron9 for explaining what I needed to do in general, and helping me evaluate StackOverflow answers to find a good one 😅)

Handling SSR with this approach was particularly tricky (shoutout to @liuyenwei for tons of help!) given that we're no longer generating items on the fly. As noted in `generateRealisticExampleItems`:
> This approach generates a finite set of pins with realistic heights — a fundamental shift from generating random pins on the fly. We have to take care to ensure that the same pins are generated on the server and the client. This is done by generating an array of random number seeds in `getServerSideProps` in docs/pages/integration-test/masonry.js, which are used to generate the pin heights. We then need to make sure we're pulling from these pin heights in a consistent way on the server and the client. This is where `randomNumberSeed` comes in: we pull the first `numberOfItems` heights from the array for SSR. For subsequent calls on the client, we use a random `randomNumberSeed` to pull random heights from the array. So SSR will always use the first `numberOfItems` heights, and we don't attempt to keep track of "where we are" in the heights array after that.

Here's what this change looks like in action:
_before_
Screen Shot 2023-04-12 at 4 03 39 PM

_after_
Screen Shot 2023-04-12 at 4 04 07 PM

The _after_ doesn't look all that similar to the grids I see on my account, but I think there's a lot of variation in pin heights depending on the topics a user is interested in. I have no idea how various heights are grouped/distributed — it may be that users see less variation in heights within their grids, but some grids are full of taller pins and some shorter pins. So the underlying data here (or how we're using that data) may need to be tweaked, but this approach of using the data to generate "realistic" pin heights seems sound.


docs/pages/integration-test/masonry.js
```diff
@@ -79,3 +121,17 @@ export default function TestPage(): Node {
     </ColorSchemeProvider>
   );
 }
+
+export async function getServerSideProps(): Promise<{|
+  props: {| randomNumberSeeds: $ReadOnlyArray<number> |},
+|}> {
+  // This is used to ensure we're using the same dataset of realistic pins on the server and client
+  const randomNumberSeeds = Array.from({ length: REALISTIC_PINS_DATASET_SIZE }).map(() =>
```
rlingineni, line 129: Is the seed to shuffle items on the client side?

rlingineni (approved):
This change makes sense to me, since I'd think Masonry would have non-uniform heights.

I'm not sure if I follow the seeding logic, and why it's needed. Why can't pin heights be random within a size

docs/pages/integration-test/masonry.js
```diff
@@ -64,12 +101,17 @@ export default function TestPage(): Node {
...
+            realisticPinHeights
+              ? generateRealisticExampleItems({ name: 'InitialPin', pinHeightsSample })
+              : generateExampleItems({ name: 'InitialPin' })
+          }
           manualFetch={booleanize(manualFetch)}
           MasonryComponent={Masonry}
           measurementStore={measurementStore}
           noScroll={booleanize(noScroll)}
           offsetTop={offsetTop}
+          pinHeightsSample={realisticPinHeights ? pinHeightsSample : undefined}
```
rlingineni, line 114: Do the heights have to be passed separately?

docs/pages/integration-test/masonry.js
```diff
@@ -64,12 +101,17 @@ export default function TestPage(): Node {
           externalCache={booleanize(externalCache)}
           finiteLength={booleanize(finiteLength)}
           flexible={booleanize(flexible)}
-          initialItems={generateExampleItems({ name: 'InitialPin' })}
+          initialItems={
+            realisticPinHeights
+              ? generateRealisticExampleItems({ name: 'InitialPin', pinHeightsSample })
+              : generateExampleItems({ name: 'InitialPin' })
```
rlingineni, line 107: How come the older example is still supported? Would we not always use realistic pin heights

docs/integration-test-helpers/masonry/MasonryContainer.js
```diff
@@ -222,13 +221,22 @@ export default class MasonryContainer extends Component<Props, State> {
...
+            previousItemCount: from,
+            randomNumberSeed: Math.random(),
+            pinHeightsSample,
+          })
+        : generateExampleItems({
+            name,
+            numberOfItems: until - from,
+            previousItemCount: from,
+            baseHeight,
+            randomNumberSeed: this.randomNumberSeed,
```
rlingineni, line 238: ditto on having real and non-real, and how come we use Math.random in one place, and the seed in another?

rlingineni (approved):
added a few more q's

docs/pages/integration-test/masonry.js
dangerismycat: "passed separately", as opposed to…?

docs/pages/integration-test/masonry.js
dangerismycat: No, the opposite — by generating the array of random seeds as part of getServerSideProps, we ensure that we'll generate the same dataset of pin heights on both the server and the client.

docs/pages/integration-test/masonry.js
rlingineni: Like, as part of the pins?

dangerismycat: We could potentially, sure. When we call [MasonryContainer.loadItems](https://github.com/pinterest/gestalt/pull/2823/files#diff-783f4d96975c1127b6aba8b861cf98da9aa9a1ff4b176e003993ffcaa8f1ebf9L198), for non-realistic pins we generate the next batch of pins right then. Since that logic still exists, I chose to keep the realistic pins as similar as possible. The one exception is that we have to have a static set of pin heights (we can't generate those on both the server and the client, or there would be a mismatch), so we pass the generated pin heights through. That allows us to call `generateRealisticExampleItems` where we would otherwise call `generateExampleItems`.

PR #2825, opened 2023-04-12
DateField, DatePicker: DateField new component (ALPHA), DatePicker moved to named export

AlbertCarreras:
#### What changed?

DateField, DatePicker: DateField new component (ALPHA), DatePicker moved to named export

https://user-images.githubusercontent.com/10593890/231777843-ccac2632-0f24-426e-834e-4fed2a08b183.MP4

Implementation from https://mui.com/x/react-date-pickers/custom-field/ 
https://github.com/mui/mui-x/blob/v6.1.0/docs/data/date-pickers/custom-field/PickerWithBrowserField.js

#### Why?

We need a DateField. 

More context https://paper.dropbox.com/doc/Gestalt-TDD-DateField--B2akCdnIbPlUBNFuCDkj2T7VAg-0xxEUdpAHiGhSCKxc7ayg

Bundle discussion https://paper.dropbox.com/doc/TDD-DateField--B2akCdnIbPlUBNFuCDkj2T7VAg-0xxEUdpAHiGhSCKxc7ayg#:uid=113131784674056085416175&h2=Bundle-increase-discussion

### Checklist


packages/gestalt-datepicker/src/InternalDateField.js
```diff
@@ -0,0 +1,285 @@
+// eslint-disable-next-line eslint-comments/disable-enable-pair
+/* eslint-disable react/display-name */
+// @flow strict-local
+import { useState, forwardRef, type Node } from 'react';
+import { DatePicker as MUIDatePicker } from '@mui/x-date-pickers';
+import { unstable_useDateField as useDateField } from '@mui/x-date-pickers/DateField';
```
rlingineni, line 6: why is this unstable? Is that how they have it defined

packages/gestalt-datepicker/src/InternalDateField.js
```diff
@@ -0,0 +1,285 @@
...
+          [styles.visuallyHidden]: labelDisplay === 'hidden',
+        })}
+        htmlFor={id}
+      >
+        <div className={styles.formLabel}>
+          <Text size="100">{label}</Text>
+        </div>
+      </label>
+      <Box position="relative" display="flex" alignItems="center">
+        <MUIDatePicker
```
rlingineni, line 199: Why is this datepicker?

packages/gestalt-datepicker/src/InternalDateField.js
```diff
@@ -0,0 +1,285 @@
...
+              {helperText ? (
+                <Text color="subtle" size="100">
+                  {helperText}
+                </Text>
+              ) : null}
+            </Flex.Item>
+          </Flex>
+        </Box>
+      ) : null}
+      {errorMessage ? (
```
rlingineni, line 265: Is this our existing input code and styling? Maybe we can have a `FieldWrapper` and the only child is the content? We can probably reusue in our other form inputs as well

AlbertCarreras: The gestalt package is not share with gestalt-datepicker so I duplicated all code. There's no really point of drying unless readability, which is also a good reason. If we move forward, I'll be a follow up.

packages/gestalt-datepicker/src/InternalDateField.js
AlbertCarreras: Implementation comes from https://github.com/mui/mui-x/blob/v6.1.0/docs/data/date-pickers/custom-field/PickerWithBrowserField.js

packages/gestalt-datepicker/src/InternalDateField.js
AlbertCarreras: From MUI docs

packages/gestalt-datepicker/src/DateField.js
```diff
@@ -0,0 +1,197 @@
...
+   * Callback triggered when the user blurs the input.
+   */
+  onBlur?: ({|
+    event: SyntheticFocusEvent<HTMLInputElement>,
+    value: string,
+  |}) => void,
+  /**
+   * DateField is a controlled component. `onChange` is the  callback triggered when the value of the input changes. Should be used to modify the controlled value. See the [controlled component example](https://gestalt.pinterest.systems/web/datefield#Controlled-component) to learn more.
+   */
+  onChange: (value: Date) => void,
```
dangerismycat, line 107: Should this match our other callbacks?
```js
onChange: ({ value: Date }) => void,
```

packages/gestalt-datepicker/src/DateField/InternalDateField.js
```diff
@@ -0,0 +1,278 @@
...
+          onChange={onChange}
+          onKeyDown={onKeyDown}
+          onMouseUp={onMouseUp}
+        />
+      </Box>
+    );
+  },
+);
+
+type CustomDateFieldProps = {
```
dangerismycat, line 102: Ideally we can make this type exact at some point (unless this is all getting thrown away)

PR #2833, opened 2023-04-14
Internal: Stop generating inline sourcemaps

liuyenwei:
#### What changed?

Stop generating inline sourcemaps for our dist bundles and generate separate .map files instead.

#### Why?
Our rollup configs are currently configured to generate inline sourcemaps, which means the sourcemap information is added to the generated bundle. This is convenient for development, but horrible for bundle size. Instead, we can change the configuration to generate separate .map files which will still allow consumers to have access to sourcemaps without the bundle size cost

Based on local testing, this should cut down bundle sizes for gestalt and gestalt-datepicker by well over 50%

### DIFF
--- a/packages/eslint-plugin-gestalt/rollup.config.js
+++ b/packages/eslint-plugin-gestalt/rollup.config.js
@@ -11,14 +11,14 @@ const rollupConfig = {
       format: 'umd',
       name: 'eslint-plugin-gestalt',
       exports: 'named',
-      sourcemap: 'inline',
+      sourcemap: false,
     },
     {
       file: 'dist/eslint-plugin-gestalt.es.js',
       format: 'es',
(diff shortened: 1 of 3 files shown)


packages/eslint-plugin-gestalt/rollup.config.js
```diff
@@ -11,14 +11,14 @@ const rollupConfig = {
       format: 'umd',
       name: 'eslint-plugin-gestalt',
       exports: 'named',
-      sourcemap: 'inline',
+      sourcemap: false,
```
liuyenwei, line 14: set this to false bc i don't think we need to generate sourcemaps for the eslint-plugin, but happy to also just set this to true

AlbertCarreras (approved):

PR #2843, opened 2023-04-14
Docs: add support for dark mode and RTL in Sandpack examples

dangerismycat:
Sandpack is awesome overall. But unfortunately due to how it renders examples in iframes, we lose the surrounding app context, most noticeably with dark mode and text direction. However, I've never been happy with needing to toggle dark mode or text direction for the entire site just to see how a single example responds — it's very disorienting, particularly text direction. 

This PR adds a couple of buttons to every Sandpack example (that we show any buttons for) to allow the user to toggle dark mode and text direction on a per-example basis. We track these clicks as with any other, so we can monitor usage. The icons used for text direction seem fine to me. The dark mode ones should be considered temporary; it'd be great to use a 🌞 / 🌚  combo as seen on most other sites.

Note that I needed to add a `fullDimensions` option to ColorSchemeProvider to set height and width to 100%. Since Sandpack only points at the current published version of Gestalt, that change is not reflected in this preview deploy or the screenshots below. But I tested by adding the styles manually to ColorSchemeProvider's div and everything looked great.

_default view_
Screen Shot 2023-04-14 at 2 48 58 PM

_dark mode_
Screen Shot 2023-04-14 at 2 48 40 PM

_RTL_
Screen Shot 2023-04-14 at 2 48 49 PM

_dark mode and RTL_
Screen Shot 2023-04-14 at 2 49 06 PM

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -125,7 +125,6 @@ module.exports = {
     'react/sort-comp': OFF,
     'react/state-in-constructor': [ERROR, NEVER],
     'react/static-property-placement': [ERROR, 'static public field'],
-    'testing-library/no-node-access': [ERROR, { 'allowContainerFirstChild': true }],
     'validate-jsx-nesting/no-invalid-jsx-nesting': ERROR,
   },
   'overrides': [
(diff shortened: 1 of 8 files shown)


AlbertCarreras (approved):

AlbertCarreras:
Thank you for this!

PR #2844, opened 2023-04-15
TileData: Adding new component

rlingineni:
### Summary
Adding an initial version of TileData

#### What changed?

Adding TileData, and moved to `InternalCheckbox` and `InternalDatapoint` to add additional customizations

image

#### Why?

This adds an initial version of TileData. 

In this case, we moved 

### Links

- [Jira](https://jira.pinadmin.com/browse/GESTALT-5248)
- [TDD](https://paper.dropbox.com/doc/Gestalt-TDD-TileData-ZuLG9VIBXAZtCRFZmR6RT)
- [Figma](https://www.figma.com/file/t0uPPhr3vjwVwVsxEVwY8D/TagData-and-TileData-components?type=design&node-id=176-4181&t=I6261EP2xk8qmGvA-0)

### Checklist

- [] Checked dark mode, responsiveness, and right-to-left support
- [] Checked stakeholder feedback (e.g. Gestalt designers)


packages/gestalt/dist/gestalt.js.flow
```diff
@@ -1,2 +0,0 @@
-// @flow strict
```
dangerismycat, line 1: I don't think there should be any `/dist` changes checked in?

packages/gestalt-design-tokens/README.md
```diff
@@ -78,6 +78,17 @@ Pretty nifty! This shows a few things happening:
 
 Now let's make a change and see how that affects things. Open up `tokens/color/base.json` and change `"#111111"` to `"#000000"`. After you make that change, save the file and re-run the build command `yarn build`. Open up the build files and take a look.
 
+## Where are the CSS Modules
```
dangerismycat, line 81: I don't think this is the correct place for this information — the design tokens are cross-platform, and this information is both web-specific and has nothing to do with design tokens.

packages/gestalt-core/build.js
```diff
@@ -57,7 +57,15 @@ const cssModules = (options = {}) => {
         scopeNames[hash] = classnameBuilder.getMinifiedClassname(hash);
       }
 
-      return scopeNames[hash];
+      // if it's dev mode, append the full class name for debugging and don't minify classnames
```
dangerismycat, line 60: Let's make this change in a separate PR. This isn't specific to adding this new component.

packages/gestalt/src/Datapoint/DataPointBaseProps.js
```diff
@@ -0,0 +1,37 @@
+/* eslint-disable eslint-comments/disable-enable-pair */
```
dangerismycat, line 1: Please don't do this — it makes it more difficult to accurately estimate the scope of disabled lint rules. If suppressions are needed, add them line-by-line.

packages/gestalt/src/Datapoint/DataPointBaseProps.js
```diff
@@ -0,0 +1,37 @@
...
+/* eslint-disable no-empty-pattern */
+/* eslint-disable react/no-unused-prop-types */
+// @flow strict
+
+/**
+ * React Prop Types File for React-DocGen to share props across similar components
```
dangerismycat, line 7: This is a lot of additional complexity to avoid repeating a few types. For now, let's just repeat the types.

packages/gestalt/src/Datapoint/DataPointBaseProps.js
```diff
@@ -0,0 +1,37 @@
...
+  /**
+   * Object detailing the trend value (change in time - e.g., +30%), and accessibilityLabel to describe the trend's icon (e.g., "Trending up").  See the [trend](https://gestalt.pinterest.systems/web/datapoint#Trend) variant to learn more.
+   */
+  trend?: TrendObject,
+  /**
+   * A visual indicator whether the trend is considered "good", "bad" or "neutral". By setting \`trendSentiment\` to \`auto\`, a positive trend will be considered "good", a negative trend will be considered "bad" and a trend of zero will be considered "neutral".  See the [trendSentiment](https://gestalt.pinterest.systems/web/datapoint#Trend-sentiment) variant to learn more.
+   */
+  trendSentiment?: 'good' | 'bad' | 'neutral' | 'auto',
+  /**
+   * The datapoint value (e.g., 1.23M)
```
dangerismycat, line 30: nit: these should all end in a period

packages/gestalt/src/Tile/Tile.css
```diff
@@ -0,0 +1,48 @@
+/* https://www.figma.com/file/t0uPPhr3vjwVwVsxEVwY8D/TagData-and-TileData-components?node-id=187-6540&t=aAjye5Ima2fHNHsh-0 */
```
dangerismycat, line 1: No need to link the Figma file here — this isn't a URL we want to commit in public code. Also it may not stay relevant with any future cleanup work or file reorganization.

packages/gestalt/src/Tile/Tile.css
```diff
@@ -0,0 +1,48 @@
...
+  composes: flex from "../Layout.css";
+  composes: paddingX4 from "../boxWhitespace.css";
+  composes: paddingY4 from "../boxWhitespace.css";
+  composes: rounding4 from "../Borders.css";
+  composes: borderBox from "../Layout.css";
+  composes: sizeLg from "../Borders.css";
+  composes: solid from "../Borders.css";
+  border-color: transparent;
+
+  /* Note: this may cause responsiveness weirdness on mobile */
```
dangerismycat, line 13: Can we solve this in a more responsive way?

packages/gestalt/src/Tile/Tile.css
```diff
@@ -0,0 +1,48 @@
+/* https://www.figma.com/file/t0uPPhr3vjwVwVsxEVwY8D/TagData-and-TileData-components?node-id=187-6540&t=aAjye5Ima2fHNHsh-0 */
+
+.tile {
+  composes: flex from "../Layout.css";
```
dangerismycat, line 4: Thanks for using composition! Can some of the other styles here be composed?

packages/gestalt/src/Tile/Tile.js
```diff
@@ -0,0 +1,161 @@
...
+import { ENTER } from '../keyCodes.js';
+import Tooltip from '../Tooltip.js';
+import useFocusVisible from '../useFocusVisible.js';
+import useInteractiveStates from '../utils/useInteractiveStates.js';
+import { FixedZIndex } from '../zIndex.js';
+import styles from './Tile.css';
+
+type Props = {|
+  /**
+   * Prop description.
```
dangerismycat, line 15: 🤔

packages/gestalt/src/Tile/Tile.js
```diff
@@ -0,0 +1,161 @@
...
+import Checkbox from '../Checkbox.js';
+import Flex from '../Flex.js';
+import { ENTER } from '../keyCodes.js';
+import Tooltip from '../Tooltip.js';
+import useFocusVisible from '../useFocusVisible.js';
+import useInteractiveStates from '../utils/useInteractiveStates.js';
+import { FixedZIndex } from '../zIndex.js';
+import styles from './Tile.css';
+
+type Props = {|
```
dangerismycat, line 13: Is this component exposed publicly? If not, there's no need for these prop descriptions except where they provide value to Future Us

packages/gestalt/src/Tile/Tile.js
```diff
@@ -0,0 +1,161 @@
...
+   * Adds a Tooltip on hover/focus of the Tile. See the With Tooltip variant to learn more.
+   */
+  tooltip?: string,
+|};
+
+function DisabledOverlay() {
+  return <div className={classnames(styles.tile, styles.disabledOverlay)} />;
+}
+
+function TooltipWrapper({
```
dangerismycat, line 56: nit: this name makes it sounds like this is a wrapper around Tooltip, instead of a component that conditionally wraps children with Tooltip. Consider something like MaybeTooltip or similar

packages/gestalt/src/Tile/Tile.js
```diff
@@ -0,0 +1,161 @@
...
+  if (!tooltip || disabled) return children;
+  return (
+    <Tooltip idealDirection="up" text={tooltip} zIndex={tooltipZIndex}>
+      {children}
+    </Tooltip>
+  );
+}
+
+/**
+ * [Tile] https://gestalt.pinterest.systems/web/tiledata component should be used for ... on the page.
```
dangerismycat, line 75: This also is unnecessary if we're not exposing this component publicly

packages/gestalt/src/Tile/Tile.js
```diff
@@ -0,0 +1,161 @@
...
+ */
+export default function Tile({
+  bgColor,
```
dangerismycat, line 78: nit: alphabetize please

packages/gestalt/src/Tile/Tile.js
```diff
@@ -0,0 +1,161 @@
...
+    if (onChange) {
+      onChange({ id, selected: !isSelected });
+    }
+  };
+
+  const generateSelectedColorStyles = () => {
+    const colorStyles: {| borderColor: string, backgroundColor: string |} = {};
+    if (!isSelected) return styles;
+
+    // the interal base component uses hex codes
```
dangerismycat, line 116: Why are we using hex values? Should we not always use tokens?

packages/gestalt/src/Tile/Tile.js
```diff
@@ -0,0 +1,161 @@
...
+  const classes = classnames(styles.tile, {
+    [styles.selected]: isSelected,
+    [styles.focused]: isFocusVisible,
+    [styles.hovered]: isHovered && !isFocusVisible,
+    [styles.disabled]: disabled,
+  });
+
+  const handleClick = () => {
+    setIsSelected(!isSelected);
+    if (onChange) {
```
dangerismycat, line 107: nit: you can avoid this sort of check with optional chaining:
```js
onChange?.( ... );
```

packages/gestalt/src/Tile/Tile.js
```diff
@@ -0,0 +1,161 @@
...
+    }
+  };
+
+  const generateSelectedColorStyles = () => {
+    const colorStyles: {| borderColor: string, backgroundColor: string |} = {};
+    if (!isSelected) return styles;
+
+    // the interal base component uses hex codes
+    // but could be passed in pre-tokenized values
+    if (borderColor && borderColor.startsWith('#')) {
```
dangerismycat, line 118: nit: optional chaining
```js
if (borderColor?.startsWith('#')) {
```

packages/gestalt/src/Tile/Tile.js
```diff
@@ -0,0 +1,161 @@
...
+  };
+
+  const handleKeyDown = (event: SyntheticKeyboardEvent<HTMLDivElement>) => {
+    if (event.keyCode === ENTER) {
+      setIsSelected(!isSelected);
+    }
+  };
+
+  return (
+    <div style={{ position: 'relative' }}>
```
dangerismycat, line 134: Is there a reason to not use Box here?

packages/gestalt/src/Tile/Tile.js
```diff
@@ -0,0 +1,161 @@
...
+      {disabled && <DisabledOverlay />}
+      <TooltipWrapper tooltip={tooltip}>
+        <div
+          tabIndex={disabled ? null : 0}
+          role="button"
```
dangerismycat, line 139: Is there a reason to not use `<button>` here?

packages/gestalt/src/Tile/Tile.js
```diff
@@ -0,0 +1,161 @@
...
+      setIsSelected(!isSelected);
+    }
+  };
+
+  return (
+    <div style={{ position: 'relative' }}>
+      {disabled && <DisabledOverlay />}
+      <TooltipWrapper tooltip={tooltip}>
+        <div
+          tabIndex={disabled ? null : 0}
```
dangerismycat, line 138: I think you want `-1` here instead of `null`

packages/gestalt/src/Tile/Tile.js
```diff
@@ -0,0 +1,161 @@
...
+  });
+
+  const handleClick = () => {
+    setIsSelected(!isSelected);
+    if (onChange) {
+      onChange({ id, selected: !isSelected });
+    }
+  };
+
+  const generateSelectedColorStyles = () => {
```
dangerismycat, line 112: This is a bit of unnecessary misdirection — this logic can just be inlined below

packages/gestalt/src/Tile/Tile.js
```diff
@@ -0,0 +1,161 @@
...
+          className={classes}
+          onBlur={handleOnBlur}
+          onClick={handleClick}
+          onMouseEnter={handleOnMouseEnter}
+          onMouseLeave={handleOnMouseLeave}
+          onKeyDown={handleKeyDown}
+          style={generateSelectedColorStyles()}
+        >
+          <Flex direction="row" gap={2}>
+            <Flex.Item>{children}</Flex.Item>
```
dangerismycat, line 150: Since you're not applying any styles or grouping multiple child components, these `Flex.Item`s don't seem necessary (they're applied automatically under the hood)

packages/gestalt/src/TileData.css.flow
```diff
@@ -0,0 +1,5 @@
+// @flow strict
```
dangerismycat, line 1: There is no TileData.css file, so this file is unnecessary

packages/gestalt/src/TileData.js
```diff
@@ -0,0 +1,104 @@
...
+  | '08'
+  | '09'
+  | '10'
+  | '11'
+  | '12';
+
+type Props = {|
+  ...DataPointBaseProps,
+  /**
+   * disables component interactivity
```
dangerismycat, line 25: Prop descriptions should all begin with a capital letter and end with a period

packages/gestalt/src/TileData.js
```diff
@@ -0,0 +1,104 @@
...
+   */
+  disabled?: boolean,
+  /**
+   * An identifier to be passed in a callback, and distinguish multiple DataPoints
```
dangerismycat, line 29: Which callback? Does this id need to be unique? If it's used for necessary disambiguation, why is it not required?

packages/gestalt/src/TileData.js
```diff
@@ -0,0 +1,104 @@
...
+   */
+  id?: string,
+  /**
+   * Handler if the item is selected
+   */
+  onChange: ({| selected: boolean, id?: string |}) => void,
```
dangerismycat, line 35: We typically always return the event in handlers, in addition to any other helpful information

packages/gestalt/src/TileData.js
```diff
@@ -0,0 +1,104 @@
...
+  /**
+   * Controls whether the tile is selected or not
+   */
+  selected?: boolean,
+  /**
+   * A valid color code from the data visualization palette
+   */
+  color?: DataVisualizationColors,
```
dangerismycat, line 43: alphabetize these props, please

packages/gestalt/src/TileData.js
```diff
@@ -0,0 +1,104 @@
+// @flow strict
+import { type Node } from 'react';
+import DataPoint from './Datapoint.js';
+import { useColorScheme } from './contexts/ColorSchemeProvider.js';
+import { type DataPointBaseProps } from './Datapoint/DataPointBaseProps.js';
+import Tile from './Tile/Tile.js';
+
+// Note: 03 is ommited because it doesn't have a corresponding dark token
```
dangerismycat, line 8: Do we not allow users to select `03` at all?

packages/gestalt/src/TileData.js
```diff
@@ -0,0 +1,104 @@
...
+  /**
+   * Handler if the item is selected
+   */
+  onChange: ({| selected: boolean, id?: string |}) => void,
+  /**
+   * Controls whether the tile is selected or not
+   */
+  selected?: boolean,
+  /**
+   * A valid color code from the data visualization palette
```
dangerismycat, line 41: This should include a link to the data viz palette page, and a reason why `03` is not included

packages/gestalt/src/TileData.js
```diff
@@ -0,0 +1,104 @@
...
+   */
+  color?: DataVisualizationColors,
+  /**
+   * Shows a checkbox. Useful when multi-select is available
+   */
+  showCheckbox?: boolean,
+  /**
+   * Adds a Tooltip on hover/focus of the Tile. See the with [Tooltip](#Tooltip) variant to learn more.
```
dangerismycat, line 49: All links in prop descriptions need to be absolute, as they're also surfaced in other contexts where the full link is needed

packages/gestalt/src/TileData.js
```diff
@@ -0,0 +1,104 @@
...
+   */
+  tooltip?: string,
```
dangerismycat, line 51: Check out how we add tooltips on other components — there are other controls that will be needed, so this will need to be an object. Badge and IconButton are good examples of this (Datapoint is very much not!)

packages/gestalt/src/TileData.js
```diff
@@ -0,0 +1,104 @@
...
+|};
+
+/**
+ * Use TileData to select a multiple categories to compare with each other in a graph or chart view, while still being able to see all of the data points
```
dangerismycat, line 55: The component description should include a link to the docs page itself. Check out other components for examples. Also this should end with a period.

packages/gestalt/src/TileData.js
```diff
@@ -0,0 +1,104 @@
...
+ */
+export default function TileData({
+  disabled = false,
```
dangerismycat, line 58: nit: alphabetize please

packages/gestalt/src/TileData.js
```diff
@@ -0,0 +1,104 @@
...
+    const shade = getColorHex(color);
+    // add an alpha channel to the hex, at 10% opacity
+    // https://gist.github.com/lopspower/03fb1cc0ac9f32ef38f4
+    const bgColor = `${shade}1A`;
+    return bgColor;
+  };
+
+  return (
+    <Tile
+      tooltip={tooltip}
```
dangerismycat, line 92: nit: alphabetize please

packages/gestalt/src/contexts/ColorSchemeProvider.js
```diff
@@ -102,8 +103,45 @@ const darkModeTheme = {
   blueActive: '#4a85c9',
 };
 
+/**
+ * Turns a token name like color-text-warning to colorTextWarning
+ */
+const cleanTokenName = (tokenName: string): string => {
```
dangerismycat, line 109: This isn't "cleaning", it's transforming from kebab-case to camelCase.

packages/gestalt/src/utils/useInteractiveStates.js
```diff
@@ -0,0 +1,61 @@
...
+  handleOnMouseUp: () => void,
+  isActive: boolean,
+  isFocused: boolean,
+  isHovered: boolean,
+|}) => {
+  const [isActive, setActive] = useState(false);
+  const [isFocused, setFocused] = useState(false);
+  const [isHovered, setHovered] = useState(false);
+
+  const handleOnBlur = () => {
```
dangerismycat, line 23: These should all use useCallback to pass down stable references

playwright/accessibility/web__tiledata.spec.mjs
```diff
@@ -0,0 +1,8 @@
+// @flow strict
+import { test } from '@playwright/test';
+import expectAccessiblePage from './expectAccessiblePage.mjs';
+
+test('About us TileData check', async ({ page }) => {
```
dangerismycat, line 5: Not sure what "About us" means here?

docs/pages/web/tiledata.js
```diff
@@ -0,0 +1,179 @@
...
+          />
+        </MainSection.Subsection>
+      </MainSection>
+
+      <MainSection name="Best practices">
+        {/* <MainSection.Subsection columns={2}>
+          <MainSection.Card
+            cardSize="sm"
+            type="do"
+            description="Use TileData paired with data visualization colors to make a visual connection between a chart and it’s related data"
```
dangerismycat, line 64: nit: "its"

docs/pages/web/tiledata.js
```diff
@@ -0,0 +1,179 @@
...
+          />
+
+          <MainSection.Card
+            cardSize="sm"
+            type="don't"
+            description="Use it to present a single option, since they don't need to be selected. Use a [Datapoint](/web/datapoint) instead."
+            defaultCode={`<TileData color="01" title="Impressions" value="2M" trend={{ value: 1, accessibilityLabel: 'Trending up' }}/> `}
+          />
+        </MainSection.Subsection>
+        <MainSection.Subsection columns={2}>
```
dangerismycat, line 89: No visual examples for these?

docs/pages/web/tiledata.js
```diff
@@ -0,0 +1,179 @@
...
+            cardSize="sm"
+            type="don't"
+            description="Use for page-level navigation, as TileData is to be used in conjunction with a chart. Use [Tabs](/web/tabs), [SegmentedControl](/web/segmentedcontrol), [SideNav](/web/sidenavigation), or another local navigation instead."
+          />
+        </MainSection.Subsection>
+      </MainSection>
+
+      <AccessibilitySection
+        name={generatedDocGen?.displayName}
+        description="Users should be able to navigate or activate a TagData using a keyboard or other input modalities. Be sure to include an `accessibilityLabel` for the screen reader within the child element."
```
dangerismycat, line 105: nit: we never use articles with component names, so this should be "Users should be able to navigate or activate _TagData_ using a…"

also: should this be "TileData"?

docs/pages/web/tiledata.js
```diff
@@ -0,0 +1,179 @@
...
+      />
+
+      <MainSection
+        name="Localization"
+        description="Make sure content that is placed in a data point is set-up to work in R-T-L languages. Be sure to localize text and `accessibilityLabel` in TileData and all subcomponents as well. When the text of the TileData reaches its Max width, either intentionally or through localization, will wrap as needed to display the full text. Keep this in mind when selecting wording for your TileData menu items. Note that localization can lengthen text by 20 to 30 percent. "
```
dangerismycat, line 110: nits:
- "…that is placed in _Datapoint_ is _set up_ to work in _RTL_ languages."
- "Be sure to localize `title`, `value`, and `trend.accessibilityLabel`."
- Which text is being referred to when wrapping? `title`? `value`? Something else? We need to be _very specific_ about these sorts of things.
- weird casing on "max width"

docs/pages/web/tiledata.js
```diff
@@ -0,0 +1,179 @@
...
+        <MainSection.Subsection columns={2}>
+          <MainSection.Card
+            cardSize="sm"
+            type="do"
+            description="Use TileData to switch between data on charts"
+          />
+          <MainSection.Card
+            cardSize="sm"
+            type="don't"
+            description="Use for page-level navigation, as TileData is to be used in conjunction with a chart. Use [Tabs](/web/tabs), [SegmentedControl](/web/segmentedcontrol), [SideNav](/web/sidenavigation), or another local navigation instead."
```
dangerismycat, line 98: "SideNav" is not a component

docs/pages/web/tiledata.js
```diff
@@ -0,0 +1,179 @@
...
+      />
+
+      <MainSection
+        name="Localization"
+        description="Make sure content that is placed in a data point is set-up to work in R-T-L languages. Be sure to localize text and `accessibilityLabel` in TileData and all subcomponents as well. When the text of the TileData reaches its Max width, either intentionally or through localization, will wrap as needed to display the full text. Keep this in mind when selecting wording for your TileData menu items. Note that localization can lengthen text by 20 to 30 percent. "
+      />
+
+      <MainSection name="Variants">
+        <MainSection.Subsection
+          description={`Use TileData's Data Visulization to display multiple colors`}
```
dangerismycat, line 115: These descriptions should all end with a period. Check out other docs pages for examples.

Also: these descriptions are very minimal. Can we provide more information to the reader about what problems they can solve using the example shown in each variant?

docs/examples/tiledata/group.js
```diff
@@ -0,0 +1,65 @@
+// @flow strict
+import { useState, type Node } from 'react';
+import { TileData, Flex } from 'gestalt';
+
+export default function Example(): Node {
+  const DataSources = [
```
dangerismycat, line 6: variables that are not React components should not use PascalCase

docs/examples/tiledata/group.js
```diff
@@ -0,0 +1,65 @@
...
+  // add all elements to our map
+  const selectedMap = DataSources.reduce((acc, tile) => {
+    acc[tile.id] = tile.id;
+    return acc;
+  }, {});
+
+  const [selectedItems, setSelectedItems] = useState(selectedMap);
+
+  const handleSelect = ({ id }) => {
+    const itemsCopy = Object.assign(selectedItems, {});
```
dangerismycat, line 39: Don't use Object.assign — use the spread operator

docs/examples/tiledata/group.js
```diff
@@ -0,0 +1,65 @@
...
+      id: 'data-4',
+      name: 'DAU',
+      value: '10M',
+      color: '04',
+      tooltip: 'Daily active users',
+    },
+  ];
+
+  // add all elements to our map
+  const selectedMap = DataSources.reduce((acc, tile) => {
```
dangerismycat, line 31: Is a map the best data structure for this? Would an array of selected ids be simpler?

docs/examples/tiledata/group.js
```diff
@@ -0,0 +1,65 @@
...
+      setSelectedItems(itemsCopy);
+    } else {
+      itemsCopy[id] = id;
+    }
+    setSelectedItems(itemsCopy);
+  };
+
+  return (
+    <Flex gap={4}>
+      {DataSources.map((tile) => (
```
dangerismycat, line 51: nit: destructure here, no need for the local `tile` variable

docs/examples/tiledata/color.js
```diff
@@ -0,0 +1,57 @@
+// @flow strict
+import { type Node, useState } from 'react';
+import { TileData, Flex } from 'gestalt';
+
+export default function Example(): Node {
+  const [selectedTileId, setSelectedTileId] = useState('1');
+
+  const isSelected = (id: string) => selectedTileId === id;
+  const handleSelection = (ev) => {
```
dangerismycat, line 9: nit: this isn't actually returning an event, and there's no need for the local `ev` variable — just destructure what's needed

docs/examples/tiledata/color.js
```diff
@@ -0,0 +1,57 @@
...
+    setSelectedTileId(ev.id);
+  };
+
+  return (
+    <Flex gap={4} wrap justifyContent="center">
+      <TileData
```
dangerismycat, line 15: Is there any reason we're showing just these five colors? Should we not show all available colors, either at-a-glance or using a SelectList?

docs/pages/web/tiledata.js
```diff
@@ -0,0 +1,179 @@
...
+        >
+          <MainSection.Card
+            sandpackExample={<SandpackExample code={group} name="Checkbox Variant" />}
+          />
+        </MainSection.Subsection>
+      </MainSection>
+
+      <QualityChecklist component={generatedDocGen?.displayName} />
+
+      <MainSection name="Related">
```
dangerismycat, line 152: What other similar components are there? Either visually similar or functionally similar?

dangerismycat (changes requested):
Not sure why there are yarn.lock changes in this PR? Those should be reverted.

docs/pages/web/tiledata.js
laylaleeal: > 

Hey Ryan! we don't have visually similar components ATM in Gestalt. but functionality is similar to checkboxes/radios

package.json
```diff
@@ -5,6 +5,7 @@
     "packages/*"
   ],
   "dependencies": {
+    "node-fetch": "^2.6.7",
```
dangerismycat, line 8: Why are we adding this dependency? Please double-check all changes introduced in PRs, _especially_ anything touching package.json or yarn.lock. This is particularly important with larger PRs like this one, where individual file changes can be missed.

package.json
```diff
@@ -83,8 +84,8 @@
     "stylelint-config-prettier": "^9.0.4",
     "stylelint-config-standard": "^23.0.0",
     "stylelint-order": "^5.0.0",
-    "xml2js": "^0.5.0",
-    "typescript": "^5.0.2"
+    "typescript": "^5.0.2",
```
dangerismycat, line 87: This change will fall off after rebasing once [this PR](https://github.com/pinterest/gestalt/pull/2932) is merged

packages/gestalt/src/Checkbox/InternalCheckbox.js
```diff
@@ -0,0 +1,260 @@
...
+import Label from '../Label.js';
+import FormErrorMessage from '../shared/FormErrorMessage.js';
+import FormHelperText from '../shared/FormHelperText.js';
+import Text from '../Text.js';
+import useFocusVisible from '../useFocusVisible.js';
+import styles from './InternalCheckbox.css';
+
+type Props = {|
+  /**
+   * Indicates whether or not Checkbox is checked. See the [state variant](https://gestalt.pinterest.systems/web/checkbox#State) to learn more.
```
dangerismycat, line 25: These comments are no longer necessary now that this is an internal component, as they will never be surfaced to users. They can all be removed, unless there are any that are particularly useful to Future Us.

packages/gestalt/src/Datapoint.js
```diff
@@ -33,25 +27,25 @@ type Props = {|
...
+   * The datapoint value (e.g., 1.23M).
    */
-  trend?: TrendObject,
+  value: string,
   /**
-   * A visual indicator whether the trend is considered "good", "bad" or "neutral". By setting \`trendSentiment\` to \`auto\`, a positive trend will be considered "good", a negative trend will be considered "bad" and a trend of zero will be considered "neutral".  See the [trendSentiment](https://gestalt.pinterest.systems#Trend-sentiment) variant to learn more.
+   * Contextual information displayed in a tooltip to describe the Datapoint. See the [tooltipText](https://gestalt.pinterest.systems#Tooltip-text) variant to learn more.
    */
-  trendSentiment?: 'good' | 'bad' | 'neutral' | 'auto',
+  tooltipText?: string,
```
dangerismycat, line 44: Please keep the props alphabetized

packages/gestalt/src/Datapoint/InternalDatapoint.js
```diff
@@ -0,0 +1,117 @@
...
+|};
+
+type TrendObject = {|
+  accessibilityLabel: string,
+  value: number,
+|};
+
+type Props = {|
+  /**
+   * Adds a badge to the title. Currently a beta feature, expect changes.
```
dangerismycat, line 24: ditto here, no need for these comments now that this is internal. Only worth calling out what's different from Datapoint

packages/gestalt/src/Tile/Tile.css
dangerismycat: ^ this question remains

packages/gestalt/src/Tile/Tile.js
dangerismycat: This conversation was marked as "resolved" but no changes were made, and there was no explanation. Please do not do that.

This question remains: are we exposing this component publicly? If not, no need for these comments.

packages/gestalt/src/Tile/Tile.js
```diff
@@ -0,0 +1,189 @@
...
+   * Shows a visible checkbox when the tile is selected state. See when using in a [group](http://gestalt.pinterest.systems/web/tiledata#Group).
+   */
+  showCheckbox?: boolean,
+  /**
+   * Adds a Tooltip on hover/focus of the Tile. See the With Tooltip variant to learn more.
+   */
+  tooltip?: TooltipProps,
+|};
+
+function ShouldUseTooltip({
```
dangerismycat, line 72: nit: "ShouldUseTooltip" sounds like a function that would return a boolean indicating if a tooltip should be used or not. This component conditionally wraps its children in a tooltip, so we'd typically use something like `MaybeTooltip`

packages/gestalt/src/Tile/Tile.js
dangerismycat: This was marked "resolved" but neither addressed nor explained.

packages/gestalt/src/Tile/Tile.js
```diff
@@ -0,0 +1,189 @@
...
+  id = '',
+  onChange,
+  selected = false,
+  showCheckbox,
+  tooltip,
+}: Props): Node {
+  const { handleOnBlur, handleOnMouseEnter, handleOnMouseLeave, isHovered } =
+    useInteractiveStates();
+
+  const [isSelected, setIsSelected] = useState(selected);
```
dangerismycat, line 112: This is a code smell. Why are we duplicating selection logic in the component? In what scenario would need to manipulate selected state internally and not rely on `selected` and `onChange`?

packages/gestalt/src/Tile/Tile.js
```diff
@@ -0,0 +1,189 @@
...
+/**
+ * Used Internally to wrap a component with a Tile View
+ */
+export default function Tile({
+  bgColor,
+  borderColor,
+  children,
+  disabled = false,
+  id = '',
+  onChange,
```
dangerismycat, line 104: This prop name isn't clear. What's changing? Is this `onClick` or `onTap`?

packages/gestalt/src/Tile/Tile.js
dangerismycat: I'm still unclear as to what this function is solving? It's only called once, and accepts no parameters, relying on the surrounding function context. If an easily testable helper function is desired, this should be moved out of the function body and be a pure function. Otherwise there's no need to wrap this logic in a function.

packages/gestalt/src/Tile/Tile.js
```diff
@@ -0,0 +1,189 @@
...
+    return colorStyles;
+  };
+
+  const handleKeyDown = ({
+    event,
+  }: {|
+    event: SyntheticKeyboardEvent<HTMLDivElement> | SyntheticKeyboardEvent<HTMLAnchorElement>,
+  |}) => {
+    if (event.keyCode === ENTER) {
+      setIsSelected(!isSelected);
```
dangerismycat, line 151: Is there a reason to not call `onChange` here? Wouldn't that lead to drift between the internal and external state?

packages/gestalt/src/TileData.js
```diff
@@ -0,0 +1,149 @@
...
+  /**
+   * A valid color code from the [data visualization palette](https://gestalt.pinterest.systems/foundations/data_visualization/palette). 03 is not supported yet, because it has no corresponding dark theme.
+   */
+  color?: DataVisualizationColors,
+  /**
+   * Indicates if TileData should be disabled. Disabled TileData is inactive and cannot be interacted with.
+   */
+  disabled?: boolean,
+  /**
+   * An optional identifier to be passed back in the onChange callback. It can be helpful to distinguish multiple DataPoints.
```
dangerismycat, line 55: s/Datapoint/TileData

packages/gestalt/src/TileData.js
```diff
@@ -0,0 +1,149 @@
...
+   */
+  id?: string,
+  /**
+   * Handler if the item selection state is changed.
+   */
+  onChange?: TileChangeHandler,
```
dangerismycat, line 61: `onChange` isn't a clear prop name. Consider `onClick`, or `onSelectionChange`

packages/gestalt/src/TileData.js
```diff
@@ -0,0 +1,149 @@
...
+  /**
+   * Controls whether the tile is selected or not.
+   */
+  selected?: boolean,
```
dangerismycat, line 65: `selected` and `onChange` must always be used together, yes? If so, that should be indicated in these descriptions.

packages/gestalt/src/TileData.js
```diff
@@ -0,0 +1,149 @@
...
+  /**
+   * Shows a visible checkbox when the tile is selected state. See when using in a [group](https://gestalt.pinterest.systems/web/tiledata#Group).
+   */
+  showCheckbox?: boolean,
```
dangerismycat, line 69: Is this prop necessary? When would we want to allow for selection and not show a checkbox? What happens if `showCheckbox` is true but neither `selected` nor `onChange` are present?

packages/gestalt/src/TileData.js
```diff
@@ -0,0 +1,149 @@
...
+  /**
+   * The header text for the component.
+   */
+  title: string,
+  /**
+   * Adds a Tooltip on hover/focus of the Tile. See the with [Tooltip](https://gestalt.pinterest.systems/web/tooltip) variant to learn more.
```
dangerismycat, line 75: s/Tile/TileData

packages/gestalt/src/TileData.js
```diff
@@ -0,0 +1,149 @@
...
+  tooltip,
+  title,
+  trend,
+  trendSentiment,
+  value,
+}: Props): Node {
+  const theme = useColorScheme();
+
+  /** We use the color hex to generate a shade. Data visualization colors are a part of theme tokens */
+  const getColorHex = (vizColor: string) => {
```
dangerismycat, line 111: Consider moving these helper functions outside of the function component body so you can write unit tests on them

packages/gestalt/src/TileData.js
dangerismycat: This was marked "resolved" and neither addressed nor commented on.

packages/gestalt/src/TileData.test.js
```diff
@@ -0,0 +1,26 @@
+// @flow strict
+import { create } from 'react-test-renderer';
+import TileData from './TileData.js';
+
+describe('TileData', () => {
```
dangerismycat, line 5: These tests are pretty minimal. Can we get better coverage of the various UI states of this component?

yarn.lock
```diff
@@ -18329,4 +18329,4 @@ zip-stream@^3.0.1:
 zwitch@^2.0.0:
   version "2.0.2"
   resolved "https://registry.yarnpkg.com/zwitch/-/zwitch-2.0.2.tgz#91f8d0e901ffa3d66599756dde7f57b17c95dce1"
-  integrity sha512-JZxotl7SxAJH0j7dN4pxsTV6ZLXoLdGME+PsjkL/DaBrVryK9kTGq06GfKrwcSOqypP+fdXGoCHE36b99fWVoA==
+  integrity sha512-JZxotl7SxAJH0j7dN4pxsTV6ZLXoLdGME+PsjkL/DaBrVryK9kTGq06GfKrwcSOqypP+fdXGoCHE36b99fWVoA==
```
dangerismycat, line 18332: There should not be any yarn.lock changes in this PR.

dangerismycat (changes requested):
Don't forget to update `docs/docs-components/siteIndex.js` so TileData appears in the side nav.

There's an issue with the current approach for selection: the checkboxes themselves do not respond to clicks. This isn't good. The entire element needs to be clickable.
https://github.com/pinterest/gestalt/assets/12059539/5cae13b8-ac89-44d5-88c5-6286ced6fd3b

dangerismycat (changes requested):

docs/pages/web/tiledata.js
laylaleeal: the team mentioned not doing visual examples on do's and don'ts if they're simple and don't need a visual for support

docs/docs-components/COMPONENT_DATA.js
```diff
@@ -2804,6 +2805,56 @@ const GENERAL_COMPONENT_LIST: $ReadOnlyArray<ListItemType> = [
...
+      accessible: {
+        summary: null,
+        a11yVisual: null,
+        a11yScreenreader: null,
+        a11yNavigation: null,
+        a11yComprehension: null,
+      },
+      badge: null,
+      deprecated: false,
+      documentation: 'ready',
```
dangerismycat, line 2841: I don't think this is true? If we do have Android/iOS docs ready, we shouldn't mark this as `'ready'` in this PR — let's wait till we actually publish those docs.

packages/gestalt/src/Tile/Tile.js
dangerismycat: I'm still seeing the prop description comments. If there is a difference between this component's props and the props of external components, we should call that out. Otherwise these comments are a second source of truth and there will likely be drift with the information surfaced in the docs.

packages/gestalt/src/Tile/Tile.js
dangerismycat: Can you address this question please? Why are we duplicating selected logic with state?

packages/gestalt/src/Tile/Tile.js
dangerismycat: I know this is just a nit, but it would still be nice to see the comment either addressed in the code or as a comment here.

packages/gestalt/src/TileData.js
dangerismycat: ?

docs/pages/web/tiledata.js
```diff
@@ -0,0 +1,148 @@
...
+        </MainSection.Subsection>
+      </MainSection>
+
+      <MainSection name="Best practices">
+        <MainSection.Subsection columns={2}>
+          <MainSection.Card
+            cardSize="sm"
+            type="do"
+            description="Always present one tile in its selected state on default."
+            defaultCode={`<TileData color="01" title="Impressions" value="2M" selected trend={{ value: 5, accessibilityLabel: 'Trending up' }} /> `}
```
dangerismycat, line 54: We should avoid defining code as inline strings like this — we don't get the benefit of linting, type checking, etc. All examples should be defined in `/examples` files.

docs/pages/web/tiledata.js
dangerismycat: @rlingineni Can we add Checkbox and RadioGroup here?

docs/pages/web/tiledata.js
```diff
@@ -0,0 +1,148 @@
...
+            sandpackExample={<SandpackExample code={disabled} name="Disabled variant" />}
+          />
+        </MainSection.Subsection>
+
+        <MainSection.Subsection
+          description={`Use checkboxes when enabling a multi-select experience. You can show a checkbox state by passing the \`showCheckbox\` prop.`}
+          title="Group"
+        >
+          <MainSection.Card
+            sandpackExample={<SandpackExample code={group} name="Checkbox Variant" />}
```
dangerismycat, line 125: I'm not seeing the checkbox when the item isn't selected. Is that intentional?
Screen Shot 2023-05-17 at 2 30 26 PM

docs/pages/web/tiledata.js
```diff
@@ -0,0 +1,148 @@
...
+            cardSize="sm"
+            type="do"
+            description="Always present one tile in its selected state on default."
+            defaultCode={`<TileData color="01" title="Impressions" value="2M" selected trend={{ value: 5, accessibilityLabel: 'Trending up' }} /> `}
+          />
+
+          <MainSection.Card
+            cardSize="sm"
+            type="don't"
+            description="Use TileData to present a single option, since TileData's don't need to be selected. Use [Datapoint](https://gestalt.pinterest.systems/web/datapoint) instead."
```
dangerismycat, line 60: I don't understand the second clause of the first sentence, "since TileDatas don't need to be selected". What are we trying to convey here?

docs/pages/web/tiledata.js
```diff
@@ -0,0 +1,148 @@
...
+            defaultCode={`<TileData color="01" title="Impressions" value="2M" trend={{ value: 1, accessibilityLabel: 'Trending up' }}/> `}
+          />
+        </MainSection.Subsection>
+        <MainSection.Subsection columns={2}>
+          <MainSection.Card
+            cardSize="sm"
+            type="do"
+            description="Use the `showCheckbox` property when multiple Tiledatas can be selected. See the [group](https://gestalt.pinterest.systems/web/tiledata#Group) variant."
```
dangerismycat, line 68: "TileData". Also, we typically tell the reader why we're directing them to a link, like "See the group variant for more details."

docs/pages/web/tiledata.js
```diff
@@ -0,0 +1,148 @@
...
+            defaultCode={`<Flex gap={2}> <TileData color="01" title="Impressions" selected value="2M" showCheckbox trend={{ value: 1, accessibilityLabel: 'Trending up' }}/> <TileData color="02" title="Impressions" selected showCheckbox value="2M" trend={{ value: 1, accessibilityLabel: 'Trending up' }}/>   </Flex>`}
+          />
+          <MainSection.Card
+            cardSize="sm"
+            type="don't"
+            description="Use Tiledata when multiple items can be selected without a visible checkbox."
```
dangerismycat, line 74: "TileData"

docs/pages/web/tiledata.js
```diff
@@ -0,0 +1,148 @@
...
+            defaultCode={`<Flex gap={2}> <TileData color="01" title="Impressions" selected value="2M"  trend={{ value: 1, accessibilityLabel: 'Trending up' }}/> <TileData color="02" title="MAU" selected  value="2M" trend={{ value: 1, accessibilityLabel: 'Trending up' }}/>   </Flex>`}
+          />
+        </MainSection.Subsection>
+      </MainSection>
+
+      <AccessibilitySection
+        name={generatedDocGen?.displayName}
+        description="Users should be able to navigate or activate TileData using a keyboard or other input modalities. Be sure to include an `accessibilityLabel` for the screen reader if you're using the `trend` or `tooltip` label."
```
dangerismycat, line 82: "…if you're using the `trend` or `tooltip` _props_."

docs/pages/web/tiledata.js
```diff
@@ -0,0 +1,148 @@
...
+      />
+
+      <MainSection
+        name="Localization"
+        description={`
+          Make sure that is placed in a Datapoint is set-up to work in RTL languages. Be sure to localize \`title\`, \`value\`, \`trend.accessibilityLabel\`, and \`tooltip.accessibilityLabel\` in TileData.
```
dangerismycat, line 88: This still references Datapoint. This page is for TileData.

And I still don't understand what this first sentence is telling the user? As mentioned previously, we don't provide any free-form (e.g. Node) props, so any RTL support should be handled within TileData itself. What specific things would the user need to do to support RTL?

docs/pages/web/tiledata.js
```diff
@@ -0,0 +1,148 @@
...
+        <MainSection.Subsection
+          description="TileData can be used along side the colors provided from the Data Visualization [Color Palette](https://gestalt.pinterest.systems/foundations/data_visualization/palette#12-Color-categorical-palette). You may use colors to distinguish different data lines."
+          title="Colors"
+        >
+          <MainSection.Card
+            sandpackExample={<SandpackExample code={color} name="Colors Variant" />}
+          />
+        </MainSection.Subsection>
+        <MainSection.Subsection
+          description="Use tooltip to display clarifying information on hover/focus. We recommend using tooltips when trying to provide the user additional context/details."
```
dangerismycat, line 103: backticks needed around `tooltip` to clarify that we're referring to the prop.

packages/gestalt/src/Datapoint.js
```diff
@@ -41,15 +35,15 @@ type Props = {|
    */
   tooltipZIndex?: Indexable,
   /**
-   * Object detailing the trend value (change in time - e.g., +30%), and accessibilityLabel to describe the trend's icon (e.g., "Trending up").  See the [trend](https://gestalt.pinterest.systems#Trend) variant to learn more.
+   * Object detailing the trend value (change in time - e.g., +30%), and accessibilityLabel to describe the trend's icon (e.g., "Trending up").  See the [trend](https://gestalt.pinterest.systems/web/datapoint#Trend) variant to learn more.
```
dangerismycat, line 38: Good catch on these URLs!

packages/gestalt/src/Datapoint/InternalDatapoint.js
```diff
@@ -0,0 +1,98 @@
...
+  title: string,
+  tooltipText?: string,
+  trend?: TrendObject,
+  trendSentiment?: 'good' | 'bad' | 'neutral' | 'auto',
+  tooltipZIndex?: Indexable,
+  value: string,
+|};
+
+/**
+ * [Datapoint](https://gestalt.pinterest.systems/web/datapoint) displays at-a-glance data for a user to quickly view key metrics.
```
dangerismycat, line 42: You can remove these comments as well.

packages/gestalt/src/Checkbox/InternalCheckbox.js
```diff
@@ -0,0 +1,217 @@
...
+   * Make the checkbox readonly. Interactivity is disabled, but it can be used as a visual indicator. Click handler events are also disabled
+   */
+  readOnly?: boolean,
+  ref?: Element<'input'>, // eslint-disable-line react/no-unused-prop-types
+  size?: 'sm' | 'md',
+  style?: {| borderColor?: string, backgroundColor?: string |},
+|};
+
+/**
+ * [Checkbox](https://gestalt.pinterest.systems/web/checkbox) is used for multiple choice selection. They are independent of each other in a list, and therefore, different from [RadioButton](https://gestalt.pinterest.systems/web/radiobutton), one selection does not affect other checkboxes in the same list.
```
dangerismycat, line 46: You can remove these comments as well.

packages/gestalt/src/Tile/Tile.js
```diff
@@ -0,0 +1,186 @@
...
+   */
+  id?: string,
+  /**
+   * Indicate if tile is in a selected state.
+   */
+  selected?: boolean,
+  /**
+   * Handler if the item selection state is changed.
+   */
+  onTap?: TileChangeHandler,
```
dangerismycat, line 60: nit: alphabetize please

packages/gestalt/src/Tile/Tile.js
```diff
@@ -0,0 +1,186 @@
...
+    | SyntheticKeyboardEvent<HTMLAnchorElement>,
+  selected: boolean,
+  id?: string,
+|}) => void;
+
+type Props = {|
+  /**
+   * The content to be wrapped by tile.
+   */
+  children?: Node,
```
dangerismycat, line 36: nit: alphabetize please

packages/gestalt/src/TileData.js
```diff
@@ -0,0 +1,154 @@
...
+  /**
+   * A valid color code from the [data visualization palette](https://gestalt.pinterest.systems/foundations/data_visualization/palette).
+   */
+  color?: DataVisualizationColors,
+  /**
+   * Indicates if TileData should be disabled. Disabled TileData is inactive and cannot be interacted with.
+   */
+  disabled?: boolean,
+  /**
+   * An optional identifier to be passed back in the onChange callback. It can be helpful to distinguish multiple TileDatas.
```
dangerismycat, line 55: This still references `onChange`

packages/gestalt/src/TileData.js
```diff
@@ -0,0 +1,154 @@
...
+  /**
+   * Handler if the item selection state is changed.
+   */
+  onTap?: TileChangeHandler,
+  /**
+   * Controls whether the tile is selected or not.
+   */
+  selected?: boolean,
+  /**
+   * Shows a visible checkbox when the tile is selected state. See when using in a [group](https://gestalt.pinterest.systems/web/tiledata#Group).
```
dangerismycat, line 67: s/"tile"/"TileData"

packages/gestalt/src/TileData.js
dangerismycat: ?

packages/gestalt/src/TileData.js
dangerismycat: ?

packages/gestalt/src/TileData.js
dangerismycat: I understand the "selection without a checkbox" use case, but what happens if the user passes `showCheckbox` and neither `selected` nor `onTap`?

rlingineni (commented):
Adding review comments

docs/pages/web/tiledata.js
```diff
@@ -0,0 +1,185 @@
...
+          />
+        </MainSection.Subsection>
+      </MainSection>
+
+      <QualityChecklist component={generatedDocGen?.displayName} />
+
+      <MainSection name="Related">
+        <MainSection.Subsection
+          description={`
+      **[Datapoint](/Datapoint)**
```
dangerismycat, line 166: this URL is incorrect: `/web/datapoint`

packages/gestalt/src/TileData.js
```diff
@@ -0,0 +1,155 @@
...
+  /**
+   * An optional identifier to be passed back in the onTap callback. It can be helpful to distinguish multiple TileDatas.
+   */
+  id?: string,
+  /**
+   * Handler if the item selection state is changed.
+   */
+  onTap?: TileChangeHandler,
+  /**
+   * Controls whether the TileData is selected or not. Use it alongside the OnTap Handler
```
dangerismycat, line 63: No need to capitalize "handler", and sentences always end with a period.

packages/gestalt/src/Tile/Tile.js
dangerismycat: The ideal state makes sense — but that would involve a lot more control that we don't currently have internally. For now this is something that will have to remain solved by devs externally (similar to pretty much all our interactions, really)

packages/gestalt/src/TileData.js
dangerismycat: We don't need to be dogmatic about always moving helper functions outside of the parent scope, but it's a good idea in general to make their dependencies more obvious and to be able to write unit tests on that logic specifically. In this case, the TileData component function now has basically no logic in it, so unit tests on it can be pretty minimal (assuming our test coverage of Tile and InternalDatapoint is pretty good). In general it's a lot faster and easier to write good unit tests on pure functions as opposed to components, so I typically advocate for pulling out as much logic as possible and spending more time testing the helper functions than the actual component.

rlingineni (commented):
Fixed and made changes

rlingineni (commented):
Fixed and made changes

rlingineni (commented):
Fixed and made changes

rlingineni (commented):
Fixed and made changes

packages/gestalt/src/Tile/Tile.js
```diff
@@ -0,0 +1,187 @@
...
+    disabled: boolean,
+    hovered: boolean,
+    selected: boolean,
+  |},
+  colorStyles: {| borderColor?: string, backgroundColor?: string |},
+) {
+  let backgroundColor = 'transparent';
+  let borderColor = 'transparent';
+
+  if (state.selected) {
```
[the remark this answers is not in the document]
dangerismycat: This works! Much much easier to read and parse the logic hierarchy.

docs/pages/web/tiledata.js
```diff
@@ -0,0 +1,185 @@
...
+          />
+        </MainSection.Subsection>
+      </MainSection>
+
+      <QualityChecklist component={generatedDocGen?.displayName} />
+
+      <MainSection name="Related">
+        <MainSection.Subsection
+          description={`
+      **[Datapoint](/web/Datapoint)**
```
dangerismycat, line 166: This is still incorrect — URLs are case-sensitive, and all our component paths are completely lowercase

dangerismycat (approved):
Looks good! Nice work!

docs/pages/web/tiledata.js
rlingineni: oops, got it! Fixed

PR #2853, opened 2023-04-18
Masonry: add experimental prop to batch paints

dangerismycat:
Currently Masonry measures and paints items sequentially, like
```
measure
paint
measure
paint
measure
paint
```
We're probably going to need to alter this behavior for future 2-col module work. @liuyenwei had the idea to test out that change independent of the modules and algorithm changes, by batching paints. So we'd instead do something like
```
measure
measure
measure
paint
paint
paint
```
in the hope that the browser can batch the paints together, which would be more efficient and performant.

This PR introduces a mechanism to experimentally batch Masonry paints, which we can run immediately to start testing out hypotheses around batching performance.

Preliminary perf output looks promising…
_no batching_
Screen Shot 2023-04-19 at 3 22 38 PM

_batching, reading clientHeight and setting immediately_
Screen Shot 2023-04-19 at 3 24 28 PM

_batching, also batching clientHeight reads_
Screen Shot 2023-04-19 at 3 28 54 PM

Looking at paints (dark purple): all three begin with a couple of paints, first `Pre-Paint` and then `Layerize`. The non-batched version also has a couple of additional paints, `Recalculate Style` and `Layout`. These additional paints appear not per item, but per set of `minCol` items. Since both versions of batching eliminate these two paints, that's possible indication of a perf win.

Batching clientHeight reads also seems to avoid the (very) tall column of `recursivelyTraverseMutationEffects`/`commitMutationEffectsOnFiber`. I'm not sure if this is something we care about, but it is noticeable in the flame chart.

### DIFF
--- a/packages/gestalt/src/Masonry.js
+++ b/packages/gestalt/src/Masonry.js
@@ -7,6 +7,7 @@ import throttle, { type ThrottleReturn } from './throttle.js';
 import { type Cache } from './Masonry/Cache.js';
 import defaultLayout from './Masonry/defaultLayout.js';
 import fullWidthLayout from './Masonry/fullWidthLayout.js';
+import MeasureItems from './Masonry/MeasureItems.js';
 import MeasurementStore from './Masonry/MeasurementStore.js';
 import ScrollContainer from './Masonry/ScrollContainer.js';
 import { getElementHeight, getRelativeScrollTop, getScrollPos } from './Masonry/scrollUtils.js';
@@ -93,6 +94,10 @@ type Props<T> = {|
    * Specifies whether or not Masonry dynamically adds/removes content from the grid based on the user's viewport and scroll position. Note that `scrollContainer` must be specified when virtualization is used.
    */
   virtualize?: boolean,
+  /**
+   * Experimental prop to batch paints for possible performance improvements. This is an experimental prop and may be removed in the future.
+   */
+  _batchPaints?: boolean,
 |};
 
 type State<T> = {|
@@ -444,6 +449,7 @@ export default class Masonry<T: { ... }> extends ReactComponent<Props<T>, State<
       minCols,
       renderItem,
       scrollContainer,
+      _batchPaints,
     } = this.props;
     const { hasPendingMeasurements, measurementStore, width } = this.state;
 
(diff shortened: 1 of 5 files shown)


packages/gestalt/src/Masonry.js
```diff
@@ -93,6 +94,10 @@ type Props<T> = {|
    * Specifies whether or not Masonry dynamically adds/removes content from the grid based on the user's viewport and scroll position. Note that `scrollContainer` must be specified when virtualization is used.
    */
   virtualize?: boolean,
+  /**
+   * Experimental prop to batch paints for possible performance improvements. This is an experimental prop and may be removed in the future.
+   */
+  _batchPaints?: boolean,
```
dangerismycat, line 100: Props prefixed with an underscore are automatically filtered out of the props tables in the docs, so this will remain undocumented

packages/gestalt/src/Masonry/MeasureItems.js
```diff
@@ -0,0 +1,69 @@
...
+  measurementStore,
+  renderItem,
+}: Props<T>): Node {
+  const measuringPositions = getPositions(items);
+  const refs = useMemo(() => new Map(), []);
+
+  useLayoutEffect(() => {
+    if (refs.size === items.length) {
+      refs.forEach((el, data) => {
+        measurementStore.set(data, el.clientHeight);
```
liuyenwei, line 33: would do some profiling and double check but my guess is that you'll want to do something like:
```
const heights = refs.map((el) => el.clientHeight);
heights.forEach((height) => {
  measurementStore.set(..)
});
```

this ensures that you're grouping all the calls to `clientHeight` together so the browser will batch the paints. my guess is that intermediate `measurementStore.set` calls would result in the browser triggering multiple paints still

packages/gestalt/src/Masonry.js
```diff
@@ -543,33 +549,43 @@ export default class Masonry<T: { ... }> extends ReactComponent<Props<T>, State<
...
-                >
-                  {renderItem({ data, itemIdx: measurementIndex, isMeasuring: true })}
-                </div>
-              );
-            })}
+            {_batchPaints ? (
+              <MeasureItems
+                baseIndex={itemsToRender.length}
+                getPositions={getPositions}
+                items={items}
```
liuyenwei, line 556: should this be `itemsToMeasure`?

packages/gestalt/src/Masonry/MeasureItems.js
dangerismycat: Ah that makes sense…that's the part I was missing from our conversation about the approach

packages/gestalt/src/Masonry.js
dangerismycat: 🤦 indeed it should

liuyenwei (approved):

PR #2860, opened 2023-04-19
Link: Fixing code not expanding vertically

eniomoura:
#### What changed?

Fixed the cases where code examples should open vertically but were opened to the side instead.

#### Why?

Some change recently may have changed the default direction of code expanding, thus cases where the direction wasn't explicit were now opening incorrectly.

### Links

- [Link docs page](https://gestalt.pinterest.systems/web/link)

### Checklist

- [ N/A ] Added unit and Flow Tests
- [ N/A ] Added documentation + accessibility tests


AlbertCarreras (approved):
lgtm

PR #2864, opened 2023-04-20
Popover, PopoverEducational: internal popover component to build different APIs for Popover & PopoverEducational

eniomoura:
#### What changed?

Created new InternalPopover component, and substituted PopoverEducational and Popover contents to utilize it

#### Why?

To expose only necessary props for usage in the intended use cases.

### Links

- [Jira](https://jira.pinadmin.com/browse/GESTALT-6017)

### Checklist

- [ Yes ] Added unit and Flow Tests
- [ N/A ] Added documentation + accessibility tests


packages/gestalt/src/Popover/InternalPopover.js
```diff
@@ -0,0 +1,157 @@
...
+import Flex from '../Flex.js';
+import InternalDismissButton from '../shared/InternalDismissButton.js';
+
+type Color = 'blue' | 'red' | 'white' | 'darkGray';
+type Size = 'xs' | 'sm' | 'md' | 'lg' | 'xl' | 'flexible' | number;
+type IdealDirection = 'up' | 'right' | 'down' | 'left';
+type Role = 'dialog' | 'listbox' | 'menu' | 'tooltip';
+
+type Props = {|
+  /**
```
AlbertCarreras, line 15: can we remove all the prop descriptions here?

we only keep the descriptions in the exposed API (unless there's a prop that we want to add more context to)

packages/gestalt/src/Popover/InternalPopover.js
```diff
@@ -0,0 +1,157 @@
...
+   * The maximum width of Popover. See the [size](https://gestalt.pinterest.systems/web/popover#Size) variant to learn more.
+   */
+  size?: Size,
+  // This property can be set when `ScrollBoundaryContainer` is set to `overflow="visible"` but therefore limits the height of the Popover-based component. Some cases require
+  __dangerouslySetMaxHeight?: '30vh',
+|};
+
+/**
+ * [Popover](https://gestalt.pinterest.systems/web/popover) is a floating view that contains a task related to the content on screen. It can be triggered when the user clicks or focuses on an element, typically [Button](https://gestalt.pinterest.systems/web/button) or [IconButton](https://gestalt.pinterest.systems/web/iconbutton). It can also be triggered automatically, as in the case of user education. Popover is non-modal and can be dismissed by interacting with another part of the screen or an item within Popover.
```
AlbertCarreras, line 80: We don't need the description here, same as prop descriptions

packages/gestalt/src/Popover/InternalPopover.js
```diff
@@ -0,0 +1,157 @@
...
+ *
+ * Popover is most appropriate for desktop screens and can contain a variety of elements, such as [Button](https://gestalt.pinterest.systems/web/button) and [Images](https://gestalt.pinterest.systems/web/image). Popover is also the container used to construct more complex elements like [Dropdown](https://gestalt.pinterest.systems/web/dropdown) and the board picker, pictured below, which allow people to choose the board to save a Pin to.
+ *
+ * ![Popover light mode](https://raw.githubusercontent.com/pinterest/gestalt/master/playwright/visual-test/Popover.spec.mjs-snapshots/Popover-chromium-darwin.png)
+ * ![Popover dark mode](https://raw.githubusercontent.com/pinterest/gestalt/master/playwright/visual-test/Popover-dark.spec.mjs-snapshots/Popover-dark-chromium-darwin.png)
+ *
+ */
+export default function Popover({
```
AlbertCarreras, line 88: Can we rename to InternalPopover here?

packages/gestalt/src/Popover/InternalPopover.js
```diff
@@ -0,0 +1,157 @@
...
+  accessibilityLabel = 'Popover',
```
AlbertCarreras, line 89: these is pretty nit.

I think these defaults props can be removed as we are setting them on the external Popover. 

If we do that let's make sure we don't lose them on PopoverEducational, unless we are overwriting them on PopoverEducational

packages/gestalt/src/Popover/InternalPopover.flowtest.js
```diff
@@ -0,0 +1,7 @@
+// @flow strict
```
AlbertCarreras, line 1: We don't really need this file. flowtest.js files are pretty boilerplater. we don't get much value in this case

packages/gestalt/src/Popover/InternalPopover.jsdom.test.js
```diff
@@ -0,0 +1,29 @@
+// @flow strict
```
AlbertCarreras, line 1: Cos we have Popover and PopoverEducational tests, we can leave this test or remove as it not adding much value as it's redundant with the other two

packages/gestalt/src/PopoverEducational.js
```diff
@@ -162,7 +162,7 @@ export default function PopoverEducational({
 
   return (
     <Box zIndex={zIndex}>
-      <Popover
+      <InternalPopover
```
AlbertCarreras, line 165: Line 174 is redundant as we set a default.  Workrth try cleaning up

packages/gestalt/src/PopoverEducational.js
```diff
@@ -162,7 +162,7 @@ export default function PopoverEducational({
...
         accessibilityLabel={accessibilityLabel}
```
AlbertCarreras, line 166: Here we lost the accessibilityLabel as we removed it from Internal.


accessibilityLabel = 'Popover', >> LOST
  positionRelativeToAnchor = true, >> HARD coded
  color = 'white', >> HARD coded
  role = 'dialog', >> DEFAULT
  shouldFocus = true, >> DEFAULT
  showCaret = false, >> HARD coded
  size = 'sm', >> >> DEFAULT

packages/gestalt/src/PopoverEducational.js
eniomoura: DONE

packages/gestalt/src/PopoverEducational.js
eniomoura: DONE

PR #2869, opened 2023-04-21
TextArea: Add MaxHeight for Rows with Tags

rlingineni:
### Summary
Set max-height to rows when using Tags with TextArea

#### What changed?

Setting the maxHeight and minHeights to be the same in the TagsWrapper, so an overflow appears.

#### Why?
We have an issue with having too many tags, and not wrapping in the TextArea. This causes the div to grow past the given area. There may be a better design, though this adds consistency to the control in respecting the rows prop

### Links

- [Jira](https://jira.pinadmin.com/secure/RapidBoard.jspa?rapidView=1936&view=planning&selectedIssue=GESTALT-6019&quickFilter=12031&issueLimit=100)
- [TDD](link to Paper doc)
- [Figma](link to Figma file)

### Checklist
- [] Verified accessibility: keyboard & screen reader interaction

### DIFF
--- a/docs/pages/web/textarea.js
+++ b/docs/pages/web/textarea.js
@@ -577,7 +577,7 @@ function TextAreaPopoverExample() {
             defaultCode={`
 function Example(props) {
   const [value, setValue] = React.useState('');
-  const [tags, setTags] = React.useState(['San Francisco', 'New York']);
+  const [tags, setTags] = React.useState(Array(100).fill("New York", 0, 100));
   const ref = React.useRef();
 
   const onChangeTagManagement = ({ value }) => {
--- a/docs/pages/web/textarea.js
+++ b/docs/pages/web/textarea.js
@@ -645,7 +645,7 @@ function Example(props) {
         <MainSection.Subsection
           title="With rows"
           description={`
-  You can use the rows property to set the max height of the Textarea. It will begin to overflow after the rows limit has been reached`}
+          The rows prop sets the number of rows shown in TextArea. The input will show a scrollbar if the content exceeds the rows limit.`}
         >
           <MainSection.Card
             defaultCode={`          
@@ -655,17 +655,18 @@ const [value, setValue] = React.useState('');
 const [rows, setRows] = React.useState(2);
 
 return (
-  <Box width="100%">
-    <Box marginBottom={2}>
+  <Flex direction="column" width="100%" gap={4}>
+    <Box width={120}>
       <NumberField label="Number of Rows" onChange={({value})=>{setRows(value)}} value={rows}/>
     </Box>
     <TextArea
+      label="Rows example"
       onChange={({value})=>{setValue(value)}}
       placeholder={"this text area has " + rows + " rows"}
       value={value}
       rows={rows}
     />
-  </Box>
+  </Flex>
 );
 }
 `}
--- a/docs/pages/web/textarea.js
+++ b/docs/pages/web/textarea.js
@@ -577,7 +577,7 @@ function TextAreaPopoverExample() {
             defaultCode={`
 function Example(props) {
   const [value, setValue] = React.useState('');
-  const [tags, setTags] = React.useState(Array(100).fill("New York", 0, 100));
+  const [tags, setTags] = React.useState(['San Francisco', 'New York']);
   const ref = React.useRef();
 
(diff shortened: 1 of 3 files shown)


AlbertCarreras:
I think you might havr to update some snapshots. 

Also, can you create an example for rows in the Docs for both text and Tags.

Can you put a Before and After example in the description?


Questions, this is set to rows={2} 
Screenshot by Dropbox Capture

This is 3 rows right now 
Screenshot by Dropbox Capture


Do we want to show 2 full lines of Tags and a little bit of the third? How should we display them? Right now

rlingineni:
I've updated the snapshot.

Re: 2 lines and 3 lines, may actually be a bigger change, because everywhere this is used today, the row height with tags might grow in size. 

I think it should be okay to leave it as is, and the consumer can add an extra row if they want. 

I've added a `with rows` example to the docs. Let me know!

docs/pages/web/textarea.js
```diff
@@ -635,6 +638,36 @@ function Example(props) {
...
   );
 }
+`}
+          />
+        </MainSection.Subsection>
+
+        <MainSection.Subsection
+          title="With rows"
+          description={`
+  You can use the rows property to set the max height of the Textarea. It will begin to overflow after the rows limit has been reached`}
```
dangerismycat, line 648: This doesn't really capture that `rows` sets both the min and max height. Consider something like
"The `rows` prop sets the number of rows shown in `TextArea`. The input will show a scrollbar if the content exceeds the `rows` limit."

docs/pages/web/textarea.js
```diff
@@ -635,6 +638,36 @@ function Example(props) {
...
+const [rows, setRows] = React.useState(2);
+
+return (
+  <Box width="100%">
+    <Box marginBottom={2}>
+      <NumberField label="Number of Rows" onChange={({value})=>{setRows(value)}} value={rows}/>
+    </Box>
+    <TextArea
+      onChange={({value})=>{setValue(value)}}
+      placeholder={"this text area has " + rows + " rows"}
```
dangerismycat, line 664: This should use `label` so it's still visible if the user types something

docs/pages/web/textarea.js
```diff
@@ -635,6 +638,36 @@ function Example(props) {
...
+            defaultCode={`          
+function Example(props) {
+
+const [value, setValue] = React.useState('');
+const [rows, setRows] = React.useState(2);
+
+return (
+  <Box width="100%">
+    <Box marginBottom={2}>
+      <NumberField label="Number of Rows" onChange={({value})=>{setRows(value)}} value={rows}/>
```
dangerismycat, line 660: Screen Shot 2023-04-25 at 6 04 49 PM

The UI looks a bit goofy having this NumberField the full width of the TextArea. Consider limiting this width to something that would accommodate two digits or so.

docs/pages/web/textarea.js
```diff
@@ -635,6 +638,36 @@ function Example(props) {
...
+          <MainSection.Card
+            defaultCode={`          
+function Example(props) {
+
+const [value, setValue] = React.useState('');
+const [rows, setRows] = React.useState(2);
+
+return (
+  <Box width="100%">
+    <Box marginBottom={2}>
```
dangerismycat, line 659: nit: use Flex for the parent and set `gap={2}` instead of using this Box

dangerismycat (commented):
Reiterating Alberto's request for before/after images in the PR summary, please 🙂

packages/gestalt/src/TextArea.js
```diff
@@ -221,6 +221,7 @@ const TextAreaWithForwardRef: AbstractComponent<Props, HTMLTextAreaElement> = fo
 
   const tagsWrapperStyle = {
     minHeight: rows * ROW_HEIGHT + INPUT_PADDING_WITH_TAGS,
```
AlbertCarreras, line 223: I'm curious about why we need both minHeight & maxHeight being the same number, aint that just the same as height?

A follow up would be calculating the space that a row of tag + top/bottom margin takes and add ~1/3 of the next row. I think we should show rows + 1/3 of the next. But that's a design design too. We can iterate on that.

AlbertCarreras (approved):
R+, but read and address Ryan's comments

PR #2870, opened 2023-04-24
Link, Button, TapArea, IconButton: fix default label to target="blank" positioning issue on Tables

AlbertCarreras:
#### What changed?

Link, Button, TapArea, IconButton: fix default label to target="blank" positioning issue on Tables

Bug 
https://jira.pinadmin.com/browse/BUG-157543?jql=project%20%3D%20BUG%20AND%20resolution%20%3D%20Unresolved%20AND%20component%20%3D%20Gestalt%20%20ORDER%20BY%20%20priority%20DESC%2C%20updated%20DESC

#### Why?

### Links

- [TDD](link to Paper doc)
- [Figma](link to Figma file)

### Checklist


packages/gestalt/src/NewTabAccessibilityLabel.js
```diff
@@ -10,7 +10,9 @@ export default function NewTabAccessibilityLabel({
 |}): Node {
   const { accessibilityNewTabLabel } = useDefaultLabelContext('Link');
   return target === 'blank' ? (
-    <Box display="visuallyHidden">{`; ${accessibilityNewTabLabel}`}</Box>
+    <Box position="relative">
```
rlingineni, line 13: Can we add a test for this, similar to the demo, so if this pops up as an issue with table, we can catch this error in case it become a bug again?

rlingineni (approved):
Approving as a fix, but I think it would be good to add a test to table so we can catch if this happens again

packages/gestalt/src/NewTabAccessibilityLabel.js
AlbertCarreras: I'll add test in Follow up

PR #2871, opened 2023-04-25
PopoverEducational: changing `accessibilityLabel` prop to optional

eniomoura:
### Summary

accessibilityLabel is now not a required prop in PopoverEducational

#### Why?

It shouldn't have been required, given in many cases there is no context to define an appropriate specific accessibilityLabel, and a default is provided by the component.

### DIFF
--- a/packages/gestalt/src/PopoverEducational.js
+++ b/packages/gestalt/src/PopoverEducational.js
@@ -29,7 +29,7 @@ type PrimaryActionType = {|
 |};
 
 function PrimaryAction({
-  accessibilityLabel = 'Popover',
+  accessibilityLabel,
   href,
   text,
   onClick,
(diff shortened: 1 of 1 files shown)


packages/gestalt/src/PopoverEducational.js
```diff
@@ -29,7 +29,7 @@ type PrimaryActionType = {|
 |};
 
 function PrimaryAction({
-  accessibilityLabel,
+  accessibilityLabel = 'Popover',
```
AlbertCarreras, line 32: hehe not here!

AlbertCarreras: This is the button

AlbertCarreras: In line 123 we should see the default. Can you make sure you rebase?

Screenshot by Dropbox Capture

eniomoura: Argh, got distracted and ended up changing the wrong thing.
Will rebase and fix.

eniomoura: Done :)

AlbertCarreras (approved):

PR #2874, opened 2023-04-26
Internal: upgrade to Flow 0.192.0 

AlbertCarreras:
Internal: upgrade to Flow 0.192.0 

Screenshot by Dropbox Capture

### DIFF
--- a/docs/docs-components/clipboardCopy.js
+++ b/docs/docs-components/clipboardCopy.js
@@ -52,5 +52,6 @@ export default function clipboardCopy(text: string): Promise<void> {
 
   return success
     ? Promise.resolve()
-    : Promise.reject(new DOMException('The request is not allowed', 'NotAllowedError'));
+    : // $FlowIssue[cannot-resolve-name] DOMException is not part of the internal flow definitions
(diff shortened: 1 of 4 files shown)


docs/docs-components/clipboardCopy.js
```diff
@@ -52,5 +52,6 @@ export default function clipboardCopy(text: string): Promise<void> {
 
   return success
     ? Promise.resolve()
-    : Promise.reject(new DOMException('The request is not allowed', 'NotAllowedError'));
+    : // $FlowIssue[cannot-resolve-name] DOMException is not part of the internal flow definitions
```
AlbertCarreras, line 55: HERE

dangerismycat (approved):

docs/docs-components/clipboardCopy.js
jackhsu978: looks like widely supported https://developer.mozilla.org/en-US/docs/Web/API/DOMException/DOMException#browser_compatibility but not defined in https://github.com/facebook/flow/blob/main/lib/dom.js

`$FlowIssue` is accurate since this is a Flow internal issue.

LGTM

jackhsu978 (approved):
LGTM

PR #2879, opened 2023-04-27
Dropdown: implement adaptive Dropdown with SheetMobile for mobile devices

AlbertCarreras:
Dropdown: implement adaptive Dropdown with SheetMobile for mobile devices

Brave Browser - Dropdown - Gestalt 2023-04-27 at 11 58 53 PM

### DIFF
--- a/packages/gestalt/src/Dropdown.js
+++ b/packages/gestalt/src/Dropdown.js
@@ -8,7 +8,10 @@ import { ESCAPE, SPACE, TAB, ENTER, UP_ARROW, DOWN_ARROW } from './keyCodes.js';
 import Layer from './Layer.js';
 import Popover from './Popover.js';
 import { type Indexable } from './zIndex.js';
+import AnimationProvider from './animation/AnimationContext.js';
+import { useDeviceType } from './contexts/DeviceTypeProvider.js';
 import { DropdownContextProvider } from './Dropdown/Context.js';
+import PartialPage from './SheetMobile/PartialPage.js';
 import { type DirectionOptionType } from './utils/keyboardNavigation.js';
 
 const KEYS = {
@@ -123,6 +126,11 @@ type Props = {|
    */
   maxHeight?: '30vh',
   /**
+   * Mobile-only prop. Callback fired when Dropdown's in & out animations end. See the [mobile variant](https://gestalt.pinterest.systems/web/dropdown#mobile) to learn more.
+   */
+  mobileOnAnimationEnd?: ({| animationState: 'in' | 'out' |}) => void,
+
+  /**
    * Callback fired when the menu is closed.
    */
   onDismiss: () => void,
@@ -149,8 +157,12 @@ export default function Dropdown({
   onDismiss,
   zIndex,
   maxHeight,
+  mobileOnAnimationEnd,
 }: Props): Node {
-  const [hoveredItem, setHoveredItem] = useState<number>(0);
+  const deviceType = useDeviceType();
+  const isMobile = deviceType === 'mobile';
+
+  const [hoveredItem, setHoveredItem] = useState<?number>(isMobile ? undefined : 0);
 
   const dropdownChildrenArray = Children.toArray(children);
   const allowedChildrenOptions = getChildrenOptions(dropdownChildrenArray);
@@ -170,7 +182,7 @@ export default function Dropdown({
     event: SyntheticKeyboardEvent<HTMLElement>,
     direction: DirectionOptionType,
   ) => {
-    const newIndex = direction + hoveredItem;
+    const newIndex = direction + (hoveredItem ?? 0);
     const optionsCount = allowedChildrenOptions.length - 1;
 
     // If there's an existing item, navigate from that position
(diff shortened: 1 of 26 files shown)


packages/gestalt/src/Dropdown.js
```diff
@@ -122,10 +125,18 @@ type Props = {|
...
+   */
+  mobileOnAnimationEnd?: ({| animationState: 'in' | 'out' |}) => void,
   /**
    * Callback fired when the menu is closed.
    */
   onDismiss: () => void,
+  /**
+   * Dropdown can adapt to mobile devices to [SheetMobile](https://gestalt.pinterest.systems/web/sheetmobile). Mobile adaptation is not enabled by default. Set to 'sheetMobile' to enable SheetMobile in mobile devices. See the [mobile variant](https://gestalt.pinterest.systems/web/dropdown#mobile) to learn more.
+   */
+  mobileAdaptation?: 'default' | 'sheetMobile',
```
dangerismycat, line 139: Is the plan to eventually make `sheetMobile` the default? If so, we may want to make this something like `disableMobileUI` and codemod all current usages to include the prop, such that any new usages won't have it — therefore encouraging new Dropdown usages to embrace the adaptive UI.

packages/gestalt/src/Dropdown.js
```diff
@@ -149,8 +160,13 @@ export default function Dropdown({
...
   zIndex,
   maxHeight,
+  mobileOnAnimationEnd,
+  mobileAdaptation = 'default',
 }: Props): Node {
-  const [hoveredItem, setHoveredItem] = useState<number>(0);
+  const deviceType = useDeviceType();
+  const isMobile = deviceType === 'mobile';
+
+  const [hoveredItem, setHoveredItem] = useState<?number>(isMobile ? undefined : 0);
```
dangerismycat, line 169: nit: this is the index of the hovered item, yes? If so, consider `hoveredItemIndex` to be more specific. We could also use an id or the item reference itself, so clarifying helps readability.

packages/gestalt/src/DropdownLink.js
```diff
@@ -41,6 +41,7 @@ type Props = {|
   onClick?: ({|
     event: SyntheticMouseEvent<HTMLAnchorElement> | SyntheticKeyboardEvent<HTMLAnchorElement>,
     dangerouslyDisableOnNavigation: () => void,
+    mobileOnDismissStart: () => void,
```
dangerismycat, line 44: Can you add documentation about this new field?

packages/gestalt/src/SheetMobile.js
```diff
@@ -104,6 +105,10 @@ type Props = {|
    * Sets the SheetMobile's height. See the [size variant](https://gestalt.pinterest.systems/web/sheetmobile#Size) for more info.
    */
   size?: 'default' | 'full' | 'auto',
+  /**
+   * An object representing the zIndex value of the Dropdown menu. Learn more about [zIndex classes](https://gestalt.pinterest.systems/web/zindex_classes)
```
dangerismycat, line 109: This should reference SheetMobile and not Dropdown, yes?

AlbertCarreras: DONE

packages/gestalt/src/DropdownLink.js
AlbertCarreras: DONE

packages/gestalt/src/Dropdown.js
AlbertCarreras: DONE

packages/gestalt/src/Dropdown.js
AlbertCarreras: DONE

Keeping disabled, I'll run codemod next and invert default as a follow up. less blocking PR

packages/gestalt/src/Dropdown.js
```diff
@@ -149,8 +160,13 @@ export default function Dropdown({
   onDismiss,
   zIndex,
   maxHeight,
+  mobileOnAnimationEnd,
+  disableMobileUI = true,
```
dangerismycat, line 164: This is an anti-pattern, leading to a confusing DX. For example, which of these yields a disabled mobile UI?
```js
<Dropdown />
<Dropdown disableMobileUI />
<Dropdown disableMobileUI={false} />
```
The first two both do, but how is a developer to know that the first example would disable the mobile UI?

We also plan to make this behavior the default in the future, requiring another breaking change in the future.

We can avoid that future breaking change and encourage all new usages of Dropdown to use the adaptive UI by setting this to our target default of `disableMobileUI = false`, and codemodding all current Dropdown usages.

dangerismycat (changes requested):

rlingineni:
I can also take a look at this tomorrow, since it seems like a big PR

packages/gestalt/src/Dropdown.js
AlbertCarreras: Ryan, 

In my comment above:

Screenshot 2023-05-03 at 9 27 29 PM

I'm going to run the codemod, independently from the PR, set all current instances to true, then set default to false

@dangerismycat

dangerismycat (approved):

PR #2909, opened 2023-05-08
Internal: add test for Masonry scroll performance

dangerismycat:
To speed up the feedback on future Masonry algorithm changes, we'd like to have internal scroll performance test(s) to give us early feedback on the viability of solutions. This PR adds a simple MVP perf test, which we may improve in the future.

Known shortcomings:
- This hits localhost, not a permanent test URL
- This does not persist timings/deltas, which will need to be tracked manually

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -30,6 +30,7 @@ module.exports = {
     'react',
     'react-hooks',
     'testing-library',
+    'ui-testing',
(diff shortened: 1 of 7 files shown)


dangerismycat:
No time to work on this PR now. Closing, might reopen in the future if we get more bandwidth.

PR #2919, opened 2023-05-10
Dropdown: changed Dropdown.Link icon to "visit"

ponori:
## What
Dropdown: changed Dropdown.Link icon from `arrow-up-right` to `visit`

### BEFORE
Screen Shot 2023-05-10 at 11 10 46 AM

### AFTER
Screen Shot 2023-05-10 at 11 12 00 AM

PR #2927, opened 2023-05-10
Dropdown: fixes visual snapshots after [#2919]

AlbertCarreras:
Dropdown: fixes visual snapshots after 2919


dangerismycat (approved):
thank you! 🙂

PR #2942, opened 2023-05-15
Internal: fixes to pass ` flow codemod annotate-implicit-instantiations --write --include-widened .` & `flow codemod annotate-implicit-instantiations --write .`

AlbertCarreras:
Internal: fixes to pass ` flow codemod annotate-implicit-instantiations --write --include-widened .` & `flow codemod annotate-implicit-instantiations --write .`

Prepping codebase to upgrade to Flow v 0.203.1

Annotate all parameters that are only required to be annotated in LTI.
- [x] flow codemod annotate-functions-and-classes --include-lti --write .
Annotate declarations to help unbreak type cycles.
- [x] flow codemod annotate-cycles --write .
Annotate underconstrained react hooks like useState(null).
- [x] flow codemod annotate-react-hooks --write .
Annotate underconstrained empty arrays.
- [x] flow codemod annotate-empty-array --write .
Annotate under-constrained type arguments.
- [x] flow codemod annotate-implicit-instantiations --write .
Annotate type arguments that might be widened. 
This is a noisy codemod that doesn't always produce what you might want.
- [x] flow codemod annotate-implicit-instantiations --write --include-widened .
Same as above, but it will annotate function returns like 
`array.map((v): T => {...})` instead of `array.map<T, _>(v => {...})`
- [x]  flow codemod annotate-implicit-instantiations --write --include-widened --annotate-special-function-return .

From https://medium.com/flow-type/local-type-inference-for-flow-aaa65d071347

Screenshot by Dropbox Capture

### DIFF
--- a/docs/pages/integration-test/masonry.js
+++ b/docs/pages/integration-test/masonry.js
@@ -13,7 +13,7 @@ import MasonryContainer from '../../integration-test-helpers/masonry/MasonryCont
 // This can get bumped up another order of magnitude or so if needed…perf drops off pretty rapidly after that
 const REALISTIC_PINS_DATASET_SIZE = 1000;
 
-const measurementStore = Masonry.createMeasurementStore();
+const measurementStore = Masonry.createMeasurementStore<{ ... }, mixed>();
 
 // This is the counterpart to `normalizeValue` in `playwright/masonry/utils/getServerURL.mjs`
 function booleanize(value: string): boolean {
--- a/packages/gestalt/src/Video/Controls.js
+++ b/packages/gestalt/src/Video/Controls.js
@@ -131,18 +131,18 @@ function VideoControls({
 
   const muted = volume === 0;
 
-  const [showFullscreenButton, setShowFullscreenButton] = useState(false);
+  const [showFullscreenButton, setShowFullscreenButton] = useState<boolean>(false);
 
   useEffect(() => {
     setShowFullscreenButton(
       typeof document !== 'undefined' &&
-        (document.fullscreenEnabled ||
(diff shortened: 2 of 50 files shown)


docs/pages/integration-test/masonry.js
```diff
@@ -13,7 +13,7 @@ import MasonryContainer from '../../integration-test-helpers/masonry/MasonryCont
 // This can get bumped up another order of magnitude or so if needed…perf drops off pretty rapidly after that
 const REALISTIC_PINS_DATASET_SIZE = 1000;
 
-const measurementStore = Masonry.createMeasurementStore();
+const measurementStore = Masonry.createMeasurementStore<{ ... }, mixed>();
```
AlbertCarreras, line 16: any thoughts here? @dangerismycat

packages/gestalt/src/Video/Controls.js
```diff
@@ -131,18 +131,18 @@ function VideoControls({
...
   const muted = volume === 0;
 
-  const [showFullscreenButton, setShowFullscreenButton] = useState(false);
+  const [showFullscreenButton, setShowFullscreenButton] = useState<boolean>(false);
 
   useEffect(() => {
     setShowFullscreenButton(
       typeof document !== 'undefined' &&
-        (document.fullscreenEnabled ||
+        (!!document.fullscreenEnabled ||
```
dangerismycat, line 139: Since the `document` checks are already wrapped in parens, you should be able to coerce the entire statement once instead of each `document` check

dangerismycat (approved):

PR #2948, opened 2023-05-17
Internal: Unminify classname when developing

rlingineni:
When rollup is run watch mode (our development version), the outputted css will also include the full classname to make it easier for debugging in the inspect tab.

**Why?**
When building components, it's hard to track the minified css or know which class to modify. This can also encourage behavior to to write new classes instead of extending current ones. The code may write the style class like this `styles.button`, but when debugging, it's up to the developer to associate that with `DUt` as a className

What changed:
Classnames when in development are run as `fullClassName___minifiedClassName`

**BEFORE**
image

**AFTER**
image

### DIFF
--- a/packages/gestalt-core/build.js
+++ b/packages/gestalt-core/build.js
@@ -57,7 +57,14 @@ const cssModules = (options = {}) => {
         scopeNames[hash] = classnameBuilder.getMinifiedClassname(hash);
       }
 
-      return scopeNames[hash];
+      // if it's dev mode, also append the full class name for debugging
+      const isDevMode = process.argv.includes('-w') || process.argv.includes('--watch');
+      let minifiedName = scopeNames[hash];
+      if (isDevMode) {
+        minifiedName = `${name}__${minifiedName}`;
+      }
+
+      return minifiedName;
     },
     getJSON: (filePath, exportTokens) => {
       Object.entries(exportTokens).forEach(([className, value]) => {
--- a/packages/gestalt-core/build.js
+++ b/packages/gestalt-core/build.js
@@ -57,14 +57,11 @@ const cssModules = (options = {}) => {
         scopeNames[hash] = classnameBuilder.getMinifiedClassname(hash);
       }
 
-      // if the DEVMODE environment variable is set, then don't minify the classnames
-      const isDevMode = process.env.DEVMODE === 'true';
-      let minifiedName = scopeNames[hash];
-      if (isDevMode) {
-        minifiedName = `${name}__${minifiedName}`;
-      }
+      // if it's not production, use the minfied name + extended name
+      const className =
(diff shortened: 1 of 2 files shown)


packages/gestalt-core/build.js
```diff
@@ -57,7 +57,14 @@ const cssModules = (options = {}) => {
         scopeNames[hash] = classnameBuilder.getMinifiedClassname(hash);
       }
 
-      return scopeNames[hash];
+      // if it's dev mode, also append the full class name for debugging
+      const isDevMode = process.argv.includes('-w') || process.argv.includes('--watch');
```
AlbertCarreras, line 61: can you explain / add a comment for this?

process.argv.includes('-w') || process.argv.includes('--watch');

AlbertCarreras: @rlingineni

packages/gestalt-core/build.js
```diff
@@ -57,7 +57,14 @@ const cssModules = (options = {}) => {
         scopeNames[hash] = classnameBuilder.getMinifiedClassname(hash);
       }
 
-      return scopeNames[hash];
+      // if the DEVMODE environment variable is set, then don't minify the classnames
+      const isDevMode = process.env.DEVMODE === 'true';
+      let minifiedName = scopeNames[hash];
```
dangerismycat, line 62: In general, if a variable only changes on a single conditional, it's best to use a ternary so we can use `const`. That prevents the variable from being reassigned later.

packages/gestalt-core/build.js
```diff
@@ -57,7 +57,14 @@ const cssModules = (options = {}) => {
         scopeNames[hash] = classnameBuilder.getMinifiedClassname(hash);
       }
 
-      return scopeNames[hash];
+      // if the DEVMODE environment variable is set, then don't minify the classnames
+      const isDevMode = process.env.DEVMODE === 'true';
```
dangerismycat, line 61: We already have `NODE_ENV`. Is there a reason to introduce another `env` field?

packages/gestalt-core/build.js
rlingineni: Ah, I did this tweak, and changing the NODE_ENV variable causes things to fail. Not sure why. It may be being somewhere internally. So it seems a tad more complex to debug. See error below:

I kept the DEVMODE env variable, but changed to a ternary for simplicity

image

packages/gestalt-core/build.js
rlingineni: Ah, re this: tweaking the NODE_ENV seems to cause a failure somewhere else. And it's not entirely obvious why

image

rlingineni (commented):
Made a few changes, lmk!

dangerismycat (approved):

PR #2952, opened 2023-05-19
Table: added "selected"and "hoverStyle" prop  to all Table.Row subcomponents

AlbertCarreras:
#### What changed?

Table: added "selected"and "hoverStyle" prop to all Table.Row subcomponents

Brave Browser - Table - Gestalt 2023-05-19 at 4 52 02 AM

### Links

- [Jira](https://jira.pinadmin.com/browse/GESTALT-4697)
- [Figma](https://www.figma.com/file/pGpBsKBAXMwl2bv1dMISfs/Documentation-Table-Selected-Row?type=design&node-id=1%3A1956&t=J4YninrFn7P6Wzyb-1)

### Checklist

### DIFF
--- a/docs/examples/table/selected.js
+++ b/docs/examples/table/selected.js
@@ -1,0 +1,136 @@
+// @flow strict
+import { useState, type Node } from 'react';
+import { Table, Checkbox, Flex, Text, Box } from 'gestalt';
+
+export default function Example(): Node {
+  const [selected, setSelectedRow] = useState<$ReadOnlyArray<string>>([]);
+
+  return (
+    <Box height="100%" width="100%" overflow="scroll">
+      <Flex alignItems="start" gap={4} height="100%" justifyContent="center" width="100%">
+        <Table accessibilityLabel="Selected row example table">
+          <Table.Header>
+            <Table.Row>
+              <Table.HeaderCell>
+                <Text />
+              </Table.HeaderCell>
+              <Table.HeaderCell>
+                <Text weight="bold">Selected</Text>
+              </Table.HeaderCell>
+              <Table.HeaderCell>
+                <Text weight="bold">Column</Text>
+              </Table.HeaderCell>
+              <Table.HeaderCell>
+                <Text weight="bold">Column</Text>
+              </Table.HeaderCell>
+              <Table.HeaderCell>
+                <Text weight="bold">Subcomponent</Text>
+              </Table.HeaderCell>
+            </Table.Row>
+          </Table.Header>
+          <Table.Body>
+            <Table.Row
+              key="row1"
+              hoverStyle="gray"
+              selected={selected.includes('row') ? 'selected' : 'unselected'}
+            >
+              <Table.Cell>
+                <Text />
+              </Table.Cell>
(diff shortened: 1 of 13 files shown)


docs/pages/web/table.js
```diff
@@ -2459,6 +2460,19 @@ function Example() {
 `}
           />
         </MainSection.Subsection>
+        <MainSection.Subsection
+          title="Selected & hovered state"
+          description={`Table.Row, Table.RowExpandable and Table.RowDrawer support hovered and selected states.
+
+If a row subcomponent is selectable, toggle the \`selected\` prop between "selected" and "unselected" to keep a constant border space in the row that is only visible when the row is selected.
+
+The \`selected\` prop should not be set, or set to "undefined", only if the row is not selectable. In this case, it does not set a border.
```
dangerismycat, line 2469: grammar nit: the first sentence here is a little awkward. Consider switching the clauses, so starting with "If the row is not selectable…" and then the rest.

packages/gestalt/dist/index.d.ts
```diff
@@ -1729,23 +1729,28 @@ interface TableSortableHeaderCellProps {
...
 
 interface TableRowExpandableProps {
   accessibilityCollapseLabel: string;
   accessibilityExpandLabel: string;
   children: Node;
   expandedContents: Node;
-  id: string;
   expanded?: string | undefined;
-  hoverStyle?: 'none' | 'gray' | undefined;
+  hoverStyle?: 'gray' | 'none' | undefined;
```
dangerismycat, line 1742: Interesting…is this correct? It's different from the Flow type. Flow is saying "this prop is optional, but if it's defined, it will be this enum of values". TS is saying "this prop is optional, but if it's defined, it will be this enum of values or `undefined`". The TS type seems to give additional flexibility we're not intending.

docs/pages/web/table.js
dangerismycat: Also, this recommendation doesn't match the Flow type. If we do want devs to be able to pass `undefined`, we need to add that to the enum of acceptable values.

dangerismycat (approved):

docs/examples/table/selected.js
```diff
@@ -0,0 +1,126 @@
+// @flow strict
+import { useState, type Node } from 'react';
+import { Table, Checkbox, Flex, Text, Box } from 'gestalt';
+
+export default function Example(): Node {
+  const [selectedRow, setSelectedRow] = useState('unselected');
```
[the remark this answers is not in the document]
AlbertCarreras: DONE

docs/pages/web/table.js
AlbertCarreras: DONE

packages/gestalt/dist/index.d.ts
AlbertCarreras: Good catch. I'll leave because I kept that pattern of optional props having undefined value, but I realized later that they are unnecessary. Technically optional, if not passed, is undefined and defaults props override undefined.

I'll clean up this issue in a follow up but for all types.

PR #2955, opened 2023-05-19
Masonry: add experimental support for two-column items

dangerismycat:
### Summary

This PR introduces a proof-of-concept to support rendering 2-column modules in the Masonry grid. See the [project doc](https://docs.google.com/document/d/1w7E3LvaWXPkMto9RjVUSOR333rUlxgyItiCFiifj8EE/edit) for a comprehensive summary of the project.

This PR adds an undocumented experimental prop to Masonry. When this prop is `true` (and using `layout` is `"basic"` or `"basicCentered"`, Masonry will use a new layout algorithm. When a batch of pins to be laid out contains a 2-col module, we switch from the classic layout algorithm ("put the item in the shortest column") to a graph-based approach. This approach takes inspiration from the seminal [Knuth+Plass text layout algorithm for TeX](https://defoe.sourceforge.net/folio/knuth-plass.html) to explore every possible layout using the 1-col items in the batch, looking for the layout that will yield two adjacent columns of very similar height. The 2-col module will be inserted in this location, and the remaining 1-col items will be laid out normally.

A major conceptual change for this is that when using the new algorithm, we cache the column heights array and item positions (by item data reference). In the classic algorithm, we throw away heights and positions on each render and lay out the entire grid from scratch each time. This depends on a major assumption: that grid items will always be laid out in the order given. Since we now break that assumption in the batch of items containing the 2-col module we need to be able to directly associate position with a specific item, not just an `items` index.

This PR does not yet implement pin leveling to minimize whitespace above the 2-col module. That technique may not be necessary for an MVP for this project — and if it is, it can be implemented in a future PR.

### Testing these changes

Since this is an experimental POC, the changes are not represented on the Masonry docs page. Here's how to test locally:
- pull down this branch locally
- run `yarn start`
- visit `http://localhost:8888/integration-test/masonry?realisticPinHeights=1&twoColItems=1` (`realisticPinHeights` is optional)

### Demo
https://github.com/pinterest/gestalt/assets/12059539/142c6338-23d3-4ade-be1a-0801b08c73d3

_A note about this demo: this is using the "realistic pin heights" option introduced back in #2823. Though based on actual pin data, the "realism" here is questionable. There seems to be a lot more variation in heights than I've ever seen in a grid. So the whitespace above 2-col modules is possibly (probably?) worse than it will be in a "real world" grid._


packages/gestalt/src/Masonry/defaultLayout.js
```diff
@@ -1,7 +1,13 @@
 // @flow strict
 import { type Cache } from './Cache.js';
```
liuyenwei, line 2: i was thinking about this and we should probably fork this into a separate layout (at least until after the experiment):
- makes it easier to run the experiment (i.e. if experiment + two column, then use defaultWithTwoColumnLayout)
- ensures we don't make any unexpected changes to the existing default Layout

dangerismycat: Yeah, that makes sense. I've been thinking about if this should rely on just the presence of 2-col modules or if there should also be an experiment prop…the experiment prop isn't technically necessary but probably safer, and does make forking this layout easier.

packages/gestalt/src/Masonry/defaultTwoColumnModuleLayout.js
```diff
@@ -0,0 +1,365 @@
...
+      const { heights: finalHeights, position: twoColItemPosition } = getTwoColItemPosition<T>({
+        item: twoColItem,
+        heights: lowestScoreNode.heights,
+        ...commonGetPositionArgs,
+      });
+
+      // Combine winning positions and 2-col item position, add to cache
+      const finalPositions = [
+        ...winningPositions,
+        { item: twoColItem, position: twoColItemPosition },
```
liuyenwei, line 337: thinking ahead, once we integrate this into Pinboard it would be good to somehow have access to this so we can log data on average number of whitespace we end up with per pass.  maybe something like a callback prop could help with that e.g. `onTwoColumnLayout(position, heights)`

packages/gestalt/src/Masonry.js
```diff
@@ -5,19 +5,24 @@ import FetchItems from './FetchItems.js';
...
+import { type Position } from './Masonry/types.js';
 import uniformRowLayout from './Masonry/uniformRowLayout.js';
 import throttle, { type ThrottleReturn } from './throttle.js';
 
 const RESIZE_DEBOUNCE = 300;
 
-const layoutNumberToCssDimension = (n: ?number) => (n !== Infinity ? n : undefined);
+// When there's a 2-col item in the most recently fetched batch of items, we need to measure more items to ensure we have enough possible layouts to find a suitable one
+// The number of items measured at a time is the number of columns * this multiplier
+const TWO_COL_ITEMS_MEASURE_BATCH_SIZE = 6;
```
liuyenwei, line 23: maybe a comment around why this is static and why we're doing 6

packages/gestalt/src/Masonry.js
```diff
@@ -64,12 +69,15 @@ type Props<T> = {|
...
-   * A function that renders the item to be displayed in the grid. This function is passed three props: the item's data, the item's index in the grid, and a flag indicating if Masonry is currently measuring the item.
+   * A function that renders the item you would like displayed in the grid. This function is passed three props: the item's data, the item's index in the grid, and a flag indicating if Masonry is currently measuring the item.
+   *
+   * If present, `heightAdjustment` indicates the number of pixels this item needs to grow/shrink to accommodate a 2-column item in the grid. Items must respond to this prop by adjusting their height or layout issues will occur.
    */
   renderItem: ({|
     +data: T,
     +itemIdx: number,
     +isMeasuring: boolean,
+    +heightAdjustment?: number,
```
liuyenwei, line 80: do we need this atm?

packages/gestalt/src/Masonry.js
```diff
@@ -531,12 +572,19 @@ export default class Masonry<T: { ... }> extends ReactComponent<Props<T>, State<
       gridBody = <div style={{ width: '100%' }} ref={this.setGridWrapperRef} />;
     } else {
       // Full layout is possible
-      const itemsToRender = items.filter((item) => item && measurementStore.has(item));
+      const itemsWithMeasurements = items.filter((item) => item && measurementStore.has(item));
+      const itemsWithPositions = items.filter((item) => item && positionStore.has(item));
+
+      const itemsWithoutPositions = items.filter((item) => item && !positionStore.has(item));
+      // $FlowFixMe[prop-missing] We're assuming `columnSpan` exists
+      const hasTwoColumnItems = itemsWithoutPositions.some((item) => item.columnSpan === 2);
```
liuyenwei, line 580: `const hasTwoColumnItems = _twoColItems && itemsWithout....`?

liuyenwei (approved):

packages/gestalt/src/Masonry.js
dangerismycat: Sorry, I'd left a comment explaining and then forgot to submit 🤦 We don't need it yet, so it could be removed. This will be used for future pin leveling work.

packages/gestalt/src/Masonry.js
dangerismycat: Ah good call, I didn't update that comment

packages/gestalt/src/Masonry.js
dangerismycat: Good catch!

packages/gestalt/src/Masonry/defaultTwoColumnModuleLayout.js
dangerismycat: Yeah I was thinking about that as well. I think I'll add that in a future PR so we can start playing around with these changes in Pinboard

PR #2959, opened 2023-05-23
GlobalEventsHandlerProvider: new provider to share global event handlers to consuming components

AlbertCarreras:
#### What changed?

GlobalEventsHandlerProvider: new provider to share global event handlers to consuming components #2959

Brave Browser - HandlerProvider - Gestalt 2023-05-23 at 10 52 54 AM

#### Why?

Read RFC above 

### Links

- [TDD](link to Paper doc)
- [Figma](link to Figma file)

### Checklist


rlingineni:
Hi, this is great!

It seems like it's up to Pinboard subscribers to individually define each element. I wasn't sure if the plan is to have this be globally added to Pinboard at the top-level, or re-used across different places. Are things getting overrided a possible concern? 

Like is this a valid case?
```
// app.js
<MyApp>
<GlobalEventsHandlerProvider button={} sheetMobile={// maybe not called } >
    <Button>
    <GlobalEventsHandlerProvider sheetMobile={} >
      <SheetMobile>
      <Button>
  <GlobalEventsHandlerProvider>
<GlobalEventsHandlerProvider>
<MyApp/>
```

It does seem like for every component, we'd have to go in and add it as a potential [prop](https://deploy-preview-2959--gestalt.netlify.app/web/utilities/globaleventshandlerprovider#Props) and keep it up to date with our events getting changed. 


What if, and not completely sure of the exact trade offs, but if the event emitting was more generic like an `onEvent` handler and not tuned by component? The caller can then handle it based on their needs. In the future, we could also expand to emit non interactive events (e.g time to paint, or mount) as part of that emitter. 

```
<MyApp>
<GlobalEventsHandlerProvider onEvent={} >
    <Button>
    <GlobalEventsHandlerProvider onEvent={} >
      <SheetMobile>
      <Button>
  <GlobalEventsHandlerProvider>
<GlobalEventsHandlerProvider>
<MyApp/>
```

And on Event could give everything like:
```js
const onEvent = (event:string, sender:name, message?:any)  => {
 console.log(`A gestalt event happened for ${sender}`)
}
```
And in all the components we call something like

```
emit('onClick', 'TextField', {value:'H'})
emit('onDismiss', 'SheetMobile', {value:'H'})
```

docs/examples/globaleventshandlerprovider/sheetMobileHandlers.js
```diff
@@ -0,0 +1,56 @@
...
+
+export default function Example(): Node {
+  const [showComponent, setShowComponent] = useState<null | boolean>(true);
+
+  const PAGE_HEADER_ZINDEX: FixedZIndex = new FixedZIndex(10);
+  const ABOVE_PAGE_HEADER_ZINDEX: CompositeZIndex = new CompositeZIndex([PAGE_HEADER_ZINDEX]);
+
+  return (
+    <GlobalEventsHandlerProvider
+      sheetMobile={{
```
dangerismycat, line 23: Should we demonstrate proper Provider value best practices in the examples, by memoizing the passed-down objects and contained callbacks?
```js
const sheetMobileOnOpen = useCallback(() => console.log('On open handler'), []);
const sheetMobileOnClose = useCallback(() => console.log('On close handler'), []);
const sheetMobileHandlers = useMemo(() => ({ onOpen: sheetMobileOnOpen, onClose: sheetMobileOnClose }), []);

...

<GlobalEventsHandlerProvider sheetMobile={sheetMobileHandlers}>
```
const sheetMobileHandlers =

docs/pages/web/sheetmobile.js
```diff
@@ -327,6 +328,27 @@ When using these render props, just pass the argument \`onDismissStart\` to your
             }
           />
         </MainSection.Subsection>
+
+        <MainSection.Subsection
+          title="External handlers"
+          description={`SheetMobile consumes external handlers from GlobalEventsHandlerProvider.
+
+Handlers:
+- onOpen: executed on useEffect
```
dangerismycat, line 337: "executed on useEffect" doesn't really mean anything. An effect will get run once when the component mounts, but then also again any time its dependencies change. I wouldn't get caught up on the use of `useEffect` here — that's really an implementation detail. What's pertinent to the dev is that `onOpen` will run when the sheet opens, and that `onClose` will run just before the sheet closes. Whether it uses `useEffect` or some other mechanism is irrelevant — and really, if the developer needs to be be aware of that mechanism, we have a leaky abstraction.

docs/pages/web/sheetmobile.js
```diff
@@ -327,6 +328,27 @@ When using these render props, just pass the argument \`onDismissStart\` to your
...
+- onClose: executed on the cleanup function of useEffect
+
+See [GlobalEventsHandlerProvider](/web/utilities/GlobalEventsHandlerProvider) for more information.`}
```
dangerismycat, line 340: This URL isn't correct — component names in URLs are always all lowercase

packages/gestalt/src/SheetMobile/FullPage.js
```diff
@@ -61,8 +62,25 @@ export default function FullPage({
...
+  // Consumes DefaultLabelProvider
   const { accessibilityLabel: defaultAccessibilityLabel } = useDefaultLabelContext('SheetMobile');
 
+  // Consumes GlobalEventsHandlerProvider
+  const { sheetMobile } = useGlobalEventsHandlerContext() ?? {
+    sheetMobile: { onOpen: () => {}, onClose: () => {} },
+  };
+
+  useEffect(() => {
+    sheetMobile?.onOpen?.();
```
dangerismycat, line 76: This reads a little wonky. Consider renaming the variable from useGlobalEventsHandlerContext to something like "sheetMobileHandlers" so this reads more like `sheetMobileHandlers?.onOpen?.()`

packages/gestalt/src/SheetMobile/FullPage.js
```diff
@@ -61,8 +62,25 @@ export default function FullPage({
...
+
+    return function cleanup() {
+      sheetMobile?.onClose?.();
+    };
+  }, [sheetMobile]);
```
dangerismycat, line 81: Passing an object to `useEffect` isn't great, as the object reference can stay the same while its properties change, and the effect won't run again. We should destructure the handlers out of `sheetMobileHandlers` and use those directly in the dependency array.

packages/gestalt/src/SheetMobile/PartialPage.js
```diff
@@ -85,10 +86,25 @@ export default function PartialPage({
...
+    sheetMobile: { onOpen: () => {}, onClose: () => {} },
+  };
+
+  useEffect(() => {
+    sheetMobile?.onOpen?.();
+
+    return function cleanup() {
+      sheetMobile?.onClose?.();
+    };
+  }, [sheetMobile]);
```
dangerismycat, line 105: ditto here on renaming and destructuring

packages/gestalt/src/contexts/GlobalEventsHandlerProvider.js
```diff
@@ -0,0 +1,40 @@
+// @flow strict
+import { createContext, type Context, type Element, type Node, useContext } from 'react';
+
+export type NoopType = () => void;
+
+type SheetMobileHandlersType = {|
```
dangerismycat, line 6: nit: we're not exporting this type or using it elsewhere in this file, so I'd just inline it in GlobalEventsHandlerContextType

playwright/accessibility/web__utilities__globaleventshandlerprovider.spec.mjs
```diff
@@ -0,0 +1,8 @@
+// @flow strict
+import { test } from '@playwright/test';
+import expectAccessiblePage from './expectAccessiblePage.mjs';
+
+test('GlobalEventsHandlerProvider Accessibility check', async ({ page }) => {
+  await page.goto('/web/utilities/GlobalEventsHandlerProvider');
```
dangerismycat, line 6: This URL isn't correct

docs/pages/web/utilities/globaleventshandlerprovider.js
```diff
@@ -0,0 +1,53 @@
...
+        type="utility"
+      />
+
+      <GeneratedPropTable generatedDocGen={generatedDocGen} />
+
+      <MainSection name="Variants">
+        <MainSection.Subsection title="GlobalEventsHandlerProvider in SheetMobile">
+          <MainSection.Card
+            cardSize="lg"
+            description={`GlobalEventsHandlerProvider has one prop for each component subscribing to the provider.
```
dangerismycat, line 26: This copy needs to be updated. It still references the previous Provider name, and is a bit vague about if this affects all usages of a single component or if it needs to be implemented at the component callsite. Consider something like
"GlobalEventsHandlerProvider is a React Context Provider that is intended to be implemented at the top level of your app or feature. Each prop corresponds to a specific Gestalt component, and allows you to add logic that will run for every instance of that component in the component tree.

For SheetMobile, we can add `onOpen` and `onClose` handlers. `onOpen` runs on component mount, when the SheetMobile first appears in the UI. `onClose` runs just before the component unmounts and is removed from the UI."

dangerismycat (approved):

docs/docs-components/COMPONENT_DATA.js
```diff
@@ -3976,6 +3977,59 @@ const UTILITIES_LIST: $ReadOnlyArray<ListItemType> = [
       figma: null,
     },
   },
+  {
+    svg: <ProviderHandlers />,
+    name: 'GlobalEventsHandlerProvider',
+    aliases: ['React.useEffect'],
```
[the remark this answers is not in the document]
AlbertCarreras: NO, good ctch

rfcs/rfcs/2023-05-effectsProvider-new-component/README.md
```diff
@@ -0,0 +1,335 @@
+- Start Date: May 2023
+- RFC PR: https://github.com/pinterest/gestalt/pull/2918
+- Authors: Alberto Carreras, acarreras
+
+# (RFC title goes here)
+
+## Summary
+
+EffectsProvider is a React.Context provider intended to share React effects (useEffect logic) with consuming components.
```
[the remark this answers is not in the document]
AlbertCarreras: yes

docs/docs-components/COMPONENT_DATA.js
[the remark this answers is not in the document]
AlbertCarreras: DONE

docs/pages/web/sheetmobile.js
AlbertCarreras: DONE

docs/pages/web/utilities/globaleventshandlerprovider.js
AlbertCarreras: DONE

packages/gestalt/src/SheetMobile/FullPage.js
AlbertCarreras: DONE

packages/gestalt/src/SheetMobile/FullPage.js
AlbertCarreras: DONE

PR #2960, opened 2023-05-24
Callout: Convert examples to use Sandpack

diyorbek:
#### What changed?

Migrated examples in the docs from the old editor to the sandpack editor for the Callout component.

#### Why?

Ensures the editor remains consistent with other components, as part of the current migration efforts.

### Links

- [Jira](https://jira.pinadmin.com/browse/GESTALT-4876)

### Checklist

- [n/a] Added unit and Flow Tests
- [n/a] Added documentation + accessibility tests
- [n/a] Verified accessibility: keyboard & screen reader interaction

### DIFF
--- a/docs/examples/callout/accessibilityExample.js
+++ b/docs/examples/callout/accessibilityExample.js
@@ -1,30 +1,34 @@
 // @flow strict
 import { type Node } from 'react';
-import { Callout } from 'gestalt';
+import { Callout, Flex, Box } from 'gestalt';
 
 export default function Example(): Node {
   return (
-    <Callout
-      dismissButton={{
-        accessibilityLabel: 'Dismiss this banner',
-        onDismiss: () => {},
-      }}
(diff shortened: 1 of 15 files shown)


AlbertCarreras:
> @AlbertCarreras @dangerismycat I hope it makes sense that examples have bigger height now. Because the editor appears on the side of the examples. Please, let me know if I should keep the heights of the examples the same as before.
> 
> Screenshot 2023-05-25 at 17 17 49 Screenshot 2023-05-25 at 17 17 55

Yeah, don't worry much about the spacing.

AlbertCarreras (approved):
tiny things
we should hide all code editors in all the examples except Main and Best practices. It's a prop (hideEditor)
For the DON'T examples, we hide the editors completely. It's a prop (hideControls)

AlbertCarreras:
We also need to hide this code in Donts
Screenshot by Dropbox Capture

PR #2964, opened 2023-05-25
TileData: add accessibility pause

AlbertCarreras:
#### What changed?

TileData: add accessibility pause

#### Why?

##BEFORE
Screenshot by Dropbox Capture

##AFTER
Screenshot by Dropbox Capture


packages/gestalt/src/Link.js
```diff
@@ -18,10 +18,11 @@ import textStyles from './Typography.css';
 import Box from './Box.js';
 import getRoundingClassName from './getRoundingClassName.js';
 import Icon from './Icon.js';
-import NewTabAccessibilityLabel, { getAriaLabel } from './NewTabAccessibilityLabel.js';
 import Text from './Text.js';
 import useFocusVisible from './useFocusVisible.js';
 import useTapFeedback, { keyPressShouldTriggerTap } from './useTapFeedback.js';
+import getAriaLabel from './accessibility/getAriaLabel.js';
+import NewTabAccessibilityLabel from './accessibility/NewTabAccessibilityLabel.js';
```
rlingineni, line 25: How come these are changed?

rlingineni (approved):
Thanks a bunch (and for catching this)! This looks awesome, and much more reusable than a fix on my end.

packages/gestalt/src/Link.js
AlbertCarreras: I  broke the file in 2

PR #2974, opened 2023-06-01
Internal: Cleaner Logging for DocGen Errors

rlingineni:
When we run `yarn start` we usually see a bunch of errors from docgen.

We were logging the entire stacktrace, and this made the terminal log output larger.  This updates the logging method to give more details on the filepath with missing docgen and the issue.

Before:
Screenshot by Dropbox Capture

After:
Screenshot by Dropbox Capture


dangerismycat (approved):
Ideally we resolve these errors at some point, but this is a great improvement for now!

PR #2985, opened 2023-06-02
Internal: fix TS declaration file

AlbertCarreras:
#### What changed?

Internal: fix TS declaration file


dangerismycat (approved):
Do we have linting that could catch these syntax errors in the future? If not, could we add that?

AlbertCarreras:
> Do we have linting that could catch these syntax errors in the future? If not, could we add that?

No, but we should definitely add it.

PR #2988, opened 2023-06-05
TileData: fix typo + copyediting

dangerismycat:
Noticed an extraneous `a` in TileData's header description. Took a quick pass of copyediting while looking at that file


rlingineni:
thanks!!

PR #2992, opened 2023-06-05
Internal: Add Url Param to toggle Dev Example Mode

rlingineni:
Sometimes we need to explicitly remind people to turn on DevExample Mode. 

This makes it a query param.

So you can say `<deployurl>/web/tiledata?devexample=true`

And have the previews run everywhere. Since we anticipate future PRs to utilize this, this well help us make sharing links easier for review.

One thing to note, passing this param once will probably always enable devmode for people in the deployment versions (probably okay) since I believe this state is actually preserved in localstorage. But also, every deploy url is different.

### DIFF
--- a/docs/docs-components/Header.js
+++ b/docs/docs-components/Header.js
@@ -78,8 +78,8 @@ function Header() {
       process.env.NODE_ENV === 'production' &&
       window?.location?.href?.startsWith('https://deploy-preview-');
 
-    if (isDeployPreviewEnvironment) setShowDevelopmentEditorSwitch(true);
-    if (process.env.NODE_ENV === 'development') setShowDevelopmentEditorSwitch(true);
+    if (isDeployPreviewEnvironment || process.env.NODE_ENV === 'development')
+      setShowDevelopmentEditorSwitch(true);
   }, [setShowDevelopmentEditorSwitch, router.pathname, router.query]);
 
   const { colorScheme, setColorScheme, devExampleMode, setDevExampleMode } = useAppContext();
--- a/docs/docs-components/Header.js
+++ b/docs/docs-components/Header.js
@@ -79,7 +79,8 @@ function Header() {
       window?.location?.href?.startsWith('https://deploy-preview-');
 
     if (isDeployPreviewEnvironment) setShowDevelopmentEditorSwitch(true);
-  }, [setShowDevelopmentEditorSwitch, router.query]);
+    if (process.env.NODE_ENV === 'development') setShowDevelopmentEditorSwitch(true);
+  }, [setShowDevelopmentEditorSwitch, router.pathname, router.query]);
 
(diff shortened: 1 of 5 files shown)


AlbertCarreras:
can we document this somewhere in the docs? 

https://gestalt.pinterest.systems/get_started/developers/contributing/development_process

rlingineni:
@AlbertCarreras - added something to the docs lmk

AlbertCarreras:
haha very very last thing, can we add the `?devexample=true` by default in the releases?

rlingineni:
> 

That'd be great, maybe in eng weekly, also just wondering, if this is something we should enable globally

docs/docs-components/Header.js
```diff
@@ -63,12 +63,24 @@ function Header() {
   );
 
   useEffect(() => {
+    const devModeSetFromUrl = router.query.devexample && router.query.devexample.length > 0;
+
+    // do not show switch if set via url
     if (
+      devModeSetFromUrl &&
+      (router.query.devexample === 'true' || router.query.devexample === 'false')
```
dangerismycat, line 71: I'm curious why we want to hide if `?devexample=false`?

docs/markdown/get_started/developers/contributing/development_process.md
```diff
@@ -178,7 +178,10 @@ git push -f origin HEAD
 
 - After a Gestalt maintainer adds a correct semver label and approves a Pull Request, the PR will be ready to merge. Coordinate with the reviewer to determine when the PR should be merged.
 
-- Check the status of your PR https://github.com/pinterest/gestalt/pull/XXXX and access the deploy preview using the built site URL https://deploy-preview-XXXX--gestalt.netlify.app from Netlify.
+- Check the status of your PR https://github.com/pinterest/gestalt/pull/XXXX and access the deploy preview using the built site URL https://deploy-preview-XXXX--gestalt.netlify.app from Netlify. 
+
+- When sharing development urls, you may have to remind viewers to toggle on development mode from the header menu. To simplify sharing, you can add a `?devexample=true` parameter to the docs url to enable the examples by default.
```
dangerismycat, line 183: This should also include a reason why dev mode needs to be turned on, something like
"…to toggle on development mode from the header menu. This is necessary to see unpublished changes in the docs code examples. To simplify sharing…"

dangerismycat (approved):

docs/docs-components/Header.js
rlingineni: Because it still overrides localhost

rlingineni: Oh, whoops, nvm, it doesn't. I thought it was a hard boolean here. I guess it doesn't matter, I can remove

docs/markdown/get_started/developers/contributing/development_process.md
rlingineni: done

PR #2995, opened 2023-06-06
Chart: new component for data visualization

AlbertCarreras:
Thanks for creating a PR! Please follow this template and delete items/sections that are not relevant to your changes, including these instructions.

Please also make sure your [PR title](https://github.com/pinterest/gestalt/#releasing) matches our format: `{ComponentName}: Description (mentioning platform if relevant)`, or `{ComponentName}, {OtherComponentName}: Description (mentioning platform if relevant)` if multiple components are affected.

#### What changed?

From a high level, what are the changes this PR introduces? (No need to recount line-by-line, we can see that.)

#### Why?

### Links

- [TDD](link to Paper doc)
- [Figma](link to Figma file)

### Checklist


packages/gestalt-charts/src/Chart.js
```diff
@@ -1,23 +1,402 @@
...
+   */
+  layout?: 'horizontal' | 'vertical',
+  /**
+   * Prop description.
+   */
+  legend?: 'auto' | 'hidden' | 'complete',
+  /**
+   * Prop description.
+   */
+  onChangeAccessibility: () => void,
```
rlingineni, line 101: Is this the accessibility mode on the top? 

Is the only thing this does is change the pattern? Wondering if can not confuse it with actual aria modes, and change it to something more explicit like `onFillPatternChange()` or `onVisualPatternChange()`

packages/gestalt-charts/src/Chart.js
```diff
@@ -1,23 +1,402 @@
...
+    stroke: string,
+    strokeOpacity: number,
+  |}>,
+  /**
+   * Prop description.
+   */
+  renderTooltip?:
+    | 'auto'
+    | 'none'
+    | (({| active: ?boolean, payload: ?{ ... }, label: string | number |}) => Node),
```
rlingineni, line 122: Do we have guidelines for a custom tooltip?

packages/gestalt-charts/src/Chart.js
```diff
@@ -1,23 +1,402 @@
...
+  /**
+   * Prop description.
+   */
+  scale?: 'time',
```
rlingineni, line 126: What is scale? Like units?

packages/gestalt-charts/src/Chart.js
```diff
@@ -1,23 +1,402 @@
...
+  /**
+   * Prop description.
+   */
+  biaxial?: 'xAxis' | 'yAxis',
+  /**
+   * Prop description.
+   */
+  data: $ReadOnlyArray<{|
+    name: string | number,
+    [string]: number,
```
rlingineni, line 40: Any reason were' using keys like this?
```
data = { name:string, values:number[]}
```

Though the format seems okay, it seems like the values represent X axis bits, and not the groups. E.g.
It seems like for the linegraph this can be an issue, because it's up to the consumer to compute the stops the x-axis scale when passing the data.

packages/gestalt-charts/src/Chart.js
```diff
@@ -1,23 +1,402 @@
...
+  |}>,
+  /**
+   * Prop description.
+   */
+  description?: string,
+  /**
+   * Prop description.
+   */
+  domain?:
```
rlingineni, line 49: what does this mean?

packages/gestalt-charts/src/Chart.js
```diff
@@ -1,23 +1,402 @@
...
+   */
+  scale?: 'time',
+  /**
+   * Prop description.
+   */
+  title?: string,
+  /**
+   * Prop description.
+   */
+  tickFormatter?: (number) => string,
```
rlingineni, line 134: Will this always be a number?

packages/gestalt-charts/src/Chart.js
```diff
@@ -1,23 +1,402 @@
...
+  /**
+   * Prop description.
+   */
+  type: 'composed' | 'line' | 'bar',
```
rlingineni, line 138: what is the composed type

rlingineni (commented):
added some api feedback

packages/gestalt-charts/src/Chart.js
AlbertCarreras: DONE

packages/gestalt-charts/src/Chart.js
AlbertCarreras: We need it for Timeseries where dates need to be translated. I'l chat with Hector about this though

packages/gestalt-charts/src/Chart.js
AlbertCarreras: this could be renamed to variant = "timeseries" it''s needed for timeseries

packages/gestalt-charts/src/ChartGraph.js
```diff
@@ -0,0 +1,586 @@
...
+                    yAxisId="left"
+                    tickFormatter={tickFormatter?.yAxisLeft}
+                  />
+                </Fragment>
+              ) : null}
+              {isHorizontalBiaxial ? (
+                <YAxis
+                  axisLine={false}
+                  domain={Array.isArray(range) ? range : range.yAxisLeft}
+                  orientation="right"
```
rlingineni, line 512: Can we write a function that takes in the layout and spits out the axis. Looks like this is super similar to the top, and just two prop changes right? 

If we're cool with prop spreading, then just tweaking the object

It can probably make this a tad more compact
```
getYAxis(layout:horizontal|horizontalBiAxial|...){

}
```

packages/gestalt-charts/src/ChartGraph.js
```diff
@@ -0,0 +1,586 @@
...
+                  domain={range}
+                  orientation="top"
+                  style={FONT_STYLE_VALUES}
+                  tickLine={false}
+                  tickCount={tickCount}
+                  type="number"
+                  xAxisId="top"
+                  tickFormatter={tickFormatter?.xAxisTop}
+                />
+              ) : null}
```
rlingineni, line 557: if these are null, can you do `{isVerticalBiaxial && <Component/>}` then I don't think you need to pass the `null`

rlingineni: Since there is no else statement here

packages/gestalt-charts/src/ChartGraph/Header.js
```diff
@@ -0,0 +1,60 @@
...
+              {description}
+            </Text>
+          </Flex>
+        </Flex.Item>
+        {visualPatternSelected === 'disabled' ? null : (
+          <IconButton
+            accessibilityLabel=""
+            dangerouslySetSvgPath={{
+              __path:
+                'M16 11h-3v2.667l2.8 3.733a1 1 0 0 1-1.6 1.2L12 15.667 9.8 18.6a1 1 0 1 1-1.6-1.2l2.8-3.733V11H8a1 1 0 0 1 0-2h8a1 1 0 0 1 0 2zm-4-7a2 2 0 1 1-.001 4.001A2 2 0 0 1 12 4zm0-4C5.373 0 0 5.373 0 12c0 6.628 5.373 12 12 12s12-5.372 12-12c0-6.627-5.373-12-12-12z',
```
rlingineni, line 44: can we just support this icon as part of this change?

packages/gestalt-charts/src/ChartGraph/Header.js
```diff
@@ -0,0 +1,60 @@
...
+  onVisualPatternChange: () => void,
+  visualPatternSelected: ?'accessible' | 'default' | 'disabled',
+|}): Node {
+  const { accessibleViewText, defaultViewText } = useDefaultLabel('ChartGraph');
+
+  return (
+    <Box
+      width="100%"
+      dangerouslySetInlineStyle={{
+        __style: { marginBottom: '20px' },
```
rlingineni, line 24: I thought box has `marginBottom`? like marginBottom={5}

packages/gestalt-charts/src/ChartGraph.js
```diff
@@ -0,0 +1,586 @@
...
+                />
+              ))}
+            </Flex>
+          </Box>
+        ) : null}
+        <div className="_gestalt" style={{ width: '100%', height: '100%', maxWidth: 960 }}>
+          <ResponsiveContainer
+            debounce={150}
+            onResize={(width, height) => {
+              setChartHeight(height);
```
rlingineni, line 449: I was looking at the `readyToRender` stuff. What if you moved this into the `ChartProvider` and then all of the children will have access to height.

It might be more re-usable as we add more components, and help us responsively adapt the other components

```
const {width, height} = useChartProvider()

if(width < 0 || height < 0) 
```

packages/gestalt-charts/src/ChartGraph/EmptyBox.js
```diff
@@ -0,0 +1,7 @@
+// @flow strict
+import { type Node } from 'react';
+import { Box } from 'gestalt';
+
+export default function EmptyBox(): Node {
+  return <Box />;
```
rlingineni, line 6: Do we need a separate component for this?

packages/gestalt-charts/src/ChartGraph/LegendIcon.js
```diff
@@ -0,0 +1,128 @@
...
+
+  const { decal: showVisualPattern } = useChartContext();
+
+  if (payloadData.referenceArea === 'default') {
+    const dimension = 16;
+
+    return (
+      <svg height={dimension} width={dimension} aria-hidden>
+        <rect
+          style={{ fill: '#ccc', fillOpacity: '0.5' }}
```
rlingineni, line 39: is this a recharts thing? how come the fill is constant

packages/gestalt-charts/src/ChartGraph/LegendIcon.js
```diff
@@ -0,0 +1,128 @@
...
+          width={dimension}
+          rx={2}
+        />
+      </svg>
+    );
+  }
+
+  const { isLegend } = payloadData;
+
+  const colorMap = Object.entries({
```
rlingineni, line 50: Can we loop through the colorScheme like you did in `usePattern.js` than use a hardcoded map we have to update

rlingineni: Not sure if the idea is to get all the colors? or you can use useHex hook you wrote earlier probably.

```
const allColors = useColorScheme()

const allDataVizColors = Object.keys(allColors).filter(key=>key.startsWith('colorDataVisualization')).map(key => allColors[key])
```

packages/gestalt-charts/src/ChartGraph/renderElements.js
```diff
@@ -0,0 +1,130 @@
...
+            />
+          )}
+          {...(isHorizontal
+            ? { yAxisId: values.axis || 'left' }
+            : { xAxisId: values.axis || 'bottom' })}
+        />
+      );
+    }
+
+    // Recharts doesn't recognize wrappers on their components, therefore, needs to be build within ChartGraph
```
rlingineni, line 92: what does wrappers mean?

rlingineni: like no customization

packages/gestalt-charts/src/ChartGraph.js
```diff
@@ -0,0 +1,586 @@
...
+  let ChartType = ComposedChart;
+
+  if (isBar) {
+    ChartType = BarChart;
+  }
+  if (isLine) {
+    ChartType = LineChart;
+  }
+  // SUBCOMPONENTS
+  const chartElements = useMemo(
```
rlingineni, line 329: Why memo the subcomponents, and not the entire thing?

packages/gestalt-core/build.js
```diff
@@ -59,7 +59,7 @@ const cssModules = (options = {}) => {
 
       // if it's not production, use the minified name + extended name
       const className =
-        process.env.DEVMODE === 'true' ? `${name}__${scopeNames[hash]}` : scopeNames[hash];
+        process.env.DEVMODE === 'false' ? `${name}__${scopeNames[hash]}` : scopeNames[hash];
```
rlingineni, line 62: is this broken?

flow-typed/npm/recharts_vx.x.x.js
```diff
@@ -0,0 +1,4 @@
+declare module 'recharts' {
+  declare module.exports: any;
```
rlingineni, line 2: you might be able to declare classes like for the `payload` prop and stuff here if they match the recharts definition and "import" from recharts

rlingineni (commented):
first round: didn't dig into the examples yet. 

Nice work! This is a lot of new stuff to piece together. Mostly feedback on code readability

docs/pages/web/chartgraph.js
```diff
@@ -0,0 +1,599 @@
...
+            sandpackExample={<SandpackExample code={title} name="Title & description" />}
+          />
+        </MainSection.Subsection>
+
+        <MainSection.Subsection
+          title="Tooltip"
+          columns={2}
+          description={`For showing more precise details on hover.
+
+Requires props:  \`renderTooltip\`
```
rlingineni, line 367: nit: I don't think we need the "requires" word, thinking just 'Props':  will do, and then a list.

docs/pages/web/chartgraph.js
```diff
@@ -0,0 +1,599 @@
...
+        <MainSection.Subsection columns={2}>
+          <MainSection.Card
+            cardSize="lg"
+            title="No tooltip"
+            sandpackExample={<SandpackExample code={noTooltip} name="No tooltip" />}
+          />
+          <MainSection.Card
+            cardSize="lg"
+            title="Custom tooltip"
+            sandpackExample={<SandpackExample code={customTooltip} name="Custom tooltip" />}
```
rlingineni, line 385: Custom tooltip looks similar to regular, am I missing something?

Maybe add a header or something, change the background, layout? 
image



Default:
image

Custom:
image

docs/pages/web/chartgraph.js
```diff
@@ -0,0 +1,599 @@
...
+            sandpackExample={<SandpackExample code={referenceArea} name="ReferenceArea" />}
+          />
+        </MainSection.Subsection>
+
+        <MainSection.Subsection
+          title="Precision in line graphs"
+          description={`To show exact and accurate data, lines should be rectilinear. When showing trends, forecasts and imprecise data, then lines should be curved to denote that these are just approximations.
+
+
+Requires props: \`elements\` and setting \`precision='estimate'\`
```
rlingineni, line 419: Maybe props should be like `elements=[{type: "line"  precision: "estimate"}]`

Also just wondering here, since the example is wonky to work with. Will this work on a bar graph? How about multiple lines? Multiple bars

docs/pages/web/chartgraph.js
```diff
@@ -0,0 +1,599 @@
...
+        >
+          <CombinationNew cardSize="xs" color={['01', '02', '03', '04', '05', '06']}>
+            {({ color }) => (
+              <Box
+                dangerouslySetInlineStyle={{
+                  __style: {
+                    backgroundColor: `var(--g-colorDataVisualization${color})`,
+                  },
+                }}
+                height={100}
```
rlingineni, line 442: Can these be smaller?

docs/pages/web/chartgraph.js
```diff
@@ -0,0 +1,599 @@
...
+            cardSize="lg"
+            sandpackExample={<SandpackExample code={colors} name="Colors" layout="column" />}
+          />
+        </MainSection.Subsection>
+
+        <MainSection.Subsection
+          title="Layout"
+          description={`ChartGraph supports 4 layouts. Legend positions automatically on each layout for easier comprehension.
+
+Single axis:
```
rlingineni, line 470: Check formatting here, indents look weird:

image

docs/pages/web/chartgraph.js
```diff
@@ -0,0 +1,599 @@
...
+            )}
+          </CombinationNew>
+          <MainSection.Card
+            cardSize="lg"
+            sandpackExample={<SandpackExample code={colors} name="Colors" layout="column" />}
+          />
+        </MainSection.Subsection>
+
+        <MainSection.Subsection
+          title="Layout"
```
rlingineni, line 467: thinking this section should be higher

docs/pages/web/chartgraph.js
```diff
@@ -0,0 +1,599 @@
...
+        >
+          <MainSection.Card sandpackExample={<SandpackExample code={range} name="Range" />} />
+        </MainSection.Subsection>
+
+        <MainSection.Subsection
+          title="Responsive"
+          description={`ChartGraph is a responsive component. ChartGraph's width adjust to the parent container. In order to render properly in requires a parent container with set dimensions.
+
+
+          For vertical layour, ChartGraph has a set height based on the amount of ticks (five or three). For horizontal layouts, ChartGraph has a set height.
```
rlingineni, line 501: Typo here and sentence issues.

This isn't intentional right:

docs/pages/web/chartgraph.js
```diff
@@ -0,0 +1,599 @@
...
+
+          When ChartGraphs are contained within small containers (under 576 px wide), set \`initialTicks={3}\` to prevent ChartGraph to flicker.
+          `}
+        >
+          <MainSection.Card
+            sandpackExample={<SandpackExample code={responsive} name="Responsive" />}
```
rlingineni, line 507: Ideally, the radiobuttons should also be wrapped in this example if we're going to have a slider that goes to 200px. or move them outside

image

docs/pages/web/chartgraph.js
```diff
@@ -0,0 +1,599 @@
...
+Requires props: \`tickFormatter.timeseries\`.
+            `}
+        >
+          <MainSection.Card
+            sandpackExample={<SandpackExample code={timeseries} name="Time series" />}
+          />
+        </MainSection.Subsection>
+      </MainSection>
+
+      <MainSection name="Writing">
```
rlingineni, line 549: I feel this should be higher because this page is sooo long

rlingineni: Maybe after Do/Don't. I feel this should be read

rlingineni (commented):
some comments on doc

docs/docs-components/data/components.js
```diff
@@ -495,32 +495,34 @@ const componentData: $ReadOnlyArray<ComponentData> = [
...
-  //       status: {
-  //         accessible: {
-  //           summary: 'notAvailable',
-  //           a11yVisual: 'notAvailable',
-  //           a11yScreenreader: 'notAvailable',
-  //           a11yNavigation: 'notAvailable',
-  //           a11yComprehension: 'notAvailable',
-  //         },
-  //         documentation: 'notAvailable',
-  //         figmaStatus: 'notAvailable',
```
hectoid, line 517: We do have Figma components for bar (both horizontal and column) and line charts. But, they need to be renamed and updated with the latest to match code. I think we can mark as Ready here; designers can mak a note in Figma that there will be updates.

hectoid (commented):
a note on Figma status

hectoid:
Very nice! Most comments/suggested changes are due to my own bad proofreading in Coda. Some updates below.
image
This should be a warning banner instead.


Pasted Graphic
Can we add the "converted line" and add more content to “Paid” to really make this unreadable? The example is still fairly easy to read. See below instead:
image


Pasted Graphic 2

Sorry, my poor editing. Remove the highlighted statement.


Pasted Graphic 3
I think all graphs with more than one color should have the accessibility button


Pasted Graphic 5
Change the highlight to “use Bar horizontal instead.”

Pasted Graphic 7
Different colors are used for different categories (say “active users / inactive users”). So, in this case, there is only 1 category, but some of that data in the category isn’t available. Instead, this should just be the blue line for both solid and dashed curved lines. Then the legend just shows one item: the forecast.


--------
Nice-to-haves that can happen later:
Pasted Graphic 6
Can be a v2 or 3, but am thinking the background of the bars needs to be white so the lines don’t show through. If that’s possible at all.

image
Can the slider follow our usual UI colors? https://www.figma.com/file/MI9IvOTCfmenj31GQ2p5aB/Handoff-ChartGraph?type=design&node-id=1395%3A10436&mode=design&t=y50QunO6FLbRS1s3-1

AlbertCarreras:
> Very nice! Most comments/suggested changes are due to my own bad proofreading in Coda. Some updates below. image This should be a warning banner instead.
> 
> Pasted Graphic Can we add the "converted line" and add more content to “Paid” to really make this unreadable? The example is still fairly easy to read. See below instead: image Pasted Graphic 2
> Sorry, my poor editing. Remove the highlighted statement.
> 
> Pasted Graphic 3 I think all graphs with more than one color should have the accessibility button Pasted Graphic 5 Change the highlight to “use Bar horizontal instead.” Pasted Graphic 7 Different colors are used for different categories (say “active users / inactive users”). So, in this case, there is only 1 category, but some of that data in the category isn’t available. Instead, this should just be the blue line for both solid and dashed curved lines. Then the legend just shows one item: the forecast.
> Nice-to-haves that can happen later: Pasted Graphic 6 Can be a v2 or 3, but am thinking the background of the bars needs to be white so the lines don’t show through. If that’s possible at all.
> 
> image Can the slider follow our usual UI colors? https://www.figma.com/file/MI9IvOTCfmenj31GQ2p5aB/Handoff-ChartGraph?type=design&node-id=1395%3A10436&mode=design&t=y50QunO6FLbRS1s3-1

@hectoid Re responsive container, I can try, it's the same one used in masonry docs. Leaving for now. But we can change

rlingineni (approved):

PR #3005, opened 2023-06-09
Docs: create new COMPONENT_DATA data structure type for scalability

dangerismycat:
We've started running into scalability issues with COMPONENT_DATA. ([relevant Slack convo](https://pinterest.slack.com/archives/G01GN7CE99S/p1686182653016419)) Specifically we need:
- To use different names for the same component on different platforms
- To better handle Figma-only components
- To use different aliases for the same component on different platforms

The original data structure has been modified several times, but these modifications were more "tacked on" than full rewrites (as you do), and there are multiple areas for worthwhile cleanup. I'll leave comments where problems were addressed and we can discuss in more depth on Slack.

For now, this PR doesn't actually refactor the component data to use the new data structure. Once we agree on the data structure, I can either add that refactor to this PR or in a follow-up PR.


docs/docs-components/data/types.js
```diff
@@ -12,3 +12,64 @@ export type DesignOverview = {|
   path: string,
   svg: SVGElement,
 |};
+
+type StatusType = 'notAvailable' | 'partial' | 'planned' | 'ready';
```
dangerismycat, line 16: no change here

docs/docs-components/data/types.js
```diff
@@ -12,3 +12,64 @@ export type DesignOverview = {|
...
+
+/**
+ * All components should be in a single list. That list can be filtered by category/platform/etc where needed.
```
dangerismycat, line 19: This is my recommended strategy for refactoring the actual component data. We don't need multiple lists — we already use a `category` field, which allows for any necessary filtering. A single list makes components easier to deal with.

docs/docs-components/data/types.js
```diff
@@ -12,3 +12,64 @@ export type DesignOverview = {|
...
+ */
+
+type ComponentAccessibility = {|
+  a11yComprehension?: StatusType,
```
dangerismycat, line 23: As a general rule, these types don't have nullable _values_, only nullable _fields_. Nullable fields are easy to deal with using optional chaining, and this simplifies downstream logic — if a field exists, its value exists.

docs/docs-components/data/types.js
```diff
@@ -12,3 +12,64 @@ export type DesignOverview = {|
...
+  a11yNavigation?: StatusType,
+  a11yScreenreader?: StatusType,
+  a11yVisual?: StatusType,
+  summary?: StatusType,
+|};
+
+type ComponentCategory =
```
dangerismycat, line 30: Removed the categories we were using for design foundations tiles (see #3004). Those weren't components anyways so were an awkward fit here.

docs/docs-components/data/types.js
```diff
@@ -12,3 +12,64 @@ export type DesignOverview = {|
...
+  | 'Loading'
+  | 'Messaging'
+  | 'Navigation'
+  | 'Overlays'
+  | 'Pins and imagery'
+  | 'Structure'
+  | 'Text'
+  | 'Utilities';
+
+type PlatformObjType<T> = {|
```
dangerismycat, line 48: The new data structure splits out most fields by platform (and we're treating Figma as another platform to allow for Figma-only components). This gives us a lot of flexibility for different values by platform.

Given that most fields _aren't_ platform-specific, there are a couple of approaches we can take:
- Since all platform fields are nullable, default to `web` values if others aren't present
- Duplicate values manually to each platform
I think the first approach is much, much better — it uses a single source of truth and prevents a lot of tedious copy/pasting (and bugs from inevitable drift).

This also cuts down on a lot of boilerplate inherent in the current data structure. If a component doesn't exist on Android, we don't need to add an object full of null values — we just don't add `android` at all in the `status` object.

docs/docs-components/data/types.js
```diff
@@ -12,3 +12,64 @@ export type DesignOverview = {|
...
+  android?: T,
+  figma?: T,
+  ios?: T,
+  web?: T,
+|};
+
+export type ComponentData = {|
+  id: string, // This is essentially the internal name
```
dangerismycat, line 56: Since components might have different names by platform, this gives us a single "name" to refer to the component by when reference component data objects.

docs/docs-components/data/types.js
```diff
@@ -12,3 +12,64 @@ export type DesignOverview = {|
...
+  name: PlatformObjType<string>,
+  visual: {|
+    svg: PlatformObjType<SVGElement>,
+    hasDarkBackground?: boolean,
```
dangerismycat, line 60: This is only relevant to the SVG, so moving it to live alongside the SVG.

docs/docs-components/data/types.js
```diff
@@ -12,3 +12,64 @@ export type DesignOverview = {|
...
+  |},
+  alias: PlatformObjType<$ReadOnlyArray<string>>,
+  description: PlatformObjType<string>,
+  category: PlatformObjType<$ReadOnlyArray<ComponentCategory>>,
+  status: PlatformObjType<{|
```
dangerismycat, line 65: The current data structure is very repetitive, particularly vis-a-vis mobile platforms. This cuts down on the repetition and ensures a single source of truth.

docs/docs-components/data/types.js
```diff
@@ -12,3 +12,64 @@ export type DesignOverview = {|
...
+    accessible: ComponentAccessibility,
+    badge?: 'New' | 'Pilot' | 'Experimental',
+    documentation: StatusType,
+    figmaOnly?: boolean,
+    figmaStatus?: StatusType,
+    mobileAdaptive?: StatusType,
+    responsive?: StatusType,
+    status: StatusType | 'deprecated',
```
dangerismycat, line 73: `'ready'` + `'deprecated'` doesn't make much sense — `'deprecated'` is really another status.

PR #3007, opened 2023-06-12
DatePicker: fixes in Flow|TypeScript types

AlbertCarreras:
### Breaking changes

Run codemod to detect DatePicker. Also, Flow|TypeScript complains could help fixing this.
```
--component=DatePicker
```

#### What changed?

DatePicker returns "Date" objects and "null", "null" value was not in the type. 

Major breaking change as it might affect useState value when implementing the upgrade

#### Why?

Flow errors

### DIFF
--- a/packages/gestalt-datepicker/src/DatePicker.js
+++ b/packages/gestalt-datepicker/src/DatePicker.js
@@ -10,9 +10,9 @@ import {
 import InternalDatePicker from './DatePicker/InternalDatePicker.js';
 
 // LocaleData type from https://github.com/date-fns/date-fns/blob/81ab18785146405ca2ae28710cdfbb13a294ec50/src/locale/af/index.js.flow
-// flowlint unclear-type:off
 // NOTE: DO NOT USE PER-LINE FLOW SUPPRESSIONS HERE
 // They will get picked up by the docgen and bork the type displayed on the docs
+// flowlint unclear-type:off
 type LocaleData = {|
   code?: string,
   formatDistance?: (...args: $ReadOnlyArray<any>) => any,
@@ -104,7 +104,7 @@ export type Props = {|
    */
   onChange: ({|
     event: SyntheticInputEvent<HTMLInputElement>,
-    value: Date,
+    value: Date | null,
   |}) => void,
   /**
    * Placeholder text shown if the user has not yet input a value. The default placeholder value shows the date format for each locale, e.g. MM/DD/YYYY.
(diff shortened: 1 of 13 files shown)


packages/gestalt-datepicker/src/DatePicker.js
```diff
@@ -113,15 +113,15 @@ export type Props = {|
   /**
    * Required for date range selection. End date on a date range selection. See the [date range example](https://gestalt.pinterest.systems/web/datepicker#Date-range) to learn more.
    */
-  rangeEndDate?: Date,
+  rangeEndDate?: Date | null,
```
rlingineni, line 116: Does the question mark take care of this?

rlingineni: Or does it explicitly have to be null?

rlingineni (approved):

packages/gestalt-datepicker/src/DatePicker.js
AlbertCarreras: rangeEndDate might not always be used, it's just for ranges. But when used the value might come as null when changed. This value is dynamic, different from minDate or maxDate

dangerismycat: (I'm assuming you're referring to the question mark after `rangeEndDate`)

Question marks in Flow generally convey "optional", but there are subtle differences when that is applied to the prop name vs the prop value.

```js
{| propName?: 'a' | 'b' |}
```
`propName` may be supplied, or it may be omitted. If it's supplied, it _must_ be either `'a'` or `'b'`. `propName={undefined}` is an error.

```js
{| propName: ?boolean |}
```
`propName` is a required prop — omitting it is an error. When `?` is applied to a prop value, it translates to `propValue | null | void` (`void` is `undefined`). So in this case `propName={undefined}` is fine, as would be `null` or a boolean.

```js
{| propName?: ?boolean |}
```
`propName` is an optional prop and can be omitted. If it's supplied, it can be `undefined`, `null`, or a boolean. This is the most flexible option but also requires the most type checking within the component logic.

I've found that generally I like optional props with strongly defined types for components — the prop can be omitted, but if it's present, we have a good idea of its type. In places like this one where we do need some sort of nil type as well, I prefer the approach Alberto is taking here: `null` never happens by accident (unlike `undefined`) so it conveys intention, and specifically using `Date | null` is stronger than `?Date`, where `undefined` would also be acceptable. (This is similar to [my preferred approach for typing objects](https://github.com/pinterest/gestalt/pull/3005#discussion_r1224878077))

PR #3008, opened 2023-06-12
Tokens: Update $selected-strong and $selected-weak values

rlingineni:
### Summary
Swapping the weak and strong token colors, re [this thread](https://pinterest.slack.com/archives/C0HUV5J93/p1680555287142499).

#### What changed?

Our current color tokens for $color-background-selected-strong and $color-background-selected-weak seem to be inverted according to our color guidelines. It seems to be a mistake. 

#### Why?

The ordering of and the actual token value was incorrect here.

### How to fix:

We didn't notice instances of these [token usages](https://sourcegraph.pinadmin.com/search?q=context:global+repo:%5Egithub%5C.com/pinternal/pinboard+file:webapp/+--color-background-selected-strong&patternType=regexp&sm=0&groupBy=repo), so a codemod was not necessary here.

To check if your code is using this, check by looking for usages of this token in areas like style classes or conditional backgrounds :

```
__style: { borderColor: 'var(--color-background00)', },
```

In the event we missed things, prior to this release version, in places you were using color `--color-background-selected-strong`, swap it out to be,`--color-background-selected-weak`, please make sure to flip their usage to get the correct assigned colors. 

Otherwise you may see the colors in reversed roles.

### Links
- [Jira](https://jira.pinadmin.com/browse/GESTALT-6032)


dangerismycat (approved):
As discussed in our meeting, before merging please update the PR summary with steps to address the breaking changes (including full token names, as those are difficult to parse from the JSON)

PR #3019, opened 2023-06-17
Tokens: New disabled token for text and icons

hectoid:
### Summary
Added disabled text and icon token and updated the Color docs.

#### What changed?

Added disabled token to the base.js file. And updated the color Usage page to reflect the new token. Also re-worded the typography section to match the style of the rest of the doc.
image
image
image
image

#### Why?

A color that looks more disabled than our subtle color is needed. Subtle is for secondary text, but there is a conflict when there is a subhead or helper text in a subtle color on the page, then also a disabled element with the same color. Android and iOS devs also need the token so they can implement on their end.

### Links

- [Jira](https://jira.pinadmin.com/browse/GESTALT-6295)
- [Figma](https://www.figma.com/file/cPXcu0NmmxRf5ywyybuxTU/Gestalt-foundations?type=design&node-id=7073%3A3468&t=f9QuM2lMLIuBBx2Q-1)

### Checklist

### DIFF
--- a/docs/pages/foundations/color/usage.js
+++ b/docs/pages/foundations/color/usage.js
@@ -236,10 +236,11 @@ export default function ColorUsagePage(): Node {
           title="Standard text colors"
           description={`
           Typography colors are used on text elements such as headings and body.
-          Use the tokens:
-          **$color-text-default** - The default text color, such as headings and body text.
-          **$color-text-subtle** - For secondary, subtle text color, such as additional info or sub-header.
-          **$color-text-shopping** - For text related to shopping products or surfaces.
+          Token uses:
+          **$color-text-default** - Use as the default text color for headings and body text.
+          **$color-text-subtle** - Use for secondary info such helper text or subheadings.
+          **$color-text-disabled** - Use for text inside of disabled UI controls.
+          **$color-text-shopping** - Use for text related to shopping products or surfaces.
 
           `}
         >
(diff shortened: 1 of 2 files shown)


AlbertCarreras:
Re a11y test, on the page add   

```
await expectAccessiblePage({
    page,
    rules: {
      'color-contrast': { enabled: false },
    },
  });
```

We should fix this in a more targeted way.

AlbertCarreras:
Re Prettier, 

Install in VScode https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode
The cmd + Shift + P ==> Format document (force)

AlbertCarreras (changes requested):
Also, can we run this through the meeting w/ engs?

Why?

We should not add a token and document it but not add it to Text, Button, etc. 

I think this kind of changes need to be passed by eng so we can evaluate the best way of implementing them.

We could merge and document this, but then there would be no way of accessing them

hectoid:
> Also, can we run this through the meeting w/ engs?
> 
> Why?
> 
> We should not add a token and document it but not add it to Text, Button, etc.
> 
> I think this kind of changes need to be passed by eng so we can evaluate the best way of implementing them.
> 
> We could merge and document this, but then there would be no way of accessing them

Let's meet first. I've added it to the agenda for next Monday. Sorry, wasn't sure if we could add tokens and then handle applying it later.

PR #3020, opened 2023-06-20
TagData: Fit and Finish on TagData 

rlingineni:
Hiding editor controls on a few docs for TagData examples:

Looks like TagData styling breaks when in prod (due to HTML CSS button default styling in devmode, but not existing in sandbox). This happens in the CodeSandbox

image

Should be:
image

This also has a few more fixes:

- [X] show border over the dismiss area
- [X] mouse is pointer on dismiss
- [X]  'x' is 10x10 ixon
- [X]  fix gray fills in examples

Also removing controls from docs site


packages/gestalt/src/TagData.js
```diff
@@ -242,16 +242,22 @@ export default function TagData({
...
             onRemove({ event, id });
           }}
         >
-          <Icon
-            accessibilityLabel={accessibilityRemoveIconLabel ?? accessibilityRemoveIconLabelDefault}
-            color={fgColor}
-            icon="cancel"
-            size={8}
-          />
+          <Box display="flex" alignItems="center" items="center" width="100%">
```
dangerismycat, line 251: I think `items` is a typo? At least it's not a prop on Box or an HTML attribute I'm familiar with

dangerismycat (approved):

packages/gestalt/src/TagData.js
rlingineni: Good catch, fixed!

Also, weird, I thought flow is supposed to catch something like that

dangerismycat: It's not on Box — Box is unique. It's a [passthrough component](https://github.com/pinterest/gestalt/blob/bdb1cdbb3b56748260de206cbd36b0fe92a3edb3/packages/gestalt/src/Box.js#L528), so aside from a handful of props we explicitly remove, anything passed to Box will go through. We do have [a lint rule](https://github.com/pinterest/gestalt/blob/master/packages/eslint-plugin-gestalt/src/no-box-disallowed-props.js) that we use in Pinboard to establish further restrictions, but we don't enable that rule within our own repo.

PR #3026, opened 2023-06-22
Docs: Add Value to Default Label Provider Docs

rlingineni:
Adding TagData Remove label to docs table

https://gestalt.pinterest.systems/web/utilities/defaultlabelprovider#Default-strings-provided


AlbertCarreras:
https://github.com/pinterest/gestalt/blob/c5ee1613bb3241203690d20f19b2338693f483e1/packages/gestalt/dist/index.d.ts#L345 


Can yo add it here as well?

AlbertCarreras (changes requested):
add type to TS file

rlingineni:
@AlbertCarreras - added, and also updated readme to remind adding ts defs

scripts/templates/README.md
```diff
@@ -103,3 +103,8 @@ yarn playwright:update-visual-test playwright/visual-test/ComponentName --update
 
 `.github/pr-title-checker-config.json`
 Add your new component name to the regex (alphabetize please!). If you're renaming, replacing, or removing a component, — _and the component also doesn't exist on any other platforms_ — remove that component name from the regex.
+
```
AlbertCarreras, line 106: See line 76 @rlingineni

packages/gestalt/src/TagData.js
```diff
@@ -44,11 +44,7 @@ export type TileChangeHandler = ({|
 |}) => void;
 
 export type OnRemoveHandler = ({|
-  event:
-    | SyntheticMouseEvent<HTMLDivElement>
-    | SyntheticKeyboardEvent<HTMLDivElement>
-    | SyntheticMouseEvent<HTMLAnchorElement>
-    | SyntheticKeyboardEvent<HTMLAnchorElement>,
+  event: SyntheticMouseEvent<HTMLButtonElement>,
```
AlbertCarreras, line 47: Why this change?

packages/gestalt/dist/index.d.ts
```diff
@@ -1864,6 +1867,29 @@ interface TagProps {
   type?: 'default' | 'error' | 'warning' | undefined;
 }
 
+interface TagDataProps {
+  accessibilityRemoveIconLabel?: string | undefined;
+  baseColor?: 'primary' | 'secondary'
+  color?: DataVisualizationColors,
+  disabled?: boolean,
+  id?: string,
+  onTap?: | AbstractEventHandler<
```
AlbertCarreras, line 1876: have forced prettier here?

the initial union is needed? `"|" Abs`

packages/gestalt/src/TagData.js
AlbertCarreras: what if it's click from the keyboard?

AlbertCarreras (approved):
R+ but look at the comments

packages/gestalt/src/TagData.js
rlingineni: Same handler/implementation as`Tag`. It's an HTML button `onClick` handler, it doesn't give the keyboard events back

scripts/templates/README.md
rlingineni: Oh lol, I just can't read

PR #3028, opened 2023-06-22
Image: Fix web hydration issue

mziemer21:
Fixing this hydration error:

> Warning: Extra attributes from the server: elementtiming
>     at img

#### What changed?

If no element timing prop is passed, use the src as a value.

#### Why?

This causes a hydration issue. I'm guessing the server adds the attribute, but the browser removes it because it is empty. This causes a mismatch in HTML, throwing a hydration error.

### Checklist


AlbertCarreras (approved):

AlbertCarreras:
@mziemer21  Can you format it? it's failing prettier

PR #3030, opened 2023-06-22
Reordering Android component items to be alphabetical

ponori:
Moved SearchField nav item into its alphabetical position.


AlbertCarreras (approved):
better not break anything!

PR #3042, opened 2023-06-27
WIP Internal: implement system props for dataTestId

dangerismycat:
*** THIS IS A WIP PROPOSAL AND SHOULD NOT BE MERGED UNTIL `dataTestId` HAS BEEN IMPLEMENTED IN EVERY COMPONENT ***

*****

## The Problem

In order to fully support needs across Pinterest, but particularly M10n, all Gestalt components and sub-components need to implement dataTestId. In addition, we need to enable easier experimentation on Gestalt components that can be controlled at the callsite. [ExperimentProvider](https://github.com/pinterest/gestalt/blob/9b2a9c38daa265d892f19696ad84bcb1a3f1eac6/packages/gestalt/src/contexts/ExperimentProvider.js#L4) only provides global experiment control, and all attempts to make it more callsite-specific have been overly-convoluted and wonky.

## A Solution

GitHub’s design system, [Primer](https://primer.style/react/), has the concept of “[system props](https://primer.style/react/system-props)”. These are common props that are applied to all components by default. While Primer uses these for styles — something I think we should not do, as it would open the floodgates to custom components — I think we could apply the same concept to both dataTestId and experiments.

### dataTestId 

We need to implement this prop on every component and sub-component. While the implementation will be specific to each component (i.e. which HTML element this is assigned to under the hood), the docs can be mostly generic.

### Experiments

[I’ve kicked around some ideas related to this](https://paper.dropbox.com/doc/Experimentation-in-Gestalt-2.0--B6_04hMIWcWPwY851WD661cRAg-RNTRK1YN01g3yUOiKy8ld) (and [Jack Hsu](mailto:jackhsu@pinterest.com) has had some good ideas as well), but haven’t implemented anything. The tl;dr of that doc is that I think we should bring back per-component experiment props, but a single prop that has a similar “register the experiment name” approach as ExperimentProvider. This prop wouldn’t need to be created for each experiment; it would exist already (on every component and sub-component), and can be used where needed for an arbitrary number of simultaneous experiments.

This idea isn’t ready for primetime yet, but it would build on the same system props approach.

## Implementation

As noted above, we will need to manually implement dataTestId in each component. We will also need to manually add any relevant code for each experiment to be run in the future. What the system props approach would do is provide a common external interface for these props.

This PR prototypes a way of adding these shared system props to every component and sub-component automatically as part of the `docGen` process. This reduces the boilerplate needed for props documentation and ensures that every component uses the same interface for these props.

### DIFF
--- a/docs/docs-components/docgen.js
+++ b/docs/docs-components/docgen.js
@@ -23,12 +23,30 @@ export type DocGen = {|
   |},
 |};
 
+function addSystemProps(docGenArg: DocGen): DocGen {
+  return {
+    ...docGenArg,
+    props: {
+      ...docGenArg.props,
+      dataTestId: {
+        defaultValue: null,
+        required: false,
+        description:
+          'A string that is used as a data attribute `data-test-id` on the underlying element.',
+        flowType: {
+          name: 'string',
+        },
+      },
+    },
+  };
+}
+
 export type DocType = {|
   generatedDocGen: DocGen,
 |};
 
 export default function docGen(componentName: string): DocGen {
-  return metadata[componentName];
+  return addSystemProps(metadata[componentName]);
 }
 
 export function multipleDocGen(componentNames: $ReadOnlyArray<string>): {|
--- a/packages/gestalt/src/Avatar.js
+++ b/packages/gestalt/src/Avatar.js
@@ -6,6 +6,7 @@ import { useColorScheme } from './contexts/ColorSchemeProvider.js';
 import Icon from './Icon.js';
 import Image from './Image.js';
 import Mask from './Mask.js';
+import { type SystemProps } from './systemProps.js';
 
 const sizes = {
   xs: 24,
@@ -16,6 +17,7 @@ const sizes = {
 };
 
 type Props = {|
+  ...SystemProps,
   /**
    * String that clients such as VoiceOver will read to describe the element. Will default to `name` prop if not provided.
    */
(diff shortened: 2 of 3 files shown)


rlingineni:
This is cool! Didn't know about system props on Github.

This approach works, wondering about Flow-Types here?  I saw the paper doc mentions use of a `commonProps`,  is the idea that we're going to update all the components to have something like a `{...commonProps,..restOfProps}` in all of the components? Or is this purely a docs change? 

It seems like we're getting a second source of truth here? 

Instead of manually appending a parsed value to the array, could the shared type actually live as a props type in the codebase? The `addSystemProps` method can read from that file

Something like this:
```
export type SystemProps = {
/*description here*/
dataTestId: string
}
```


I think this a good approach for `checkExperiment` as well, but as a different convo, I think I have a few more questions on experiments in general.

dangerismycat:
@rlingineni Ah sorry, I forgot to implement the other side of this. Pushing up a mock implementation using Avatar right now, which I think (hope?) will better explain things

packages/gestalt/src/Avatar.js
```diff
@@ -16,6 +17,7 @@ const sizes = {
 };
 
 type Props = {|
+  ...SystemProps,
```
dangerismycat, line 20: This won't get picked up by the docgen function. It's not that it gets added with a bad type, it just gets completely ignored.

docs/docs-components/docgen.js
```diff
@@ -23,12 +23,30 @@ export type DocGen = {|
...
+  };
+}
+
 export type DocType = {|
   generatedDocGen: DocGen,
 |};
 
 export default function docGen(componentName: string): DocGen {
-  return metadata[componentName];
+  return addSystemProps(metadata[componentName]);
```
dangerismycat, line 49: Rather than add the system props types at each component page, we can add them once here

PR #3045, opened 2023-06-28
Internal: implement new component data type (w/helpers)

dangerismycat:
This PR implements the new component data structure proposed in #3005. I modified the data structure slightly — whereas previously each field required an object that took values for each platform, the entire data structure is now split by platform. This seemed to make things a bit easier to deal with. 

I also added a couple of helpers to get components by platform and by platform+category, with tests.

Note that this PR doesn't actually switch to _consume_ the new component data. I wanted to merge this change and let other folks double-check the data to make sure it's correct (I'm pretty confident the web data is fine, but less so on Android/iOS, and much less so for Figma). So this doesn't represent any changes that would be visible in the docs yet.

### DIFF
--- a/docs/docs-components/data/components-new.js
+++ b/docs/docs-components/data/components-new.js
@@ -1,0 +1,2606 @@
+// @flow strict
+import { type Element } from 'react';
+import { type ComponentData } from './types.js';
+import Box from '../../graphics/building-blocks/Box.svg';
+import Column from '../../graphics/building-blocks/Column.svg';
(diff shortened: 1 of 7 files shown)


docs/docs-components/data/components-new.js
```diff
@@ -0,0 +1,2606 @@
+// @flow strict
```
dangerismycat, line 1: Note that this is a different file from `docs/docs-components/data/components.js`. That file remains untouched, so no changes here will be reflected in the docs yet.

PR #3050, opened 2023-06-30
WIP Masonry: Add support for pin leveling

dangerismycat:
After talking through pin leveling with @jportella93, I was inspired to take a crack at it. It's hard. 😅 

This PR does a few things:
- Combines the existing positionStore and a new heightAdjustmentStore into a single itemAttributeStore. We should probably fold the measurementStore into this eventually.
- Attempts to identify the two items just above the 2-col module and passes along the necessary heightAdjustment based on the additionalWhitespace
- Adjusts the layout position of the 2-col module to take up the additional whitespace

But there are problems. I'm not sure if this will be technically possible given the current constraint that item heights cannot change after the first render. This is leads to overlapping items. There are a couple of possible solutions I see to this:
- Introduce some way for items to communicate changes in their height to Masonry. This feels gross.
- Only lay out a 2-col module in a large enough batch of items that it includes the items just above the module. We would adjust the layout of the target items _and_ update their heights in the measurementStore, and optimistically assume that the item height was adjusted correctly in userland. This feels brittle.

I don't really like either of these solutions. 😕 But I'm also not sure what would be a better approach.

_Overlapping pins, I think because they're still honoring their original heights_
Screen Shot 2023-06-30 at 4 42 45 PM

_the whitespace here was so large that an additional pin could fit in it, leading to targeting the wrong item_
Screen Shot 2023-06-30 at 4 42 37 PM

### DIFF
--- a/docs/integration-test-helpers/masonry/ExampleGridItem.js
+++ b/docs/integration-test-helpers/masonry/ExampleGridItem.js
@@ -5,10 +5,16 @@ type Props = {|
   // $FlowFixMe[unclear-type]
   data: Object,
   expanded: boolean,
+  heightAdjustment?: number,
   itemIdx: number,
 |};
 
-export default function ExampleGridItem({ data = {}, itemIdx, expanded }: Props): Element<'div'> {
+export default function ExampleGridItem({
+  data = {},
+  heightAdjustment,
+  itemIdx,
+  expanded,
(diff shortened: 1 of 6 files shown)


AlbertCarreras:
I'm going to close it and sav the PR for future reference.

PR #3055, opened 2023-07-03
Internal: upgrade Flow to 0.210.2

dangerismycat:
I still want those nifty utility types I'd referenced in #3053. The Flow docs claim they were available in `≥0.209.0`, which seems to be a lie. Upgrading to the latest version to try to get `$Exclude`!

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -1,5 +1,5 @@
 [version]
-0.209.0
+0.210.2
 
 [ignore]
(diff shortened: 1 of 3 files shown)


dangerismycat:
Update: that utility type was probably present in `0.209.0`. Unintuitively, the new crop of utility types are _not_ prefixed with `$`, unlike all previous utility types. No idea why.

PR #3059, opened 2023-07-06
Box: add support for `<p>` in `as`

dangerismycat:
The [`<p>` tag](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/p) is useful for semantic text. We should support it!

### DIFF
--- a/docs/examples/box/as.js
+++ b/docs/examples/box/as.js
@@ -16,31 +16,42 @@ export default function Example(): Node {
           borderStyle="sm"
         >
           <Text color="light" weight="bold">
-            Top Nav Menu: as=&quot;nav&quot;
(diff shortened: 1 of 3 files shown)


docs/examples/box/as.js
```diff
@@ -16,31 +16,42 @@ export default function Example(): Node {
...
-            <Text>{'<article>'}</Text>
-            <Text>{'<h2>Article 1</h2>'}</Text>
-            <Text>{'</article>'}</Text>
+            <Text>
+              HTML output:
+              <br />
+              <code>
+                {'<article>'}
+                <br />
+                {'  <h2>Article 1</h2>'}
```
dangerismycat, line 50: The leading spaces are non-breaking, so they're respected and not collapsed

PR #3060, opened 2023-07-06
Column, Label, Letterbox, Mask: migrate docs examples to Sandpack

dangerismycat:


.eslintrc.js
```diff
@@ -142,6 +142,7 @@ module.exports = {
       'rules': {
         // Allows us to avoid directing the user off our docs site in examples
         'jsx-a11y/anchor-is-valid': OFF,
+        '@next/next/no-img-element': OFF,
```
dangerismycat, line 145: This isn't relevant for docs examples

PR #3071, opened 2023-07-11
Docs: Header toggle button placement

hectoid:
### Summary
Updated dark mode and dev mode button placement

#### What changed?

Moved the buttons to the right

#### Why?

Current placement was conflicting with the logo; moving to the right makes more sense as these are no longer primary buttons; size now also matches with size of the search field
image
image
image

### Links

- [Jira](https://jira.pinadmin.com/browse/GESTALT-6213)

- [Figma](https://www.figma.com/file/MilnJEDb9c8lM1N7hqxa5Y/Gestalt-site-Header-toggles?type=design&node-id=1%3A546&mode=design&t=GCjA2I7PzxyhJsGg-1)

### Checklist


AlbertCarreras:
@hectoid Thoughts about just making the IconButton selected?

hectoid:
> @hectoid Thoughts about just making the IconButton selected?

For both? I thought about it for the code one, but it would be weird to have one that toggles and another one that doesn't. But, the dev one is internal anyway, so probably fine. Just don't know if making the dark mode one also selected would make sense. It would be more closely aligned with our patterns in product.

AlbertCarreras (approved):

PR #3078, opened 2023-07-18
Button: fixes ` name` prop documentation and Flow

AlbertCarreras:
### Breaking change

Breaking change because before you could pass name to Button role="link" which was dead code. Now you cant.

```
--component=Button
--prop=name
```
We can use the above codemod to detect Button with name used on them and see if they are also role="link" but it's much faster to rely on Flow complaining.

#### What changed?

Button: fixes ` name` prop documentation and Flow

#### Why?

Previously missing

### DIFF
--- a/docs/pages/web/button.js
+++ b/docs/pages/web/button.js
@@ -117,6 +117,15 @@ export default function DocsPage({ generatedDocGen }: DocType): Node {
             ],
           },
           {
+            name: 'name',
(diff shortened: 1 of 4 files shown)


rlingineni (approved):
What does this prop do? Is it for forms, and is it required when a role link is set?

it was already there passed to buttons, does nothing to links. it was not previously passed.

PR #3079, opened 2023-07-18
Popover: Deprecating showcaret and color=blue

eniomoura:
#### What changed?

Renamed the Popover showcaret prop and the color blue for Popover, to allow for deprecation of remaining Popover instances in Pinboard still using those.

#### Why?

The remaining deprecation cases may break instances of Popover in Pinboard if migrated straight to PopoverEducational, so we decided to mark these for deprecation in Pinboard instead, and keep the remaining Popovers unchanged for now, while still allowing deprecation of the component.

### Links

- [Jira](https://jira.pinadmin.com/browse/GESTALT-6752)

### Checklist

- [ Yes ] Updated unit and Flow Tests
- [ Yes ] Updated documentation + accessibility tests


packages/gestalt/src/Popover.js
```diff
@@ -95,7 +95,7 @@ export default function Popover({
   color = 'white',
   role = 'dialog',
   shouldFocus = true,
-  showCaret = false,
+  _deprecatedShowCaret,
```
AlbertCarreras, line 98: you removed the `= false`

AlbertCarreras: this is prob failing the snapshot. i'd bring it back

AlbertCarreras (approved):
R+ with that default prop change pending revert

packages/gestalt/src/Popover.js
eniomoura: DONE

PR #3095, opened 2023-07-25
Text: Add CJK wrapping to Text

rlingineni:
### Summary

Adding CJK support to wrap `Text`.

#### What changed?

Introduces a css rule:
```
*:lang(zh) .Text,
*:lang(ko) .Text,
*:lang(ja) .Text {
  word-break: keep-all;
}
```

CJK languages should use a `keep-all` word-break to preserve the words. 

See the #Word-Breaks-and-CJK-languages section.

#### Why?

We noticed that in small width containers the text was not getting wrapped as expected. This creates a poor user experience.

Before:
image

After:
image

We still need to verify the downstream changes and verify if these changes break surfaces

### Links

- [TDD](link to Paper doc)
- [Figma](link to Figma file)

### Checklist

### DIFF
--- a/docs/examples/text/cjkText.js
+++ b/docs/examples/text/cjkText.js
@@ -1,0 +1,102 @@
+// @flow strict
+import React from 'react';
+import { Box, Flex, Text } from 'gestalt';
+
+export default function CJK(): Node {
+  // eslint-disable-next-line react/no-unstable-nested-components
+  function NarrowShell({ text, useKeepAll }: {| text: string, useKeepAll: boolean |}) {
+    return (
+      <Box width={24} color="dark">
+        <Text overflow="normal" size="200" color="inverse">
+          {useKeepAll ? <div style={{ wordBreak: 'keep-all' }}>{text}</div> : text}
+        </Text>
+      </Box>
+    );
+  }
+
+  const languages = [
+    {
+      id: 'English',
+      code: 'en',
+      lang: 'Your account options and more',
+    },
+    {
+      id: 'Chinese',
+      code: 'zh',
+      lang: '帐户和更多选项',
+    },
+    {
+      id: 'Korean',
+      code: 'ko',
+      lang: '계정 및추 가옵션',
+    },
+  ];
+
+  return (
+    <Box
+      marginTop={10}
+      padding={4}
+      width="100%"
+      height="100%"
+      display="flex"
+      justifyContent="center"
+      alignItems="center"
+      direction="column"
+    >
+      <Flex direction="column" gap={8}>
+        <Box>
+          <Box marginBottom={2}>
+            <Text weight="bold">With Keep-All</Text>
+          </Box>
+          <Flex direction="row" gap={4}>
+            {languages.map(({ id, lang, code }) => (
+              <Box key={id}>
+                <Box marginBottom={2}>
+                  <Text size="200">{id}</Text>
+                </Box>
+                <div lang={code}>
+                  <Box key={id} width={120} color="dark" rounding={2} padding={2}>
+                    <NarrowShell text={lang} />
+                  </Box>
+                </div>
+              </Box>
+            ))}
+          </Flex>
+        </Box>
+
+        <Box>
+          <Box marginBottom={2}>
+            <Text weight="bold">Without Keep-All</Text>
+          </Box>
+          <Flex direction="row" gap={4}>
+            {languages.map(({ id, lang }) => (
+              <Box key={id}>
+                <Box marginBottom={2}>
+                  <Text size="200">{id}</Text>
+                </Box>
+                <Box key={id} width={120} color="dark" rounding={2} padding={2}>
+                  <NarrowShell text={lang} />
+                </Box>
+              </Box>
+            ))}
+          </Flex>
+        </Box>
+      </Flex>
+    </Box>
+  );
+
+  /*
+  return (
+    <Box width="100%" height="100%" display="flex" direction="column" alignItems="center">
+      <Box padding={2}>
+        <Box rounding={2} padding={2}>
+          <Flex direction="column" gap={8}>
+            <NarrowToolTipShell text={kor} />
+            <NarrowToolTipShell text={eng} />
+            <NarrowToolTipShell useKeepAll text={kor} />
+          </Flex>
+        </Box>
+      </Box>
+    </Box>
+  ) */
+}
--- a/docs/examples/text/cjkText.js
+++ b/docs/examples/text/cjkText.js
@@ -3,16 +3,6 @@ import React, { type Node } from 'react';
 import { Box, Flex, Text } from 'gestalt';
 
 export default function cjkText(): Node {
-  function NarrowShell({ text, useKeepAll }: {| text: string, useKeepAll?: boolean |}) {
-    return (
-      <Box width={24} color="dark">
-        <Text overflow="normal" size="200" color="inverse">
-          {useKeepAll ? <div style={{ wordBreak: 'keep-all' }}>{text}</div> : text}
-        </Text>
-      </Box>
-    );
-  }
-
   const languages = [
     {
       id: 'English',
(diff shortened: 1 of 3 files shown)


docs/pages/web/text.js
```diff
@@ -185,6 +186,15 @@ export default function TextPage({ generatedDocGen }: {| generatedDocGen: DocGen
           title="Text-wrapping and hyphenation"
           description="Hyphenation on iOS is turned off by default to avoid incorrect word breaks when strings of text wrap to the next line. This is especially helpful for international languages where an incorrect word break can greatly change the meaning of a word or sentence."
         />
+
+        <MainSection.Subsection
+          title="Word Breaks and CJK languages"
+          description={`Chinese, Japanese and Korean languages and may not be broken in the middle of a word. This creates text-wrapping issues and tall containers. To avoid this, the Text component uses the CSS property \`word-break:keep-all\` to keep words on CJK languages together when CJK is detected.
```
AlbertCarreras, line 192: languages **and** may

typo

AlbertCarreras: the Text component > Text

docs/pages/web/text.js
```diff
@@ -185,6 +186,15 @@ export default function TextPage({ generatedDocGen }: {| generatedDocGen: DocGen
           title="Text-wrapping and hyphenation"
           description="Hyphenation on iOS is turned off by default to avoid incorrect word breaks when strings of text wrap to the next line. This is especially helpful for international languages where an incorrect word break can greatly change the meaning of a word or sentence."
         />
+
+        <MainSection.Subsection
+          title="Word Breaks and CJK languages"
```
AlbertCarreras, line 191: This isn't necessarily a variant right? It's more like a _test_ example.

Would "Word breaks" be enough if we add it under localization and we specify in the description what CJK languages stands for?

packages/gestalt/src/Text.css
```diff
@@ -1,3 +1,9 @@
 .Text {
   composes: antialiased sansSerif from "./Typography.css";
 }
+
+*:lang(zh) .Text,
```
AlbertCarreras, line 5: can we add a comment above explaining what this is preveinting?

docs/pages/web/text.js
AlbertCarreras: I kinda fill this should go on the Quality Checklist under "Internationalization", kinda like RTL.

packages/gestalt/src/Text.css
AlbertCarreras: Same for Headings?

docs/pages/web/text.js
```diff
@@ -185,6 +186,15 @@ export default function TextPage({ generatedDocGen }: {| generatedDocGen: DocGen
           title="Text-wrapping and hyphenation"
```
AlbertCarreras, line 186: this section is pretty much irrelevant right>

          title="Text-wrapping and hyphenation" Hyphenation on iOS is turned off by default...

docs/pages/web/text.js
AlbertCarreras: Or even here: Foundations > Typography

https://gestalt.pinterest.systems/foundations/typography/guidelines#Best-practices

AlbertCarreras: I think the examples would be great in best practices here.

If Text handles it all internally, it shuoldn't be a variant.

docs/pages/web/text.js
```diff
@@ -185,6 +186,15 @@ export default function TextPage({ generatedDocGen }: {| generatedDocGen: DocGen
...
           description="Hyphenation on iOS is turned off by default to avoid incorrect word breaks when strings of text wrap to the next line. This is especially helpful for international languages where an incorrect word break can greatly change the meaning of a word or sentence."
         />
+
+        <MainSection.Subsection
+          title="Word Breaks and CJK languages"
+          description={`Chinese, Japanese and Korean languages and may not be broken in the middle of a word. This creates text-wrapping issues and tall containers. To avoid this, the Text component uses the CSS property \`word-break:keep-all\` to keep words on CJK languages together when CJK is detected.  
+          <br /> Be sure to consider text overflow, and making sure a container is wide enough to support a CJK term.
+          `}
+        >
+          <SandpackExample code={cjkText} name="CJK Text" hideEditor hideControls />
```
AlbertCarreras, line 196: consider:

1. making the container sandbox taller previewHeight higher
2. breaking the example in a side by side 2 column example

docs/examples/text/cjkText.js
```diff
@@ -0,0 +1,86 @@
...
+    {
+      id: 'Korean',
+      code: 'ko',
+      lang: '계정 및추 가옵션',
+    },
+  ];
+
+  return (
+    <Box
+      marginTop={10}
```
AlbertCarreras, line 36: why marginTop={10}?

docs/examples/text/cjkText.js
```diff
@@ -0,0 +1,86 @@
...
+      padding={4}
+      width="100%"
+      height="100%"
+      display="flex"
+      justifyContent="center"
+      alignItems="center"
+      direction="column"
+    >
+      <Flex direction="column" gap={8}>
+        <Box>
```
AlbertCarreras, line 46: line 46 can we do Flex here with gap?

docs/examples/text/cjkText.js
```diff
@@ -0,0 +1,86 @@
...
+          <Box marginBottom={2}>
```
AlbertCarreras, line 47: marginBottom={2} might not be needed with Flex'sgap

docs/examples/text/cjkText.js
```diff
@@ -0,0 +1,86 @@
...
+          </Box>
+          <Flex direction="row" gap={4}>
+            {languages.map(({ id, lang, code }) => (
+              <Box key={id}>
+                <Box marginBottom={2}>
+                  <Text size="200">{id}</Text>
+                </Box>
+                <div lang={code}>
+                  <Box key={id} width={120} color="dark" rounding={2} padding={2}>
+                    <NarrowShell text={lang} />
```
AlbertCarreras, line 58: this narrow shell doens't seem to save much, I'd inline for easier readeability

docs/pages/web/text.js
rlingineni: Agree! I'm not even sure what the best practice here is though. It's just a consideration maybe?

docs/pages/web/text.js
rlingineni: Hard for me to say, because it's been around for a long time.

docs/pages/web/text.js
rlingineni: Since this `keep-all` originates in the text component, I think it's good to call out here

docs/pages/web/text.js
rlingineni: fixed

docs/examples/text/cjkText.js
rlingineni: removed

docs/pages/web/text.js
rlingineni: fixed

packages/gestalt/src/Text.css
rlingineni: Anywhere the .Text class is applied

rlingineni: Let's test this on an alpha release actually!

docs/examples/text/cjkText.js
rlingineni: Fixed

docs/examples/text/cjkText.js
rlingineni: removed with above change

rlingineni (commented):
Made changes, yeah re-where this should live. I think with `Text` mainly because it should.

Also, re-thinking this. I think since this can break localization, this is a good alpha-candidate, and I can find someone from ii8n to test it as well

AlbertCarreras (approved):

PR #3103, opened 2023-07-26
DateField: Add locale format translations

rlingineni:
### Summary

Not sure what the impact of the namespace import here could be.

#### What changed?

We take the code from localeData, and make it so that we can get it from the MUI locales object, and pass it in to `localeText`. A light transformation is needed between the two.

#### Why?

For birthday picking, translations are an issue.

Before:
image

After:
image

### Checklist

### DIFF
--- a/packages/gestalt-datepicker/src/DateField/InternalDateField.js
+++ b/packages/gestalt-datepicker/src/DateField/InternalDateField.js
@@ -246,6 +246,9 @@ function InternalDateField({
   value,
 }: InternalDateFieldProps): Node {
   let translations;
+
+  // converts date-fns localeData.code (e.g. es-EN) from to the format expected by the MUI Locale (esEN)
+  // https://mui.com/x/react-date-pickers/localization/
   if (localeData && localeData.code) {
     // turns en-US to enUS
     const split = localeData.code.split('-');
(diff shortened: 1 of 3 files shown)


packages/gestalt-datepicker/src/DateField/InternalDateField.js
```diff
@@ -243,9 +245,28 @@ function InternalDateField({
   readOnly = false,
   value,
 }: InternalDateFieldProps): Node {
+  let translations;
+  if (localeData && localeData.code) {
+    // turns en-US to enUS
+    const split = localeData.code.split('-');
+    if (split.length === 1) {
+      // turns 'es' into 'enES'
```
AlbertCarreras, line 253: Can you elaborater more on why this? Are there discrepancies in locale codes?

AlbertCarreras (approved):

packages/gestalt-datepicker/src/DateField/InternalDateField.js
```diff
@@ -243,9 +245,28 @@ function InternalDateField({
   readOnly = false,
   value,
 }: InternalDateFieldProps): Node {
+  let translations;
+  if (localeData && localeData.code) {
+    // turns en-US to enUS
+    const split = localeData.code.split('-');
```
AlbertCarreras, line 251: coudl we do here just 

`.replace("-", "")`

PR #3111, opened 2023-07-27
Popover: add "forceDown" to `idealDirection`

AlbertCarreras:
#### What changed?

Popover: add "forceDown" to `idealDirection`\

AFTER THIS ILL DO A FOLLOW-UP PR CLEANING UP POPOVER EXAMPLES

#### Why?

Some edge cases require forcing Popover's position down


packages/gestalt/dist/index.d.ts
```diff
@@ -1411,7 +1413,7 @@ interface PopoverProps {
   children?: Node | undefined;
   color?: 'deprecatedBlue' | 'red' | 'white' | 'darkGray' | undefined;
   id?: string | undefined;
-  idealDirection?: FourDirections | undefined;
+  idealDirection?: FiveDirections | undefined;
```
rlingineni, line 1416: lol, can we just call this PopoverDirections?

packages/gestalt/src/utils/positioningUtils.js
```diff
@@ -160,13 +160,13 @@ export function getPopoverDir({
   const skipNoEdgeCondition =
     isScrollBoundaryContainer && (nonTopEdge || nonBottomEdge) && (nonLeftEdge || nonRightEdge);
 
-  if (!skipNoEdgeCondition && (nonTopEdge || nonBottomEdge)) {
+  if (idealDirection !== 'forceDown' && !skipNoEdgeCondition && (nonTopEdge || nonBottomEdge)) {
```
rlingineni, line 163: might be possible to wrap this condition on the outside of both statements

```
if(idealDir !=== forceDown){
// do the internal logic
}

File: packages/gestalt/src/utils/positioningUtils.js [hunk 3]
```diff
@@ -176,9 +176,16 @@ export function getPopoverDir({
...
+  if (
+    idealDirection &&
+    idealDirection !== 'forceDown' &&
+    spaces[DIR_INDEX_MAP[idealDirection]] > 0
+  ) {
     // user pref
     popoverDir = idealDirection;
+  } else if (idealDirection && idealDirection === 'forceDown') {
+    // user pref
+    popoverDir = 'down';
```
rlingineni, line 188: if this is the only logic, then can you just do if(idealDirection === 'forceDown') can you just return at the top. No need for all the other logic right?

packages/gestalt/src/Caret.js
```diff
@@ -10,15 +10,15 @@ const PATHS = {
 };
 
 type Props = {|
-  direction: 'up' | 'right' | 'down' | 'left',
+  direction: 'up' | 'right' | 'down' | 'left' | 'forceDown',
```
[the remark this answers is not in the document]
AlbertCarreras: Caret is not a component in TS, it's internal. TS only covers external APIs, Types declaratoins

packages/gestalt/dist/index.d.ts
AlbertCarreras: DONE

packages/gestalt/src/utils/positioningUtils.js
AlbertCarreras: this was a refactor, done but kinda OOScope

AlbertCarreras: DONE

packages/gestalt/src/utils/positioningUtils.js
AlbertCarreras: DONE

packages/gestalt/src/utils/positioningUtils.js
```diff
@@ -160,35 +160,41 @@ export function getPopoverDir({
...
+  if (maybeEdgeCondition && (nonLeftEdge || nonRightEdge)) {
     up = 0;
     down = 0;
   }
 
   // Choose the main direction for the popover based on available spaces & user preference
   const spaces = [up, right, down, left];
 
-  let popoverDir;
+  if (idealDirection && idealDirection === 'forceDown') {
```
rlingineni, line 179: If you just put this at the top, do you need the `maybeEdgeCondition` case at all

rlingineni (approved):

PR #3115, opened 2023-07-28
DatePicker: support controlled/uncontrolled component

AlbertCarreras:
#### What changed?

DatePicker: support controlled/uncontrolled component

#### Why?

DatePicker was neither a fully controlled/uncontrolled component

https://jira.pinadmin.com/browse/BUG-154540?jql=project%20%3D%20BUG%20AND%20resolution%20%3D%20Unresolved%20AND%20component%20%3D%20Gestalt%20%20ORDER%20BY%20%20priority%20DESC%2C%20updated%20DESC

### DIFF
--- a/docs/pages/web/datepicker.js
+++ b/docs/pages/web/datepicker.js
@@ -98,10 +98,9 @@ const localeMap = {
   'zh-CN': { localeData: zhCN, lang: 'Chinese (Simplified)' },
   'zh-TW': { localeData: zhTW, lang: 'Chinese (Traditional)' },
 };
-const PREVIEW_HEIGHT = 420;
-const PREVIEW_HEIGHT_HELPERTEXT = 450;
+const PREVIEW_HEIGHT = 480;
 
-export default function DatePickerPage({ generatedDocGen }: {| generatedDocGen: DocGen |}): Node {
+export default function DocsPage({ generatedDocGen }: {| generatedDocGen: DocGen |}): Node {
   return (
     <Page title="DatePicker">
       <PageHeader name="DatePicker" description={generatedDocGen?.description}>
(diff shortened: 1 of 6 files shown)


packages/gestalt-datepicker/src/DatePicker/InternalDatePicker.js
```diff
@@ -38,26 +38,21 @@ const InternalDatePickerWithForwardRef: AbstractComponent<Props, HTMLInputElemen
...
-    value: dateValue,
+    value: controlledValue,
   }: Props,
   ref,
 ): Element<'div'> {
   const innerInputRef = useRef<null | HTMLInputElement>(null);
   useImperativeHandle(ref, () => innerInputRef.current);
 
-  const [selected, setSelected] = useState<?Date>(dateValue);
+  const [uncontrolledValue, setUncontrolledValue] = useState<?Date>(null);
```
rlingineni, line 48: nit: I think we should use the original variable naming, since this feels very generic and value seems like it can mean anything

docs/pages/web/datepicker.js
```diff
@@ -140,9 +139,20 @@ export default function DatePickerPage({ generatedDocGen }: {| generatedDocGen:
...
+DatePicker is controlled when \`value\` is not "undefined". When \`value\` is "undefined", it stays uncontrolled.
+        `}
+        >
           <MainSection.Card
             title="Empty input"
+            description={`
+If DatePicker doesn't present pre-selected date values, initialize \`value\` with "null" so the component is controlled.
+`}
             sandpackExample={
               <SandpackExample
```
rlingineni, line 157: Can we hide the code controls on this example or add a header inside the example?

rlingineni (commented):
This one, I'm not sure what's really happening. 

Since with the old example we were able achieve null state and it also had controlled

packages/gestalt-datepicker/src/DatePicker/InternalDatePicker.js
AlbertCarreras: Disagree. 

controlled and uncontrolled components have `selected` values or dates. 

Here i'm trying to differentiate the values used when the component works as controlled  and uncontrolled

AlbertCarreras: and it's naming. I could go _uncontrolledSelectedValue_ and _controlledSelectedValue_.

value is the standard for the value we pass to fields

docs/pages/web/datepicker.js
AlbertCarreras: 1. code is shown w=in variants
2. we never hide controls
3. This is out of scope, I'll let you decide in the docs comprehension improvement if we should bring to the top the titles of the examples. Otherwise I'm breking consistency

rlingineni (approved):
thanks!

PR #3116, opened 2023-07-28
Table: Headercell take full width

rlingineni:
Bug fix:

### Summary
Changes the header cell to take the full width, so end-align works as expected.

#### What changed?

Adding width 100%.

Getting rid of row-reverse since it's not accessible, and replacing with Icon rendering.

#### Why?

Before:

image

After:
image

### Links

- [TDD](link to Paper doc)
- [Figma](link to Figma file)

### Checklist

### DIFF
--- a/docs/examples/table/sortableTableCells.js
+++ b/docs/examples/table/sortableTableCells.js
@@ -18,6 +18,11 @@ export default function SortableHeaderExample(): Node {
   return (
     <Box height="100%" display="flex" justifyContent="center" alignItems="center">
       <Table accessibilityLabel="Sortable header cells">
+        <colgroup>
+          <col span="1" style={{ width: '10%' }} />
+          <col span="1" style={{ width: '10%' }} />
+          <col span="1" style={{ width: '80%' }} />
+        </colgroup>
         <Table.Header>
           <Table.Row>
             <Table.SortableHeaderCell
@@ -53,9 +58,7 @@ export default function SortableHeaderExample(): Node {
             <Text>Snax</Text>
           </Table.Cell>
           <Table.Cell>
-            <Box width={100}>
-              <Text align="end">$50</Text>
-            </Box>
+            <Text align="end">$50</Text>
           </Table.Cell>
         </Table.Row>
       </Table>
--- a/docs/examples/table/sortableTableCells.js
+++ b/docs/examples/table/sortableTableCells.js
@@ -53,7 +53,9 @@ export default function SortableHeaderExample(): Node {
             <Text>Snax</Text>
           </Table.Cell>
           <Table.Cell>
-            <Text align="end">$50</Text>
+            <Box width={100}>
+              <Text align="end">$50</Text>
+            </Box>
           </Table.Cell>
         </Table.Row>
       </Table>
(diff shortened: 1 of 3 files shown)


packages/gestalt/src/TableSortableHeaderCell.js
```diff
@@ -102,30 +130,24 @@ export default function TableSortableHeaderCell({
...
-            display="flex"
-            alignItems="center"
-            justifyContent={align}
-            dangerouslySetInlineStyle={{
-              __style: { flexDirection: align === 'end' ? 'row-reverse' : 'row' },
-            }}
-          >
+          <Box display="flex" alignItems="center" justifyContent={align}>
+            {align === 'end' && (
+              <SortIcon
```
rlingineni, line 135: Had to do this because `row-reverse` isn't accessible

docs/examples/table/sortableTableCells.js
```diff
@@ -53,7 +53,9 @@ export default function SortableHeaderExample(): Node {
             <Text>Snax</Text>
           </Table.Cell>
           <Table.Cell>
-            <Text align="end">$50</Text>
+            <Box width={100}>
```
AlbertCarreras, line 56: instead of doing width={100} ca we use 

```
        <colgroup>
          <col span="1" style={{ width: '60%' }} />
          <col span="1" style={{ width: '15%' }} />
          <col span="1" style={{ width: '15%' }} />
          <col span="1" style={{ width: '15%' }} />
        </colgroup>

```

packages/gestalt/src/TableSortableHeaderCell.js
AlbertCarreras: Can you add a comments in the descripton about what not accessible means.

packages/gestalt/src/TableSortableHeaderCell.js
```diff
@@ -59,6 +59,34 @@ type Props = {|
...
+  return (
+    <Box
+      marginStart={align === 'start' ? 2 : undefined}
+      marginEnd={align === 'end' ? 2 : undefined}
+      dangerouslySetInlineStyle={{
+        __style: { visibility },
+      }}
+    >
+      <Icon
+        accessibilityLabel=""
```
AlbertCarreras, line 82: Why is accessibilityLabel="" empty?

Shouldn't it be translated?

AlbertCarreras: Check video re accessility, I think a11y is broken. https://capture.dropbox.com/HB4Talz6UVIkYEXZ

AlbertCarreras (approved):
Lzgtm with nits and a out of scope issue to explore

packages/gestalt/src/TableSortableHeaderCell.js
rlingineni: Agree, it's not perfect. It's been like that for a while, and it might be a separate item. I actually noticed empty accessbility labels on a couple of other icons as well like in SideNav btw

Also, I took a look at video.  But some of the w3 examples (e.g. mute) seem to behave similar:
https://www.w3.org/TR/2016/WD-wai-aria-practices-1.1-20160317/examples/button/button.html

I think we need to do this two -fold:
1. Add role toggle button to TapArea
2. Put the label in the accessbility provider, but it should be dynamic, so I feel like it won't work there.

PR #3121, opened 2023-08-04
WONT MERGE ActivationCard, Callout, OverlayPanel, PopoverEducational, SheetMobile, SlimBanner, Toast, Upsell: extended dataTestId from Button

AlbertCarreras:
#### What changed?

Component: extended dataTestId

#### Why?

Extend Button's dataTestId to components using Button internally

### DIFF
--- a/packages/gestalt/src/OverlayPanel/ConfirmationPopover.js
+++ b/packages/gestalt/src/OverlayPanel/ConfirmationPopover.js
@@ -17,6 +17,7 @@ type Props = {|
   onDismiss: () => void,
   primaryAction?: {|
     accessibilityLabel?: string,
+    dataTestId?: string,
     text?: string,
     onClick?: ({|
       event:
@@ -28,6 +29,7 @@ type Props = {|
   |},
   secondaryAction?: {|
     accessibilityLabel?: string,
+    dataTestId?: string,
     text?: string,
     onClick?: ({|
       event:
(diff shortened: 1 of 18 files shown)


packages/gestalt/src/OverlayPanel/ConfirmationPopover.js
```diff
@@ -17,6 +17,7 @@ type Props = {|
   onDismiss: () => void,
   primaryAction?: {|
     accessibilityLabel?: string,
+    dataTestId?: string,
```
rlingineni, line 20: Is this public?

rlingineni: Oh I see it's passed as part of the sub props?

rlingineni (approved):
Seems okay to me. I just wish there was a more scalable approach here. Is the plan to add this to everything and everywhere?

Is there a plan to add some notion of "GlobalProps" that we can have as a base everywhere

AlbertCarreras:
@rlingineni I'm going to revisit Ryan's PR for testing ids. I remember looking at it. I'm not sure if there might be a more scalable approach. 

I don't think this is for all components, mostly. I'd like to do some evaluation first. Have some strategy. I did Button and IconButton (Button was already done) to help me prototype for SterlingTheme project, cos I needed to remove the wrapped for testing first.

But originally I was thinking to write a proposal for a better testing strategy.

AlbertCarreras:
https://github.com/pinterest/gestalt/pull/3042

PR #3122, opened 2023-08-07
GlobalEventsHandlerProvider, DatePicker, DateRange, DateField: add `onMount` handlers

AlbertCarreras:
#### What changed?

GlobalEventsHandlerProvider, DatePicker, DateRange, DateField: add `onMount` handlers

Documented GlobalEventsHandlerProvider (external handlers) in some components pages

#### Why?

Enable impression logging tests

### DIFF
--- a/packages/gestalt-datepicker/src/DatePicker.js
+++ b/packages/gestalt-datepicker/src/DatePicker.js
@@ -4,9 +4,11 @@ import {
   type Element,
   forwardRef,
   type Node,
+  useEffect,
   useImperativeHandle,
   useRef,
 } from 'react';
+import { useGlobalEventsHandler } from 'gestalt';
 import InternalDatePicker from './DatePicker/InternalDatePicker.js';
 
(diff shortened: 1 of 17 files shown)


rlingineni (approved):

packages/gestalt-datepicker/src/DatePicker.js
```diff
@@ -173,6 +175,13 @@ const DatePickerWithForwardRef: AbstractComponent<Props, HTMLInputElement> = for
 ): Node {
   const innerInputRef = useRef<null | HTMLInputElement>(null);
   useImperativeHandle(ref, () => innerInputRef.current);
+  const { datePickerHandlers } = useGlobalEventsHandler() || {
+    datePickerHandlers: undefined,
+  };
+
+  useEffect(() => {
+    if (datePickerHandlers?.onMount) datePickerHandlers?.onMount();
```
rlingineni, line 183: This may get called in non-`onMount` situations right, e.g if another handler is changed?

rlingineni: should this just be. Not sure if onMount should be changeable 
```
useEffect()=>{
 if (datePickerHandlers?.onMount) datePickerHandlers?.onMount();
},[])
```

PR #3125, opened 2023-08-08
DateField: fix locales in MM/DD/YYYY format, upgraded package, added unit tests

AlbertCarreras:
#### What changed?

DateField: fix locales in MM/DD/YYYY format, upgraded package, added unit tests

Assessment: https://coda.io/d/_d2LeXkQ1kVX/DateField_sukKr#_luoyc

BEFORE
Screenshot by Dropbox Capture

AFTER
Screenshot by Dropbox Capture

### DIFF
--- a/packages/gestalt-datepicker/src/DateField/InternalDateField.js
+++ b/packages/gestalt-datepicker/src/DateField/InternalDateField.js
@@ -212,6 +212,47 @@ type LocaleData = {|
   |},
 |};
 
+const getTranslationsFromMUIJS: (?LocaleData) => ?{|
+  fieldYearPlaceholder: (params: {| digitAmount: number |}) => string,
+  fieldMonthPlaceholder: (params: {| contentType: string |}) => string,
+  fieldDayPlaceholder: () => string,
+|} = (localeData) => {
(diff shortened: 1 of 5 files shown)


packages/gestalt-datepicker/src/DateField/InternalDateField.js
```diff
@@ -263,6 +285,17 @@ function InternalDateField({
     }
```
rlingineni, line 285: I think we should split out into separate functions for readability

```
let translations = getTranslationsFromMUIJS(code);
 if(!locale) {
  translations = getLocalTranslations(code)
 }
```

rlingineni (approved):
Works, some comments on possible re-ordering.

Can you also add a test for a MUI JS locale and non MUI-JS, and incorrect locale code

packages/gestalt-datepicker/src/DateField/InternalDateField.js
AlbertCarreras: DONE

AlbertCarreras:
> Works, some comments on possible re-ordering.
> 
> Can you also add a test for a MUI JS locale and non MUI-JS, and incorrect locale code

speaking of tests ha

PR #3129, opened 2023-08-10
ButtonLink: PageHeader fix

eniomoura:
#### What changed?
Fixed PageHeader type list

#### Why?
Allowing ButtonLinks to be used as action types for PageHeader, since it behaves as a Button.

### Links
https://jira.pinadmin.com/browse/GESTALT-6806

### DIFF
--- a/packages/gestalt/src/PageHeader.js
+++ b/packages/gestalt/src/PageHeader.js
@@ -3,6 +3,7 @@ import { type Element, type Node } from 'react';
 import Box from './Box.js';
 import { type Dimension } from './boxTypes.js';
 import Button from './Button.js';
+import ButtonLink from './ButtonLink.js';
 import Dropdown from './Dropdown.js';
 import Flex from './Flex.js';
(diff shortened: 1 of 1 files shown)


AlbertCarreras (approved):

packages/gestalt/src/PageHeader.js
```diff
@@ -22,7 +23,12 @@ import Text from './Text.js';
 import Tooltip from './Tooltip.js';
 
 export type ActionType = Element<
-  typeof Button | typeof IconButton | typeof Link | typeof Text | typeof Tooltip,
+  | typeof Button
```
AlbertCarreras, line 26: fix typescript too

AlbertCarreras (approved):

packages/gestalt/src/PageHeader.js
eniomoura: DONE

AlbertCarreras (approved):

PR #3133, opened 2023-08-14
Docs: Use TableOfContents component in ToC 

diyorbek:
### Summary

Replaced custom table of contents component with the new `TableOfContents` component in docs site `ToC`.

BEFORE:
ezgif com-video-to-gif (1)

AFTER:
ezgif com-video-to-gif

### Links

- [Jira](https://jira.pinadmin.com/browse/GESTALT-6267)

### Checklist

- [n/a] Added unit and Flow Tests
- [n/a] Added documentation + accessibility tests

### DIFF
--- a/docs/docs-components/Toc.js
+++ b/docs/docs-components/Toc.js
@@ -72,8 +72,16 @@ function useThrottledOnScroll(callback: null | (() => void), delay: number) {
 type ToCItem = {|
   id: string,
   label: string,
-  // eslint-disable-next-line flowtype/no-mutable-array
-  children: Array<{| id: string, label: string |}>,
+  children: $ReadOnlyArray<{| id: string, label: string |}>,
+|};
+
+type HandleClickParameters = {|
+  hash: string,
+  event:
+    | SyntheticMouseEvent<HTMLDivElement>
+    | SyntheticKeyboardEvent<HTMLDivElement>
+    | SyntheticMouseEvent<HTMLAnchorElement>
+    | SyntheticKeyboardEvent<HTMLAnchorElement>,
 |};
 
 type Props = {|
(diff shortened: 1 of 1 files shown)


docs/docs-components/Toc.js
```diff
@@ -69,6 +69,13 @@ function useThrottledOnScroll(callback: null | (() => void), delay: number) {
   }, [throttledCallback]);
 }
 
+type ToCItem = {|
+  id: string,
+  label: string,
+  // eslint-disable-next-line flowtype/no-mutable-array
```
AlbertCarreras, line 75: Can we fix this?

 $ReadOnlyArray<....>

docs/docs-components/Toc.js
```diff
@@ -122,7 +129,11 @@ export default function Toc({ cards }: Props): Node {
 
   const handleClick = (
```
AlbertCarreras, line 130: can we use named arguments? `({ hash,  event })`

docs/docs-components/Toc.js
diyorbek: If we use readonly array then the algorithm [here](https://github.com/pinterest/gestalt/pull/3133/files#diff-8c97d01b6c6c7a51b95de107d690111f4e5485389b5166c8a2ff19135fc7fe4cR179) will not make sense. I tried making it readonly and I don't think coming up with "less straightforward" approach just to satisfy Flow is a good idea.

docs/docs-components/Toc.js
diyorbek: yes

AlbertCarreras (approved):

docs/docs-components/Toc.js
AlbertCarreras: give it a couple thoughts, we can merge and revisit this later on as a clean up

PR #3136, opened 2023-08-15
Text: Add breakAll option to Text

rlingineni:
### Summary

Text didn't support a `BreakAll` option. 

And lineClamp by default only supported `breakWord` It was in the added in [this PR](https://github.com/pinterest/gestalt/pull/1565), and then later changed to became a [break-word](https://github.com/pinterest/gestalt/pull/1577).

We do have a few instances in Pinboard where LineClamp is used with `overflow="normal"`, but the word-break is actually `break-word` because it's used alongside lineClamp. This PR makes the `breakWord` its own class, and it is added by default, alongside lineClamp to preserve existing behavior.

However, it also adds support for `break-all` as an option to be used alongside `lineClamp`

#### Why?

Request here in [Slack](https://pinterest.slack.com/archives/C13KLG5P0/p1690853063014349). To support `lineClamp` with breakAll

Before:
image

After:
image

### Links

- [TDD](link to Paper doc)
- [Figma](link to Figma file)

### Checklist

### DIFF
--- a/packages/gestalt/src/Text.js
+++ b/packages/gestalt/src/Text.js
@@ -100,11 +100,18 @@ const TextWithForwardRef: AbstractComponent<Props, HTMLElement> = forwardRef<Pro
   ): Element<As> {
     const colorClass = semanticColors.includes(color) && colors[`${color}Text`];
 
-    // by default, lineClamp should use breakWord
-    const lineClampStyles = cx({
-      [typography.lineClamp]: isNotNullish(lineClamp),
-      [typography.breakWord]: isNotNullish(lineClamp) && overflow !== 'breakAll',
-    });
+    const getWordBreakStyle = (): string | void => {
+      if (overflow === 'breakAll') {
+        return typography.breakAll;
+      }
+
+      // default to breakWord if lineClamp is set
+      if (overflow === 'breakWord' || isNotNullish(lineClamp)) {
+        return typography.breakWord;
+      }
+
+      return undefined;
+    };
 
     const cs = cx(
       styles.Text,
@@ -116,14 +123,13 @@ const TextWithForwardRef: AbstractComponent<Props, HTMLElement> = forwardRef<Pro
       align === 'end' && typography.alignEnd,
       align === 'forceLeft' && typography.alignForceLeft,
       align === 'forceRight' && typography.alignForceRight,
-      overflow === 'breakWord' && typography.breakWord,
-      overflow === 'breakAll' && typography.breakAll,
+      getWordBreakStyle(),
       overflow === 'noWrap' && typography.noWrap,
       italic && typography.fontStyleItalic,
       underline && typography.underline,
       weight === 'bold' && typography.fontWeightSemiBold,
       weight === 'normal' && typography.fontWeightNormal,
-      lineClampStyles,
+      isNotNullish(lineClamp) && typography.lineClamp,
     );
 
     const Tag: As = inline ? 'span' : 'div';
--- a/packages/gestalt/src/Text.js
+++ b/packages/gestalt/src/Text.js
@@ -11,7 +11,7 @@ function isNotNullish(val: ?number): boolean {
 }
 
 type As = 'span' | 'div';
-type Overflow = 'normal' | 'breakWord' | 'breakAll' | 'noWrap';
+type Overflow = 'normal' | 'breakAll' | 'breakWord' | 'noWrap';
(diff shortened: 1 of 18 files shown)


packages/gestalt/src/Typography.css
```diff
@@ -75,7 +75,11 @@
 /* overflow */
 
 .breakWord {
-  word-wrap: break-word;
```
AlbertCarreras, line 78: word-wrap: break-word; & word-break: break-word; are equivalent. can we add that to the description?

https://www.w3schools.com/cssref/playdemo.php?filename=playcss_word-break&preval=break-word
https://www.w3schools.com/cssref/playdemo.php?filename=playcss_word-wrap&preval=break-word

packages/gestalt/src/Text.js
```diff
@@ -111,12 +117,13 @@ const TextWithForwardRef: AbstractComponent<Props, HTMLElement> = forwardRef<Pro
...
       align === 'forceRight' && typography.alignForceRight,
       overflow === 'breakWord' && typography.breakWord,
+      overflow === 'breakAll' && typography.breakAll,
       overflow === 'noWrap' && typography.noWrap,
       italic && typography.fontStyleItalic,
       underline && typography.underline,
       weight === 'bold' && typography.fontWeightSemiBold,
       weight === 'normal' && typography.fontWeightNormal,
-      isNotNullish(lineClamp) && typography.lineClamp,
+      lineClampStyles,
```
AlbertCarreras, line 126: could we refactor to prevent overrides?

if overflow === 'breakWord' and isLineClamp technically we would be setting breakWord twice
overflow === 'breakWord' && typography.breakWord
&
[typography.breakWord]: isNotNullish(lineClamp) && overflow !== 'breakAll',

Maybe?
`
const isLineClamp = isNotNullish(lineClamp)
...
overflow === 'breakAll' && typography.breakAll,      
{
[typography.breakWord]: overflow === 'breakWord' ||  (isLineClamp && overflow !== 'breakAll')
[typography.lineClamp]: isLineClamp
}
`

or 

...
overflow === 'breakAll' && typography.breakAll, 
(overflow === 'breakWord' ||  (isLineClamp && overflow !== 'breakAll')) && typography.breakWord,
isLineClamp && typography.lineClamp
`

packages/gestalt/src/Text.js
```diff
@@ -11,7 +11,7 @@ function isNotNullish(val: ?number): boolean {
 }
 
 type As = 'span' | 'div';
-type Overflow = 'normal' | 'breakWord' | 'noWrap';
+type Overflow = 'normal' | 'breakAll' | 'breakWord' | 'noWrap';
```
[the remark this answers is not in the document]
AlbertCarreras: Mostly bacause of the issue shown 

Line clamp truncationExample in CodeSandbox; Opens a new tabWhen the "lineClamp" prop is set to 1, text with long strings of characters might be truncated incorrectly.
--


[Example in CodeSandbox; Opens a new tab](https://codesandbox.io/s/development-mode-forked-44qkwk?file=/example.js)
When the "lineClamp" prop is set to 1, text with long strings of characters might be truncated incorrectly.


Can we chek if we are fixing this issue? and maybe removing the known issue ?

AlbertCarreras (approved):
LGTM but left several comments, but overall looks good

packages/gestalt/src/Text.js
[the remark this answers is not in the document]
rlingineni: Yes, we should

packages/gestalt/src/Typography.css
rlingineni: I don't think we should use word wrap, it's actually a deprecated property:
https://stackoverflow.com/questions/17259916/difference-between-overflow-wrap-and-word-break

packages/gestalt/src/Text.js
rlingineni: I agree, the overrides can be hard to follow. I was thinking through this, but I thought it was super hard to read/think through.

I refactored to a method like this:
```js
 const getOverflowStyle = (): string | void => {
      if (overflow === 'noWrap') {
        return typography.noWrap;
      }

      if (overflow === 'breakAll') {
        return typography.breakAll;
      }

      // default to breakWord if lineClamp is set
      if (overflow === 'breakWord' || isNotNullish(lineClamp)) {
        return typography.breakWord;
      }

      return undefined;
    };
```

PR #913, opened 2020-06-10
[DatePicker] New DatePicker component in new gestalt-datepicker package

AlbertCarreras:
This PR creates new component: `DatePicker`.

DatePicker introduces new dependencies: `react-datepicker`. To maintain the lightness of `gestalt`, `DatePicker` is introduced in a new package `gestalt-datepicker`.

To use it, it must be installed and imported: `import Datepicker from "gestalt-datepicker"`

This PR has these major changes:
- refactors `rollup-config.js` to allow multiple gestalt packages.
- hook up to automatic releases
- creates a `gestalt-datepicker` package with the DatePicker component, flow-typed coverage, and tests.
- updates `gestalt/docs` to include a `beta` DatePicker component and multiple use cases.

## TODO

- [x] Accessibility checkup
- [x] Update documentation
- [x] Add/update Tests 
- [x] [Format PR title](https://github.com/pinterest/gestalt/#releasing): `ComponentName: Description`.

PR #3143, opened 2023-08-18
Chart: new gestalt-chart package setup

AlbertCarreras:
#### What changed?

Internal: new gestalt-chart package setup

#### DOCUMENTATION

When should you create a new gestalt-<component> package instead of adding a new component to the existing gestalt?

Gestalt is a lightweight package with only 2 dependencies: "classnames" and "prop-types". By using the minimal amount of dependencies, we can maintain the size of the bundle small. If a new component requires external packages, it's worth isolating it in a single package so that importing Gestalt in our codebase keeps a small bundle.

When adding a new package, these are the three areas where you must do changes

image

####FIRST: BASIC SETUP

####Gestalt packages

* create a new folder within `/packages` following the pattern `gestalt-<component_name>`
* create 2 subfolders: `/dist` and `/src`
* create files:
  * `/`: package.json*,  and rollup.config.js*
  * `/src`: index.js*, <component>.css, <component>.js, and <component>.jsdom.js.
  * `/dist`: gestalt-<component_name>.js.flow*

* Use `/package/gestalt` or`/package/gestalt-datepicker`  files as templates (copy & paste).

image

#####Gestalt configuration

* Add the new package to the following files
  * .gitignore: Important step so that build files inside `/dist` do not get pushed to Github.
image
  * .eslintignore: Important step to prevent getting eslint errors in build files inside `/dist`.
image  
* gestalt/package.json: Important step so that your package gets built when running `cd ~/gestalt && yarn build` as well when you build the Gestalt Docs while developing when running `cd ~/gestalt && yarn start`.
image
  * gestalt/package.json: Important step so that Netlify can build your new package when pushed to Github.
image

####SECOND: DEVELOP YOUR COMPONENT

#####Gestalt Packages

* Configure the new package/<NewComponent>/package.json to correctly setup dependencies: 
image
* Configure the new package/<NewComponent>/rollup.config.js to correctly build your package: 
image
* Develop your new component: <NewComponent>.js and <NewComponent>.css.
  * If your component is using external dependencies (external components), it might come with its own CSS styles.
  * A good approach is to not import the CSS file from the dependency (small bundles) and create a new file that only styles the classnames on the features used in the new component. 
  * This approach will help you prevent styling while overriding the imported CSS.
  * The Gestalt library uses CSS modules. If your external component allows injecting your own classnames, use CSS modules. If not, work with global classes; to use global classes, wrap your component with:

<div className="_gestalt">
   <Component />
</div>

image
image

* Test your new component with @react-testing-library
* Run `yarn build`
image

####THIRD: CODE EDITORS & SANDBOXES

Add the new package to all code editors and sandboxes so they can work with your new package

Original datepicker package PR: https://github.com/pinterest/gestalt/pull/913/files#diff-528119d81e5a10738ae49f5e20d333ebc37cea63657eb9069281693e6dae0ad3


rlingineni (approved):
Seems Okay!

docs/pages/web/callout copy.js
diyorbek, line 1: What's this file for?

diyorbek (approved):

diyorbek:
LGTM

PR #3150, opened 2023-08-22
Internal: move gestalt-charts to end of release list

AlbertCarreras:
#### What changed?

Internal: move gestalt-charts to end of release list


rlingineni (approved):
Can we make this a minor

PR #3158, opened 2023-08-24
Heading: Add BreakAll support to Header

rlingineni:
### Summary
Adding BreakAll as a prop to align with Text, and re-use the same logic.

#### What changed?

Added `BreakAll`
image

#### Why?

Aligns with recent `Text` changes, and also adds a Header


packages/gestalt/dist/index.d.ts
```diff
@@ -1967,7 +1967,7 @@ interface TextProps {
   inline?: boolean | undefined;
   italic?: boolean | undefined;
   lineClamp?: number | undefined;
-  overflow?: 'normal' | 'breakWord' | 'noWrap' | undefined;
+  overflow?: 'normal' | 'breakWord' | 'breakAll' | 'noWrap' | undefined;
```
rlingineni, line 1970: this updates `TextProps` accidentally did `HeadingProps` in the other one

packages/gestalt/src/Heading.js
```diff
@@ -85,6 +85,19 @@ export default function Heading({
   overflow = 'breakWord',
   size = '600',
 }: Props): Node {
+  const getWordBreakStyle = (): string | void => {
```
AlbertCarreras, line 88: could we reuse this between Text and HEading? helper function?

AlbertCarreras (approved):

packages/gestalt/src/Heading.js
rlingineni: Agree, I think for now, it's okay to dupe, since it also imports from the CSS class, sometimes importing CSS outside of a React context doesn't always work as well

rlingineni: They may also diverge at some point

PR #3175, opened 2023-08-30
Docs: Migrate foundation examples sandpack

diyorbek:
### Summary

- Migrated foundation pages' examples to Sandpack. 
- Removed deprecated `defaultCode` prop from `PageHeader`


docs/pages/foundations/iconography/usage.js
```diff
@@ -518,50 +542,60 @@ Generally 16px and 24px icons should be used in mobile interfaces. When icons ar
...
-            <Flex
-              gap={{
-                row: 1,
-                column: 0,
-              }}
-              alignItems="center"
-            >
-              <ReactionLove />
-              <Text>2.8k</Text>
-            </Flex>
```
diyorbek, line 548: @AlbertCarreras @rlingineni Do you think it would be a good idea to keep "inline" examples (instead of putting them inside Sandpack)?

docs/pages/foundations/typography/guidelines.js
```diff
@@ -7,6 +7,17 @@ import Markdown from '../../../docs-components/Markdown.js';
 import Page from '../../../docs-components/Page.js';
 import PageHeader from '../../../docs-components/PageHeader.js';
 import PrincipleItem from '../../../docs-components/PrincipleItem.js';
+import SandpackExample from '../../../docs-components/SandpackExample.js';
+import alignAndSpaceTextForReadability from '../../../examples/iconography/alignAndSpaceTextForReadability.js';
+import dontCenterTextWithTightLeading from '../../../examples/iconography/dontCenterTextWithTightLeading.js';
+import dontTruncateParagraphText from '../../../examples/iconography/dontTruncateParagraphText.js';
```
AlbertCarreras, line 13: should we move this examples to examples/typography?

AlbertCarreras (approved):
with comment

diyorbek:
@AlbertCarreras @rlingineni I wanted to remove everything related to old code examples but then I realized we need them cuz of "Dev Mode". Which is quite a big topic itself. I think we can come back to it later. Something tells me that we don't actually need "Dev Mode" althogether. We can address the problems with Sandpack directly.

docs/pages/foundations/typography/guidelines.js
diyorbek: yes

PR #3181, opened 2023-09-01
TapAreaLink, IconButtonLink: new components

AlbertCarreras:
#### What changed?

TapAreaLink, IconButtonLink: new components

#### Why?

We are deprecating  TapArea, IconButton role="link"


packages/gestalt/src/IconButtonLink.js
```diff
@@ -0,0 +1,214 @@
...
+): Node {
+  const innerRef = useRef<null | HTMLAnchorElement>(null);
+
+  // When using both forwardRef and innerRef, React.useimperativehandle() allows a parent component
+  // that renders <IconButton ref={inputRef} /> to call inputRef.current.focus()
+  useImperativeHandle(ref, () => innerRef.current);
+
+  const [isActive, setActive] = useState(false);
+  const [isFocused, setFocused] = useState(false);
+  const [isHovered, setHovered] = useState(false);
```
diyorbek, line 139: why not `useInteractiveStates`?

diyorbek:
Maybe we should add some snapshot tests?

packages/gestalt/src/IconButtonLink.js
AlbertCarreras: Good catch, I didn't know about this hook. Much cleana'

diyorbek (approved):

PR #3183, opened 2023-09-05
Video: Support RTL layout in scrubber

diyorbek:
### Summary

In RTL language context the scrubber of Video component was moving from left to right (instead of right to left).

**Also** I updated the RTL preview of Sandpack examples, so that examples responds to `html[dir="rtl"]` CSS selector (not to container div's `dir` attribute).

#### What changed?

 - Removed redundant html elements and styles to render Playhead component.
 - Added logic to Playhead, so that 
   - it can detect text direction and respond to RTL language layout.
   - it correctly reacts to `seek()` when clicked.
 - Fixed unreponsive Video component examples. The controls were not working at all.

### Links

- [Jira](https://jira.pinadmin.com/browse/BUG-169925)

### Checklist

### DIFF
--- a/docs/docs-components/SandpackExample.js
+++ b/docs/docs-components/SandpackExample.js
@@ -243,11 +243,14 @@ export default function SandpackExample({
           import { Box, ColorSchemeProvider } from 'gestalt';
           import App from "./App";
 
+          const html = document.querySelector('html');
+          html.setAttribute('dir', '${exampleTextDirection}');
+
           const root = createRoot(document.getElementById("root"));
           root.render(
             <StrictMode>
               <ColorSchemeProvider colorScheme="${exampleColorScheme}" fullDimensions>
-                <Box color="default" height="100%" width="100%" dir="${exampleTextDirection}">
+                <Box color="default" height="100%" width="100%">
                   <App />
                 </Box>
               </ColorSchemeProvider>
--- a/packages/gestalt/src/Video/Playhead.js
+++ b/packages/gestalt/src/Video/Playhead.js
@@ -32,8 +32,10 @@ export default class VideoPlayhead extends PureComponent<Props, State> {
       const { duration, seek } = this.props;
       const { left, right, width } = this.playhead.getBoundingClientRect();
 
-      const isRTL = window.getComputedStyle(this.playhead).direction === 'rtl';
-      const difference = Math.abs(clientX - (isRTL ? right : left));
+      // As a convention, text direction is defined in `dir` attribute of `html` tag of the document.
+      // The following check is done under the assuption of that convention.
+      const isRTL = document.querySelector('html')?.getAttribute('dir') === 'rtl';
+      const difference = isRTL ? right - clientX : clientX - left;
 
       const percent = Math.max(0, Math.min(difference / width, 1));
       const newTime = percent * duration;
--- a/packages/gestalt/src/Video/Playhead.js
+++ b/packages/gestalt/src/Video/Playhead.js
@@ -114,7 +114,6 @@ export default class VideoPlayhead extends PureComponent<Props, State> {
               width={16}
               height={16}
               color="light"
-              // FIXME
               marginStart={-2}
               dangerouslySetInlineStyle={{ __style: { marginTop: -6 } }}
             />
--- a/packages/gestalt/src/Video/Playhead.js
+++ b/packages/gestalt/src/Video/Playhead.js
@@ -30,9 +30,14 @@ export default class VideoPlayhead extends PureComponent<Props, State> {
   seek: (clientX: number) => void = (clientX) => {
     if (this.playhead) {
       const { duration, seek } = this.props;
-      const { left, width } = this.playhead.getBoundingClientRect();
-      const percent = Math.max(0, Math.min((clientX - left) / width, 1));
+      const { left, right, width } = this.playhead.getBoundingClientRect();
+
+      const isRTL = window.getComputedStyle(this.playhead).direction === 'rtl';
+      const difference = Math.abs(clientX - (isRTL ? right : left));
+
+      const percent = Math.max(0, Math.min(difference / width, 1));
       const newTime = percent * duration;
+
       seek(newTime);
     }
   };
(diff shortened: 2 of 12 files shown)


packages/gestalt/src/Video/Playhead.js
```diff
@@ -93,20 +99,22 @@ export default class VideoPlayhead extends PureComponent<Props, State> {
...
+            color="secondary"
+            display="flex"
           >
+            <Box color="light" rounding={2} height="100%" width={width} />
             <Box
               rounding="circle"
               width={16}
               height={16}
               color="light"
+              // FIXME
```
AlbertCarreras, line 117: what's this fix me?

diyorbek:
I think I will have to add snapshot tests for this.

packages/gestalt/src/Video/Playhead.js
diyorbek: oops, i'll remove it

docs/docs-components/SandpackExample.js
```diff
@@ -243,11 +243,14 @@ export default function SandpackExample({
           import { Box, ColorSchemeProvider } from 'gestalt';
           import App from "./App";
 
+          const html = document.querySelector('html');
```
AlbertCarreras, line 246: nice fix

diyorbek: I tried to override `index.html` but looks like Sandpack doesn't allow it. So I had to improvise 😬

packages/gestalt/src/Video/Playhead.js
```diff
@@ -30,9 +30,14 @@ export default class VideoPlayhead extends PureComponent<Props, State> {
   seek: (clientX: number) => void = (clientX) => {
     if (this.playhead) {
       const { duration, seek } = this.props;
-      const { left, width } = this.playhead.getBoundingClientRect();
-      const percent = Math.max(0, Math.min((clientX - left) / width, 1));
+      const { left, right, width } = this.playhead.getBoundingClientRect();
+
+      const isRTL = window.getComputedStyle(this.playhead).direction === 'rtl';
```
AlbertCarreras, line 35: my concern here is that if this is server-rendered there's no window

you probably have to do `if (window) {}`

AlbertCarreras: Why Math.abs()?, that's a new calculation here. jut curious

packages/gestalt/src/Video/Playhead.js
```diff
@@ -93,15 +99,16 @@ export default class VideoPlayhead extends PureComponent<Props, State> {
           role="progressbar"
           tabIndex="-1"
         >
-          <Box left right position="absolute" color="secondary" rounding={2} height={4}>
```
AlbertCarreras, line 96: note: thhis two were merged into 1

AlbertCarreras: nested boxes were merged into 1

packages/gestalt/src/Video/Playhead.js
diyorbek: Valid concern. I overlooked that because I thought to myself "what are the chances of `seek()` method being called on the server-side.

diyorbek: But still it is good to be on the safe side. I will add a check 👍

packages/gestalt/src/Video/Playhead.js
AlbertCarreras: Sorry this were comments for myself ahah

packages/gestalt/src/Video/Playhead.js
diyorbek: Good question. `Math.abs` because i wanted to aviod extra conditional statement. In RTL we have to calculate`right - clientX`, in contrast to LTR where we have to do `clientX - left`. Because if we do `clientX - right`, it gives us a negative value. So using `Math.abs` we can preserve the position of `clientX` in the formula.

diyorbek: Alternatively, we could do this:

```
const difference = isRTL ? clientX - left : right - clientX;
```
@AlbertCarreras what do you think?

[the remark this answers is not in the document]
AlbertCarreras: Did some googling and seems a fine solution https://stackoverflow.com/questions/35170581/how-to-access-styles-from-react

AlbertCarreras (approved):
R+ for me, let's wait for Ravi to also take a look!

packages/gestalt/src/Video/Playhead.js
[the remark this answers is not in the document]
diyorbek: @AlbertCarreras yes, we can omit `window` but I wanted to make it more explicit that the method is a global one.

packages/gestalt/src/Video/Playhead.js
```diff
@@ -93,15 +99,16 @@ export default class VideoPlayhead extends PureComponent<Props, State> {
...
-            <Box color="light" rounding={2} height="100%" width={width} />
-          </Box>
           <Box
+            left
+            right
             position="absolute"
             rounding={2}
             height={4}
-            dangerouslySetInlineStyle={{ __style: { left: width } }}
```
AlbertCarreras, line 103: cn you guide me through the new logic. We removed `__style: { left` how are we making it move now? It's just `width` rn

AlbertCarreras: Got it, it's positioned to the right on rtl... so it grows that way! got it

packages/gestalt/src/Video/Playhead.js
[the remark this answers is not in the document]
rlingineni: I think perf wise `getComputedStyle()` has a cost to calculate everytime we seek. 

Since we have a pattern of `rtl` on the HTML tag, could we just read this from the html root element attribue without computing the style

something like:
```
const isRtl = document.getElement('html')?.getAttribute("dir") === "rtl"
```

rlingineni: I like the second one, just because it's a little clearer to read/debug without thinking about it

rlingineni (approved):

PR #3187, opened 2023-09-06
Tooltip: Move to Internal Tooltip with a disabled state

rlingineni:
### Summary
Sidenavigation primary action dropdown is not positioned correctly. This is due to `MaybeTooltip` re-rendering the children when the tooltip should be disabled.

#### What changed?
This is a broken change introduced in this [PR](https://github.com/pinterest/gestalt/pull/3134). The TapArea is the innerRef, but when the tooltip is disabled, the dropdown placement is initially messed up because the TapArea is re-rendered. TapArea uses `useId` and the ref likely changes on every render.

~~This moves the `MaybeTooltip` to inside the tap area, so the outer ref for the dropdown is not changed.~~

This change does two things:
- Moves tooltip to `InternalTooltip`, and adds a new disabled prop. When disabled, the tooltip will never open.
-  `MaybeTooltip` will override set the tooltip behavior to be "disabled" 

#### Why?
See [Slack](https://pinterest.slack.com/archives/C13KLG5P0/p1694022120118259)

Brave Browser - SideNavigation - Gestalt 2023-09-07 at 1 22 19 PM

This introduces a bug:
image

### Links

- [TDD](link to Paper doc)
- [Figma](link to Figma file)

### Checklist

### DIFF
--- a/packages/gestalt/src/Tooltip/InternalTooltip.js
+++ b/packages/gestalt/src/Tooltip/InternalTooltip.js
@@ -1,5 +1,5 @@
 // @flow strict
-import { type Node, useReducer, useRef } from 'react';
+import { type Node, useEffect, useReducer, useRef } from 'react';
 import Box from '../Box.js';
 import Controller from '../Controller.js';
 import Layer from '../Layer.js';
@@ -14,8 +14,12 @@ const initialState = { hoveredIcon: false, hoveredText: false, isOpen: false };
 
 const reducer = (
   state: {| hoveredIcon: boolean, hoveredText: boolean, isOpen: boolean |},
-  action: {| type: 'hoverInIcon' | 'hoverInText' | 'hoverOutIcon' | 'hoverOutText' |},
+  action: {|
+    type: 'hoverInIcon' | 'hoverInText' | 'hoverOutIcon' | 'hoverOutText',
+    disabled?: boolean,
+  |},
 ) => {
+  if (action.disabled) return { ...state, isOpen: false, hoveredIcon: false };
   switch (action.type) {
     case 'hoverInIcon':
       return {
@@ -78,10 +82,12 @@ export default function InternalTooltip({
 
   const mouseLeaveDelay = link ? TIMEOUT : 0;
 
+  useEffect(() => {
+    dispatch({ type: 'hoverOutIcon', disabled });
+  }, [disabled]);
+
   const handleIconMouseEnter = () => {
-    if (!disabled) {
-      dispatch({ type: 'hoverInIcon' });
-    }
+    dispatch({ type: 'hoverInIcon' });
   };
 
   const handleIconMouseLeave = useDebouncedCallback(() => {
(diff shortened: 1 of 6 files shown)


AlbertCarreras (approved):

packages/gestalt/src/Tooltip/InternalTooltip.js
```diff
@@ -0,0 +1,168 @@
...
+        hoveredText: false,
+        isOpen: !state.hoveredIcon ? false : state.isOpen,
+      };
+    default:
+      throw new Error();
+  }
+};
+
+type Props = {|
+  /**
```
AlbertCarreras, line 50: can we remove all prop descriptions? Duplicated from tooltip.

I only keep external API props descriptions

packages/gestalt/src/Tooltip/InternalTooltip.js
```diff
@@ -0,0 +1,168 @@
...
+              {Boolean(link) && <Box marginTop={1}>{link}</Box>}
+            </Box>
+          </Controller>
+        </Layer>
+      )}
+    </Box>
+  );
+}
+
+InternalTooltip.displayName = 'InternalTooltip';
```
AlbertCarreras, line 168: Is this necessary?

packages/gestalt/src/Tooltip/InternalTooltip.js
```diff
@@ -0,0 +1,168 @@
...
+  link,
+  idealDirection = 'down',
+  inline,
+  text,
+  zIndex,
+}: Props): Node {
+  const [state, dispatch] = useReducer(reducer, initialState);
+  let { isOpen } = state;
+
+  if (disabled) isOpen = false;
```
AlbertCarreras, line 97: not sure, but should we override isOpen within the reducer?

or something like 

const handleIconMouseEnter = () => {
    !disabled && dispatch({ type: 'hoverInIcon' });
  };

packages/gestalt/src/Tooltip/InternalTooltip.js
```diff
@@ -0,0 +1,168 @@
...
+   */
+  zIndex?: Indexable,
+|};
+
+export default function InternalTooltip({
+  accessibilityLabel,
+  children,
+  disabled,
+  link,
+  idealDirection = 'down',
```
AlbertCarreras, line 89: this `down` default is unnecessary as it comes from Tooltip. 

Screenshot by Dropbox Capture

packages/gestalt/src/Tooltip/InternalTooltip.js
```diff
@@ -0,0 +1,168 @@
...
+    default:
+      throw new Error();
+  }
+};
+
+type Props = {|
+  /**
+   * Label to provide more context around the Tooltip’s function or purpose. By default `text` is used but this prop allows you to override it. Learn more about when to override it in the [Accessibility](https://gestalt.pinterest.systems/web/tooltip#Labels) section.
+   */
+  accessibilityLabel?: string,
```
AlbertCarreras, line 53: Here we could do 

```
type Props = {|
  ...ElementConfig<typeof Tooltip>,
  disabled: boolean
|}
```

AlbertCarreras: It might run into circular dependencies though...

PR #3189, opened 2023-09-09
Internal: Re-update playwright snapshots to match master

rlingineni:
### Summary
This adds a new `yarn:playwright visual-test` command that enables playwright.

I need it for visual testing on density and border token changes, so visual tests will be helpful.

#### What changed?

1. Updated version of playwright. Now we can use the `--ui` flag to see side-by-side comparisons and diffs better.

2. Updating snapshots of failing tests. It could be the case a few components may have unexpected changes, but as of now, we may not be able to find them out since our snapshots are not up-to-date

3. Decreases the pixelmatch threshold to `.7` as mentioned in this [thread](https://github.com/microsoft/playwright/issues/20097#issuecomment-1384419426). It's very tolerant, but it helps us skim over font rendering issues that exist across machines and architectures. Might be worth a try

4. Updates our `test.sh` to only run playwright in non-CI environments and when the component library changes

Recommendation: 

If we update these snapshots, in a next PR, I can try enabling the CI playwright to see if they all pass as expected. If we realize things are too flaky, we can turn it off again.

Ideally, we should use a standard docker container locally and in CI.  For now, we can try the threshold changes to see if they work as expected

### Links

- [TDD](link to Paper doc)
- [Figma](link to Figma file)

### Checklist


AlbertCarreras (approved):
LGTM Thanks!

diyorbek (approved):

PR #3194, opened 2023-09-12
Docs: Generate component metadata only from library exposed components

diyorbek:
### Summary

React-docgen was parsing unnecessary files such as test files, internal components. Some of the files didn't have data to parse and causing irrelevant error messages.

The main motivation of this PR is to make the script work without generating error messages, which are just noises given the current state of it. What makes them just noises? Because the files where the errors are coming from, are not supposed to have metadata, so they doesn't need to be parsed by react-docgen.

So how do we exclude the unwanted files? We can add all of them to excluded paths list but they are a lot of them. Also, they are scattered among the component files and hard to identify by file name or folder they are in.

But we can identify the files that needs to be parsed easily! We can just look at the index.js file! That's why I think this approach is more accurate.

The previous logic read every single file in the given folder. Yes, we had a filter on css files and test files (but flowtest files were still slipping through, but this was an easy fix). However, there are files that are not components, that are internal components or just utilities.

#### What changed?

- `generteMetadata.js` script was updated to parse only exported component and their subcomponents from `index.js` of the packages. 
- Bumped `react-docgen` to v6.0.4
  - `parse` method now return array of data
- Ignoring node_modules/react-docgen/node_modules in `.flowconfig` as flow was checking the test files of the library (that supposed to have syntax error) and failing CI

### List of components that were removed after the update

 - DefaultAvatar
 - Foundation
 - AddCollaboratorsButton
 - CollaboratorAvatar
 - CollaboratorsCount
 - HoverOverlay
 - PositioningWrapper
 - Backdrop
 - Caret
 - InternalCheckbox
 - Collection
 - Item
 - Contents
 - Controller
 - InternalDatapoint
 - Trend
 - OptionItem
 - InternalLabel
 - InternalLink
 - InternalList
 - Message
 - MeasureItems
 - ScrollContainer
 - Action
 - Header
 - ExpandableItem
 - Title
 - ConfirmationPopover
 - InternalOverlayPanel
 - InternalPopover
 - InternalScrollBoundaryContainerWithForwardRef
 - ContentContainer
 - FullPage
 - PartialPage
 - PrimaryAction
 - GroupContent
 - GroupMobile
 - Mobile
 - PrimaryActionIconButton
 - TableOfContentsAnchor
 - TableOfContentsItemList
 - InternalTextField
 - InternalTextFieldIconButton
 - Controls
 - Playhead
 - AccessibilityPause
 - NewTabAccessibilityLabel
 - VisuallyHidden
 - AnimationContext
 - RequestAnimationFrameContext
 - OutsideEventBehavior
 - StopScrollBehavior
 - TrapFocusBehavior
 - ListProvider
 - NestingProvider
 - ScrollBoundaryContainerProvider
 - SideNavigationProvider
 - TableContext
 - FormErrorMessage
 - FormHelperText
 - FormHelperTextCounter
 - FormLabel
 - InternalDismissButton
 - MaybeTooltip
 - wrapWithComponent
 - InternalDateField
 - InternalDatePicker
 - TextInput

### Links

- [Jira](https://jira.pinadmin.com/browse/GESTALT-6949)
- [TDD](link to Paper doc)
- [Figma](link to Figma file)

### Checklist


scripts/generateMetadata.js
```diff
@@ -18,16 +26,15 @@ function logSuccess(message) {
...
+  if (excludedPaths.includes(filePath)) {
     return null;
   }
 
   const contents = await fs.promises.readFile(filePath, 'utf-8');
 
   // Not all files have data to parse
   try {
-    const parsed = reactDocs.parse(contents);
+    const [parsed] = (await reactDocs).parse(contents);
```
diyorbek, line 37: react-docgen v6 outputs array of objects. Here I am assuming there's only going to be one component in a given file.

scripts/generateMetadata.js
```diff
@@ -46,32 +53,55 @@ async function docgen(filePath) {
...
 
-  return files.filter((file) => file.length).flat();
-};
+  return subcomponentPaths;
+}
+
+/**
+ * Gets components that are imported in `index.js` of given directory (assuming they're all exported).
+ */
+function getExposedFilesFromDirectory(directoryPath) {
```
rlingineni, line 91: What benefit is this providing us over the previous approach?  Can you add in the PR description

scripts/generateMetadata.js
```diff
@@ -46,32 +53,55 @@ async function docgen(filePath) {
...
+  const indexFile = fs.readFileSync(path.join(root, directoryPath, 'index.js'), 'utf-8');
+  const importMatches = indexFile.matchAll(/from '(?<path>.+)'/g);
+  const filePaths = [...importMatches].map((match) => path.join(directoryPath, match.groups.path));
+  const subcomponentPaths = filePaths.map(getSubcomponentPaths).flat();
```
rlingineni, line 95: Just wondering, not a requirement, but does this solve the issue of let's say we have Button, and button imports from '../SharedButtonProps.js', this can output `SharedButtonProps`as part of Button?

 That was a problem we had in the past.

rlingineni (commented):
Thanks Diyor! I noticed that this does a little more than just exclude the folders. Mind adding  more as what the changes are/improvements. How it works.

Can you verify there are no regressions? How do we know we didn't miss anything

package.json
```diff
@@ -72,7 +72,7 @@
     "postcss-modules": "^4.3.0",
     "postcss-preset-env": "^7.2.0",
     "prettier": "2.8.0",
-    "react-docgen": "^5.4.0",
+    "react-docgen": "^6.0.4",
```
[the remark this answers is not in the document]
diyorbek: v5 was generating syntax error related to `this` keyword, also, it was throwing error when there're multiple component definitions. Upgrading solved those problems.

But the files where those errors were happening are now excluded. So, if we consider that, there's no urge to do that. But I think newer version is usually better, because of the bugfixes and stuff.

scripts/generateMetadata.js
diyorbek: Basically, we're generating metadata only for the components imported in `index.js` of the packages. Previously we generated it for all the components, even for the docs site components, css and test files, which are unecessary. This was not what I actually wanted to do. I wanted to get rid of the DocGen errors polluting the terminal. So, I came up with this approach.

diyorbek: I updated the PR title. What do you think?

scripts/generateMetadata.js
diyorbek: Do you mean, can it generate data for inherited props?

diyorbek: If that's what you meant, then no it doesn't I think.

diyorbek:
> Thanks Diyor! I noticed that this does a little more than just exclude the folders. Mind adding more as what the changes are/improvements. How it works.
> 
> Can you verify there are no regressions? How do we know we didn't miss anything

Good point! I actually have snapshot of the only metadata. I compared them and put the list of removed components in the PR description. To be more confident with the output, I can do a deep comparison of every field and text. Hopefully, they'll be exactly the same :D .

scripts/generateMetadata.js
rlingineni: Yeah!

rlingineni: Aw man :(

scripts/generateMetadata.js
rlingineni: No the actual summary, what's the high-level overview of how this algo works?

scripts/generateMetadata.js
diyorbek: I think we can address that in another PR. It would truly be great to have that feature.

.flowconfig
```diff
@@ -4,6 +4,7 @@
 [ignore]
 <PROJECT_ROOT>/packages/gestalt-codemods/.*/__testfixtures__
 <PROJECT_ROOT>/scripts/templates/
+<PROJECT_ROOT>/node_modules/.*
```
AlbertCarreras, line 7: what's this for?

what were you trying to solve? curious

package.json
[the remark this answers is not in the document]
AlbertCarreras: It's good to keep dependencies upgraded thanks!

scripts/generateMetadata.js
```diff
@@ -2,11 +2,19 @@
...
 const path = require('path');
 const chalk = require('chalk');
-const reactDocs = require('react-docgen');
+
+const reactDocs = import('react-docgen'); // v6 forces to use `import`
 
 const root = path.join(__dirname, '../');
 const docsPath = path.join(root, '/docs');
 
+const excludedPaths = [
```
AlbertCarreras, line 11: why are we excluding this files? trying to follow here.

AlbertCarreras: Oh cos they =dont have prop tables!

Can we add a comment explaining that?

scripts/generateMetadata.js
```diff
@@ -46,32 +53,55 @@ async function docgen(filePath) {
...
+    path.join(componentPath, '../', match.groups.path),
+  );
 
-  return files.filter((file) => file.length).flat();
-};
+  return subcomponentPaths;
+}
+
+/**
+ * Gets components that are imported in `index.js` of given directory (assuming they're all exported).
```
AlbertCarreras, line 89: One caveat is that we cannot have new component pages that aren't listed in the index. Now for development we must add them to the index file. 

I don't think it's a big issues seeing how we work, but good having that in mind when setting up new components. Maybe add a comment to `scripts/templates/README.md`

scripts/generateMetadata.js
```diff
@@ -46,32 +53,55 @@ async function docgen(filePath) {
   }
 }
 
-const getComponentNameFromFile = (file) => file.replace(/^.*[\\/]/, '').replace(/\.js/, '');
```
AlbertCarreras, line 49: also, out of curiosity, 

why did we add so many more lines of code? what did change in the logic that now we require  all that new logic? 


before it seems there was some recursive logic, and now it seems more inlined.

can you add a little context on this change of approach? just to follow better your decision making process here

scripts/generateMetadata.js
diyorbek: @AlbertCarreras I don't think it is a caveat. `generateComponent` script automatically adds new components to `index.js`. We import components from `index.js` to the docs site, we never import from the component file directly.

Though, I agree adding a comment to README

scripts/generateMetadata.js
diyorbek: The main motivation of this PR is to make the script work without generating error messages, which are just noises given the current state of it. What makes them just noises? Because the files where the errors are coming from, are not supposed to have metadata, so they doesn't need to be parsed by react-docgen.

So how do we exclude the unwanted files? We can add all of them to excluded paths list but they are a lot of them. Also, they are scattered among the component files and hard to identify by file name or folder they are in.

But we can identify the files that needs to be parsed easily! We can just look at the `index.js` file! That's why I think this approach is more accurate. 

The previous logic read every single file in the given folder. Yes, we had a filter on css files and test files (but flowtest files were still slipping through, but this was an easy fix). However, there are files that are not components, that are internal components or just utilities. 

Yes, I agree, the more code can introduce more confusion or bugs (that's tried to comment as much as I can 😬). Also, I am assuming and relying on the current conventions we have in the codebase. For example, the file/folder structures and certain syntax rules. For example, subcomponents are static fields of a component: `Component.Subcomponent`. But I don't think we are going to change in the foreseeable future.

diyorbek:
> > Thanks Diyor! I noticed that this does a little more than just exclude the folders. Mind adding more as what the changes are/improvements. How it works.
> > Can you verify there are no regressions? How do we know we didn't miss anything
> 
> Good point! I actually have snapshot of the only metadata. I compared them and put the list of removed components in the PR description. To be more confident with the output, I can do a deep comparison of every field and text. Hopefully, they'll be exactly the same :D .

I checked the deep comparison of the new and old metadata. Aside from unnecessary components, nothing was removed. Instead, more fields were added. Like interfaces/types that are in the file, not just Props.

.flowconfig
```diff
@@ -4,6 +4,7 @@
 [ignore]
 <PROJECT_ROOT>/packages/gestalt-codemods/.*/__testfixtures__
 <PROJECT_ROOT>/scripts/templates/
+<PROJECT_ROOT>/node_modules/react-docgen/node_modules
```
AlbertCarreras, line 7: flow-typed

rlingineni (approved):

PR #3196, opened 2023-09-13
Tokens: Density Tokens Update

rlingineni:
Spacing token values are changing to a new scale. This updates old and also adds new values.

### Summary
In order to maintain consistency with our Figma designs and offer more flexibility in spacing, we're using spacing on a 4pt scale. 

All visual tests are passing :)

### Links

- [Figma](https://www.figma.com/file/Y88VHmN8jtWDhWnrSZ0JuW/Density-theme-handoff?type=design&node-id=2-535&mode=design&t=ncvVsgPB8M18xIrj-0)

### Checklist


AlbertCarreras:
Can we add a table in the description of tokens before after equivalencies?

AlbertCarreras:
could we add a quick comment that we follow 4px (or _**boint**_ in gestalt terms!)

packages/gestalt-design-tokens/tokens/space/base.json
```diff
@@ -3,17 +3,27 @@
...
-      "300": { "value": "-16px" },
-      "400": { "value": "-24px" },
-      "500": { "value": "-32px" },
-      "600": { "value": "-64px" }
+      "300": { "value": "-12px" },
+      "400": { "value": "-16px" },
+      "500": { "value": "-20px" },
+      "600": { "value": "-24px" },
+      "700": { "value": "-28px" },
+      "800": { "value": "-32px" },
```
AlbertCarreras, line 25: why are we skipping negative from 900 to 1600? Shouldn't all mirror? Can we document / comment in the code if there's a reason why?

packages/gestalt/src/RadioButton.css
```diff
@@ -1,6 +1,6 @@
 :root {
   --g-border-sm: 6px;
```
AlbertCarreras, line 2: what happens with out of the system px? `6px;`

AlbertCarreras (approved):
R+ but left a major comment about missing neg tokens? in case it wasnt intentional

packages/gestalt/src/RadioButton.css
rlingineni: removed

packages/gestalt-design-tokens/tokens/space/base.json
rlingineni: Good catch, added them back for consistency after chatting with Hector. Thinking was because there was no usage of the negative tokens, but we can revisit that later in the docs and keep the tokens

PR #3197, opened 2023-09-14
Toast: Fix toast prop "variant" to "type"

zifahm:
### Summary
Fix toast typescript type to match [props table](https://gestalt.pinterest.systems/web/toast)

#### What changed?
prop "variant" changed to prop "type"

### DIFF
--- a/packages/gestalt/dist/index.d.ts
+++ b/packages/gestalt/dist/index.d.ts
@@ -2176,7 +2176,7 @@ interface ToastProps {
     | { avatar: React.ReactElement<typeof Avatar> }
     | { icon: React.ReactElement<typeof Icon> }
     | undefined;
-  variant?: 'default' | 'success' | 'error' | 'progress' | undefined;
+  type?: 'default' | 'success' | 'error' | 'progress' | undefined;
 }
 
 interface TooltipProps {


rlingineni:
Thanks for the catch @zifahm

zifahm:
@rlingineni by the way, I'm unable to see the issue tab for Gestalt. I suppose it is hidden? Could you please enable it or add me as a member so I can access the issue tab, please?
Thanks very much.

AlbertCarreras (approved):

zifahm:
@AlbertCarreras 
Hey Idk where else to post this, the issue tab is missing for me.

The WashAnimated has issue an issue with left and bottom margin/padding. not sure what caused this, maybe the new token update

https://gestalt.pinterest.systems/web/washanimated

If you hove over the card you can see the WashAnimated right padding/margin and bottom padding does not align well with the left padding or top padding. ***Right and bottom had gotten wider***.

rlingineni:
@zifahm sorry, I had missed your first comment! We closed the public issues tab [[see readme](https://github.com/pinterest/gestalt#issues)], because we have a lot of things going on at the moment and we can't address external issues at this time. 

Also, thank you! I took a look, and that is something that changed with the token updates. WashAnimated seems to have a change as a result of the spacing token changes. I made a PR to fix the changes.

Our visual tests didn't catch an issue, so something I'm also looking into as well. 

There is value of having public issues as a communication channel! So we may re-discuss on issues/how to better reach out.

zifahm:
@rlingineni

In several other GitHub repositories, I've come across a clever idea that involves restricting the issues tab exclusively to bug reports. You can achieve this by specifying an issue template.

Here's an example from the Remix bug report [issue template](https://github.com/remix-run/remix/blob/main/.github/ISSUE_TEMPLATE/bug_report.yml). Implementing this template will compel users to post issues strictly related to bug reports rather than general questions.

Furthermore, to address questions related to Gestalt, you can enable the discussion tab. This will allow other external Gestalt developers to assist with questions posted in the discussion tab, providing an effective solution for users who have inquiries.

You can also include a note in the bug report issue template like this: "If you have any questions or suggestions, please post them in the discussion tab."

I believe this approach would resolve many open-source concerns for your team. It would enable you to prioritize internal support tickets while external discussion questions could be handled by external developers.


I think this would solve a lot of open source concers for you guys, and you guys can prioratis only internal support tieckets, and the external discussion questions would be fullfiled by external developers.

PR #3202, opened 2023-09-19
Video: Fix flickering video controls on iOS browsers

diyorbek:
### Summary

This PR fixes a bug that is present only in iOS browsers (mobile Safari and Chrome). Apparently, rotated SVGs on top of video element causes the video element appear on top. Sometimes, it causes SVG and its container to flicker then disappear (or to partially render).

This bug was caught in RTL mode because in RTL some of the icons are flipped.

#### What changed?

`rotateY(180deg)` CSS rule was replaced by `scale(-1, 1)` as it seems to be more stable/reliable.

#### Why?

Original bug was reported here https://jira.pinadmin.com/browse/BUG-170571

### Links

- [Jira](https://jira.pinadmin.com/browse/BUG-171048)
- [TDD](link to Paper doc)
- [Figma](link to Figma file)

### Checklist

### DIFF
--- a/packages/gestalt/src/Video.css
+++ b/packages/gestalt/src/Video.css
@@ -27,6 +27,7 @@
   composes: right0 from "./Layout.css";
   composes: paddingX2 from "./boxWhitespace.css";
   composes: paddingY2 from "./boxWhitespace.css";
+  z-index: 0;
   background-image: linear-gradient(
     to bottom,
     rgb(181 181 181 / 0),
--- a/packages/gestalt/src/Video.css
+++ b/packages/gestalt/src/Video.css
@@ -34,6 +34,4 @@
     rgb(0 0 0 / 0.5) 66%,
     rgb(0 0 0 / 0.7)
   );
-  /* Workaround for the bug in iOS browsers with rendering transformed SVGs over video */
-  z-index: 0;
 }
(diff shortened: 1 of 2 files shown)


AlbertCarreras (approved):

packages/gestalt/src/Video.css
```diff
@@ -34,4 +34,6 @@
     rgb(0 0 0 / 0.5) 66%,
     rgb(0 0 0 / 0.7)
   );
+  /* Workaround for the bug in iOS browsers with rendering transformed SVGs over video */
```
AlbertCarreras, line 37: curious is it happens with a scale flip, instead of a rotate
.arrow {
  transform: scale(-1, 1);
  transform-origin: center;
}

AlbertCarreras: solution by Dave

AlbertCarreras (approved):
Before merging, can you add a couple videos of before and after. Before can be from bug, add quick visual on how it looks when it's fixed

diyorbek:
**Reporoduced version of the bug:**
ezgif-1-0da7991978

diyorbek:
**After the fix:**
Screenshot_20230915-160139

diyorbek:
**After the fix:**

ezgif-3-5355bcfd42

PR #3204, opened 2023-09-19
TagData, TileData: Multi-line Tooltips

rlingineni:
Making tooltips more rich for dataviz components. This enables rich text.

#### What changed?

Continuing to build on Internal Tooltip. This adds support for the `string[]` element to enable paragraphs and more spacing.

#### Why?

The m1On teams needs this to enable denser views and add more clarity around these elements.

image

### Links

- [Original thread](https://pinterest.slack.com/archives/C13KLG5P0/p1694560363775429?thread_ts=1674168008.120529&cid=C13KLG5P0)
- [Design Discussion](https://pinterest.slack.com/archives/GD6AMGH9C/p1694562105160289)

### Checklist

### DIFF
--- a/docs/examples/tiledata/tooltip.js
+++ b/docs/examples/tiledata/tooltip.js
@@ -11,7 +11,7 @@ export default function Example(): Node {
           text: [
             'Monthly Active Users',
             'The total monthly users over the last 30 days',
-            ' MAU has gone up by 10% over the last 30 days',
+            'MAU has gone up by 10% over the last 30 days',
           ],
         }}
         title="MAU"
--- a/docs/examples/tiledata/tooltip.js
+++ b/docs/examples/tiledata/tooltip.js
@@ -1,11 +1,26 @@
 // @flow strict
 import { type Node } from 'react';
-import { Flex, TileData } from 'gestalt';
+import { Flex, Text, TileData } from 'gestalt';
 
 export default function Example(): Node {
   return (
-    <Flex justifyContent="center" alignItems="center" width="100%" height="100%">
-      <TileData tooltip={{ text: 'Weekly Active Users' }} title="WAU" value="1.25M" selected />{' '}
+    <Flex justifyContent="center" alignItems="center" width="100%" height="100%" gap={2}>
+      <TileData tooltip={{ text: 'Weekly Active Users' }} title="WAU" value="1.25M" />{' '}
+      <TileData
+        tooltip={{
+          text: (
+            <Text>
+              <strong>Monthly Active Users</strong>
+              <p>The total monthly users over the last 30 days</p>
+              <p>MAU has gone up by 10% over the last 30 days</p>
+            </Text>
+          ),
+        }}
+        title="MAU"
+        value="2.25M"
+        selected
+        trend={{ value: 10, accessibilityLabel: 'Increased by 10%' }}
+      />{' '}
     </Flex>
   );
 }
--- a/docs/examples/tiledata/tooltip.js
+++ b/docs/examples/tiledata/tooltip.js
@@ -1,6 +1,6 @@
 // @flow strict
 import { type Node } from 'react';
-import { Flex, Text, TileData } from 'gestalt';
+import { Flex, TileData } from 'gestalt';
 
 export default function Example(): Node {
   return (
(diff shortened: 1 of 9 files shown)


AlbertCarreras (approved):

packages/gestalt/src/Tooltip/InternalTooltip.js
```diff
@@ -60,7 +68,7 @@ type Props = {|
   idealDirection?: 'up' | 'right' | 'down' | 'left',
   inline?: boolean,
   link?: Node,
-  text: string,
+  text: string | Element<typeof Text>,
```
AlbertCarreras, line 71: Do we wanna add an example like https://gestalt.pinterest.systems/web/toast#Message

Does this require any guidelines?

AlbertCarreras:
Before merging, please, see first my comment here https://pinterest.slack.com/archives/GD6AMGH9C/p1695215078507769?thread_ts=1694562105.160289&cid=GD6AMGH9C


A couple concerns:
- Links in tooltips should soon be deprecated in favor of HelpButton

 - Opening it to text means now we can add Links within Text, so we are opening the door to  misusages. In general, I think most components can open to Text, in this case were links are unaccessible... Should we provide guidance or maybe find an alternative, such as an array of strings rather than Text that get passed to Text internally or an object `{ title: xxxx, text: [text, text]} `

AlbertCarreras (commented):
as per feedback w hector

packages/gestalt/src/Tooltip/InternalTooltip.js
rlingineni: removed to not use string array

AlbertCarreras (approved):
Can we add new types to TS files?

docs/examples/tiledata/tooltip.js
```diff
@@ -4,8 +4,21 @@ import { Flex, TileData } from 'gestalt';
...
-    <Flex justifyContent="center" alignItems="center" width="100%" height="100%">
-      <TileData tooltip={{ text: 'Weekly Active Users' }} title="WAU" value="1.25M" selected />{' '}
+    <Flex justifyContent="center" alignItems="center" width="100%" height="100%" gap={2}>
+      <TileData tooltip={{ text: 'Weekly Active Users' }} title="WAU" value="1.25M" />{' '}
+      <TileData
+        tooltip={{
+          text: [
+            'Monthly Active Users',
+            'The total monthly users over the last 30 days',
+            ' MAU has gone up by 10% over the last 30 days',
```
AlbertCarreras, line 14: nit: dead space

packages/gestalt/src/Tooltip/InternalTooltip.js
```diff
@@ -102,6 +102,26 @@ export default function InternalTooltip({
     dispatch({ type: 'hoverOutText' });
   }, mouseLeaveDelay);
 
+  const getTooltipText = () => {
+    if (Array.isArray(text)) {
+      const lines = [];
+      // first and last line should not have a <p> tag, (adds extra padding)
+      text.map((line, idx) =>
+        lines.push(
```
AlbertCarreras, line 110: yo don't really need the push into line right?

you can just return and use the map  const a = text.map()

packages/gestalt/src/Tooltip/InternalTooltip.js
```diff
@@ -102,6 +102,26 @@ export default function InternalTooltip({
     dispatch({ type: 'hoverOutText' });
   }, mouseLeaveDelay);
 
+  const getTooltipText = () => {
```
AlbertCarreras, line 105: any benefit in memoizing this?

AlbertCarreras (approved):

packages/gestalt/src/Tooltip/InternalTooltip.js
diyorbek: @rlingineni maybe we should also check the type of `line`? When someone disables flow type checking and passes a JSX or smth, should it be our concern?

packages/gestalt/src/Tooltip/InternalTooltip.js
rlingineni: It's not a complex operation

PR #3219, opened 2023-09-26
ChartGraph: Tabular representation feature 

AlbertCarreras:
### Breking change

DefaultLabelProvider is expanded

#### What changed?

ChartGraph: Tabular representation feature 

Brave Browser - ChartGraph - Gestalt 2023-09-28 at 7 22 32 PM

#### Why?

[Accessibility. See https://github.com/pinterest/gestalt/pull/3219](https://deploy-preview-3219--gestalt.netlify.app/web/chartgraph#Tabular-representation)


packages/gestalt-charts/src/ChartGraph/TabularDataModalTable.js
```diff
@@ -0,0 +1,76 @@
...
+import { Table, Text, useDefaultLabel } from 'gestalt';
+import { type TransformedTabularDataType } from './useTabularData.js';
+
+type Props = {|
+  title: string,
+  isHorizontalLayout: boolean,
+  transformedTabularData: TransformedTabularDataType,
+  onSortChange: ('series' | 'x' | 'y') => void,
+  sortOrder: 'desc' | 'asc',
+  sortCol: null | 'series' | 'x' | 'y',
```
diyorbek, line 12: Can we extract these to union type declaration? They have been repeated a lot.

AlbertCarreras: DONE

packages/gestalt-charts/src/ChartGraph/useTabularData.js
```diff
@@ -0,0 +1,139 @@
...
+  isHorizontalLayout,
+}) => {
+  const { tableSeriesText, tableXAxisText, tableYAxisText } = useDefaultLabel('ChartGraph');
+
+  return `${tableSeriesText},${isHorizontalLayout ? tableXAxisText : tableYAxisText},${
+    isHorizontalLayout ? tableYAxisText : tableXAxisText
+  }\n${transformedTabularData
+    .map((x) => Object.values(x))
+    .map((e) => e.join(','))
+    .join('\n')}`;
```
diyorbek, line 30: nit: Do you think it's a good idea to wrap it with `useMemo`? I am not sure how often this hook is used.

packages/gestalt-charts/src/ChartGraph/useTabularData.js
```diff
@@ -0,0 +1,139 @@
...
+
+const useTabularData: UseTabularDataProps = ({
+  data,
+  filterId,
+  filterOrder,
+  labelMap,
+  tickFormatter,
+  isHorizontalLayout,
+}) => {
+  const transformedTabularData = data
```
diyorbek, line 92: nit: Same concern here. These calculations are triggered in every render wherever the hook is present. Please, correct me if I'm not understanding something.

packages/gestalt-charts/src/ChartGraph/useTabularData.js
AlbertCarreras: It´s a good concern. I can try to memoize. I think this is only called on the TabularDataModal component so there shouldn't be many rerenders in general after the mounting of Modal. If it were within ChartGraph, I think there could be more unrelated rerenders
@diyorbek

AlbertCarreras: DONE

packages/gestalt-charts/src/ChartGraph/useTabularData.js
AlbertCarreras: DONE

diyorbek (approved):

PR #3220, opened 2023-09-26
Tooltip: Move to Internal Tooltip

rlingineni:
### Summary

Moving to an internal tooltip. Original change description is in #3187.  It was reverted in #3209 due to failing tests.

#### What changed?

This fixes extra rendering that kept getting called in the useEffect 

Before:

```
useEffect(() => {
      dispatch({ type: 'hoverOutIcon', disabled });
  }, [disabled]);
 ```
  
This would dispatch the event every time on first render. So when a lot of conditional renders were happening, this would cause a `too many renders` issue.

E.g. in something like
```
{text && <Tooltip> <Child> </Tooltip}
```
Every time text would change, the element would fall into re-rendering.

  
 Now:
 ```
 useEffect(() => {
     if(disabled === true){
           dispatch({ type: 'hoverOutIcon', disabled });
     }
  }, [disabled]);
 ```
  
 This only dispatches the disabled action if the item in the disabled prop is true.
 
Also, added disabled to all of the other actions as a prop

### DIFF
--- a/packages/gestalt/src/Tooltip.js
+++ b/packages/gestalt/src/Tooltip.js
@@ -1,50 +1,7 @@
 // @flow strict
-import { type Node, useReducer, useRef } from 'react';
-import Box from './Box.js';
-import Controller from './Controller.js';
-import Layer from './Layer.js';
-import Text from './Text.js';
-import useDebouncedCallback from './useDebouncedCallback.js';
+import { type Node } from 'react';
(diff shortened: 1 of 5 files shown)


diyorbek (approved):

AlbertCarreras (approved):

AlbertCarreras:
Can you dociment, take a screenshot of the useEffect and some context of what was happening?

PR #3227, opened 2023-09-27
Internal: Update local test check

rlingineni:
#### What changed?
Visual tests were not always running locally as expected. This updates old Bash script.

The old git diff command was paginating, and the grep command was not always finding instances.

This adds a `--name-only` flag to the git diff so we get a complete list of the files, and greps everything at once. 

 It also wraps the function an if else statement to continue executing due to `set -euo pipefail` requirements

### DIFF
--- a/scripts/test.sh
+++ b/scripts/test.sh
@@ -47,11 +47,14 @@ then
   exit 1
 fi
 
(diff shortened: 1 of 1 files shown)


AlbertCarreras:
Can you give more context on the change? What's different?

diyorbek (approved):

PR #3251, opened 2023-10-12
Tokens: Move to Border radius tokens

rlingineni:
This PR introduces Border Radius Tokens and updates Gestalt to use them

### Summary

We're introducing border-radius tokens so we can use them alongside with our density work.

#### What changed?

Adding the following token values in increments of 4
```
    "0": { "value": "0" },
    "100": { "value": "4px" },
    "200": { "value": "8px" },
    "300": { "value": "12px" },
    "400": { "value": "16px" },
    "500": { "value": "20px" },
    "600": { "value": "24px" },
    "700": { "value": "28px" },
    "800": { "value": "32px" },
    "pill": { "value": "999px" }
```
#### Why?

We need to use radius tokens for density.  CSS compose statements cannot be used in non singular classes when writing like this:

```
SegmentedControl.sm {
  padding: var(--space-100);
  composes: rounding3 from "./Borders.css";
}

.SegmentedControl.md {
  padding: var(--space-100);
  composes: rounding3 from "./Borders.css";
}

``` 

### Links

- [TDD](link to Paper doc)
- [Figma](link to Figma file)

### Checklist


packages/gestalt-design-tokens/tokens/rounding/base.json
```diff
@@ -0,0 +1,14 @@
...
+    "0": { "value": "0" },
+    "100": { "value": "4px" },
+    "200": { "value": "8px" },
+    "300": { "value": "12px" },
+    "400": { "value": "16px" },
+    "500": { "value": "20px" },
+    "600": { "value": "24px" },
+    "700": { "value": "28px" },
+    "800": { "value": "32px" },
+    "pill": { "value": "999px" }
```
AlbertCarreras, line 12: we are doing pill but not circle? circle is 50%

packages/gestalt/src/Borders.css
```diff
@@ -68,43 +68,43 @@ html[dir="rtl"] .borderLeft {
 }
 
 .pill {
-  border-radius: 999px;
+  border-radius: var(--rounding-pill);
```
AlbertCarreras, line 71: we should also tokenize border-radius: 50%, no?

packages/gestalt/src/Tag.css
```diff
@@ -6,11 +6,11 @@
 }
 
 html[dir="rtl"] .button {
-  border-radius: 8px 0 0 8px;
+  border-radius: var(--rounding-200) 0 0 var(--rounding-200);
```
AlbertCarreras, line 9: could we tokenize this as alias? ... just wondering. 

rounding_container_rtl
rounding_container_ltr

packages/gestalt-design-tokens/tokens/rounding/base.json
rlingineni: Added!

packages/gestalt/src/Borders.css
```diff
@@ -68,43 +68,43 @@ html[dir="rtl"] .borderLeft {
...
 }
 
 .rounding0 {
   border-radius: 0;
```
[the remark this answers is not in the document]
rlingineni: Yeah, I went ahead and updated this to use it also

packages/gestalt/src/Borders.css
rlingineni: Yeah I wasn't sure if percents are okay, but seems fine

AlbertCarreras (approved):

PR #3254, opened 2023-10-13
Internal:  run `yarn flow-codemod removeExplicitlyExactObjectTypeSyntax`

AlbertCarreras:
More context: 
Exact object types by default, by default
https://medium.com/flow-type/exact-object-types-by-default-by-default-cc559af6f69


.eslintrc.js
```diff
@@ -45,7 +45,7 @@ module.exports = {
     'flowtype/define-flow-type': ERROR,
     'flowtype/no-mutable-array': ERROR,
     'flowtype/no-types-missing-file-annotation': ERROR,
-    'flowtype/require-exact-type': [ERROR, ALWAYS],
+    'flowtype/require-exact-type': OFF,
```
AlbertCarreras, line 48: Flow is already set to `exact_by_default` to `true` in `.flowconfig`, so this was redundant syntax {| |}

jackhsu978 (approved):

PR #3267, opened 2023-10-18
SelectList: Add fullWidth prop to selectlist

rlingineni:
This adds a fullWidth property to SelectList so it can take the entire space of a container. I

t's supported in image

We can't enable `width=100%` on all SelectLists because it may begin to overflow existing containers and break UIs, and it's hard to detect these instances from code-alone.

### Summary
This adds a `fullWidth` prop that sets the width of the selectList to be more responsive.

This also aligns with the current ComboBox behavior where the outer container can have a width
```
<Box width={320}>
<Combobox />
</Box>
```

#### What changed?

image

From a high level, what are the changes this PR introduces? (No need to recount line-by-line, we can see that.)

### Links

- [TDD](link to Paper doc)
- [Figma](link to Figma file)

### Checklist

### DIFF
--- a/docs/examples/selectlist/fullWidthExample.js
+++ b/docs/examples/selectlist/fullWidthExample.js
@@ -1,0 +1,29 @@
+// @flow strict
+import { type Node } from 'react';
+import { Box, SelectList } from 'gestalt';
+
+export default function Example(): Node {
+  return (
+    <Box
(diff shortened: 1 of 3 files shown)


rlingineni:
Abandon, we haven't heard any further requests for this.

PR #3279, opened 2023-10-24
Docs: Stackblitz Support

rlingineni:
Wouldn't it be easier for designers and developers to tweak the docs in their browser and avoid the command line and all the app switching?

After merging this PR, I can make a video with a demo of contributing docs.

[Stackblitz](https://developer.stackblitz.com/guides/user-guide/what-is-stackblitz) is an alternative to Codespaces, it uses local machine and Web Containers to compile code.

This PR adds a script to lower the memory load of the startup script, and lets us develop on Gestalt in the browser. It's not meant for developing components, but can be used for quick doc updates in a reliable way.

### Summary

This enables two easy workflows:

1. Edit single file markdown in a [quick editor](https://stackblitz.com/~/github.com/stackblitz/docs/edit/main/docs/codeflow/content-updates-with-web-publisher.md), and see live changes

2. Edit multiple files like this in the [pr.new/github.com/stackblitz/docs](https://pr.new/github.com/stackblitz/docs) like this and easily make a PR

image

#### What changed?

Adding a start script so it doesn't have so many watchers and the webcontainer doesn't crash

#### Why?

The context switch, opening VS Code making changes, is often a pain point with building and editing our code for designers. This should make onboarding easier, and minimize the friction. 

Unlike CodeSandbox or GH Codespaces, this is a free solution, and uses local web compute, not remote services, so it should be worth a try since it doesn't require any changes on our end. It's not always the most reliable environment, but we can try it out.

### Links

- [TDD](link to Paper doc)
- [Figma](link to Figma file)

### Checklist


.vscode/tasks.json
```diff
@@ -3,7 +3,9 @@
   "tasks": [
     {
       "type": "shell",
-      "command": "yarn start",
+      "osx": {
```
rlingineni, line 6: We don't want to auto-run this in the cloud VSCode environment, so changing the setting to only run automatically in our MacOS environment (should be no change on our end)

package.json
```diff
@@ -193,5 +193,8 @@
     "type": "git",
     "url": "https://github.com/pinterest/gestalt"
   },
+  "stackblitz": {
```
rlingineni, line 196: start command when opened in stackblitz.com

scripts/stackblitz.sh
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env sh
+
+# Simplified start script for the stack blitz environment, which can't handle multiple watchers
+set -x
+set -e
+
+(cd packages/gestalt-design-tokens && yarn build)
```
rlingineni, line 7: Notice these are "build" commands, no watcher is enabled to save web memory usage

scripts/stackblitz.sh
```diff
@@ -0,0 +1,13 @@
...
+(cd packages/gestalt && yarn build)
+(cd packages/gestalt-charts && yarn build)
+(cd packages/gestalt-datepicker && yarn build)
+(cd docs && yarn build-metadata)
+(cd docs && (npx next dev))
```
rlingineni, line 12: Only docs are live-reloaded when running in the stackblitz env. Other commands can be run as needed

AlbertCarreras (approved):

PR #3283, opened 2023-10-27
Module: Adding Density Sizing

rlingineni:
Adding Density sizing to module. The current and default module is `lg`. This introduces to smaller sizes `md` and `small`

### Summary
Original:
image

Two new variants:

sm:
image

md:
image

#### What changed?

Keeping the values in the `applyModuleDensity` function method because most of this is built with Box, and that function is reused across expandable, and regular box

### Links

- [TDD](link to Paper doc)
- [Figma](https://www.figma.com/file/Y88VHmN8jtWDhWnrSZ0JuW/Density-theme-handoff?type=design&node-id=7-2760&mode=design&t=48zJl9yQWS6BFqsA-4)

### Checklist


packages/gestalt/src/Module/applyModuleDensity.js
```diff
@@ -0,0 +1,26 @@
+// @flow strict
+import Box from '../Box.js';
+
+type Gap = 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10 | 11 | 12;
+
+const applyModuleDensityStyle = (
+  size: 'sm' | 'md' | 'lg',
+): {
+  gap: Gap,
```
AlbertCarreras, line 9: this is not really 0-12 it's more like 2 | 4 | 6

That or `$ElementType<React$ElementConfig<typeof Flex>, 'gap'>`

packages/gestalt/src/ModuleExpandable.js
```diff
@@ -63,12 +68,16 @@ export default function ModuleExpandable({
...
   items,
   onExpandedChange,
+  size = 'lg',
 }: Props): Node {
   const [expandedId, setExpandedId] = useState<?number>(getExpandedId(expandedIndex));
 
   const { name: colorSchemeName } = useColorScheme();
   const isDarkMode = colorSchemeName === 'darkMode';
 
+  const density = applyModuleDensityStyle(size);
```
AlbertCarreras, line 78: const { rounding } = applyModuleDensityStyle(size);

packages/gestalt/src/Module.js
```diff
@@ -47,6 +52,10 @@ type Props = {
    * If set to `error`, displays error icon and changes title to red text. Be sure to provide an `iconAccessibilityLabel` when set to `error`. See the [error variant](https://gestalt.pinterest.systems/web/module#Static-Error) for more details.
    */
   type?: 'error' | 'info',
+  /**
```
AlbertCarreras, line 55: delete this section, repeated

packages/gestalt/src/Module/__snapshots__/ExpandableItem.test.js.snap
```diff
@@ -42,7 +42,7 @@ exports[`ModuleExpandableItem renders correctly 1`] = `
               className="box xsCol12"
             >
               <div
-                className="Flex rowGap2 columnGap0 xsDirectionRow xsItemsCenter"
+                className="Flex rowGap3 columnGap0 xsDirectionRow xsItemsCenter"
               >
                 <div
```
AlbertCarreras, line 47: rowGap2 > rowGap3 expected?

docs/pages/web/module.js
```diff
@@ -190,6 +191,17 @@ export default function DocsPage({
...
           />
         </MainSection.Subsection>
+        <MainSection.Subsection
+          title="Sizes"
+          description={`Module can have different sizes. The default size is large with a padding of 24px (\`$space-600\`). For a dense variant, use the \`sm\` size with a padding of 8px (\`$space-200\`).`}
+        >
+          <MainSection.Card
+            cardSize="lg"
+            sandpackExample={
+              <SandpackExample name="Expandable With Error Type" code={sizesExample} />
```
AlbertCarreras, line 201: Expandable With Error Type > Density example

packages/gestalt/src/Module.js
```diff
@@ -39,6 +40,10 @@ type Props = {
    * Unique id to identify this Module
    */
   id: string,
+  /**
+   * Sets the size of the Module. See the [size variant](https://gestalt.pinterest.systems.com/web/module#size) to learn more.
+   */
+  size?: 'sm' | 'md' | 'lg',
```
AlbertCarreras, line 46: Curious. Is this density or size?

Could we just have 2 variants? default | dense

What's the md for? Whats the use case?

In the example we only cover sm for compact density

I think here we could move to density prop. With sizes we would have the same issue with Buttons, how much the sm gets reduced. Or maybe the Density provider overrides all sizes to sm instead of moving values down.

If there's no use case for md, I'd advocate to have default  | compact density variants to avoid inconsistencies.

Screenshot by Dropbox Capture

AlbertCarreras (approved):

hectoid:
@rlingineni I don't see any issues!

packages/gestalt/src/Module.js
rlingineni: I think we had this convo here:
https://pinterest.slack.com/archives/G01GN7CE99S/p1699984230644209?thread_ts=1699974863.546079&cid=G01GN7CE99S

packages/gestalt/src/Module/applyModuleDensity.js
rlingineni: This causes flow errors.

The Gap prop on Flex supports two kinds of values, a row/col object, or numerical Gap. In this case, I only want the numerical type

 <Flex direction="column" gap={{ column: gap, row: 0 }}>

PR #3284, opened 2023-10-27
Tag: Add Density Variants

rlingineni:
Adding density sizes to tag.  Default size is `medium`

Still two issues:

- Gap between remove button border radius and outer container. See small variant hover
- A tad extra spacing between the remove button and the end of text content. See truncated example

#### What changed?
image

Added fixed density sizing to various areas. The remove button is now absolutely positioned like TagData so we don't have to size it.

#### Why?

### Links

- [TDD](link to Paper doc)
- [Figma](https://www.figma.com/file/Y88VHmN8jtWDhWnrSZ0JuW/Density-theme-handoff?type=design&node-id=7-3566&mode=design&t=48zJl9yQWS6BFqsA-4)

### Checklist


rlingineni:
@hectoid @laylaleeal 

WIP Tag:

https://deploy-preview-3284--gestalt.netlify.app/web/tag?devexample=true#Size

Still has some issues. Could I get a mockup of the end button hover state and it's size? That's causing some problems.

 I ended up dropping the font size on the smaller variant. Also tweaking the icon size.

hectoid:
@rlingineni a separate issue, but the Warning tag does not have a dismiss button hover color change like the other two variants: https://deploy-preview-3284--gestalt.netlify.app/web/tag#Warning

hectoid:
@rlingineni close icon div is squished to 12x10; should be 12x12. 
image

The icon should also be 12x12, not 10x10. to be consistent with our standard icon sizes and 4pt grid.

We need more padding in the SM variant between text and close button. It's hitting the edge and looks bad. Go with 4px padding. 
image

hectoid:
These aren't density issues, so maybe a separate PR, but @rlingineni in the small tag, text is getting cut off, and the "x" doesn't center-align vertically with the left icon and text. 
image

hectoid:
@rlingineni rounding in the hover state doesn't match rounding of the tag: 
image

rlingineni:
@hectoid  - when you get a chance, can I get a 👍🏾  on this

hectoid:
@rlingineni can you add a large version of the warning or error tag so we can see what it looks like with an icon? Also, suggest labeling: "Large warning tag" and "Medium error tag" to be clear that this depicts size.
image

I think this can then be seen as "approved" without needing another review.

hectoid:
sorry, @rlingineni per the above, the labels can look as shown, and if possible, laid out this way so that users can see each type of tag as small, medium and large:
image

hectoid:
Spacing and sizing all looks good to me! I give that part a 👍🏼.

packages/gestalt/src/Tag.css
```diff
@@ -1,24 +1,71 @@
 .button {
   composes: noBorder from "./Borders.css";
-  composes: paddingX2 from "./boxWhitespace.css";
-  composes: paddingY3 from "./boxWhitespace.css";
   composes: pointer from "./Cursor.css";
+
+  /*
+  Since the button is on an already transparent bg, it appears darker. This adjusts the background color to be a tad brighter since it's layered on top.
+  */
+  backdrop-filter: brightness(1.1);
```
diyorbek, line 8: @rlingineni it feels hacky to me. Would changing the colors instead work in this case? 🤔

rlingineni: Agree, it is a little hacky. It was something I figured out w/ TagData.

https://github.com/pinterest/gestalt/blob/646c4107ec3283f449001574f14a03d800bb8df9/packages/gestalt/src/TagData.css#L36

Since we're overlaying the buttons, and bg color can be dynamic, it uses this

rlingineni: Actually, I removed it, and it seemed to work. I got rid of it. Good catch. I kept it for warning base to retain a darker bg

packages/gestalt/src/Tag.js
```diff
@@ -52,6 +55,44 @@ type Props = {
   type?: 'default' | 'error' | 'warning',
 };
 
+const applyDensityTheme = (s: 'sm' | 'md' | 'lg') => {
```
dangerismycat, line 58: nit: no need to be cryptic, `size` would be a more expressive variable name here 🙂

packages/gestalt/src/TagData.js
```diff
@@ -251,7 +251,7 @@ export default function TagData({
               }
               color={fgColor}
               icon="cancel"
-              size={10}
+              size={8}
```
dangerismycat, line 254: This change appears unrelated? Should this be in a separate PR?

dangerismycat (approved):
Is the close icon really the same size in the PR summary examples? If so this is a dang good optical illusion 😅

diyorbek (approved):

PR #3296, opened 2023-11-13
Internal: Renamed `type Node` to  `type Node as ReactNode`

AlbertCarreras:
#### What changed?

Internal: Renamed `type Node` to  `type Node as ReactNode`

#### Why?

Flow to TypeScript codemods can't diferentiate between React and DOM Node types. We need to specify `type Node as ReactNode` so our custom codemods to migrate to TS can detect React Node, migrate it properly and remove it when it's the return in a function component.


rlingineni (approved):
Should the non packaged Gestalt stuff be changed? E.g. docs components/etc. Or only the stuff in the packages folder

AlbertCarreras:
> Should the non packaged Gestalt stuff be changed? E.g. docs components/etc. Or only the stuff in the packages folder

Everything that the Jack's codemod will move to TS which is everything.

PR #3297, opened 2023-11-13
Internal: Changed config in .eslintrc.js to set import/extensions to never require .js extension

AlbertCarreras:
#### What changed?

Internal: Changed config in .eslintrc.js to set import/extensions to never require .js extension

#### Why?

Preparing for Flow to TypeScript migration. Because in TS we'll have .ts and .tsx and codemod is not properly moving .js to .ts and .tsx and it's creating bugs.


.eslintrc.js
```diff
@@ -57,7 +57,7 @@ module.exports = {
     'flowtype/space-after-type-colon': [ERROR, ALWAYS, { 'allowLineBreak': true }],
     'flowtype/space-before-type-colon': [ERROR, NEVER],
     'flowtype/type-import-style': ERROR,
-    'import/extensions': [ERROR, ALWAYS, { 'ignorePackages': true }],
+    'import/extensions': [ERROR, 'ignorePackages', { 'js': 'never' }],
```
AlbertCarreras, line 60: config change

PR #3311, opened 2023-11-28
SideNavigation: group `expanded` & `onExpand` prop added

AlbertCarreras:
SideNavigation: group `expanded` & `onExpand` prop added

Screenshot by Dropbox Capture
Screenshot by Dropbox Capture

### DIFF
--- a/docs/pages/web/sidenavigation.js
+++ b/docs/pages/web/sidenavigation.js
@@ -398,9 +398,11 @@ To prevent visual overload, do not include counters in the parent if the childre
 1. static: Group is expanded by default and there isn't expanding/collapsing behavior
 2. expandable: Group is expandable and all items are initially collapsed except groups with active children.
 
-When we set \`expand\` to a boolean value, SideNavigation.Group and SideNavigation.NestedGroup become a controlled component. If not passed or set to "undefined", it stays uncontrolled.
+When \`display='expandable'\`, we can manage the state of each group's collapsing/expanded state by setting SideNavigation.Group and SideNavigation.NestedGroup as controlled components.
 
-When \`display='expandable'\`, we can manage the state of each group's collapsing/expanded state by setting  SideNavigation.Group and SideNavigation.NestedGroup as controlled components.
+To work as controlled components, set \`expand\` prop to a boolean value. If not passed or set to "undefined", they stay uncontrolled.
+
+Beware that when controlled, the list path to the active item is not automatically expanded.
 `}
         >
           <MainSection.Card


docs/pages/web/sidenavigation.js
```diff
@@ -383,8 +386,46 @@ To prevent visual overload, do not include counters in the parent if the childre
...
+            sandpackExample={<SandpackExample code={nestedExample} name="Nested example" />}
+          />
+        </MainSection.Subsection>
+
+        <MainSection.Subsection
+          title="Group display"
+          columns={2}
+          description="Group display can be \:
+1. static: group items are visible and there isn't expanding/collapsing behavior
+2. expandable: group items aren't visible un;ess one children is active and there is expanding/collapsing behavior
```
rlingineni, line 398: typo

packages/gestalt/src/SideNavigation/GroupContent.js
```diff
@@ -12,7 +12,7 @@ import Text from '../Text';
 import { type Indexable } from '../zIndex';
 
 type IconType = $Keys<typeof icons> | { __path: string };
-type Display = 'expandable' | 'static';
+type Display = 'expandable' | 'expanded' | 'static';
```
rlingineni, line 15: Wouldn't it be cleaner if expanded was it's own prop? 

Like `<Sidenavigation.Group type="expandable"  expanded />`

Expandable vs Expanded seems a little confusing, and the states don't seem that different no?

rlingineni: Expanded seems like a forced override essentially

rlingineni (approved):
r+ but comments on api

packages/gestalt/src/SideNavigation/GroupContent.js
AlbertCarreras: I agree w ```<Sidenavigation.Group type="expandable"  expanded />```

My only caveat is ```<Sidenavigation.Group type="static"  expanded />``` this now becomes possible.

rlingineni: That's ok right? It's expanded anyways. It's a dead prop at that point

rlingineni: Since all static displays are expanded by default

AlbertCarreras: i think i prefer 

`'static' | 'expandable' | 'expandableExpanded'`


1. We avoid booleans
2. We avoid dead code
3. expandableExpanded is more explicit

I thought about expandable > expandableCollapsed but... that's a breaking change. but we could discuss.

rlingineni: That seems confusing, are you sure?

It seems to me, just from looking at the examples, that expandable means -> collapse icons are showing vs they are not showing. 

expandableExpanded seems like a lot,

`Expanded` is a state variable. Essentially, we're making it a controlled component. It's like `checked`.  In static mode, there are no "expand" states, so it's not that important. Maybe there's a flow type to disable that case

AlbertCarreras: we are not making it a controlled component, we are just setting a new initial state that controls if a group should be be initialized (collapsed vs expanded). But the user can still go and change that and it's still uncontrolled. 

The engineer shouldn't keep the state collapsed/expanded outside, it's uncontrolled to that respect.

AlbertCarreras: Actually I see... some other improvement here.

AlbertCarreras: I see your point now. With expanded prop we can control them. I thought just have them open from the beginning, but posterior changes don't persist. so there's no real differences between that and static.

AlbertCarreras: it probably makes more sense to control it. There are many implementations to control that. Then we need on onExpand too there.

packages/gestalt/src/SideNavigationGroup.js
```diff
@@ -78,9 +86,11 @@ export default function SideNavigationGroup({
   children,
   counter,
   display = 'expandable',
+  expanded: groupExpanded,
```
rlingineni, line 89: it's not clear groupExpanded is the prop. What if we called it `expandedProp`?

docs/examples/sidenavigation/displayExpanded.js
```diff
@@ -0,0 +1,63 @@
...
+import { Box, DeviceTypeProvider, SideNavigation } from 'gestalt';
+
+export default function Example(): ReactNode {
+  const [expandedElements, setExpandedElements] = useState<$ReadOnlyArray<string>>([
+    'Christmas',
+    'Classic Christmas',
+  ]);
+
+  return (
+    <DeviceTypeProvider deviceType="mobile">
```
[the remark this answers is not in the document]
AlbertCarreras: this is a leftover from ttesting , good catch

PR #3312, opened 2023-11-29
Popover: Add experiment hooks to components (rollout phase 1)

diyorbek:
### Summary

Following components are under their own experimentation name:

- ComboBox 
  - web_gestalt_popover_v2_combobox
  - mweb_gestalt_popover_v2_combobox
- Dropdown 
  - web_gestalt_popover_v2_dropdown
  - mweb_gestalt_popover_v2_dropdown
- HelpButton 
  - web_gestalt_popover_v2_helpbutton
  - mweb_gestalt_popover_v2_helpbutton
- OverlayPanel 
  - web_gestalt_popover_v2_confirmationpopover
  - web_gestalt_popover_v2_confirmationpopover
- Popover 
  - web_gestalt_popover_v2
  - mweb_gestalt_popover_v2
- PopoverEducational 
  - web_gestalt_popover_v2_popovereducational
  - web_gestalt_popover_v2_popovereducational
 
#### What changed?

Above listed components now can be switched between old and new implementation of Popover using `ExperimentProvider` by providing correspoding experiment name.

#### Why?

This is the 1st phase of releasing Popover based components. The components in this PR have relative less instances and less chance of introducing major bugs.

### Links

- [Jira](https://jira.pinadmin.com/browse/GESTALT-7089)
- [Jira](https://jira.pinadmin.com/browse/GESTALT-7090)
- [Jira](https://jira.pinadmin.com/browse/GESTALT-7091)
- [Jira](https://jira.pinadmin.com/browse/GESTALT-7092)
- [Jira](https://jira.pinadmin.com/browse/GESTALT-7093)
- [TDD](https://paper.dropbox.com/doc/Gestalt-TDD-Popover-V2--CEo8OkcRGejqo8Kh_nkTx6n9Ag-T6gxnHxXGWEJin2TkENXr#:uid=212219304881499687738554&h2=1st-phase-of-rollout)
- [Figma](link to Figma file)

### Checklist


packages/gestalt/src/Popover.js
```diff
@@ -108,15 +155,39 @@ export default function Popover({
...
-      __dangerouslySetMaxHeight={__dangerouslySetMaxHeight}
+      scrollBoundary={scrollBoundary}
+      hideWhenReferenceHidden={hideWhenReferenceHidden}
     >
       {children}
     </InternalPopover>
   );
 }
+
+export function ExperimentalPopover({
```
AlbertCarreras, line 172: The ExperimentProvider goes in Pinboard. We read from within the component, but we set the provider at the app level

docs/docs-components/contexts/DocsExperimentProvider.js
```diff
@@ -9,7 +9,20 @@ import { useAppContext } from '../appContext';
...
+  OverlayPanel: [
+    'web_gestalt_popover_v2_confirmationpopover',
+    'mweb_gestalt_popover_v2_confirmationpopover',
+  ],
+  Popover: ['web_gestalt_popover_v2', 'mweb_gestalt_popover_v2'],
+  PopoverEducational: [
+    'web_gestalt_popover_v2_popovereducational',
+    'mweb_gestalt_popover_v2_popovereducational',
+  ],
+};
```
AlbertCarreras, line 25: The Provider below acts as the one in Pinboard but for our docs

packages/gestalt/src/Popover/InternalPopover.js
```diff
@@ -56,12 +56,12 @@ export default function InternalPopover({
 
   useEffect(() => {
     dismissButtonRef.current?.focus();
+    console.log('new pop');
```
AlbertCarreras, line 59: this might complain. we shoudn't ship this

scripts/generateMetadata.js
```diff
@@ -36,10 +36,13 @@ async function docgen(filePath) {
   }
 
   const contents = await fs.promises.readFile(filePath, 'utf-8');
+  const reactDocs = await reactDocsImport;
+  const resolver = new reactDocs.builtinResolvers.FindExportedDefinitionsResolver();
 
   // Not all files have data to parse
   try {
-    const [parsed] = (await reactDocs).parse(contents);
+    // Take only the first exported component
```
AlbertCarreras, line 44: we might not need this if we remove the experiment provider

packages/gestalt/src/Popover.js
diyorbek: Yes. The experiments for Popover and its dependent components are separate. 

For example, ComboBox has its own experiment and Popover has its own too. When ComboBox experiment is switched on, it has to switch on the Popover experiment for the underlying Popover.

I could just add a new prop directly to Popover but it would pollute the API.

AlbertCarreras: Oh I see. Just add a prop to Popover _experimentalPopover, an internal prop basically rather than triggering the experiment. We've done this in the past. https://gestalt.pinterest.systems/get_started/developers/contributing/experimentation#Experimental-prop 

Here you won't be able to set (easily) 2 experiments that simultaneously get the same audience in the experiment. 

Let's control the experiment from the dependent components and have them internally trigger the change in Popover.

diyorbek: If I understand you correcly, internally we should control the experiment via `_experimentalPopover` prop and not use `ExperimentProvider`. And in Pinboard we will use `ExperimentProvider` to control the experiment. Right?

AlbertCarreras: Yes
in Pinboard we will use ExperimentProvider to control 

- ComboBox
  - web_gestalt_popover_v2_combobox
  - mweb_gestalt_popover_v2_combobox
- Dropdown
  - web_gestalt_popover_v2_dropdown
  - mweb_gestalt_popover_v2_dropdown
- HelpButton
  - web_gestalt_popover_v2_helpbutton
  - mweb_gestalt_popover_v2_helpbutton
- OverlayPanel
  - web_gestalt_popover_v2_confirmationpopover
  - web_gestalt_popover_v2_confirmationpopover
- PopoverEducational
  - web_gestalt_popover_v2_popovereducational
  - web_gestalt_popover_v2_popovereducational

Then internally, we pass a hidden prop from each component to Popover via _experimentalPopover communicating that Popover is in the experiment

AlbertCarreras: This will allow you to "also" test Popover manually on individual instances in Pinboard. You can do _experimentalPopover={true} in a Popover and see what happens.

AlbertCarreras:
@diyorbek There's some weird behavior in the Docs when you activate the experiment and open/close popover it scrolls all the way back to the top. Any thoughts?

diyorbek:
@AlbertCarreras That behavior with unexpected scrolling was happening only in Dropdown component. So, before `focus()`ing an element inside Popover, we should wait for `isPositioned` property from floating-ui to be true. [Source](https://github.com/floating-ui/floating-ui/pull/2001). Otherwise it causes focusing to an element outside of viewport which just causes scrolling to the top of the page.

I added `__onPositioned` internal prop to Popover to listen to `isPositioned`.

AlbertCarreras:
https://deploy-preview-3312--gestalt.netlify.app/web/dropdown Dropdown still behaves weirdly... like it jumps all around... can we fix this?

packages/gestalt/src/Dropdown.js
```diff
@@ -279,6 +291,7 @@ export default function Dropdown({
       shouldFocus
       size="xl"
       __dangerouslySetMaxHeight={maxHeight}
+      __onPositioned={() => setIsPopoverPositioned(true)}
```
AlbertCarreras, line 294: we could merge this for now, but I think I'd be better to use some InternalPopover rather than the exposed Ppopover here in the next PRs... we can keep some props internal in the InternalPopover and we don't have to expose them.

AlbertCarreras (approved):

PR #3320, opened 2023-12-05
Module: renaming Module to Accordion

AlbertCarreras:
Accordion: renaming Module to Accordion

### Breaking change

1. Rename

Run
```
yarn codemod renameComponent ~/path/to/your/code
  --componentName=Module
  --nextComponentName=Accordion
```

Also, manually detect subcomponents 

```
--component=Module
--subcomponent=Expandable
--prop= id
```

2. DefaultLabelProvider

Update accordingly to add Accordion default strings https://gestalt.pinterest.systems/web/utilities/defaultlabelprovider


rlingineni:
I want to merge this one first since it's Module related:

https://github.com/pinterest/gestalt/pull/3283

hectoid:
@AlbertCarreras text updated suggestion for the component description. instead of: "all times, or expand and collapse as individual modules or a group of modules"
change to:
"all items, or expand and collapse as an individual item or a group of items." So we avoid the term "module".

Do we know how often people are using the non-collapsible accordion? I realize that if somethng doesn't collapse or expand, it's no longer an accordion…

hectoid:
@AlbertCarreras  Suggest updating this language on the Component overview as it says nothing of what makes the container that holds content special. 
image

Suggestion: Accordion is a container that can be expanded and collapsed to show content about a single subject.

rlingineni (approved):
Seems okay 👍🏾

PR #3321, opened 2023-12-07
PopoverEducational: Add `disablePortal`

diyorbek:
### Summary

PopoverEducational should always be rendered near its anchor, not outside the root through portal.

#### Why?

When `disablePortal` is not set, Popover is rendered through portal and `zIndex` prop of PopoverEducational will have no effect.

### Links

- [TDD](link to Paper doc)
- [Figma](link to Figma file)

### Checklist

### DIFF
--- a/packages/gestalt/src/PopoverEducational.js
+++ b/packages/gestalt/src/PopoverEducational.js
@@ -210,6 +210,7 @@ export default function PopoverEducational({
         id={id}
         idealDirection={idealDirection}
         onDismiss={onDismiss}
+        disablePortal
         showCaret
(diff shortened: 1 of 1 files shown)


rlingineni:
Should the docs also be updated with this information

diyorbek:
@rlingineni I don't think so. It is internal and with the old Popover implementation it had `positionRelativeToAnchor` prop. With the new Popover the same behavior is achived by `disablePortal` prop of InternalPopover. I kinda fixed a bug here.

rlingineni (approved):

PR #3322, opened 2023-12-11
Toast: reduce side margins from 32 to 16 px

AlbertCarreras:
#### What changed?

Toast: reduce side margins from 32 to 16 px

#### Why?

Designs need more Toast space in mobile devices and small screens.


rlingineni:
Any snapshot updates?

rlingineni (approved):

PR #3325, opened 2023-12-12
Internal: test to fix darkmode

AlbertCarreras:
Dark mode is broken in Gestalt Docs: components using React Portal is set outside the ColorSquemeProvider and doesn't  get the darkmode changes.

In Pinboard, we set Gestalt CSS in the page header. I was trying to replicate that... as it was mentioned as the reason why Pinboard doesn't have this same issue.

I'm testing this solution

Internal: test to fix darkmode
Screenshot by Dropbox Capture


diyorbek:
@AlbertCarreras I did not understand the purpose of this PR. Could you write more in the description?

PR #3332, opened 2023-12-13
BannerOverlay: New Component

eniomoura:
#### What changed?

New component - BannerOverlay was added.
image

#### Why?

[BannerOverlays](https://gestalt.pinterest.systems/web/banneroverlay) displays short educational messages when users have performed actions that indicate some intent, such as a related pin tap or idea pin swipe. It is scrim-less, meaning users can scroll content underneath without having a wash layer on top of the content.

### Links

- [Jira](https://jira.pinadmin.com/browse/GESTALT-6096)
- [TDD](https://docs.google.com/document/d/1YqwRkeMyRFKWByexkMQoKDnmCV09Za04jbMO1wXeZKk/edit)
- [Figma](https://www.figma.com/file/FgT8AK1jcMno2mWSUmvjLI/Gestalt-for-web-(Copy)?type=design&node-id=29448-27585&mode=design&t=DsEKpJjBrarepNFR-0)

### Checklist

- [ Yes ] Added unit and Flow Tests
- [ Yes ] Added documentation + accessibility tests


eniomoura:
Turned into Major Release due to change in DefaultLabelProvider API.

packages/gestalt/src/BannerOverlay.js
```diff
@@ -0,0 +1,276 @@
...
+};
+
+type Props = {
+  /**
+   * Adds a dismiss button to BannerOverlay. See the [Dismissible variant](https://gestalt.pinterest.systems/web/banneroverlay#Dismissible) for more info.
+   * The `accessibilityLabel` should follow the [Accessibility guidelines](https://gestalt.pinterest.systems/web/banneroverlay#Accessibility).
+   *
+   */
+  dismissButton?: {
+    accessibilityLabel?: string,
```
AlbertCarreras, line 41: accessibilityLabel?: string >> Can we just remove this one and use a default string from provider?  "Dismiss banner"

AlbertCarreras: i think we can just do onDismiss

packages/gestalt/src/BannerOverlay.js
```diff
@@ -0,0 +1,276 @@
...
+   */
+  message: string | Element<typeof Text>,
+  /**
+   * Distance (in pixels) from the viewport edge (offsetTop will be used, if desktop, offsetBottom will be used, if mobile).
+   */
+  offsetTop?: number,
+  /**
+   * Distance (in pixels) from the viewport edge (offsetTop will be used, if desktop, offsetBottom will be used, if mobile).
+   */
+  offsetBottom?: number,
```
AlbertCarreras, line 55: could we do `offset: { bottom: x, top: x}`

Also, can we just do 0 as default values?

packages/gestalt/src/BannerOverlay.js
```diff
@@ -0,0 +1,276 @@
...
+   */
+  primaryAction?:
+    | {
+        accessibilityLabel: string,
+        href: string,
+        label: string,
+        onClick?: $ElementType<ElementConfig<typeof ButtonLink>, 'onClick'>,
+        rel?: $ElementType<ElementConfig<typeof Link>, 'rel'>,
+        role: 'link',
+        size?: $ElementType<ElementConfig<typeof Button>, 'size'>,
```
AlbertCarreras, line 83: typeof ButtonLink

packages/gestalt/src/BannerOverlay.js
```diff
@@ -0,0 +1,276 @@
...
+   * The accessibilityLabel should follow the Accessibility guidelines.
+   * See the Primary action variant to learn more.
+   */
+  primaryAction?:
+    | {
+        accessibilityLabel: string,
+        href: string,
+        label: string,
+        onClick?: $ElementType<ElementConfig<typeof ButtonLink>, 'onClick'>,
+        rel?: $ElementType<ElementConfig<typeof Link>, 'rel'>,
```
AlbertCarreras, line 81: typeof ButtonLink

packages/gestalt/src/BannerOverlay.js
```diff
@@ -0,0 +1,276 @@
...
+        role: 'link',
+        size?: $ElementType<ElementConfig<typeof Button>, 'size'>,
+        target?: $ElementType<ElementConfig<typeof Link>, 'target'>,
```
AlbertCarreras, line 84: typeof ButtonLink

packages/gestalt/src/BannerOverlay.js
```diff
@@ -0,0 +1,276 @@
...
+    text: string,
+    accessibilityLabel: string,
+    href: string,
+    onClick?: ({
+      event: SyntheticMouseEvent<HTMLAnchorElement> | SyntheticKeyboardEvent<HTMLAnchorElement>,
+      dangerouslyDisableOnNavigation: () => void,
+    }) => void,
+  },
+  /**
+   * Adds an optional primary button for user interaction.
```
AlbertCarreras, line 69: See `OnLinkNavigationProvider`... add link here

packages/gestalt/src/contexts/DefaultLabelProvider.js
```diff
@@ -131,6 +134,9 @@ export const fallbackLabels: DefaultLabelContextType = {
     accessibilityCollapseLabel: 'Collapse section',
     accessibilityExpandLabel: 'Expand section',
   },
+  BannerOverlay: {
+    accessibilityDismissButtonLabel: 'Dismiss',
```
AlbertCarreras, line 138: Dismiss banner

packages/gestalt/src/BannerOverlay/subcomponents.js
```diff
@@ -0,0 +1,148 @@
...
+import Icon from '../Icon';
+import Image from '../Image';
+import Link from '../Link';
+import Mask from '../Mask';
+import Text from '../Text';
+
+const SIZE_THUMBNAIL = 40;
+const SIZE_ICON = 40;
+
+export function BannerOverlayMessage({
```
AlbertCarreras, line 24: can we reuse this component from Toast directly?

packages/gestalt/src/BannerOverlay/subcomponents.js
```diff
@@ -0,0 +1,148 @@
...
+          >
+            {helperLink?.text}
+          </Link>
+        </Text>
+      ) : null}
+    </Fragment>
+  );
+}
+
+export function BannerOverlayImageThumbnail({
```
AlbertCarreras, line 120: Could we reuse these subcomponents from Toast directly as well?

Maybe instead of under Toast, they can be under /shared

packages/gestalt/src/BannerOverlay/PrimaryAction.js
```diff
@@ -0,0 +1,54 @@
...
+    }
+  | {
+      accessibilityLabel: string,
+      label: string,
+      onClick: $ElementType<ElementConfig<typeof Button>, 'onClick'>,
+      role: 'button',
+      size?: $ElementType<ElementConfig<typeof Button>, 'size'>,
+    };
+
+export default function PrimaryAction({
```
AlbertCarreras, line 25: can we reuse this too?

packages/gestalt/src/BannerOverlay.js
```diff
@@ -0,0 +1,276 @@
...
+ * BannerOverlay is a sticky component triggered by scroll, on tap or long-press, placed at the bottom of the screen.
+ * ![BannerOverlay light mode](https://raw.githubusercontent.com/pinterest/gestalt/master/playwright/visual-test/BannerOverlay.spec.mjs-snapshots/BannerOverlay-chromium-darwin.png)
+ */
+export default function BannerOverlay({
+  message,
+  title,
+  dismissButton,
+  helperLink,
+  primaryAction,
+  offsetTop = 130,
```
AlbertCarreras, line 122: i think maybe the defaults are 0? to the engineer to find the correct offset?

packages/gestalt/src/BannerOverlay.js
```diff
@@ -0,0 +1,276 @@
...
+   * An optional thumbnail to display next to the text.
+   */
+  thumbnail?:
+    | { image: Element<typeof Image> }
+    | { avatar: Element<typeof Avatar> }
+    | { icon: Element<typeof Icon> },
+  /**
+   *  Heading of BannerOverlay. Content should be [localized](https://gestalt.pinterest.systems/web/banneroverlay#Localization). See the [Text variant](https://gestalt.pinterest.systems/web/banneroverlay#Text) to learn more.
+   */
+  title?: string | Element<typeof Text>,
```
AlbertCarreras, line 103: why do we allow title to be Text? I'm thinking that title should always be a string no? 

No need for links or bold in the title.

packages/gestalt/src/BannerOverlay.js
```diff
@@ -0,0 +1,276 @@
...
+  if (typeof title === 'string') {
+    messageTextElement = title;
+  }
+
+  // If `message` is a Text component, we need to override any text colors within to ensure they all match
+  const messageIsTextNode =
+    typeof message !== 'string' &&
+    Children.only<Element<typeof Text>>(message).type.displayName === 'Text';
+
+  if (messageIsTextNode) {
```
AlbertCarreras, line 146: we should dry one day this code! ahaha one day

packages/gestalt/src/BannerOverlay.js
```diff
@@ -0,0 +1,276 @@
...
+              onClick={primaryAction.onClick}
+              role="button"
+              size="sm"
+            />
+          ))
+        : null}
+    </Flex.Item>
+  );
+  return (
+    <Layer>
```
AlbertCarreras, line 211: This component is just not ging to be accessible... but I don't know how to it at the moment. It's goig to be at the bottom of DOM tree. And it's rendered since the page loads... so it's kind of unreachable.

Can we make some research on how to make floating Banners keyboard accessible?

packages/gestalt/src/BannerOverlay.js
```diff
@@ -0,0 +1,276 @@
...
+      <Box
+        zIndex={zIndex}
```
AlbertCarreras, line 213: should we put the xIndex in the Layer?

docs/pages/web/banneroverlay.js
```diff
@@ -0,0 +1,150 @@
...
+import Page from '../../docs-components/Page';
+import PageHeader from '../../docs-components/PageHeader';
+import QualityChecklist from '../../docs-components/QualityChecklist';
+import SandpackExample from '../../docs-components/SandpackExample';
+import main from '../../examples/banneroverlay/main';
+
+export default function DocsPage({ generatedDocGen }: { generatedDocGen: DocGen }): ReactNode {
+  return (
+    <Fragment>
+      <BannerOverlay
```
AlbertCarreras, line 18: can we just leave one banneroverlay? haha it gets confusing with the main example one or place them at tdifferent heights

packages/gestalt/src/BannerOverlay.js
```diff
@@ -0,0 +1,276 @@
...
+  primaryAction,
+  offsetTop = 130,
+  offsetBottom = 24,
+  thumbnail,
+  zIndex,
+}: Props): ReactNode {
+  const { name: colorSchemeName } = useColorScheme();
+  const isDarkMode = colorSchemeName === 'darkMode';
+
+  const isMobileWidth = useDeviceType() === 'mobile';
```
AlbertCarreras, line 130: isMobileDevice instead? isMobileWidth confuses me with media query

packages/gestalt/src/BannerOverlay.js
```diff
@@ -0,0 +1,276 @@
...
+    messageTextElement = <span className={textColorOverrideStyles}>{message}</span>;
+  }
+
+  const { accessibilityDismissButtonLabel: accessibilityDismissButtonLabelDefault } =
+    useDefaultLabelContext('BannerOverlay');
+
+  const { lightModeBackground, darkModeBackground, textColor } = DEFAULT_COLORS;
+
+  const dismissButtonComponent = dismissButton && (
+    <Flex.Item flex="none" alignSelf={isMobileWidth ? 'end' : 'center'}>
```
AlbertCarreras, line 160: consider narrow viewports web... meaning when is desktop BUT we are in narrow window. You might want to play with 

display: none smDisplay: flex etc.

packages/gestalt/src/BannerOverlay.js
```diff
@@ -0,0 +1,276 @@
...
+            </Flex.Item>
+          ) : null}
+
+          {!!thumbnail?.avatar &&
+          Children.only<Element<typeof Avatar>>(thumbnail.avatar).type.displayName === 'Avatar' ? (
+            <Flex.Item flex="none">
+              <BannerOverlayAvatarThumbnail thumbnail={thumbnail.avatar} />
+            </Flex.Item>
+          ) : null}
+          {messageComponent}
```
AlbertCarreras, line 267: can we inline messageComponent?

packages/gestalt/src/BannerOverlay.js
```diff
@@ -0,0 +1,276 @@
...
+              accessibilityLabel={primaryAction.accessibilityLabel}
+              href={primaryAction.href}
+              label={primaryAction.label}
+              onClick={primaryAction.onClick}
+              rel={primaryAction?.rel}
+              role="link"
+              size="sm"
+              target={primaryAction?.target}
+            />
+          ),
```
AlbertCarreras, line 197: Im curious about this comma here... I've never done this myself before hahah

packages/gestalt/src/BannerOverlay.js
```diff
@@ -0,0 +1,276 @@
...
+          primaryAction.role !== 'link' && (
+            <PrimaryAction
+              accessibilityLabel={primaryAction.accessibilityLabel}
+              label={primaryAction.label}
+              onClick={primaryAction.onClick}
+              role="button"
+              size="sm"
+            />
+          ))
+        : null}
```
AlbertCarreras, line 207: this null is unreachable right?

packages/gestalt/src/BannerOverlay.js
AlbertCarreras: could this work? 

```
primaryAction && primaryAction?.accessibilityLabel && primaryAction?.label (  <Flex.Item flex={isMobileWidth ? 'shrink' : 'none'}>
      {
        primaryAction.role === 'link'  ? (
            <PrimaryAction
              accessibilityLabel={primaryAction.accessibilityLabel}
              href={primaryAction.href}
              label={primaryAction.label}
              onClick={primaryAction.onClick}
              rel={primaryAction?.rel}
              role="link"
              size="sm"
              target={primaryAction?.target}
            />
          )
        : (  <PrimaryAction
              accessibilityLabel={primaryAction.accessibilityLabel}
              label={primaryAction.label}
              onClick={primaryAction.onClick}
              role="button"
              size="sm"
            /> ) }
    </Flex.Item>)

```

AlbertCarreras:
I would start building the examples but I would have a toggle on/off to show them as they all overlap at the top on development mode. Thoughts?

packages/gestalt/src/BannerOverlay.js
eniomoura: DONE

packages/gestalt/src/BannerOverlay.js
eniomoura: DONE

packages/gestalt/src/BannerOverlay.js
eniomoura: DONE

packages/gestalt/src/BannerOverlay.js
eniomoura: DONE

packages/gestalt/src/BannerOverlay.js
eniomoura: DONE

packages/gestalt/src/BannerOverlay.js
eniomoura: DONE

packages/gestalt/src/BannerOverlay.js
eniomoura: You're correct, had checked this before, but just forgot to change the type.
DONE

packages/gestalt/src/BannerOverlay.js
eniomoura: One day! heheh
Any suggestions for now?

packages/gestalt/src/BannerOverlay.js
eniomoura: I'll try researching a bit, but I assume this is more of an issue with Layer. Any time it is used, the children in the tree are not accessible. We should consider a solution for this considering the issue with Layer itself later, though.

packages/gestalt/src/BannerOverlay.js
eniomoura: DONE

docs/pages/web/banneroverlay.js
eniomoura: Yeah, this is just there while documentation work is still pending. I'll leave this unresolved to remind myself to remove it before final approval.

packages/gestalt/src/BannerOverlay.js
eniomoura: DONE

packages/gestalt/src/BannerOverlay.js
eniomoura: DONE

packages/gestalt/src/BannerOverlay.js
eniomoura: Didn't understand this code syntax... Is this first line supposed to be a condition for a nested ternary?
Tried pasting it and vscode didn't understand it either, it look like a function signature heheh
The commas are the same behavior as an array, this has the same behavior and should be more familiar:
`const CTAComponent = primaryAction && (
    <Flex.Item flex={isMobileDevice ? 'shrink' : 'none'}>
      {primaryAction?.accessibilityLabel && primaryAction?.label
        ? [
            primaryAction.role === 'link' && (
              <PrimaryAction
                accessibilityLabel={primaryAction.accessibilityLabel}
                href={primaryAction.href}
                label={primaryAction.label}
                onClick={primaryAction.onClick}
                rel={primaryAction?.rel}
                role="link"
                size="sm"
                target={primaryAction?.target}
              />
            ),
            primaryAction.role !== 'link' && (
              <PrimaryAction
                accessibilityLabel={primaryAction.accessibilityLabel}
                label={primaryAction.label}
                onClick={primaryAction.onClick}
                role="button"
                size="sm"
              />
            ),
          ]
        : null}
    </Flex.Item>
  );`

packages/gestalt/src/BannerOverlay.js
eniomoura: Yeah, I'll remove this ternary.

packages/gestalt/src/BannerOverlay/PrimaryAction.js
eniomoura: DONE

packages/gestalt/src/BannerOverlay/subcomponents.js
eniomoura: DONE

packages/gestalt/src/BannerOverlay/subcomponents.js
eniomoura: DONE

packages/gestalt/src/BannerOverlay.js
eniomoura: DONE

packages/gestalt/src/BannerOverlay.js
eniomoura: Got rid of the Layer component entirely, opting instead for a css-based approach.

packages/gestalt/src/BannerOverlay.js
eniomoura: I separated that part of the code as a function so it's a least modular.

eniomoura: I think this one probably refers to the thumbnail code down a bit from here, but just in case you meant the actual text node check, i separated that as a function so it's a bit more modular heheh

packages/gestalt/src/BannerOverlay.js
eniomoura: This probably refers to some old handler for Link that got pasted here during documentation. I'll double check with Cintia, but I believe this can just be removed.

docs/pages/web/banneroverlay.js
eniomoura: DONE

eniomoura:
> I would start building the examples but I would have a toggle on/off to show them as they all overlap at the top on development mode. Thoughts?

I'm using the 'showComponent' state to know whether I'm in dev mode or sandpack. If it's sandpack, it'll render the BannerOverlay normally, since the viewport is limited to the sandpack box, but if it's dev mode it'll start with a button instead, that can be pressed so each BannerOverlay renders only when their buttons are pressed.

docs/examples/banneroverlay/main.js
```diff
@@ -0,0 +1,43 @@
...
+    <Button
+      onClick={() => {
+        setShowComponent(true);
+      }}
+      text="Show BannerOverlay"
+    />
+  ) : (
+    <BannerOverlay
+      zIndex={new FixedZIndex(100)}
+      offset={{ top: 130, bottom: 24 }}
```
rlingineni, line 18: I'm not sure if we should expose offset. In Pinboard, we have a "ToastManager" for positioning. Is it like toast at the same location every time across all surfaces?

I think if we let people set this, they'll start putting it in funky places

packages/gestalt/src/BannerOverlay.css
```diff
@@ -0,0 +1,17 @@
+.textColorOverrideLight * {
+  /* stylelint-disable-next-line declaration-no-important */
+  color: var(--color-text-dark) !important;
```
rlingineni, line 3: why are these ovverrides?

packages/gestalt/src/BannerOverlay.js
```diff
@@ -0,0 +1,262 @@
...
+
+  // If `message` is a Text component, we need to override any text colors within to ensure they all match
+  const checkTextNode = () => {
+    const messageIsTextNode =
+      typeof message !== 'string' &&
+      Children.only<Element<typeof Text>>(message).type.displayName === 'Text';
+
+    if (messageIsTextNode) {
+      const textColorOverrideStyles = isDarkMode
+        ? styles.textColorOverrideDark
```
rlingineni, line 137: no need for this, you can get it from `--text-color-default`. Color scheme provider automatically flips the text I think?

packages/gestalt/src/BannerOverlay.js
```diff
@@ -0,0 +1,262 @@
...
+        : styles.textColorOverrideLight;
+
+      messageTextElement = <span className={textColorOverrideStyles}>{message}</span>;
```
rlingineni, line 140: I think I kind of follow what's happening here. You can set the color prop on the Text node by doing. `const updatedTextElement = cloneElement(TextNode, {color:'inverse'})` this will enable you set the props on the TextElement

packages/gestalt/src/BannerOverlay.js
```diff
@@ -0,0 +1,262 @@
...
+          size="sm"
+        />
+      )}
+    </Flex.Item>
+  );
+
+  const isMessageTextNode = checkTextNode();
+  return (
+    <Box
+      color={isDarkMode ? darkModeBackground : lightModeBackground}
```
rlingineni, line 189: I think just using "background" will work. Color scheme provider will take care

rlingineni: Oh, unless this isn't a 1:1 flip, in that case, I guess it makes sense?

packages/gestalt/src/BannerOverlay.js
```diff
@@ -0,0 +1,262 @@
...
+      direction={isMobileDevice ? 'column' : 'row'}
+      smPaddingY={4}
+      fit
+      maxWidth={isMobileDevice ? 348 : 900}
+      width="100%"
+      zIndex={zIndex}
+    >
+      {isMobileDevice && dismissButtonComponent}
+      <Flex alignItems="center" gap={4}>
+        {!!thumbnail?.image &&
```
rlingineni, line 226: nit: Totally, stylistic, but would it be cleaner to make these a single BannerImage Component?

Then we can also re-use this logic in Toast as a helper

```
<BannerImage thumbnail={thumbnail}>
```

````
const ThumbnailImage = () => {

if(thumbnail.image &&  
Children.only<Element<typeof Image>>(thumbnail.image).type.displayName === 'Image') 
return <ToastImageThumbnail thumbnail>

if(thumbnail.avatar &&  
Children.only<Element<typeof Image>>(thumbnail.image).type.displayName === 'Avatar') 
return <ToastAvatarThumbnail thumbnail>

return null;

}
 
```

rlingineni: I guess you can add it here and re-use in both places? 

[packages/gestalt/src/Toast/subcomponents.js](https://github.com/pinterest/gestalt/blob/39211e6b3d6b47a09bb4a840b98dd5147712602f/packages/gestalt/src/Toast/subcomponents.js)

Again, optional

packages/gestalt/src/BannerOverlay.js
```diff
@@ -0,0 +1,262 @@
...
+      <IconButton
+        accessibilityLabel={accessibilityDismissButtonLabelDefault}
+        icon="cancel"
+        iconColor="darkGray"
+        onClick={onDismiss}
+        size="xs"
+      />
+    </Flex.Item>
+  );
+  const CTAComponent = primaryAction && (
```
rlingineni, line 161: Keep the conditional in the main logic for readability, because it's hard to follow if this will render or not

```
<Flex direction="row" alignSelf={isMobileDevice ? 'end' : 'center'} gap={4}>
        {primaryAction && CTAComponent}
        {!isMobileDevice && dismissButtonComponent}
      </Flex>

```

packages/gestalt/src/BannerOverlay.js
```diff
@@ -0,0 +1,262 @@
...
+    }
+    return messageIsTextNode;
+  };
+
+  const { accessibilityDismissButtonLabel: accessibilityDismissButtonLabelDefault } =
+    useDefaultLabelContext('BannerOverlay');
+
+  const { lightModeBackground, darkModeBackground, textColor } = DEFAULT_COLORS;
+
+  const dismissButtonComponent = !!onDismiss && (
```
rlingineni, line 150: Keep the conditional in the main logic for readability, because it's hard to follow if this will render or not

so {!!onDismiss && dismissButtonComponent}

packages/gestalt/src/BannerOverlay.js
rlingineni: Or maybe a new file right next to it

packages/gestalt/src/BannerOverlay.js
```diff
@@ -0,0 +1,262 @@
...
+  const isMessageTextNode = checkTextNode();
+  return (
+    <Box
+      color={isDarkMode ? darkModeBackground : lightModeBackground}
+      paddingX={4}
+      paddingY={3}
+      rounding={4}
+      borderStyle="shadow"
+      dangerouslySetInlineStyle={
+        isMobileDevice
```
rlingineni, line 195: You could add the conditional in the property and compact is some more like:

```
 __style: {
                position: 'fixed',
                bottom: isMobileDevice ? offset.bottom : 'unset';
                top: !isMobileDevice ? offset.top : 'unset'
                left: '50%',
                transform: 'translateX(-50%)',
              },

```

rlingineni (commented):
Hey nice work, it seems to work fine. I have some feedback on the API, and some code structure

docs/examples/banneroverlay/main.js
eniomoura: I was initially against this too, but after considering the options with Alberto this seemed like the best tradeoff (the final prop version was suggested by him heheh).
This could be also implemented in Pinboard with a gestaltExtension redefining the API and implementing something like a BannerOverlayManager as well.

packages/gestalt/src/BannerOverlay.js
eniomoura: This is following the Toast pattern to accept Text Nodes as content without breaking the style for the component, won't color:inverse fail to deal with anything not black/white?

packages/gestalt/src/BannerOverlay.js
eniomoura: I'll create a follow-up for this, since it will also affect Toast

packages/gestalt/src/BannerOverlay.js
eniomoura: I fiddled with this a bit and decided to follow Toast's pattern. The issue I had was that the default background wasn't suitable for dark mode elevation floating, but when I used 'elevationFloating' instead, the color for light mode background was transparent, not white. I remember this being mentioned in a Gestalt meeting by Hector as an issue once.

packages/gestalt/src/BannerOverlay.js
eniomoura: Done!

packages/gestalt/src/BannerOverlay.js
eniomoura: Done!

packages/gestalt/src/BannerOverlay.js
eniomoura: Neat! Done :)

packages/gestalt/src/BannerOverlay.css
eniomoura: See line 140 comment

packages/gestalt/src/BannerOverlay.js
eniomoura: See line 140 comment

docs/pages/web/banneroverlay.js
```diff
@@ -0,0 +1,206 @@
...
+- To help users learn more about a specific idea or interest
+- To support users when they have performed actions that indicate medium or high intent (related pin tap, idea pin swipe)
+        `}
+          />
+          <MainSection.Card
+            cardSize="md"
+            type="don't"
+            title="When not to use"
+            description={`
+- If there is a need to block the content underneath Use Modal instead
```
rlingineni, line 51: sentence casing

docs/pages/web/banneroverlay.js
```diff
@@ -0,0 +1,206 @@
...
+- To replace Toast
+- For lengthy messages, forms, or blocks of information. Consider OverlayPanel or new page instead
+        `}
+          />
+        </MainSection.Subsection>
+      </MainSection>
+      <MainSection name="Best Practices">
+        <MainSection.Subsection columns={1}>
```
rlingineni, line 59: This should be 2 columns right? Like our other docs wdyt

rlingineni: Or is for size wrapping it's like this?

packages/gestalt/src/BannerOverlay.js
rlingineni: Ah I see, if it's a pattern we have then I think it'll be okay to keep

packages/gestalt/src/Shared/ToastSubcomponents.js
```diff
@@ -139,6 +139,14 @@ export function ToastImageThumbnail({
...
 
+export function BannerOverlayIconThumbnail({
+  thumbnail,
+}: {
+  thumbnail: Element<typeof Icon>,
+}): ReactNode {
+  return <Box aria-hidden>{cloneElement(thumbnail, { size: SIZE_ICON })}</Box>;
+}
+
 export function ToastIconThumbnail({ thumbnail }: { thumbnail: Element<typeof Icon> }): ReactNode {
```
rlingineni, line 150: Can you add a prop to this? It looks like the exact same thing

docs/pages/visual-test/BannerOverlay-dark.js
```diff
@@ -0,0 +1,40 @@
+// @flow strict
+import { type Node as ReactNode } from 'react';
+import { BannerOverlay, ColorSchemeProvider, FixedZIndex, Image, Link, Text } from 'gestalt';
+
+export default function Example(): ReactNode {
+  return (
+    <ColorSchemeProvider colorScheme="dark">
+      <BannerOverlay
```
rlingineni, line 8: Can you reference the light snapshot here, since they are the same? This is a new pattern. Take a look at SegmentedControl. That way we only have to update one

docs/pages/visual-test/BannerOverlay.js
```diff
@@ -0,0 +1,38 @@
+// @flow strict
+import { type Node as ReactNode } from 'react';
+import { BannerOverlay, FixedZIndex, Image, Link, Text } from 'gestalt';
+
+export default function Example(): ReactNode {
+  return (
+    <BannerOverlay
```
rlingineni, line 7: Can you also add other variants like secondary button

docs/pages/web/banneroverlay.js
```diff
@@ -0,0 +1,206 @@
...
+
+On desktop devices, the BannerOverlay should appear at the top of the screen (below navigation). The BannerOverlay shouldn't block navigation.
+On mobile devices, the BannerOverlay should appear at the bottom of the screen. The 'offset' prop can be used to adjust the position of the BannerOverlay.`}
+          />
+          <SandpackExample code={mobile} name="Main BannerOverlay example" hideEditor />
+          <SandpackExample code={desktop} name="Main BannerOverlay example" hideEditor />
+        </MainSection.Subsection>
+      </MainSection>
+
+      <AccessibilitySection name={generatedDocGen?.displayName}>
```
rlingineni, line 147: We generally have the accessbility section up at the top

docs/pages/web/banneroverlay.js
```diff
@@ -0,0 +1,206 @@
...
+        <MainSection.Subsection
+          title="Labels"
+          description={`
+          \`dismissButton\` and \`primaryAction\`, require a short, descriptive label for screen readers, which should also be localized.
+
+          Icons and thumbnails on BannerOverlay are purely decorative, and can therefore have an empty string as the \`accessibilityLabel\`. The thumbnail (Image) or Icon should supply an alt or accessibilityLabel, respectively, if the Image or Icon supplies extra context or information.`}
+        />
+        <MainSection.Subsection
```
rlingineni, line 155: Localization is typically it's own larger Main Section

rlingineni (approved):
Looks good, but see comments. Some of the Docs organization likely needs to be fixed

docs/pages/web/banneroverlay.js
eniomoura: Done!

docs/pages/web/banneroverlay.js
eniomoura: Yeah, I set it as 1 column because the text will wrap differently from actual dialog usage if cramped in columns. Similar to SlimBanner examples.

packages/gestalt/src/Shared/ToastSubcomponents.js
eniomoura: Done!

docs/pages/web/banneroverlay.js
eniomoura: Done! I was following the doc at https://coda.io/d/UpsellOverlay-component-doc_d0osQGIFXif/UpsellOverlay_suVqw#_lubMu
Put it after best practices, same as the other ones.

docs/pages/web/banneroverlay.js
eniomoura: Done!

docs/pages/visual-test/BannerOverlay-dark.js
eniomoura: Done!

docs/pages/visual-test/BannerOverlay.js
eniomoura: Done!

docs/pages/web/banneroverlay.js
```diff
@@ -0,0 +1,226 @@
...
+
+<b>Please note</b>: On desktop devices, the BannerOverlay should appear at the top of the screen (below navigation). The BannerOverlay shouldn't block navigation.
+On mobile devices, the BannerOverlay should appear at the bottom of the screen. The 'offset' prop can be used to adjust the position of the BannerOverlay.`}
+          />
+          <MainSection.Card
+            type="don't"
+            sandpackExample={
+              <SandpackExample code={dontStack} name="Don't - Stack" hideEditor hideControls />
+            }
+            description="Stack multiple UpsellOverlays; only one UpsellOverlay should appear on the screen per time."
```
rlingineni, line 110: @eniomoura - looks like the naming here is wrong

docs/pages/web/banneroverlay.js
```diff
@@ -0,0 +1,226 @@
...
+            type="do"
+            sandpackExample={<SandpackExample code={doConcise} name="Do - Concise" hideEditor />}
+            description="Be concise when writing the content. The BannerOverlay is intended to display short messages. Ideally max of 3 lines.
+Please consider localization. "
+          />
+          <MainSection.Card
+            type="don't"
+            sandpackExample={
+              <SandpackExample
+                code={dontLong}
```
rlingineni, line 89: Hey in this example:

image

Is there a way to not mess-up the button with a min width on that area or something? 

Let's say in the bad case there's some poor localization and someone does do this. This can create a breaking experience. Maybe we should grow the container size, or wrap the text into new lines. wdyt?

rlingineni: Also, I get it's a don't example lol, so we don't have to do this

docs/pages/web/banneroverlay.js
eniomoura: Fixed!

docs/pages/web/banneroverlay.js
```diff
@@ -0,0 +1,223 @@
...
+            description={`You can optionally have the Call to Action dismiss the BannerOverlay by not setting the dismissButton prop.
+
+<b>Please note:</b> Make sure to either set the CTA to dismiss the BannerOverlay or set the dismissButton prop. Otherwise, the BannerOverlay will be stuck on the screen.`}
+          />
+        </MainSection.Subsection>
+      </MainSection>
+      <MainSection name="Responsive">
+        <MainSection.Subsection>
+          <MainSection.Card
+            description={`UpsellOverlay is responsive to different devices. Therefore, OverlayUpsell dweb is set to a max width of 900 px to preserve a great usability experience and consistency with other components.
```
rlingineni, line 169: One more instance, calling it UpsellOverlay

eniomoura: Fixed!

docs/pages/web/banneroverlay.js
eniomoura: Added a nowrap style to the CTA, to prevent this from happening :)

PR #3335, opened 2023-12-17
Docs: Cosmetic fixes

hectoid:
### Summary
Fixed background colors, rounding and some descriptions

#### What changed?

Bumped up the background of component illustrations because it's hard to see for people (like me) with degrading vision and doesn't match the saturation/brightness of our other backgrounds.

Before:
image

After:
image

Increased spacing between component cards so that grouping is better.
Before:
image

After:
image

Matched illustration card rounding to the rest of the images, graphics and examples on the site:
image

Updated some descriptions to make them shorter and fix the use of hyphens:
image

Moved SheetMobile under Overlays, since it's not really a code building block.
image

#### Why?

General changes to improve consistency and visual quality. See above for explanations

### Checklist


.vscode/settings.json
```diff
@@ -1,8 +1,8 @@
 {
   "css.lint.validProperties": ["composes"],
   "editor.codeActionsOnSave": {
-    "source.fixAll.eslint": true,
-    "source.fixAll.shellcheck": true
+    "source.fixAll.eslint": "explicit",
+    "source.fixAll.shellcheck": "explicit"
```
rlingineni, line 5: I'm not sure if this was intended, mind switching it back to `true`

rlingineni (approved):
looks good!

Also related, but not necessarily in scope here, it seems the segmented control on the overview page wraps on mobile.

image

Would you be able to update the minWidth in this [line](https://github.com/AlbertCarreras/gestalt/blob/66d6be180a4dc3695b7026b6dd3e945ab4549d0a/docs/docs-components/Overview.js#L75) to be something like 250?

.vscode/settings.json
hectoid: thanks! that wasn't intended. not sure how it happened. yes, will switch back. good catch on the segmented control. wonder if changing to the "small" variant will fix.

hectoid: @rlingineni i think I fixed it

rlingineni: Hey, these are still showing up as "explicit"

I think you can run a command like this to reset a file back to how it is in master
```
git checkout master -- .vscode/settings.json
```

PR #3340, opened 2024-01-02
List: Add Sizing to List

rlingineni:
Add multiple sizes for list sizing.

### Summary
We support the 'Text' Element in List, but not different font sizes. This adds more font sizing support to List to align with `Text` and passes it to the rest of the children.

Before:
image

AFter:
image

#### What changed?

Adding support for a `size` prop to for varying and uniform font-sizes

#### Why?
List is used alongside paragraphs and other sets of text. Since it is primarily a text element, aligning with a text sizing here seems good.

### Links

- [TDD](link to Paper doc)
- [Figma](link to Figma file)

### Checklist

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,6 +1,7 @@
 {
   "css.lint.validProperties": ["composes"],
   "editor.codeActionsOnSave": {
+    "source.organizeImports": false,
     "source.fixAll.eslint": true,
     "source.fixAll.shellcheck": true
   },
--- a/docs/examples/list/fontSize.js
+++ b/docs/examples/list/fontSize.js
@@ -1,0 +1,37 @@
+// @flow strict
+import { type Node as ReactNode, useState } from 'react';
+import { Box, List, SegmentedControl, Text } from 'gestalt';
+
+export default function Example(): ReactNode {
+  const [selectedItemIndex, setSelectedItemIndex] = useState(0);
+  const items = ['100', '200', '300', '400', '500', '600'];
+  return (
+    <Box
+      padding={8}
+      height="100%"
+      display="flex"
+      alignItems="center"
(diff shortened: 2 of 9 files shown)


.vscode/settings.json
```diff
@@ -1,6 +1,7 @@
 {
   "css.lint.validProperties": ["composes"],
   "editor.codeActionsOnSave": {
+    "source.organizeImports": false,
```
diyorbek, line 4: what's the motivation for this change?

rlingineni: VSCode's sorting and es-lint autofix was was breaking my changes. Every time on auto-save, vscode would override the eslint auto-fix to sort imports.

So I added this to disable the vscode import organization.

diyorbek: So is this a temporary change then? Without this rule we won't have sorted imports for the future changes. Also, I think i didn't understand how exactly import sorting is breaking your changes. Could you elaborate more?

rlingineni: Nope, this is a permanent fix. Sorry, let me clarify.

We have an eslint rule called [sort-imports](https://eslint.org/docs/latest/rules/sort-imports). On save, vscode automatically runs `source.fixAll.eslint` so it sorts the order, and fixes problem.

VSCode, also ships a feature with `source.organizeImports`. When I save my file after auto-sorting my imports. VSCode overrides that before saving, and uses its own formatting.

Correct:
image

Becomes this after save:
image

If I don't make this change here, I can't ship my code, because my editor is broken otherwise.

Hopefully that clarifies

diyorbek: Oh, now I see it. Thanks for clarifying 👍

diyorbek (approved):

PR #3343, opened 2024-01-08
Internal: fix Flow suppressions in docs PropTable

dangerismycat:
It's been a while. Just verifying that I remember our processes 😅


diyorbek (approved):
Looks good 👍

PR #3347, opened 2024-01-09
Internal: fixes to .ts files

AlbertCarreras:
Internal: fixes to .ts files

Added "scissors" icon + formatting


packages/gestalt/dist/index.d.ts
```diff
@@ -253,6 +253,7 @@ type Icons =
   | 'save-outline'
   | 'saved'
   | 'scale'
+  | 'scissors'
```
AlbertCarreras, line 256: main change

PR #3349, opened 2024-01-10
BannerOverlay: Fixing responsiveness and margin for small desktop viewports

eniomoura:
### Summary

This is a follow-up to BannerOverlay component creation, fixing issues with small screens on desktop not being responsive as intended.

#### What changed?

On small desktop screens, the banner is going to mantain margins against the edges of the screen even if it becomes smaller than 900px, and on small breakpoints, the mobile variant is going to be used for desktop.

Screenshot by Dropbox Capture

#### Why?

To prevent issues with rendering the CTAs on small desktop viewports.


eniomoura:
> with `import InternalDismissButton from '../shared/InternalDismissButton';`?

Done!

AlbertCarreras (approved):

PR #3365, opened 2024-01-18
ESLint plugin: fix adding responsive props missing

AlbertCarreras:
Eslint plugin: fix adding responsive props missing


dangerismycat (approved):
good catch!

PR #3381, opened 2024-01-24
AvatarGroup: Fix rerenders on hover

diyorbek:
### Summary

AvatarGroup had a serious performance bug that it rerendered Avatars on hover. I was causing because the items were regenerated everytime when state changed.
Even though the Avatars had the same key, AvatarStackGroup component was being created remounted each time when its props have changed. Becuase it was wrapped with useCallback!

Before:
ezgif-3-ac71b8ab60

After:
ezgif-3-0a8bd338a8

### Links

- [TDD](link to Paper doc)
- [Figma](link to Figma file)

### Checklist

- [-] Added unit and Flow Tests
- [-] Added documentation + accessibility tests
- [-] Checked stakeholder feedback (e.g. Gestalt designers, relevant feature teams)


packages/gestalt/src/AvatarGroup.js
```diff
@@ -131,58 +131,48 @@ const AvatarGroupWithForwardRef: AbstractComponent<Props, UnionRefs> = forwardRe
...
-        <AddCollaboratorsButton
-          hovered={hovered}
-          key={`collaboratorStack-addButton-${collaborators.length}`}
-          pileCount={pileCount}
-          size={size}
-        />,
-      ];
-    }
-
-    const AvatarGroupStack = useCallback(
```
diyorbek, line 175: It's so curious that this was actually causing the bug, not the array operations above!

dangerismycat: Did you figure out which of `accessibilityLabel, collaboratorStack, isDisplayOnly, isFitSize` was changing?

diyorbek: It was `collaboratorStack` changing on every render. It's kinda weird that even when I wrapped it with useMemo, the bug still persisted.

dangerismycat: Did you wrap it in `useMemo` before or after you refactored to define it using a `const` and fragments?

diyorbek: I tried useMemo before refactoring. I think the root of the problem is the `.slice` method that is used to create `displayedCollaborators`. Because it creates a new array instance. The dependecy list of the hooks are compared by reference.
But how does it work without any hooks? The idea is to convert the array data into components. React does quite a good job of "diff"ing component trees (than the list of dependencies) and prevents unecessary renders.

dangerismycat: Ah, yep. That would be expected then. We were creating new array references all over the place before.

And correct, while the new logic does create a new variable full of components on each render, React is smart enough to not update the DOM if nothing in those components is changing.

Nice fix!

dangerismycat (approved):

PR #3386, opened 2024-01-26
Internal: add displayName to all top-level components

dangerismycat:
As part of tracking adoption for Visual Refresh, we need a way to identify Gestalt components in prod code. This is possible if the component has an explicitly-set `displayName` property. This PRs adds `displayName`s to all top-level components. We can address sub-components in a future PR if tracking their usage is necessary.


rlingineni (approved):
Nice!

PR #3388, opened 2024-01-26
Internal: add test for Masonry scroll performance

manvoid:
### Summary

This pr creates a new test for masonry performance on scroll. This test only calculates avg fps at the moment.

#### Why?

We plan to make a lot of changes and improvement on the Masonry project this year, we want to be sure that the performance stays in a healthy range moving forward.

#### How does it work?

The script uses the `/integration-test/masonry` test page. We connect to the chrome devtools protocol (CDP) to gather frame data from the Tracing events. The page scrolls x number of times waiting on each scroll for certain number of elements to be rendered, on each scroll we set a tag to the tracing events. After all the scroll we gather the data from the tracing data and calculate the avg fps for all the frames between each scroll.

Since we don't have a solid 'base' number to compare the fps, the test creates a file with the collected data that we can connect to other process inside github actions so we can gather more data and analyze those numbers.

#### Limitations

* The number of pins that are rendered on each scroll changes with each scroll because the heights increase, so in some point the script freezes waiting for more elements that the available on the page with a high scroll count (more than 15)
* Sometimes we have very high fps numbers (higher that 100 and up to 1000 fps)
* We still don't have a good "base" metric, that has to be defined
* We need some time at the end of the test so the Tracing events completes, currently I added a wait but this could be improved

#### Next steps

* Get the best default values and base numbers for healthy ranges for performance
* Integrate with ci to show developers the impact of the pr on the performance


performance-tests/index.mjs
```diff
@@ -0,0 +1,141 @@
...
+const PINS_COUNT = 5;
+const PAGE_HEIGHT = 1000;
+
+/**
+ * Args:
+ * --scroll-count: number
+ *  Number of times to scroll the page and fetch items
+ * --debug: flag
+ *  Open browser with dev tools, more verbose logging, and don't close browser when done
+ * --verbosity: 'silent' | 'default' | 'verbose'
```
liuyenwei, line 18: i think this is fine as a utility for testing, but let's split some of this code out so we can reuse it to write an actual integration test.

also, we should likely move this folder under `playwright` instead of creating a new top level dir

performance-tests/index.mjs
```diff
@@ -0,0 +1,141 @@
...
+      ? console.log(message)
+      : undefined;
+
+  // Launch browser and page
+  const browser = await playwright.chromium.launch(launchOptions);
+  const page = await browser.newPage();
+
+  const client = await page.context().newCDPSession(page);
+  await client.send('Overlay.setShowFPSCounter', {
+    show: true,
```
liuyenwei, line 62: this is an interesting approach. this certainly works and gets us the average FPS (although in testing locally, i was running into issues with the OCR actually getting data from the screenshot)

there are a few downsides with this approach imo:

- we are limited to average FPS (vs being able to get data on longest frame, shortest frame, etc)
- this will always be the average FPS of the entire session
- we have to do this weird dance of capturing a screenshot, saving it, reading it using via ocr library

Alternatively, we could hook into Chrome's tracing information (i.e. the data you get when you do a performance trace) since this includes information about frames being drawn as well as the timestamps for when the event happens.

Was playing around with the CDP protocol and we should be able to get this information directly through the [Tracing](https://chromedevtools.github.io/devtools-protocol/tot/Tracing/) API. something like
```
const session = await page.context().newCDPSession(page);
// start tracing
await session.send('Tracing.start');
//scroll scroll scroll
// listen for trace events
session.on('Tracing.dataCollected', (data) => {
  const events = data.value.filter((e) => {
    // i think the event we want here is `DrawFrame`
    return e.name === 'DrawFrame'
});
//end tracing
await session.send('Tracing.end');
```

The general idea here is that we should be able to take this data and use that to derive data such as longest frame, avg fps, etc

Let me know what you think about this approach

flow-typed/npm/playwright_vx.x.x.js
```diff
@@ -0,0 +1,45 @@
...
+ *   'playwright'
+ *
+ * Fill this stub out by replacing all the `any` types.
+ *
+ * Once filled out, we encourage you to share your work with the
+ * community by sending a pull request to:
+ * https://github.com/flowtype/flow-typed
+ */
+
+declare module 'playwright' {
```
liuyenwei, line 16: thank you for adding this

dangerismycat: +1, thank you Manuel!

package.json
```diff
@@ -69,6 +69,7 @@
     "lottie-react": "^2.3.1",
     "netlify-cli": "^7.1.0",
     "nodemon": "^2.0.12",
+    "playwright": "1.37.1",
```
dangerismycat, line 72: If we're adding this as a dependency, it may be worth revisiting how we trigger other integration tests runs (see the `playwright:{foo}` scripts below). (You may also want to add a script to specifically start a run of just the perf tests, for dev convenience when working on them)

performance-tests/index.mjs
```diff
@@ -0,0 +1,141 @@
+// @flow strict
+// import url from 'url';
```
dangerismycat, line 2: this should be removed

performance-tests/selectors.mjs
```diff
@@ -0,0 +1,23 @@
+// @flow strict
+
+const selectors = {
```
dangerismycat, line 3: Do these need to be repeated from `/playwright/masonry/utils`? Can we use those selectors instead of duplicating them here?

performance-tests/waitForRenderedItems.mjs
```diff
@@ -0,0 +1,230 @@
+// @flow strict
```
dangerismycat, line 1: ditto with this file

performance-tests/index.mjs
manvoid: I moved the file to the playwright directory and changed it to be a normal test.

flow-typed/npm/playwright_vx.x.x.js
manvoid: I removed this meantime because the "playwright" dependency is not required right now, but I could add it if it add value

package.json
manvoid: I changed the test to be a standard test instead of a script, I saw that the playwright documentation always uses `npx playwright ...`, if it makes sense to change it we can do it but maybe on another pr.

yarn.lock
```diff
@@ -13517,8 +13517,6 @@ minizlib@^1.3.3:
   version "1.3.3"
   resolved "https://registry.yarnpkg.com/minizlib/-/minizlib-1.3.3.tgz#2290de96818a34c29551c8a8d301216bd65a861d"
   integrity sha512-6ZYMOEnmVsdCeTJVE0W9ZD+pVnE8h9Hma/iOwwRDsdQoePpoX56/8B6z3P9VNwppJuBKNRuFDRNRqRWexT9G9Q==
-  dependencies:
```
dangerismycat, line 13520: This change seems unexpected?

playwright/masonry/scroll-performance.spec.mjs
```diff
@@ -0,0 +1,122 @@
...
+        targetItemsGTE: PINS_COUNT * (i + 1),
+      });
+
+      performanceTimings.push(performance.now());
+    }
+
+    await client.send('Tracing.end');
+
+    // We need this to wait for the tracing data to be completed
+    await page.waitForTimeout(2000);
```
dangerismycat, line 117: I believe you should be able to add a suppression comment here to allow waitForTimeout

playwright/masonry/scroll-performance.spec.mjs
```diff
@@ -0,0 +1,122 @@
...
+        'playwright/test-results/scroll-performance-timings.csv',
+        eventsToCsv(fpsEvents)
+      );
+    });
+
+    await page.setViewportSize({ width: 1600, height: PAGE_HEIGHT });
+    await page.goto(getServerURL());
+
+    for (let i = 1; i < scrollCount; i += 1) {
+      // await page.waitForTimeout(300);
```
dangerismycat, line 95: nit: can we remove this commented-out line?

playwright/masonry/scroll-performance.spec.mjs
```diff
@@ -0,0 +1,122 @@
...
+      await client.send('Tracing.recordClockSyncMarker', {
+        syncId: 'scrollTimingMark',
+      });
+
+      await page.evaluate(
+        // $FlowFixMe[incompatible-use]
```
dangerismycat, line 101: Can you add a note about why this suppression is needed?

playwright/masonry/scroll-performance.spec.mjs
```diff
@@ -0,0 +1,122 @@
+// @flow strict
```
dangerismycat, line 1: Can you add comments throughout explaining some of the more cryptic logic?

dangerismycat (approved):

playwright/masonry/scroll-performance.spec.mjs
```diff
@@ -0,0 +1,122 @@
...
+  test('scroll test', async ({ page }) => {
+    const performanceTimings = [performance.now()];
+    const events /*: Array<{ name: string, ts: number }> */ = [];
+    let fpsEvents = [];
+    let fpsAvgPerScroll = [];
+
+    const scrollCount = DEFAULT_SCROLL_COUNT;
+
+    const client = await page.context().newCDPSession(page);
+    await client.send('Tracing.start');
```
liuyenwei, line 77: based on https://chromedevtools.github.io/devtools-protocol/tot/Tracing/#method-start we should be able to configure https://chromedevtools.github.io/devtools-protocol/tot/Tracing/#type-TraceConfig and specify `includedCategories` so we don't get _all_ tracing data. wonder if that might be a good thing to do (not blocking)

liuyenwei (changes requested):
requesting changes since the yarn.lock change seems suspicious and the waitForTimeout should be changed

playwright/masonry/scroll-performance.spec.mjs
```diff
@@ -0,0 +1,122 @@
...
+};
+
+const getFpsAvgPerScroll = (events) => {
+  const fpsAvgPerScroll = [];
+  let fpsSum = 0;
+  let count = 0;
+
+  for (let i = 0; i < events.length; i += 1) {
+    if (events[i].name.includes('clock_sync')) {
+      fpsAvgPerScroll.push(fpsSum / count);
```
liuyenwei, line 45: ohh clever. so instead of starting/stopping the tracing, we're just marking the scroll events and using that to indicate when we should reset our fps calculations?

playwright/masonry/scroll-performance.spec.mjs
manvoid: I added the filter, still we need to filter by name but having the category filtered is better

playwright/masonry/scroll-performance.spec.mjs
manvoid: Yes, I tried stopping/starting the tracing but I would have to wait each time for all the events to finish and this can make the test behave in strange ways, I find this approach better

playwright/masonry/scroll-performance.spec.mjs
```diff
@@ -0,0 +1,122 @@
...
+    const scrollCount = DEFAULT_SCROLL_COUNT;
+
+    const client = await page.context().newCDPSession(page);
+    await client.send('Tracing.start');
+
+    client.on('Tracing.dataCollected', createDataCollectedHandler(events));
+
+    client.on('Tracing.tracingComplete', () => {
+      fpsEvents = getFpsFromEvents(events);
+      fpsAvgPerScroll = getFpsAvgPerScroll(fpsEvents);
```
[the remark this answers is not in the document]
manvoid: I added longest and shortest

playwright/masonry/scroll-performance.spec.mjs
manvoid: It is not needed, thanks for flagging

playwright/masonry/scroll-performance.spec.mjs
```diff
@@ -0,0 +1,96 @@
...
+      );
+    });
+
+    const tracingCompleteEvent = new Promise((resolve) => {
+      client.on('Tracing.tracingComplete', () => {
+        resolve();
+      });
+    });
+
+    await page.setViewportSize({ width: 1600, height: PAGE_HEIGHT });
```
liuyenwei, line 54: this should go before we start the tracing session. we want to open the page first and then start the trace, otherwise the trace will just be the initial page load

playwright/masonry/scroll-performance.spec.mjs
```diff
@@ -0,0 +1,96 @@
...
+    await page.goto(getServerURL());
```
liuyenwei, line 55: not sure if it makes a difference but should we also add a `waitForRenderedItems` here to wait for the initial render.

also, just noticed that based on where this is currently placed we likely are also capturing `DrawFrame` events from the initial render which could be skewing the actual FPS calculations

liuyenwei (approved):

PR #3391, opened 2024-01-30
Masonry: Cleanup code around batch paints

liuyenwei:
#### What changed?

This PR cleans up the experimental `_batchPaints` prop that was first introduced in https://github.com/pinterest/gestalt/pull/2853 and removes all associated code. 

#### Why?
This was originally added as a way to separately validate the impact of batching Masonry's item paints. The relevant experiment code has since been deleted in Pinboard so we no longer need this prop.


packages/gestalt/src/Masonry/MeasureItems.js
```diff
@@ -1,81 +0,0 @@
...
-  items: $ReadOnlyArray<T>,
-  measurementStore: Cache<T, number>,
-  renderItem: ({
-    +data: T,
-    +itemIdx: number,
-    +isMeasuring: boolean,
-  }) => ReactNode,
-};
-
-export default function MeasureItems<T>({
```
dangerismycat, line 20: Isn't this the version of item measuring we want to keep using?

liuyenwei: can you elaborate? this code is only being hit if `__batchPaints` is passed in

dangerismycat: Right — isn't batching paints the behavior we want to keep?

liuyenwei: i don't think so (correct me if i'm wrong). `__batchPaints` will mean that all measurements are always grouped together vs just batching the paints for the pages that contain multi column items.

Currently, we have it configured to increase the batch size if the current list of items to measure contains a two column item: https://github.com/pinterest/gestalt/blob/master/packages/gestalt/src/Masonry.js#L597. i suspect we'll need to tune this as we increase support to N columns but we likely don't need to always batch paints

dangerismycat: Ah yeah I think you are correct. A bummer to lose a little bit of code modularization with the MeasureItems component but not a big deal — we'll do a lot more of that cleanup in the future.

dangerismycat (approved):

PR #3394, opened 2024-01-31
Tokens: change color-text-subtle &  color-text-icon-subtle  value to match Figma + Token changelog

AlbertCarreras:
Changelog: [packages/gestalt-design-tokens/CHANGELOG](https://github.com/pinterest/gestalt/blob/master/packages/gestalt-design-tokens/CHANGELOG.md)

### Major

| name                   | old value                     | new value                     |
| ---------------------- | ----------------------------- | ----------------------------- |
| color-text-subtle      | color.gray.roboflow.550.value | color.gray.roboflow.500.value |
| color-text-icon-subtle | color.gray.roboflow.550.value | color.gray.roboflow.500.value |

Before 
Screenshot by Dropbox Capture

After
Screenshot by Dropbox Capture

#### Why

Reconcile with Figma values

#### Breaking change notes

It might break visual snapshots

### DIFF
--- a/packages/gestalt-design-tokens/CHANGELOG.md
+++ b/packages/gestalt-design-tokens/CHANGELOG.md
@@ -1,0 +1,24 @@
+This file is updated manually
+
+## 134.0.0 https://github.com/pinterest/gestalt/pull/3394
+
+### Major
+
+| name                   | old value                     | new value                     |
+| ---------------------- | ----------------------------- | ----------------------------- |
+| color-text-subtle      | color.gray.roboflow.550.value | color.gray.roboflow.500.value |
+| color-text-icon-subtle | color.gray.roboflow.550.value | color.gray.roboflow.500.value |
+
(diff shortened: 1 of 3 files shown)


packages/gestalt-design-tokens/CHANGELOG.md
```diff
@@ -0,0 +1,24 @@
+This file is updated manually
```
dangerismycat, line 1: We should look into updating this file automatically with the PR summary whenever changes are detected within `/packages/gestalt-design-tokens`

dangerismycat (approved):

packages/gestalt-design-tokens/CHANGELOG.md
AlbertCarreras: Yes... I think we should create a tokens package with separate release and create the changelog from there.

PR #3411, opened 2024-02-07
Docs, TagData, TileData: Docs, TagData, Tile to consume directly from tokens rather than ColorSchemeProvider and legacy colors

AlbertCarreras:
Internal: deprecated custom colors in Docs consumed from ColorSchemeProvider 

TileData BEFORE & AFTER
Screenshot by Dropbox Capture


dangerismycat (approved):

packages/gestalt/src/utils/datavizcolors/getDataVisualizationColor.js
```diff
@@ -1,13 +1,17 @@
 // @flow strict
-import { type Theme } from '../../contexts/ColorSchemeProvider';
+import darkColorDesignTokens from 'gestalt-design-tokens/dist/json/variables-dark.json';
```
AlbertCarreras, line 2: @rlingineni Here's the most significant change

dangerismycat (approved):

PR #3413, opened 2024-02-07
Tokens, Modal, OverlayPanel, SheetMobile, Popover: consume from overlay component tokens

AlbertCarreras:
Tokens, Modal, OverlayPanel, SheetMobile, Popover: consume from overlay component tokens

### Minor

NEW

| name                     | light                          | dark                                          |
| ------------------------ | ------------------------------ | --------------------------------------------- |
| overlay-color-background | color.background.default.value | color.background.elevation.floating.darkValue |

REVERT

| name                                      | old value                                    | new value                                        |
| ----------------------------------------- | -------------------------------------------- | ------------------------------------------------ |
| color-background-wash-dark                | rgb(0 0 0 / 0.8)                             | rgba(0, 0, 0, 0.8)                               |
| color-background-wash-light               | rgb(255 255 255 / 0.9)                       | rgba(255, 255, 255, 0.9)                         |

#### Why

Tokenize overlay component background colors. Overlay components require a white background but a floating elevation on darkmode. We needed a custom color pairing.

#### Breaking change notes

Reverting token changes in from https://github.com/pinterest/gestalt/pull/3407 This PR introduces a bug as reported by Android where alpha values `color_background_wash_dark` and `color_background_wash_light` and now that alpha value is gone, leaving the color just black and white without opacity value assigned.


packages/gestalt/src/Colors.css
```diff
@@ -316,6 +317,10 @@
   color: var(--g-colorRed100);
 }
 
+.redBg {
```
AlbertCarreras, line 320: putting this back as it broke red popovers

eniomoura (approved):

PR #3417, opened 2024-02-09
Masonry: Fix integration tests and run in CI

liuyenwei:
#### What changed?

This PR fixes the suite of Masonry integration tests and also updates the `playwright:test` command so they run in CI.

#### Why?
These integration tests were added to Gestalt last year however, 
- they don't all fully run
- they don't actually run in CI.

In preparation for a number of changes we'll be making to Masonry this half, it is vital that we have regression testing in CI to catch potential issues not found through manual testing.

#### Next Steps
- Validate that the tests are not flakey
- Write more tests (specifically around existing two column functionality)


playwright/masonry/fetch-more.spec.mjs
```diff
@@ -28,8 +28,8 @@ test.describe('Masonry: fetch more', () => {
   });
 
   // TODO[@rjames]: fix this test, currently timing out
```
dangerismycat, line 30: I assume we can remove this TODO now?

playwright/masonry/utils/getServerURL.mjs
```diff
@@ -1,6 +1,6 @@
 // @flow strict
 
-const BASE_DOMAIN = 'http://localhost:8888';
```
dangerismycat, line 3: Ah interesting, this still runs on a local build? Should it run on the Netlify preview deploy instead? That would delay the tests running a bit. Don't know if it matters either way really

dangerismycat (approved):

playwright/masonry/utils/getServerURL.mjs
liuyenwei: the playwright tests are configured to run against localhost:3000 currently: https://github.com/pinterest/gestalt/blob/master/playwright/playwright.config.js#L31
 
actually, we can just remove `BASE_DOMAIN` here altogether and just use a relative URL since that will resolve to whatever is configured in the playwright config.

dangerismycat: ah cool, even better!

PR #3425, opened 2024-02-13
RadioGroup: Add Badge prop

eniomoura:
#### What changed?

RadioGroup now has a prop to add a Badge next to the label.

#### Why?

To satisfy the use case where users need both a Badge and a helperText with a RadioGroup.RadioButton option, as described in the component request: https://jira.pinadmin.com/browse/GESTALT-7505

### Links

- [Jira](https://jira.pinadmin.com/browse/GESTALT-7505)

### DIFF
--- a/docs/pages/web/radiogroup.js
+++ b/docs/pages/web/radiogroup.js
@@ -8,7 +8,7 @@ import MainSection from '../../docs-components/MainSection';
 import Page from '../../docs-components/Page';
 import PageHeader from '../../docs-components/PageHeader';
 import SandpackExample from '../../docs-components/SandpackExample';
-import addingABadgeExample from '../../examples/radiogroup/addingABadgeExample';
+import badge from '../../examples/radiogroup/addingABadgeExample';
 import addingAPopoverExample from '../../examples/radiogroup/addingAPopoverExample';
 import directionExample from '../../examples/radiogroup/directionExample';
 import dontUseToSelectMultipleItems from '../../examples/radiogroup/dontUseToSelectMultipleItems';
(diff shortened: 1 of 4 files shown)


packages/gestalt/src/RadioGroupButton.js
```diff
@@ -1,6 +1,7 @@
 // @flow strict
 import { type AbstractComponent, forwardRef, type Node as ReactNode, useState } from 'react';
 import classnames from 'classnames';
+import Badge from './Badge';
```
AlbertCarreras, line 4: Can we add an example in the Docs so we can see how it looks like?

docs/pages/web/radiogroup.js
```diff
@@ -288,6 +289,18 @@ export default function DocsPage({
...
         </MainSection.Subsection>
+        <MainSection.Subsection
+          title="Adding a Badge"
+          description={`
+    The \`badge\` prop can be used to add a badge to the RadioButton.
+  `}
+        >
+          <MainSection.Card
+            sandpackExample={
+              <SandpackExample name="Adding a popover example" code={addingABadgeExample} />
```
AlbertCarreras, line 300: Can we rename this to just code={badge}.   ??

AlbertCarreras: name="With Badge"

docs/pages/web/radiogroup.js
```diff
@@ -288,6 +289,18 @@ export default function DocsPage({
             }
           />
         </MainSection.Subsection>
+        <MainSection.Subsection
+          title="Adding a Badge"
```
AlbertCarreras, line 293: title="With Badge"

packages/gestalt/src/RadioGroupButton.js
eniomoura: Done

docs/pages/web/radiogroup.js
eniomoura: Done

AlbertCarreras (approved):
LGTM

PR #3430, opened 2024-02-14
Internal: Update tests for v2 of Popover and Popover-based components

diyorbek:
### Summary

Popover and Popover-based component tests now test v2. This changes were made separately from removing v2 experiment code to reduce PR size.

#### What changed?

Added test utils for testing component with experiments. They will be removed once Popover is fully released.

### Checklist


packages/gestalt/src/utils/testing/renderWithExperiment.js
```diff
@@ -0,0 +1,15 @@
+// Remove after Dropdown is fully released
```
diyorbek, line 1:
```suggestion
```

packages/gestalt/src/__snapshots__/Popover.jsdom.test.js.snap
```diff
@@ -1,161 +1,156 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`PopoverEducational does not render when the anchor is null 1`] = `null`;
+exports[`PopoverEducational does not render when the anchor is null 1`] = `<div />`;
```
diyorbek, line 3: There is an empty div because the test takes a snapshot of the container.

PR #3440, opened 2024-02-21
Popover: Add `hideWhenReferenceHidden` prop

diyorbek:
### Summary

- Added `hideWhenReferenceHidden` prop to Popover, Tooltip, ComboBox, Dropdown, HelpButton and PopoverEducational.
- Added internal `__disableFocusTrap` prop to Popover, ComboBox, Dropdown and HelpButton.
- Fixed double portalling issue with Dropdown. [Slack discussion](https://pinterest.slack.com/archives/C13KLG5P0/p1708465918425379).

#### What changed?

The default value of `hideWhenReferenceHidden` (true) is now passed in at the component props level. Previously it was passed at `Contents` component level. This change makes it more obvious how the components behave. The default value now also displayed in the docs.

#### Why?

`__disableFocusTrap` prop was added mainly to control the components during unit tests. By default Popover traps the focus and makes other elements inaccessible. `testing-library/react` queries do not work with inaccessible elements.

`hideWhenReferenceHidden` prop also helps to control test cases as `testing-library/react` doesn't provide actual DOM behaviors for autohide feature to work properly.

### Links

- [Jira](https://jira.pinadmin.com/browse/GESTALT-7570)

### Checklist

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -10,5 +10,6 @@
   "grunt.autoDetect": "off",
   "javascript.validate.enable": false,
   "npm.packageManager": "yarn",
-  "typescript.validate.enable": true
+  "typescript.validate.enable": true,
+  "testing.openTesting": "neverOpen"
 }
--- a/packages/gestalt/src/ComboBox.js
+++ b/packages/gestalt/src/ComboBox.js
@@ -151,6 +151,12 @@ type Props = {
    * An object representing the zIndex value of the ComboBox list box. Learn more about [zIndex classes](https://gestalt.pinterest.systems/web/zindex_classes)
    */
   zIndex?: Indexable,
+  /**
+   * *EXPERIMENTAL:* Whether to hide ComboBox when reference element gets out of viewport.
+   */
+  hideWhenReferenceHidden?: boolean,
+  // Whether to trap focus inside ComboBox when opened.
+  __disableFocusTrap?: boolean,
 };
 
 /**
@@ -188,6 +194,8 @@ const ComboBoxWithForwardRef: AbstractComponent<Props, HTMLInputElement> = forwa
     selectedOption,
     tags,
     zIndex,
+    hideWhenReferenceHidden = true,
+    __disableFocusTrap,
   }: Props,
   ref,
 ): ReactNode {
@@ -507,6 +515,8 @@ const ComboBoxWithForwardRef: AbstractComponent<Props, HTMLInputElement> = forwa
             shouldFocus={false}
             role="dialog"
             color="white"
+            hideWhenReferenceHidden={hideWhenReferenceHidden}
+            disableFocusTrap={__disableFocusTrap}
           >
             <Box
               aria-expanded={showOptionsList}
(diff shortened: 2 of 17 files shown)


packages/gestalt/src/ComboBox.js
```diff
@@ -188,6 +194,8 @@ const ComboBoxWithForwardRef: AbstractComponent<Props, HTMLInputElement> = forwa
     selectedOption,
     tags,
     zIndex,
+    hideWhenReferenceHidden = true,
+    __disableFocusTrap,
```
AlbertCarreras, line 198: can we set it to be boolean?

packages/gestalt/src/ComboBox.js
```diff
@@ -151,6 +151,12 @@ type Props = {
    * An object representing the zIndex value of the ComboBox list box. Learn more about [zIndex classes](https://gestalt.pinterest.systems/web/zindex_classes)
    */
   zIndex?: Indexable,
+  /**
+   * *EXPERIMENTAL:* Whether to hide ComboBox when reference element gets out of viewport.
+   */
+  hideWhenReferenceHidden?: boolean,
+  // Whether to trap focus inside ComboBox when opened.
```
dangerismycat, line 158: If this is intended for use only in unit tests, we should note that here

packages/gestalt/src/ComboBox.js
```diff
@@ -151,6 +151,12 @@ type Props = {
    * An object representing the zIndex value of the ComboBox list box. Learn more about [zIndex classes](https://gestalt.pinterest.systems/web/zindex_classes)
    */
   zIndex?: Indexable,
+  /**
+   * *EXPERIMENTAL:* Whether to hide ComboBox when reference element gets out of viewport.
```
dangerismycat, line 155: If this prop is experimental, we should add a prefix indicating that. Something like `EXPERIMENTAL_hideWhenReferenceHidden` or similar.

Or we could also make this a hidden prop for now (`_hideWhenReferenceHidden`) and remove the underscore once we deem it no longer experimental.

packages/gestalt/src/ComboBox.js
```diff
@@ -151,6 +151,12 @@ type Props = {
...
+   */
+  hideWhenReferenceHidden?: boolean,
+  // Whether to trap focus inside ComboBox when opened.
+  __disableFocusTrap?: boolean,
```
dangerismycat, line 159: nit: this only needs a single underscore prefix to be excluded from the published props table

packages/gestalt/src/ComboBox.jsdom.test.js
```diff
@@ -114,6 +114,8 @@ describe('ComboBox', () => {
         selectedOption={selectedOption}
         size={size}
         tags={tags}
+        hideWhenReferenceHidden={false} // Disabling as Popover uses IntersectionObserver which is not supported by testing-library
+        __disableFocusTrap // Disabling for testing purposes as focus trap makes other elements inaccessible
```
dangerismycat, line 118: This is the comment that should be where the prop is defined. It can be included here as well if you like, but the prop definition is more central

.vscode/settings.json
```diff
@@ -10,5 +10,6 @@
   "grunt.autoDetect": "off",
   "javascript.validate.enable": false,
   "npm.packageManager": "yarn",
-  "typescript.validate.enable": true
+  "typescript.validate.enable": true,
+  "testing.openTesting": "neverOpen"
```
dangerismycat, line 14: Does this need to be saved as a universal setting, or is this more of a personal preference?

PR #3443, opened 2024-02-22
ComboBox, Dropdown, HelpButton: Disable internal Popover portal

diyorbek:
### Summary

Popover v2 has built-in portalling feature. But it doesn't support z-index yet.
ComboBox, Dropdown and HelpButton control portalling themselves with Layer which supports z-index.

Basically here we were having 2 nested React.Portals. (Floating UI) Popover's internal portal is conflicting with Gestalt Layer and rendering itself outside of Layer (which leaves Layer empty and useless). That's why Layer's zIndex prop has no effect on the rendered Popover, as it would be in a different Portal.

This confusion will be eliminated as soon as the legacy Popover code is removed. The purpose of introducing internal Portal for Popover was to deprecate using Layer + Popover combination for internal usages.

#### Why?

Bug was reported in [Slack](https://pinterest.slack.com/archives/C13KLG5P0/p1708465918425379)

### Links

- [Popover TDD](https://docs.google.com/document/d/1GPzNVK9X-A06PgOh5Ha3ncRFb27Cxl0RRBckauFRaZ8/edit?usp=sharing)

### Checklist

### DIFF
--- a/packages/gestalt/src/ComboBox.js
+++ b/packages/gestalt/src/ComboBox.js
@@ -509,7 +509,7 @@ const ComboBoxWithForwardRef: AbstractComponent<Props, HTMLInputElement> = forwa
             onKeyDown={handleKeyDown}
             idealDirection="down"
             onDismiss={handleOnDismiss}
-            positionRelativeToAnchor={false}
+            disablePortal
             size="flexible"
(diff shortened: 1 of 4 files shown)


AlbertCarreras (approved):
LGTM!

PR #3444, opened 2024-02-22
Popover, Dropdown, ComboBox, Tooltip: Disable focus trap for Tooltip + fixed a11y in Dropdown, ComboBox

diyorbek:
### Summary

There was an overlooked case during Floating-UI integration. Tooltip didn't need FloatingFocusManager. Tooltip is not an accessible component.

Default values of new props `hideWhenReferenceHidden` and `disableFocusTrap` are now passed at the component props level. Previously the defaults set at `Contents` component level. This change makes it more obvious how the components behave. The default value now also displayed in the docs.

For the context, this is the table of components classified by accessiblity features:
Screenshot 2024-02-22 at 16 04 01

### Checklist

### DIFF
--- a/packages/gestalt/src/Popover/Contents.js
+++ b/packages/gestalt/src/Popover/Contents.js
@@ -143,10 +143,13 @@ export default function Contents({
     </div>
   );
 
-  return role === 'tooltip' ? (
-    contents
-  ) : (
-    <FloatingFocusManager context={context} returnFocus={false} modal={shouldTrapFocus ?? false}>
+  return (
+    <FloatingFocusManager
+      disabled={role === 'tooltip'}
+      context={context}
+      returnFocus={false}
+      modal={shouldTrapFocus ?? false}
+    >
       {contents}
     </FloatingFocusManager>
   );
--- a/packages/gestalt/src/Popover/Contents.js
+++ b/packages/gestalt/src/Popover/Contents.js
@@ -88,61 +88,6 @@ export default function Contents({
     return () => window.removeEventListener('keydown', onKeyDown);
   }, [onKeyDown]);
 
-  const contents = (
-    <div
-      ref={refs.setFloating}
-      tabIndex={-1}
-      className={classnames(
-        styles.container,
-        rounding === 2 && borders.rounding2,
-        rounding === 4 && borders.rounding4,
-        styles.contents,
-        styles.maxDimensions,
-        width !== null && styles.minDimensions,
-      )}
-      style={{ ...floatingStyles, visibility }}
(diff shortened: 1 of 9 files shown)


packages/gestalt/src/Popover/Contents.js
```diff
@@ -86,60 +88,66 @@ export default function Contents({
...
-          }}
-        >
-          {children}
-        </div>
+        {children}
       </div>
+    </div>
+  );
+
+  return role === 'tooltip' ? (
```
AlbertCarreras, line 146: Could we just set when this is a tooltip = -1
Screenshot by Dropbox Capture

diyorbek: FloatingFocusManager adds inert attribute to all the other elements which is not desirable. That's unconditional. That's why removing FloatingFocusManager is necessary for tooltip. Besides it doesn't make sense to have focus manager for a component that should not have focus.

AlbertCarreras: Screenshot by Dropbox Capture

packages/gestalt/src/Popover/Contents.js
```diff
@@ -86,60 +88,66 @@ export default function Contents({
...
+    contents
+  ) : (
+    <FloatingFocusManager context={context} returnFocus={false} modal={shouldTrapFocus ?? false}>
```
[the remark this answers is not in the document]
AlbertCarreras: https://floating-ui.com/docs/floatingfocusmanager#modal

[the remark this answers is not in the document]
AlbertCarreras: Can we create a table in description where we specify which components require this modal trapping and doublecheck combobox as focus stays in textfield.

packages/gestalt/src/Popover/Contents.js
AlbertCarreras: https://floating-ui.com/docs/floatingfocusmanager#disabled

packages/gestalt/src/Popover/Contents.js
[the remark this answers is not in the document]
diyorbek: I confirm that focus stays in the combobox textfield when popover is opened
Screenshot 2024-02-22 at 16 19 56

AlbertCarreras:
For context
https://github.com/pinterest/gestalt/assets/10593890/e2e2e16c-66c5-4c4c-8f8e-6dc9a4be8585

AlbertCarreras (approved):
Awesome

PR #2726, opened 2023-03-14
ModalAlert, Button: add dataTestId

dangerismycat:
#### What changed?

Adding `dataTestId` to the actions on ModalAlert, which necessitates also adding that prop to Button.

_(note the URL, linking to RTL docs trying to convince people not to use dataTestIds for unit tests)_
Screen Shot 2023-03-13 at 6 56 50 PM

Screen Shot 2023-03-13 at 6 57 12 PM

#### Why?

Request from a team refactoring a custom modal to use ModalAlert. Needed for integration tests, where we prefer to use stable test ids rather than labels that may change.

### Links

- [Jira](https://jira.pinadmin.com/browse/GESTALT-5886)

PR #3446, opened 2024-02-23
TextField: Add dataTestId prop

eniomoura:
#### What changed?

Added a dataTestId prop to TextField. Used #2726 as a model.

#### Why?

To allow for deprecating the Sterling TextField component, that implements their own as a Box around the TextField.

### Links

- [Jira]https://jira.pinadmin.com/browse/GESTALT-7558


AlbertCarreras (approved):
LGTM

PR #3449, opened 2024-02-26
Internal: Add initial unit test for masonry two column module

manvoid:
### Summary

This PR creates an inital unit test with playwright for the two modules feature on masonry. I'm adding two tests, one that only verifies that the module is rendered and other to insert a two column element on a specific place to test the correct positioning.

#### What changed?

- The inclusion of the tests.
- Fixed a bug on the `generateExampleItems` where the two column item was not added when the random number is 0. 
- Added new logic for the `set-masonry-items` event to specify in what index to add the new items

#### Why?

To add coverage for the two column module tests.

### DIFF
--- a/docs/integration-test-helpers/masonry/MasonryContainer.js
+++ b/docs/integration-test-helpers/masonry/MasonryContainer.js
@@ -97,9 +97,13 @@ export default class MasonryContainer extends Component<Props<{ ... }>, State> {
     });
 
     window.addEventListener('set-masonry-items', (e) => {
-      this.setState({
-        items: e.detail.items,
-      });
+      const index = e?.detail.index;
+
+      this.setState(({ items: prevItems }) => ({
+        items: index
+          ? [...prevItems.slice(0, index), ...e.detail.items, ...prevItems.slice(index)]
+          : e.detail.items,
+      }));
     });
 
     window.ERROR_COUNT = window.ERROR_COUNT || 0;
--- a/playwright/masonry/two-module.spec.mjs
+++ b/playwright/masonry/two-module.spec.mjs
@@ -1,0 +1,58 @@
+// @flow strict
+import { expect, test } from '@playwright/test';
+import getServerURL from './utils/getServerURL.mjs';
+import selectors from './utils/selectors.mjs';
+import waitForRenderedItems from './utils/waitForRenderedItems.mjs';
+
+test.describe('Masonry: two colums module', () => {
+  test('render the two column module', async ({ page }) => {
+    await page.setViewportSize({ width: 1200, height: 900 });
+    await page.goto(
+      getServerURL({
+        manualFetch: true,
+        twoColItems: true,
+      })
+    );
+
+    // The two modules items appear after the 50th pin, we add at least 60 items
+    const addItemsButton = await page.locator(selectors.addItems);
+    await addItemsButton.click();
+    await addItemsButton.click();
+    await addItemsButton.click();
+
+    await expect(page.getByText('columnSpan: 2')).toBeVisible();
+  });
+
+  test('position two module on specific place', async ({ page }) => {
+    await page.setViewportSize({ width: 1200, height: 900 });
+    await page.goto(
+      getServerURL({
+        manualFetch: true,
+        twoColItems: true,
+      })
+    );
+
+    const addItemsButton = await page.locator(selectors.addItems);
+    await addItemsButton.click();
+    await addItemsButton.click();
+
+    const masonryItemData = [
+      { name: 'fake1', height: 100, color: '#f00' },
+      { name: 'fake2', height: 200, color: '#f00' },
+      { name: 'fake3', height: 300, color: '#f00', columnSpan: 2 },
+    ];
+
+    await page.evaluate((proxiedItemData) => {
+      window.dispatchEvent(
+        new CustomEvent('set-masonry-items', {
+          detail: {
+            items: proxiedItemData,
+            index: 5,
+          },
+        })
+      );
+    }, masonryItemData);
+
+    await expect(page.getByText('columnSpan: 2')).toBeVisible();
+  });
+});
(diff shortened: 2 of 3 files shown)


packages/gestalt/src/Masonry/defaultTwoColumnModuleLayout.test.js
```diff
@@ -4,7 +4,223 @@ import HeightsStore from './HeightsStore';
 import MeasurementStore from './MeasurementStore';
 import { type Position } from './types';
 
+type Item = {
+  id: string,
+  columnSpan?: number,
+};
+
+const stubCache = (measurements?: { [item: string]: number, ... } = {}) => {
```
liuyenwei, line 12: we can just use MeasurementStore directly (see my example below)

packages/gestalt/src/Masonry/defaultTwoColumnModuleLayout.test.js
```diff
@@ -4,7 +4,223 @@ import HeightsStore from './HeightsStore';
...
+      cache[item.id] = value;
+    },
+    reset() {
+      cache = {};
+    },
+  };
+};
+
 describe('two column layout test cases', () => {
+  test('empty', () => {
```
liuyenwei, line 51: nit: let's call out that these are copied over from the defaultLayout (e.g. `describe('defaultLayout test cases')`).
The ultimate plan is to merge this back into the defaultLayout but these tests are good to ensure we don't break the default layout test cases

packages/gestalt/src/Masonry/defaultTwoColumnModuleLayout.test.js
```diff
@@ -4,7 +4,223 @@ import HeightsStore from './HeightsStore';
...
+
+    expect(justifyCenter).toEqual([
+      { top: 0, left: 300, width: 100, height: 100 },
+      { top: 0, left: 400, width: 100, height: 120 },
+      { top: 0, left: 500, width: 100, height: 80 },
+      { top: 0, left: 600, width: 100, height: 100 },
+    ]);
+  });
+
+  test.only('two column module', () => {
```
liuyenwei, line 179: should remove this

packages/gestalt/src/Masonry/defaultTwoColumnModuleLayout.test.js
```diff
@@ -4,7 +4,223 @@ import HeightsStore from './HeightsStore';
...
+      positionCache: positionsStubCache(positions),
+      justify: 'start',
+      rawItemCount: items.length,
+      width: 986,
+    });
+
+    layout(items);
+    const result = layout(items);
+
+    console.log(result);
```
liuyenwei, line 209: should remove the console logs too

docs/integration-test-helpers/masonry/MasonryContainer.js
```diff
@@ -89,28 +91,36 @@ export default class MasonryContainer extends Component<Props<{ ... }>, State> {
   componentDidMount() {
     window.TEST_FETCH_COUNTS = 0;
 
-    window.addEventListener('trigger-reflow', () => {
-      if (this.gridRef.current) {
-        this.gridRef.current.reflow();
-        this.forceUpdate();
-      }
-    });
+    if (!eventsSet) {
```
liuyenwei, line 94: instead of doing this, could we just remove the event listener on unmount?

```
componentDidMount() {
  window.addEventListener('trigger-reflow')...
}
componentWilllUnmount() {
  window.removeEventListener('trigger-reflow')...

```

still unexpected that this is happening twice though

scripts/setupJest.js
```diff
@@ -10,5 +10,5 @@ configure({ testIdAttribute: 'data-test-id' });
 
 // Fail tests on `console` calls
 failOnConsole({
-  shouldFailOnLog: true,
+  shouldFailOnLog: false,
```
liuyenwei, line 13: should clean this up too

performance-tests/eng.traineddata
liuyenwei, line 1: did you mean to add this?

playwright/masonry/two-module.spec.mjs
```diff
@@ -0,0 +1,24 @@
...
+      })
+    );
+
+    // The two modules items appear after the 50th pin, we add at least 60 items
+    const addItemsButton = await page.locator(selectors.addItems);
+    await addItemsButton.click();
+    await addItemsButton.click();
+    await addItemsButton.click();
+
+    await expect(page.getByText('columnSpan: 2')).toBeVisible();
```
liuyenwei, line 22: is there a more useful test we could write here? asserting that the two column module shows up is good but it doesn't do much to help guard against breakages in terms of the actual layout (although the unit test should help with that). 

if this is something you'd prefer to come back to after we do more investigation around optimizing the placement, that's fine too.

performance-tests/eng.traineddata
manvoid, line 1: I think this got added by error when doing the OCR tests for the fps count

playwright/masonry/two-module.spec.mjs
manvoid: Yeah, I wanted to do a more precise test but since currently we don't have a way to set a specific set of items on a second layout calculation (have some items first and then add a custom set of items with two column), the two column items are random so I can't set a specific position. I want to add that feature to the masonry performance test but meanwhile this at least test the render part.

packages/gestalt/src/Masonry/defaultTwoColumnModuleLayout.test.js
```diff
@@ -4,6 +4,212 @@ import HeightsStore from './HeightsStore';
...
+
+    expect(justifyCenter).toEqual([
+      { top: 0, left: 300, width: 100, height: 100 },
+      { top: 0, left: 400, width: 100, height: 120 },
+      { top: 0, left: 500, width: 100, height: 80 },
+      { top: 0, left: 600, width: 100, height: 100 },
+    ]);
+  });
+});
+
```
liuyenwei, line 212: there was a twoCol test case here previously i think. did you want to add it back in?

PR #3450, opened 2024-02-26
SelectList: fix for hidden arrow-down Icon

AlbertCarreras:
Fix.

BEFORE
Screenshot by Dropbox Capture

AFTER with BUG
Screenshot by Dropbox Capture

AFTER with FIX
Screenshot by Dropbox Capture

We need move the background color below the Icon position so it's visible again.

### DIFF
--- a/packages/gestalt/src/SelectList.js
+++ b/packages/gestalt/src/SelectList.js
@@ -98,9 +98,13 @@ function SelectList({
   const classes = classnames(
     styles.select,
     formElement.base,
-    disabled ? formElement.disabled : formElement.enabled,
-    errorMessage ? formElement.errored : formElement.normal,
     size === 'md' ? layout.medium : layout.large,
+    {
+      [formElement.normal]: !errorMessage,
+      [formElement.enabledTransparent]: !disabled,
+      [formElement.disabled]: disabled,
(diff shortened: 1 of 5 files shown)


rlingineni (approved):

packages/gestalt/src/SelectList.js
```diff
@@ -119,11 +123,13 @@ function SelectList({
     <Box>
       {label && <FormLabel id={id} label={label} labelDisplay={labelDisplay} />}
       <Box
-        color={disabled ? 'secondary' : 'default'}
         display="flex"
         position="relative"
         rounding={4}
         width="100%"
+        dangerouslySetInlineStyle={{
```
rlingineni, line 130: Why is the override needed?

rlingineni: What was it before?

rlingineni (approved):

packages/gestalt/src/SelectList.js
AlbertCarreras: It's the actual color for tokens. 

Before it was just default color from box which isn't the same as the other form fields. They were different in dark mode. Different gray.

I had to move the background color below the icon to surface it. The icon cannot go above as it affects the hovering behavior and the pointer.

PR #3452, opened 2024-02-27
Popover: Remove autofocusing dismiss button

diyorbek:
### Summary

Auto focusing dismiss button when Popover is mounted causing the content jump. 
When Popover is mounted, it is not yet positioned correctly and it is at `top: 0; left: 0`. That's why focusing it scrolls the page up.

Focusing the dismiss button is actually redundant and `FloatingFocusManager` takes care of focusing the 1st focusable element inside floating element, i.e. Popover.

Also, for Dropdown `shouldFocus={false}` was set to Popover becasue Dropdown's item always gets the focus. That's why focusing on Popover itself is redundant.

### Links

- [Bug in Jira](https://jira.pinadmin.com/browse/BUG-180930)

### Checklist

### DIFF
--- a/packages/gestalt/src/Popover/InternalPopover.js
+++ b/packages/gestalt/src/Popover/InternalPopover.js
@@ -1,5 +1,5 @@
 // @flow strict
-import { type Node as ReactNode, useEffect, useRef } from 'react';
+import { type Node as ReactNode, useRef } from 'react';
 import Controller from './Controller';
 import Box from '../Box';
 import { useDefaultLabelContext } from '../contexts/DefaultLabelProvider';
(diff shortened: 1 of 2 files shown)


packages/gestalt/src/Popover/InternalPopover.js
```diff
@@ -58,13 +58,10 @@ export default function InternalPopover({
 
   const dismissButtonRef = useRef<null | HTMLAnchorElement | HTMLButtonElement>(null);
 
-  useEffect(() => {
-    dismissButtonRef.current?.focus();
```
AlbertCarreras, line 62: should we not focus this conditionally to the experiment? @diyorbek

AlbertCarreras (approved):

PR #3455, opened 2024-02-27
TextField: Adding new autocomplete strings

eniomoura:
#### What changed?

Added all MDN and W3C (as of now) valid strings for the input autocomplete attribute.

#### Why?

To allow for cleanup of FlowFixes related to unsupported types currently present on Pinboard.

### Links

- [Jira](https://jira.pinadmin.com/browse/GESTALT-7558)
- [Meeting doc](https://docs.google.com/document/d/1Wki_fh5qhGiom5ykmNVNKwrnSYzjqY8s0qA9gxQZLrw/edit#heading=h.myivmu4qfg1h)


packages/gestalt/src/TextField/InternalTextField.js
```diff
@@ -25,6 +25,67 @@ import typography from '../Typography.css';
 
 type SizeType = 'sm' | 'md' | 'lg';
 
+type autoCompleteType =
```
willsmanley, line 28: Can we use a shared definition between these since they are the same? Otherwise LGTM

dangerismycat: This is a known issue with the tool we use to autogenerate props tables from types — any types defined outside of a given file will show up as the type name rather than the type value. The gain of autogen types and a single source of truth for a component's types was worth the tradeoff of some type duplication.

That said, that would be an awesome problem to solve someday.

dangerismycat: Oh, but actually we can dedupe this type. The type needs to be defined in TextField.js to be picked up by the autogen tool. That type can be exported and consumed by InternalTextField.js. (Philosophically that's kinda weird/gross, but it might be preferable to repeating the type and the eventual drift that could happen)

eniomoura: This was done this way because our docs are autogenerated from the component's props, and this autogeneration breaks when importing in the types. However since in the new approach we are overriding this autogeneration, I used the shared definition within InternalTextField :)

packages/gestalt/src/TextField.js
```diff
@@ -23,7 +23,7 @@ type Props = {
   /**
    * Indicate if autocomplete should be available on the input, and the type of autocomplete. Autocomplete values are implemented upon request. [Reach out to the Gestalt team](https://gestalt.pinterest.systems/team_support/get_help#Slack-channels) if you need [additional autocomplete values](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete#values) to be supported.
```
AlbertCarreras, line 24: Can we add the description here with all props?

Something like:


_Indicate if autocomplete should be available on the input, and the type of autocomplete. All [autocomplete values](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete#values) are supported._

packages/gestalt/src/TextField.js
```diff
@@ -23,7 +23,7 @@ type Props = {
...
    */
-  autoComplete?: 'bday' | 'current-password' | 'email' | 'new-password' | 'on' | 'off' | 'username',
+  autoComplete?: autoCompleteType,
```
AlbertCarreras, line 26: I think this is fine. Importing on this file.

packages/gestalt/src/TextField.js
eniomoura: DONE

docs/pages/web/textfield.js
```diff
@@ -501,6 +501,78 @@ The first example shows an empty Textfield with \`maxLength\` set to 20 characte
 export async function getServerSideProps(): Promise<{
   props: { generatedDocGen: DocGen },
 }> {
+  const generatedDocGen = await docGen('TextField');
+
+  generatedDocGen.props.autoComplete = {
+    defaultValue: null,
+    required: false,
+    flowType: {
```
AlbertCarreras, line 509: Screenshot by Dropbox Capture


I think you can just do that:

`  generatedDocGen.props.color.flowType.raw = '"bday" | "current-password" | "email" | "new-password" | "on" | "off" | "username"';
`

packages/gestalt/src/TextField.js
eniomoura: Yes, so it only ever has to be updated in one place, and it's ok since the docs are overriden.

AlbertCarreras:
Dont forget to updated the .ts file!

docs/pages/web/textfield.js
eniomoura: Ravi suggested the raw type to have all types, so it's properly copied
https://pinterest.slack.com/archives/G01GN7CE99S/p1709073511530249?thread_ts=1709064809.271199&cid=G01GN7CE99S

AlbertCarreras: We don't copy the whole Flowtype from the Table, just:

$ElementType<React$ElementConfig<typeof TextField>, 'autoComplete'>

So Ravi's suggestion is not necessary unless I'm missing something

eniomoura: Makes sense, will edit!

packages/gestalt/src/TextField/InternalTextField.js
```diff
@@ -25,6 +25,67 @@ import typography from '../Typography.css';
 
 type SizeType = 'sm' | 'md' | 'lg';
 
+export type autoCompleteType =
+  | 'on'
```
AlbertCarreras, line 29: nit but can we order alphabetically?

eniomoura: done!

rlingineni (approved):

PR #3459, opened 2024-02-29
Internal: migration to token constants

AlbertCarreras:
Internal: migration to token constants

Major due to 
Screenshot by Dropbox Capture

### DIFF
--- a/docs/docs-components/AppLayout.js
+++ b/docs/docs-components/AppLayout.js
@@ -2,6 +2,10 @@
 import { Fragment, type Node as ReactNode, useEffect, useState } from 'react';
 import { useRouter } from 'next/router';
 import { Box, DeviceTypeProvider, Divider, FixedZIndex, Flex } from 'gestalt';
+import {
+  TOKEN_COLOR_GRAY_ROBOFLOW_700,
+  TOKEN_COLOR_ORANGE_FIRETINI_0,
+} from 'gestalt-design-tokens/dist/js/constants';
 import { useAppContext } from './appContext';
 import { DocsConfigProvider, useDocsConfig } from './contexts/DocsConfigProvider';
 import DocsSideNavigation, { MIN_NAV_WIDTH_PX } from './DocsSideNavigation';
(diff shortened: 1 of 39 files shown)


packages/gestalt/src/SideNavigationTopItem.js
```diff
@@ -25,8 +26,8 @@ import Text from './Text';
 import { type Indexable } from './zIndex';
 
 export const NESTING_MARGIN_START_MAP = {
-  '0': 'var(--space-400)',
-  '1': 'var(--space-1200)',
+  '0': TOKEN_SPACE_400,
```
rlingineni, line 29: will this have `px` added? Or does it need that?

rlingineni (approved):

docs/docs-components/AppLayout.js
```diff
@@ -2,6 +2,10 @@
 import { Fragment, type Node as ReactNode, useEffect, useState } from 'react';
 import { useRouter } from 'next/router';
 import { Box, DeviceTypeProvider, Divider, FixedZIndex, Flex } from 'gestalt';
+import {
+  TOKEN_COLOR_GRAY_ROBOFLOW_700,
+  TOKEN_COLOR_ORANGE_FIRETINI_0,
+} from 'gestalt-design-tokens/dist/js/constants';
```
rlingineni, line 8: Would it make sense to update the package exports to the js file instead of the css file. I'm not sure if the variables.css file is consumed directly

What if you change `package.json` in design-tokens
```
main": "dist/css/variables.css" -> main": "dist/js/constants.js"
```

So it becomes like this

```
import  {TOKEN_COLOR_GRAY} from `gestalt-design-tokens`
```

rlingineni (approved):

docs/docs-components/AppLayout.js
AlbertCarreras: yep, that´s a good point.

packages/gestalt/src/SideNavigationTopItem.js
AlbertCarreras: TOKEN_SPACE_400 is equivalent to 'var(--space-400)'

Yes, it will refer to the value with px.

docs/docs-components/AppLayout.js
AlbertCarreras: DONE

PR #3465, opened 2024-03-04
Masonry: seeded random values for deterministic integration tests

willsmanley:
#### What changed?

Masonry integration tests now have randomized values that are deterministic from one test run to the next due to the addition of a custom pseudo random number generator that accepts a seed value.

#### Why?

This allows us to write integration tests to assert things like "this two column element should be rendered on the third row, which is the optimal position given these item sizes".

### Links

- [Jira](https://jira.pinadmin.com/browse/GESTALT-7638)

### Checklist

### DIFF
--- a/docs/pages/integration-test/masonry.js
+++ b/docs/pages/integration-test/masonry.js
@@ -5,6 +5,7 @@ import { useRouter } from 'next/router';
 import { ColorSchemeProvider, Masonry } from 'gestalt';
 import generateExampleItems from '../../integration-test-helpers/masonry/items-utils/generateExampleItems';
 import generateRealisticExampleItems from '../../integration-test-helpers/masonry/items-utils/generateRealisticExampleItems';
+import getRandomNumberGenerator from "../../integration-test-helpers/masonry/items-utils/getRandomNumberGenerator";
 import pinHeights, {
   type PinHeight,
 } from '../../integration-test-helpers/masonry/items-utils/pinHeights';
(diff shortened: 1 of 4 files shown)


docs/pages/integration-test/masonry.js
```diff
@@ -137,13 +137,26 @@ export default function TestPage({
...
+}
+
 export async function getServerSideProps(): Promise<{
   props: { randomNumberSeeds: $ReadOnlyArray<number> },
 }> {
   // This is used to ensure we're using the same dataset of realistic pins on the server and client
   const randomNumberSeeds = Array.from({
     length: REALISTIC_PINS_DATASET_SIZE,
-  }).map(() => Math.random());
+  }).map(createSeededRandom(12345));
```
liuyenwei, line 159: this is helpful for the realistic (e.g. random) pin heights examples however, looking through the test code, I think [this](https://github.com/pinterest/gestalt/blob/master/docs/integration-test-helpers/masonry/MasonryContainer.js#L280) would need to be updated too (not sure why seed is random there vs being passed in 🤔 )

@manvoid was this the issue you were running into? i also see that we have a random here for placement for the actual two column item [here](https://github.com/pinterest/gestalt/blob/master/docs/integration-test-helpers/masonry/items-utils/generateExampleItems.js#L30)

willsmanley: good finds i can catch those cases

willsmanley: i updated those two sites and additionally found an existing generator that ryan made, so i removed my new one and standardized on the one that already exists.

manvoid: As Yen mentioned this is for the "realisticPinHeights" seed that calculated the pin heights. The logic that positions the two column modules is inside the `generateExampleItems`/`generateRealisticExampleItems` helper functions. In fact we would need a way to set in a deterministic way both sets, the pins before the two column module and the two column module itself. 

This could solve the first part setting a seed that positions the same pins but still we would need a way to exactly position the two column item.

Other option I was thinking wad add the capacity of passing the pins data we want to render on the test similar to how we are doing the unit tests. That was the idea I had with the 'set-masonry-items' improvements but my solution was not correctly working because that didn't include the items on the first render. But that would be more powerful so we can set exactly the items we want to test.

willsmanley: do my new changes solve it for those two functions?

liuyenwei (approved):

manvoid (approved):

liuyenwei:
@willsmanley FYI i think CI is failing bc of prettier

PR #3471, opened 2024-03-06
Tokens: generate commonJS constant file

AlbertCarreras:
Tokens: generate commonJS constant file
Screenshot by Dropbox Capture

To prevent this issue in Pinboard 
image

the issue is because the file is using export vs the cjs module.exports

In Pinboard, we ignore _node_modules_ by default for Jest

i.e. for gestalt we ship both ecjs and esm equivalents, from a consistency perspective that would make more sense to have both (but if we have to pick one we should stick with esm. Its more ideal for treeshaking on the pinboard side
since you can't treeshake cjs)


packages/gestalt-design-tokens/build.js
```diff
@@ -150,6 +150,39 @@ StyleDictionary.registerFormat({
...
+        // `getReferences()` methods. `usesReference()` will return true if
+        // the value has a reference in it. `getReferences()` will return
+        // an array of references to the whole tokens so that you can access their
+        // names or any other attributes.
+        if (dictionary.usesReference(token.original.value)) {
+          // Note: make sure to use `token.original.value` because
+          // `token.value` is already resolved at this point.
+          const refs = dictionary.getReferences(token.original.value);
+          refs.forEach((ref) => {
+            value = value.replace(ref.value, () => `${ref.name}`);
```
diyorbek, line 169: Let's make it simpler
```suggestion
            value = value.replace(ref.value, ref.name);
```

packages/gestalt-design-tokens/build.js
```diff
@@ -150,6 +150,39 @@ StyleDictionary.registerFormat({
...
+          });
+        }
+        return `  TOKEN_${token.path
+          .join('_')
+          .toUpperCase()
+          .replace('-', '_')}: 'var(--${token.path.join('-')})',`;
+      })
+      .join(`\n`);
+
+    return `${fileHeader({
```
diyorbek, line 179: can we move this part to a variable? to make it more readable.

diyorbek (approved):

packages/gestalt-design-tokens/build.js
AlbertCarreras: DONE

packages/gestalt-design-tokens/build.js
AlbertCarreras: DONE

PR #3473, opened 2024-03-07
Masonry: Enable two column module on secod batch

manvoid:
### Summary

This PR fixes an issue where the page would freeze trying to calculate the position of a two column module when there are many items that have the height already measured but not the position. This is required to enable the use of a two column module on the first page (with the items comming from the server).

#### Detail

When Masonry is first rendered it measures the heights of all the items at the same time, after that it calculates the positions. This was not a problem for one column items because the process to calculate the positions is very fast, but for two column items this is really hard because it has to create the graph for all the items not limited to the batch size, depending on the amount of items this can freeze the page completely. 

To fix it we can force the batch size when there is a two column module taking into account the items that don't have positions and limiting those to the batch size.

This PR also fixes a bug on the integration test page were the event are added twice.

### DIFF
--- a/docs/integration-test-helpers/masonry/MasonryContainer.js
+++ b/docs/integration-test-helpers/masonry/MasonryContainer.js
@@ -86,31 +86,43 @@ export default class MasonryContainer extends Component<Props<{ ... }>, State> {
 
   randomNumberSeed: number = 0;
 
+  triggerReflow: () => void = () => {
+    if (this.gridRef.current) {
+      this.gridRef.current.reflow();
+      this.forceUpdate();
+    }
+  };
+
+  // $FlowFixMe[unclear-type]
+  setMasonryItems: (e: { detail: { items: $ReadOnlyArray<Object> } }) => void = (e) => {
+    this.setState({
+      items: e.detail.items,
+    });
+  };
+
+  // eslint-disable-next-line class-methods-use-this
+  countError: () => void = () => {
+    window.ERROR_COUNT += 1;
+  };
+
   componentDidMount() {
     window.TEST_FETCH_COUNTS = 0;
-
-    window.addEventListener('trigger-reflow', () => {
-      if (this.gridRef.current) {
-        this.gridRef.current.reflow();
-        this.forceUpdate();
-      }
-    });
-
-    window.addEventListener('set-masonry-items', (e) => {
-      this.setState({
-        items: e.detail.items,
-      });
-    });
-
     window.ERROR_COUNT = window.ERROR_COUNT || 0;
-    window.addEventListener('error', () => {
-      window.ERROR_COUNT += 1;
-    });
+
+    window.addEventListener('trigger-reflow', this.triggerReflow);
+    window.addEventListener('set-masonry-items', this.setMasonryItems);
+    window.addEventListener('error', this.countError);
 
     // Trigger a re-render in case we need to render /w scrollContainer.
     setTimeout(() => {
       this.setState({ mounted: true }); // eslint-disable-line react/no-unused-state
     });
+  }
+
+  componentWillUnmount() {
+    window.removeEventListener('trigger-reflow', this.triggerReflow);
+    window.removeEventListener('set-masonry-items', this.setMasonryItems);
+    window.removeEventListener('error', this.countError);
   }
 
   handleToggleScrollContainer: () => void = () => {
--- a/packages/gestalt/src/Masonry.js
+++ b/packages/gestalt/src/Masonry.js
@@ -594,8 +594,6 @@ export default class Masonry<T: { ... }> extends ReactComponent<Props<T>, State<
         ? Math.max(...positions.map((pos) => pos.top + pos.height))
         : 0;
 
-      console.log(positions);
-
       gridBody = (
         <div style={{ width: '100%' }} ref={this.setGridWrapperRef}>
           <div className={styles.Masonry} role="list" style={{ height, width }}>
--- a/packages/gestalt/src/Masonry.js
+++ b/packages/gestalt/src/Masonry.js
@@ -577,7 +577,7 @@ export default class Masonry<T: { ... }> extends ReactComponent<Props<T>, State<
       gridBody = <div style={{ width: '100%' }} ref={this.setGridWrapperRef} />;
     } else {
       // Full layout is possible
-      const itemsWithMeasures = items.filter((item) => item && measurementStore.has(item));
+      const itemsWithMeasurements = items.filter((item) => item && measurementStore.has(item));
       const itemsWithoutPositions = items.filter((item) => item && !positionStore.has(item));
       const hasTwoColumnItems =
         // $FlowFixMe[prop-missing] We're assuming `columnSpan` exists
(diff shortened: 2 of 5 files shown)


docs/integration-test-helpers/masonry/MasonryContainer.js
```diff
@@ -86,33 +86,45 @@ export default class MasonryContainer extends Component<Props<{ ... }>, State> {
...
 
-    window.addEventListener('set-masonry-items', (e) => {
-      this.setState({
-        items: e.detail.items,
-      });
-    });
+  // eslint-disable-next-line class-methods-use-this
+  countError: () => void = () => {
+    window.ERROR_COUNT += 1;
+  };
```
willsmanley, line 106: just curious, why do we need to count errors here?

manvoid: Not completely sure but I see that we have two integration test on playwright that check for that variable. `playwright/masonry/null-items.spec.mjs` and `playwright/masonry/handle-update-during-insertions.spec.mjs`.

willsmanley: ah i see thanks. well PR LGTM but i cant approve it.

packages/gestalt/src/Masonry.js
```diff
@@ -589,6 +589,17 @@ export default class Masonry<T: { ... }> extends ReactComponent<Props<T>, State<
         .filter((item) => item && !measurementStore.has(item))
         .slice(0, itemsToMeasureCount);
 
+      // If there is a two column module be sure that we only calculate new positions
+      // for batch size number of items
+      const itemsToRender = hasTwoColumnItems
```
liuyenwei, line 594: this works since it effectively limits the number of items we use to calculate new positions, but the downside here is that it will take more render cycles to complete, depending on where the two column item is positioned.

example:
assuming 16 items, where the two col module is in position 9 (your example in the test):
- first pass render 6 items
- second pass render 6 additional items
- third pass render 6 additional items (this one has the two column module)
- fourth pass render the remaining 4

alternatively, what if we continue to render/layout all  items but leave it up to the layout on how many items it wants to factor into the graph by limiting the size of this array here: https://github.com/pinterest/gestalt/blob/0a517310803467aaf2bc1f017d88ad38405dee26/packages/gestalt/src/Masonry/defaultTwoColumnModuleLayout.js#L266. This should achieve a similar effect since it limits the number of items that are factored into the graph but allow us to do a full layout sooner

Going off the previous example, we would render all 16 items in one pass but the two column layout would only take X amount of items into account

packages/gestalt/src/Masonry/defaultTwoColumnModuleLayout.js
```diff
@@ -278,14 +278,54 @@ const defaultTwoColumnModuleLayout = <T>({
...
+        ) {
+          currentBatch = itemsWithoutPositions.slice(
+            index,
+            index + TWO_COL_ITEMS_MEASURE_BATCH_SIZE,
+          );
+          // $FlowFixMe[incompatible-use] We're assuming `columnSpan` exists
+          if (currentBatch.some((item) => item.columnSpan > 1)) {
+            batchWithTwoColumnItem = [...currentBatch];
+          } else {
+            prevOneColumnItems.push(...currentBatch);
```
liuyenwei, line 302: this also has the effect of "pushing up" later one column items (again, very dependent on what the incoming items looks like)

assuming we have 24 items (so 4 batches) and the two column item is in the second batch, this would result in batches 1, 3 and 4 being positioned before the second batch.

instead i think what we want to do is split into three batches:
- all items before two column batch
- batch containing two column item
- everything else

first batch would then be `prevOneColumnItems`, second batch would be `batchWithTwoColumnItems` and third batch would be handled already at the end via https://github.com/pinterest/gestalt/pull/3458

packages/gestalt/src/Masonry/defaultTwoColumnModuleLayout.js
```diff
@@ -278,14 +278,54 @@ const defaultTwoColumnModuleLayout = <T>({
     };
 
     if (hasTwoColumnItems) {
+      const prevOneColumnItems = [...itemsWithPositions];
```
liuyenwei, line 281: the naming for some of these variables is now pretty confusing/inaccurate imo (e.g. `prevOneColumnItems` here could actually contain two column items)

liuyenwei (approved):

playwright/masonry/two-module.spec.mjs
```diff
@@ -21,4 +21,52 @@ test.describe('Masonry: two columns module', () => {
...
+      { name: 'Pin 14', height: 200, color: '#E230BA' },
+      { name: 'Pin 15', height: 202, color: '#FAB032' },
+    ];
+
+    const toogleMountButton = await page.locator(selectors.toggleMount);
+    await toogleMountButton.click();
+
+    await page.evaluate((proxiedItemData) => {
+      window.dispatchEvent(
+        new CustomEvent('set-masonry-items', {
```
liuyenwei, line 60: okay for a followup - ideally we'd update the integration test harness to specify initial items and test the initial page scenario here

manvoid: Yeah, my current solution is to set the items with this event and after that toggle the mount of the component off and on and it works but we could find a more straightforwad solution for that.

manvoid: I'll create a backlog item for that.

packages/gestalt/src/Masonry/defaultTwoColumnModuleLayout.js
```diff
@@ -278,14 +278,41 @@ const defaultTwoColumnModuleLayout = <T>({
...
         getOneColumnItemPositions({
-          items: itemsWithPositions,
+          items: [...itemsWithPositions, ...pre],
           heights,
           ...commonGetPositionArgs,
         });
 
+      // Adding the extra prev column items to the position cache
+      if (paintedItemPositions.length > itemsWithPositions.length) {
+        paintedItemPositions.forEach(({ item, position }) => {
```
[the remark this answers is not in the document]
manvoid: We need to add the positions to the cache, this come from the `paintedItemPositions` but that is an array of `$ReadOnlyArray<{ item: T, position: Position }>`, we would have to search for each prev item on that array, adding everything seems easier although I understand your comment. We can double check this on our sync.

packages/gestalt/src/Masonry/defaultTwoColumnModuleLayout.js
```diff
@@ -5,6 +5,10 @@ import { type HeightsStoreInterface } from './HeightsStore';
 import mindex from './mindex';
 import { type NodeData, type Position } from './types';
 
+// When there's a 2-col item in the most recently fetched batch of items, we need to measure more items to ensure we have enough possible layouts to minimize whitespace above the 2-col item
+// This may need to be tweaked to balance the tradeoff of delayed rendering vs having enough possible layouts
+export const TWO_COL_ITEMS_MEASURE_BATCH_SIZE = 6;
```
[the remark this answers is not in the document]
manvoid: Yes. I'll create a backlog ticket for this.

packages/gestalt/src/Masonry/defaultTwoColumnModuleLayout.js
manvoid: As suggested used `prev` instead.

packages/gestalt/src/Masonry/defaultTwoColumnModuleLayout.js
manvoid: Changed to a more straightforward solution creating only three parts, `pre` before the two column batch, the two column batch and the rest.

PR #3474, opened 2024-03-07
Tokens: theming folders and config for automatic token builds

AlbertCarreras:
Tokens: theming folders

For each theme folder, we pick up the name and prefix it automatically:
Screenshot by Dropbox Capture

We add logic to process themes. 
We output theme files with theme prefix.
Screenshot by Dropbox Capture

In Android we transform the name with a prefix for easier experimentation.
Screenshot by Dropbox Capture

## NOTE

Current vrTheme  folder is just a duplication from mainTheme for explorations purposes and it shouldn't be used beyond experimentation purposes.


AlbertCarreras:
@ Nice! This looks nice.  Looks like build.js has a lot changing inside of it.
Could we have it be direct 1:1 as "vr_theme" -> vr_

DONE

AlbertCarreras:
@rlingineni  Also, any reason dataVisualization is it's own theme? How is that expected to be consumed, since it's used in main 

I dont think it's its own theme... it's more like a...  subset of the base tokens and alias ones. 

I'd add the tokens in the base... I can look at refactoring that as a follow up

AlbertCarreras:
@rlingineni This is an organizational thing, but:
would it make sense to instead have the "theme"  as the root folder, and everything under there can have the color, opacity rounding structure?
Like for example, instead of color/mainTheme,
"mainTheme/color/token1"
"vrTheme/color/token1
It may not affect the output, but it gives us scope on what tokens exist and not in one theme and another.
For example, dataviz is just a color theme


A dense theme will only apply to some files. The theme now it’s just for interim experimentation I think we can just leave as is for now and reevaluate file structure in the future as we might be preoptimizing

rlingineni (approved):
Approved - but the "data-visualization" seems like an odd one

PR #3476, opened 2024-03-11
SideNavigation: Collapsible SideNavigation

diyorbek:
### Summary

Collapsible SideNavigation variant. 

#### What changed?

Collapsible SideNavigation is a *controlled component*. It adds "expand/collapse" button on top of the component. Which is sticky.

Added 3 new props to SideNavigation:

1. `collapsed` - controls wheter the component is collapsed
2. `onCollapsed` - callback is fired when expand/collapse is clicked
3. `onPreview` -  callback is fired when the component is hover over for preview

Created a set of new files for controlling the collapsible bahavior of the SideNavigation.

1. `flattenChildren.js` - a utility file for removing fragments from children. keyed flattening is copied from [react-keyed-faltten-children](https://github.com/grrowl/react-keyed-flatten-children)
2. `navigationChildrenUtils.js` - for analyzing SideNavigation children, contains logic for rendering items in the collapsed state
3. `ItemEllipsis.js` - new internal SideNavigation item for rendering collapsed items without icons.

#### Icons
icons

#### Header
header

#### Mixed
mixed

#### Empty
wmpty

### Links

- [Jira ticket](https://jira.pinadmin.com/browse/GESTALT-5594)
- [TDD](link to Paper doc)
- [Figma specs](https://www.figma.com/file/0GGe3tDQDspiHFOmsiro28/Handoff---SideNav-collapsible?type=design&node-id=0%3A1&mode=design&t=mEihTuQgsDZLL5W6-1)

### Checklist


docs/pages/web/sidenavigation.js
```diff
@@ -454,6 +459,53 @@ When building SideNavigation, we might want to render different combinations of
           />
         </MainSection.Subsection>
 
+        <MainSection.Subsection title="Collapsible" description="">
```
diyorbek, line 462: @hectoid Could you suggest anything to write as a description here?

packages/gestalt/src/SideNavigation.css
```diff
@@ -14,13 +15,39 @@
 
 .fullHeight {
   height: 100%;
-  overflow: scroll;
+  overflow-y: scroll;
+}
+
+.contentWidthTransition {
+  transition: width 0.2s;
```
diyorbek, line 22: `.2s` is used also in Switch and Masony. I took the value from there to make it consistent.
/cc @hectoid

packages/gestalt/src/SideNavigation.css
```diff
@@ -14,13 +15,39 @@
...
 }
 
 .section {
+  margin-bottom: var(--space-600);
```
diyorbek, line 26: `Section` now has both top and bottom margin. It was done mainly for the purpose of visially splitting `Section` from `TopItem` and `Group` (these don't have any kind of margin). 

It is for the case where `Section` comes before `TopItem` or `Group`. If `Section`s are stacked, the top and bottom margins collapse, and the gap between them stays consistent.

packages/gestalt/src/SideNavigation.css
```diff
@@ -14,13 +15,39 @@
...
+.section:first-child {
   margin-top: 0;
 }
+
+.section:last-child {
+  margin-bottom: 0;
+}
+
+.sectionStartDivider {
+  display: none;
```
diyorbek, line 39: Section dividers appear when `SideNavigation` is collapsed. 
There are 2 dividers for each section, top (sectionStartDivider) and bottom (sectionEndDivider).

Bottom divider is always displayed, *except* the section is the last child.

Top divider is always hidden. Because if there are 2 sections stacked, in the middle there appears 2 dividers (1st one's bottom divider and 2nd's top divider).

Top divider appears only when `TopItem` or `Group` comes before `Section`.

packages/gestalt/src/SideNavigation.css
```diff
@@ -14,13 +15,39 @@
...
+  margin-bottom: var(--space-600);
+}
+
+.sectionEndDivider {
+  margin-top: var(--space-600);
+}
+
+.liItem:not(.section) + .section .sectionStartDivider {
+  display: block;
```
diyorbek, line 48: `.liItem:not(.section)` represents `TopItem` or `Group`.

For example, `TopItem` is the previous *adjacent* sibling of `Section`, the section's top divider should appear:

```jsx
<TopItem />
<Section /> // top divider appears
```

packages/gestalt/src/SideNavigation.js
```diff
@@ -48,6 +41,18 @@ export type Props = {
...
+   */
+  collapsed?: boolean,
+  /**
+   * Callback fired when the SideNavigation is collapsible and the collapse/expand button is clicked. This functionality is not supported in mobile.
+   */
+  onCollapse?: (boolean) => void,
+  /**
+   * Callback fired when the *collapsed* SideNavigation is hovered and opened in preview mode. This functionality is not supported in mobile.
+   */
+  onPreview?: (boolean) => void,
```
diyorbek, line 55: This callback had to be introduced so that the (internal) `preview` state is communicated. This is necessary when `header` and `footer` needs to appear the same in both expanded and preview states. Take a look at the `collapsibleHeaderExample.js` file for the usage.

We could also expose `preview` state and make it controlled as well. However, I think it shouldn't be, for the simplicity of the API.

packages/gestalt/src/SideNavigation.js
```diff
@@ -67,15 +72,15 @@ export default function SideNavigation({
...
 }: Props): ReactNode {
-  const navigationChildren = getChildrenToArray({
-    children,
-    filterLevel: 'main',
-  });
   const { accessibilityDismissButtonLabel } = useDefaultLabelContext('SideNavigation');
   const id = useId();
   const deviceType = useDeviceType();
   const isMobile = deviceType === 'mobile';
+  const collapsible = collapsed != null;
```
diyorbek, line 83: As long as `collapsed` prop is a boolean value, the `SideNavigation` is collapsible.

packages/gestalt/src/SideNavigation.js
```diff
@@ -102,46 +107,29 @@ export default function SideNavigation({
...
-            </Box>
-          </div>
-        </Box>
+    <SideNavigationProvider
+      collapsible={collapsible}
+      collapsed={collapsed}
+      onCollapse={onCollapse}
+      onPreview={onPreview}
+    >
+      <ScrollBoundaryContainer overflow={collapsible ? 'visible' : undefined}>
```
diyorbek, line 124: We have to make overflow visible, because if not, box shadow in preview state will not be visible.

Also `ScrollBoundaryContainer` is soon to be deprecated and removed.

packages/gestalt/src/SideNavigation.js
```diff
@@ -102,46 +107,29 @@ export default function SideNavigation({
...
+        <SideNavigationContent
+          accessibilityLabel={accessibilityLabel}
+          footer={footer}
+          header={header}
+          showBorder={showBorder}
+        >
+          {children}
+        </SideNavigationContent>
```
diyorbek, line 132: I created a new file for the desktop `SideNavigation` contents not to bloat the file with more logic.

packages/gestalt/src/SideNavigation/Collapser.js
```diff
@@ -0,0 +1,38 @@
...
+        <IconButton
+          icon={collapsed ? 'arrow-double-forward' : 'arrow-double-back'}
+          accessibilityLabel={
+            collapsed
+              ? 'Navigation collapsed. Click button to expand.'
+              : 'Navigation expanded. Click button to collapse.'
+          }
+          size="xs"
+          onClick={() => {
+            onCollapse?.(!collapsed);
```
diyorbek, line 30: This is where the collapse state is toggled.

packages/gestalt/src/SideNavigation/Collapser.js
```diff
@@ -0,0 +1,38 @@
...
+
+            if (collapsed) setOverlayPreview(false);
```
diyorbek, line 32: Makes sure when collapsed preview state is reset.

packages/gestalt/src/SideNavigation/GroupContent.js
```diff
@@ -92,61 +101,93 @@ export default function SideNavigationGroupContent({
     }
   }, [hovered, focused, primaryAction, forceIconButton, isMobile, showIconButton]);
 
+  const hasBorder = sideNavigationCollapsed
+    ? hasActiveChild
+    : expanded && selectedItemId === itemId;
+
+  const collapsed = sideNavigationCollapsed && !overlayPreview;
```
diyorbek, line 108: Here `collapsed` means fully collapsed and not in preview state.

packages/gestalt/src/SideNavigation/GroupContent.js
```diff
@@ -92,61 +101,93 @@ export default function SideNavigationGroupContent({
...
+          aria-label={notificationAccessibilityLabel}
+          height={8}
+          width={8}
+          rounding="circle"
+          color="primary"
+          role="status"
+          position="absolute"
+          dangerouslySetInlineStyle={{ __style: { top: 4, right: 4 } }}
+        />
+      ) : null}
```
diyorbek, line 142: The above is the notification indicator on the top-right of the icon. It is displayed only is collapsed state.

packages/gestalt/src/SideNavigation/GroupContent.js
```diff
@@ -92,61 +101,93 @@ export default function SideNavigationGroupContent({
...
+
       <Flex gap={{ row: 2, column: 0 }} height="100%" width="100%">
         {icon ? (
           <Flex.Item alignSelf="center">
             <Box aria-hidden>
               {typeof icon === 'string' ? (
-                <Icon inline icon={icon} accessibilityLabel="" color="default" />
+                <Icon size={20} inline icon={icon} accessibilityLabel="" color="default" />
```
diyorbek, line 149: Icon size had to be `20` since the SideNavigation v1.

packages/gestalt/src/SideNavigation/getChildrenToArray.js
```diff
@@ -1,19 +1,16 @@
...
 import { Children, Fragment, type Node as ReactNode } from 'react';
+import { ALLOWED_CHILDREN_MAP } from './navigationChildrenUtils';
+import { type ReactChildArray } from '../utils/flattenChildren';
 
-const ALLOWED_CHILDREN_MAP = {
-  main: ['SideNavigation.Section', 'SideNavigation.TopItem', 'SideNavigation.Group'],
-  nested: ['SideNavigation.NestedItem', 'SideNavigation.NestedGroup'],
-};
-
+/** @deprecated */
```
diyorbek, line 6: This function does 3 things:

1. Checks whether subcomponent composition is valid
2. Flattens subcomponents
3. Limits `Fragment` nesting up to 2 levels

Deprecated, because I created separate functions for the first 2 things. Removed `Fragment` nesting limitation.

It is also used in `GroupdMobile` file and I didn't want to make unrelated changes.

packages/gestalt/src/SideNavigation/navigationChildrenUtils.js
```diff
@@ -0,0 +1,113 @@
...
+  return flattenChildren(children).reduce(
+    (count, child) =>
+      child.type.displayName === 'SideNavigation.Section'
+        ? count + countItemsWithIcon(child.props.children)
+        : count + Number(Boolean(child.props.icon)),
+    0,
+  );
+}
+
+export function getChildrenActiveProp(
```
diyorbek, line 43: Basically does a BFS on sidenavigation children.

packages/gestalt/src/SideNavigation/navigationChildrenUtils.js
```diff
@@ -0,0 +1,113 @@
...
+    return (
+      // eslint-disable-next-line react/no-array-index-key
+      <li key={i} className={classnames(styles.liItem)}>
+        <ItemsEllipsis {...item} />
+      </li>
+    );
+  });
+}
+
+export function groupIconlessChildren(children: ReactChildArray): $ReadOnlyArray<ReactNode> {
```
diyorbek, line 77: This function is a bit crazy one. It is for the items that are not grouped in a `Section`. 

Creates ellipses for each portion of "non-section" items (separated by `Section`s) if there are items without icons. 
That means, If there are "non-section" items before and after `Section`, each will have an ellipsis for iconless items.

Take a look at the [collapsibleWithMixedIconsExample.js](https://github.com/pinterest/gestalt/pull/3476/files#diff-f0f42099ab0c6407c1cfe8fb580d5f27ebb11d5a4004e28493796639be948fc7) file for the usecase.

packages/gestalt/src/SideNavigation/navigationChildrenUtils.js
```diff
@@ -0,0 +1,113 @@
...
+    if (lastEllipsisIndex === undefined || lastEllipsisIndex < lastSectionIndex) {
+      const ellipsis: EllipsisProps = {};
+      lastEllipsisIndex = index;
+      acc.push(ellipsis);
+    }
+
+    const lastEllipsis = acc.at(lastEllipsisIndex);
+
+    if (lastEllipsis) {
+      lastEllipsis.notificationAccessibilityLabel ||= notificationAccessibilityLabel;
```
diyorbek, line 100: if there are items with notification labels, take the 1st one and set it to the ellipsis.

packages/gestalt/src/SideNavigation/navigationChildrenUtils.js
diyorbek: Also, if there are items with 

1. notification labels
2. `active` prop

algorithms takes the 1st one's infor and sets it to the ellipsis.

packages/gestalt/src/SideNavigationGroup.js
```diff
@@ -105,37 +106,19 @@ export default function SideNavigationGroup({
...
-      .map((child) =>
-        getChildrenToArray({
-          children: child?.props?.children,
-          filterLevel: 'nested',
-        }).find(hasActiveChildCallback),
-      )
-      .filter(Boolean);
-  }
-
-  const hasAnyActiveChild =
```
diyorbek, line 137: moved this and related logic to the `navigationChildrenUtils.js`

packages/gestalt/src/SideNavigationGroup.js
```diff
@@ -105,37 +106,19 @@ export default function SideNavigationGroup({
...
-    !!hasActiveChildren || (!!hasActiveGrandChildren && !!hasActiveGrandChildren[0]);
```
diyorbek, line 138: moved this and related logic to the `navigationChildrenUtils.js`

packages/gestalt/src/SideNavigationGroup.js
```diff
@@ -198,6 +185,10 @@ export default function SideNavigationGroup({
               } else {
                 onExpand?.({ expanded: !isExpanded });
               }
+
+              if (collapsed) {
+                setOverlayPreview(true);
```
diyorbek, line 190: opens preview when group item is tapped in collapsed state.

packages/gestalt/src/SideNavigationGroup.js
```diff
@@ -235,9 +227,10 @@ export default function SideNavigationGroup({
             display={display}
             primaryAction={primaryAction}
             setCompression={setCompression}
+            hasActiveChild={hasAnyActiveChild}
           />
         )}
-        {isExpanded ? (
+        {!collapsed && isExpanded ? (
```
diyorbek, line 233: Hide nested items in collapsed state

packages/gestalt/src/SideNavigationSection.js
```diff
@@ -21,18 +29,54 @@ type Props = {
...
+  const { collapsed: sideNavigationCollapsed, overlayPreview } = useSideNavigation();
+
+  const navigationChildren = flattenChildrenWithKeys(children);
+
+  validateChildren({ children: navigationChildren, filterLevel: 'main' });
+
+  const collapsed = sideNavigationCollapsed && !overlayPreview;
+
+  const shouldCollapseAsEllipsis =
+    collapsed && countItemsWithIcon(navigationChildren) !== navigationChildren.length;
```
diyorbek, line 41: Sections collapse as ellipses when there is at least one item without an icon.

packages/gestalt/src/contexts/SideNavigationProvider.js
```diff
@@ -33,8 +42,14 @@ type Props = {
     onDismiss: () => void,
     id: string,
   },
+  collapsible?: boolean,
+  collapsed?: boolean,
+  onCollapse?: (boolean) => void,
+  onPreview?: (boolean) => void,
 };
 
+const COLLAPSE_TRANSITION_DURATION = 200; // .2s
```
diyorbek, line 51: We have to somehow keep this value consistent with the transition rule in `SideNavigation.css`

docs/docs-components/AppLayout.js
```diff
@@ -109,11 +109,11 @@ export default function AppLayout({ children, colorScheme }: Props): ReactNode {
           <Box
             display="none"
             mdDisplay="block"
-            position="fixed"
```
AlbertCarreras, line 112: Q: am curious what is this fixing?

packages/gestalt/src/SideNavigation.css
AlbertCarreras: @hectoid eventually we should convert this to time tokens

packages/gestalt/src/SideNavigation.css
AlbertCarreras: Screenshot by Dropbox Capture

packages/gestalt/src/SideNavigation.css
AlbertCarreras: nit Some of these comments could go in the code itself for future immediate reference. 

reading 
`.liItem:not(.section) + .section .sectionStartDivider {` isn't the most obvious so a comment would be nice

docs/pages/web/sidenavigation.js
```diff
@@ -454,6 +459,53 @@ When building SideNavigation, we might want to render different combinations of
...
+              <SandpackExample
+                code={collapsibleWithMixedIconsExample}
+                name="Collapsible mixed icons example"
+                previewHeight={500}
+              />
+            }
+          />
+          <MainSection.Card
+            cardSize="lg"
+            title="Collapsible header"
```
AlbertCarreras, line 498: nit I'd add some descriptions here to make it more relevant. For example, highlight the importance of onPreview for the Collapsible header example. 

Each example has nuances that might be worth highlighting. It can be a follow up PR.

docs/pages/web/sidenavigation.js
```diff
@@ -454,6 +459,53 @@ When building SideNavigation, we might want to render different combinations of
...
+            sandpackExample={
+              <SandpackExample
+                code={collapsibleHeaderExample}
+                name="Collapsible header example"
+                previewHeight={500}
+              />
```
AlbertCarreras, line 504: Can we add an example for mobile? Is any of this supported on mobile? If not, we should describe what;s not supported on the Mobile section.

packages/gestalt/src/SideNavigation.js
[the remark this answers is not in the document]
AlbertCarreras: Another option could be to follow how we usually we've been doing this events, with named args

Screenshot by Dropbox Capture


`onExpand?: ({ expanded: boolean, type: "default" | "preview" }) => void 
`

Said that, I think preview should be uncontrolled as we are delaying the collapse when on preview. If we had to support controlled previews... we would need to differenciate collapsed from previewed right? To add a different classname with the transition.

I think it's fine as it is. If we get requests for controlled we can change it in the future.

I guess the question is do we want the preview behavior as default? If we add a preview state prop that behaves similarly to collapsed then it could not be implemented and just be optional.

What's the actual use case of `onPreview`? Because I think the intention is to communicate state to header and footer right? Could all be just onCollapse (or onExpand if we change it?) Not sure the difference of this prop...

packages/gestalt/src/SideNavigation.js
AlbertCarreras: I think it's ok. But we need to be aware we might run into issues:

Brave Browser - SideNavigation - Gestalt 2024-03-13 at 11 42 14 AM

packages/gestalt/src/SideNavigation/Collapser.js
AlbertCarreras: This is the Collapser 
Screenshot by Dropbox Capture

packages/gestalt/src/SideNavigation/Collapser.js
```diff
@@ -0,0 +1,38 @@
...
+        marginBottom={-2}
+        padding={2}
+        color={raised ? 'default' : undefined}
+        borderStyle={raised ? 'raisedTopShadow' : undefined}
+      >
+        <IconButton
+          icon={collapsed ? 'arrow-double-forward' : 'arrow-double-back'}
+          accessibilityLabel={
+            collapsed
+              ? 'Navigation collapsed. Click button to expand.'
```
AlbertCarreras, line 25: This should be a DefaultLabelProvider string.

hectoid:
very nice @diyorbek ! the collapse button is missing a tooltip on hover. The shortcut key's are not a requirement for v1, but otherwise, [here](https://www.figma.com/file/0GGe3tDQDspiHFOmsiro28/Handoff---SideNav-collapsible?type=design&node-id=80-5632&mode=design&t=HGUC3gB8QYrZ0mrx-11) is the language. Sorry, I didn't specify that a tooltip is needed (as per all our icon buttons) on hover, not just on keyboard nav.

packages/gestalt/src/SideNavigation/GroupContent.js
```diff
@@ -66,11 +53,33 @@ export default function SideNavigationGroupContent({
     dropdownItems?: $ReadOnlyArray<Element<typeof Dropdown.Item>>,
   },
   setCompression: ('compress' | 'none') => void,
-}): ReactNode {
+  hasActiveChild?: boolean,
```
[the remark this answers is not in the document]
AlbertCarreras: Screenshot by Dropbox Capture

packages/gestalt/src/SideNavigation/GroupContent.js
AlbertCarreras: Screenshot by Dropbox Capture

packages/gestalt/src/SideNavigation/ItemsEllipsis.js
```diff
@@ -0,0 +1,59 @@
+// @flow strict
+import { type Node as ReactNode } from 'react';
+import Box from '../Box';
+import { useSideNavigation } from '../contexts/SideNavigationProvider';
+import Icon from '../Icon';
+import { type Props as TopItemProps } from '../SideNavigationTopItem';
+import TapArea from '../TapArea';
+
+export type Props = {
+  ...Pick<TopItemProps, 'active' | 'notificationAccessibilityLabel'>,
```
AlbertCarreras, line 10: Screenshot by Dropbox Capture

packages/gestalt/src/SideNavigation/ItemsEllipsis.js
```diff
@@ -0,0 +1,59 @@
...
+  accessibilityLabel?: string,
+};
+
+export default function ItemsEllipsis({
+  active,
+  accessibilityLabel = 'Collapsed navigation items. Expand for more options',
```
AlbertCarreras, line 16: DefaultLabelProvider here.

packages/gestalt/src/SideNavigation/ItemsEllipsis.js
```diff
@@ -0,0 +1,59 @@
...
+  notificationAccessibilityLabel,
+}: Props): ReactNode {
+  const { setOverlayPreview } = useSideNavigation();
+
+  const itemColor = active === 'page' ? 'selected' : undefined;
```
AlbertCarreras, line 21: nit, can we inline this? it's easier to read inlined unless reused.

packages/gestalt/src/SideNavigation/ItemsEllipsis.js
```diff
@@ -0,0 +1,59 @@
...
+  const textColor = active === 'page' ? 'inverse' : 'default';
```
AlbertCarreras, line 22: iconColor?

packages/gestalt/src/SideNavigation/ItemsEllipsis.js
```diff
@@ -0,0 +1,59 @@
...
+            width={8}
+            rounding="circle"
+            color="primary"
+            role="status"
+            position="absolute"
+            dangerouslySetInlineStyle={{ __style: { top: 4, right: 4 } }}
+          />
+        ) : null}
+
+        <Box>
```
AlbertCarreras, line 53: Do we need this empty Box?

packages/gestalt/src/icons/index.js
```diff
@@ -300,6 +302,8 @@ const icons = Object.freeze({
   'arrow-circle-forward': arrowCircleForward,
   'arrow-circle-up': arrowCircleUp,
   'arrow-counter-clockwise': arrowCounterClockwise,
+  'arrow-double-forward': arrowDoubleForward,
```
AlbertCarreras, line 305: OK, if we are adding icons 

Can we do this on a separate PR first?

We need to also update 

1. TS file for icons
2. We need to also add it to the Android folder.
3. We need to update the visual snapshot for icons.

I think a separate PR would be nice for this.

AlbertCarreras:
We need to update the TS file.

packages/gestalt/src/SideNavigation.css
AlbertCarreras, line 22: Neat. Follow up explorations: can we make consistent the slide in/out transitions across components?

import AnimationProvider from './animation/AnimationContext';

packages/gestalt/src/SideNavigation/NavigationContent.js
```diff
@@ -0,0 +1,149 @@
...
+  const {
+    collapsible,
+    collapsed: sideNavigationCollapsed,
+    onCollapse,
+    overlayPreview,
+    setOverlayPreview,
+    transitioning,
+  } = useSideNavigation();
+
+  const scrollContainer = useRef<HTMLDivElement | null>(null);
```
AlbertCarreras, line 45: Follow up exploration: can we make consistent shadows in scrolling? See Modal.  Are we considering here resizing not just scrolling?

Screenshot by Dropbox Capture

packages/gestalt/src/SideNavigation/NavigationContent.js
```diff
@@ -0,0 +1,149 @@
...
+  const [isScrolled, setIsScrolled] = useState(false);
+  const previewTimeoutRef = useRef<?Timeout>();
+
+  useEffect(() => {
+    const element = scrollContainer.current;
+    const scrollHandler = () => setIsScrolled(!!element?.scrollTop);
+
+    const mouseEnterHandler = () => {
+      if (sideNavigationCollapsed && !transitioning) {
+        clearTimeout(previewTimeoutRef.current);
```
AlbertCarreras, line 55: clearTimeout here is not really doing much right?
 

Screenshot by Dropbox Capture

packages/gestalt/src/SideNavigation/NavigationContent.js
```diff
@@ -0,0 +1,149 @@
...
+        setOverlayPreview(true);
+      }
+    };
+
+    const mouseLeaveHandler = () => {
+      if (sideNavigationCollapsed) {
```
AlbertCarreras, line 61: sideNavigationCollapsed if it's collapsed... how is that we are closing it?

packages/gestalt/src/SideNavigation/NavigationContent.js
```diff
@@ -0,0 +1,149 @@
...
+  return (
+    <Box
+      minWidth={collapsible ? undefined : normalWidth}
+      width={collapsible ? wrapperWidth : undefined}
+      height="100%"
+      as="nav"
+      aria-label={accessibilityLabel}
+      color="default"
+      position="relative"
+      zIndex={overlayPreview ? new FixedZIndex(1) : undefined}
```
AlbertCarreras, line 103: can we foresee any   zIndex issues if it's just 1? In the examples might work but what if we have some elements below?

packages/gestalt/src/SideNavigation/getChildrenToArray.js
```diff
@@ -1,19 +1,16 @@
...
 const getChildrenToArray = ({
   children,
   filterLevel,
 }: {
   children: ReactNode,
   filterLevel: 'main' | 'nested',
-  // $FlowFixMe[unclear-type] ALBERTO TO FIX FLOW TYPE HERE
-}): $ReadOnlyArray<any> => {
+}): ReactChildArray => {
```
AlbertCarreras, line 13: Thanks!

packages/gestalt/src/SideNavigation/navigationChildrenUtils.js
AlbertCarreras: can we add comments in the code so it's easier to read the code?

docs/docs-components/AppLayout.js
diyorbek: Not a fix but improvement. I started testing in docs navigation first and If we ever wanted it to be collapsible, fixed position wouldn't work. Because sidenav should push the main page content when its width changes. it won't behave like that if is a fixed element.

Also It's inside a flexbox but fixed position makes it non-flex item.

docs/pages/web/sidenavigation.js
diyorbek: I added the note in the props 😬. Can we make a follow up PR for this?

packages/gestalt/src/SideNavigation/navigationChildrenUtils.js
```diff
@@ -0,0 +1,113 @@
...
+      throw new Error(
+        `Gestalt ${child.type.displayName} cannot be used at ${
+          isTopLevel ? 'the top' : 'a nested'
+        } level`,
+      );
+    }
+  });
+}
+
+export function countItemsWithIcon(children: ReactChildArray): number {
```
AlbertCarreras, line 33: why do we need this function?

packages/gestalt/src/SideNavigation/ItemsEllipsis.js
diyorbek: I copied them from TopItem 😄

packages/gestalt/src/SideNavigationTopItem.js
```diff
@@ -176,30 +183,58 @@ const SideNavigationTopItemWithForwardRef: AbstractComponent<Props, HTMLLIElemen
...
+                  __style: {
+                    paddingInlineStart: nestingMargin,
+                    paddingInlineEnd: '16px',
+                  },
+                }
+          }
         >
-          <Flex gap={{ row: 2, column: 0 }} height="100%" width="100%">
+          {collapsed && icon && notificationAccessibilityLabel ? (
+            <Box
```
AlbertCarreras, line 212: I wonder if we could dry this into a Subcomponents <NotificationDot />

packages/gestalt/src/utils/flattenChildren.js
```diff
@@ -0,0 +1,45 @@
...
+    } else if (typeof node === 'string' || typeof node === 'number') {
+      // $FlowFixMe[incompatible-call]
+      acc.push(node);
+    }
+
+    return acc;
+  }, []);
+}
+
+export default function flattenChildren(children: ReactNode): ReactChildArray {
```
AlbertCarreras, line 38: when do we use flattenChildren vs  flattenChildrenWithKeys?

packages/gestalt/src/SideNavigation/navigationChildrenUtils.js
AlbertCarreras: nit can you add a description to each helper function?

packages/gestalt/src/SideNavigation/NavigationContent.js
```diff
@@ -0,0 +1,149 @@
...
+type Props = { ...SideNavigationProps };
+type Timeout = ReturnType<typeof setTimeout>;
+
+export default function NavigationContent({
+  accessibilityLabel,
+  children,
+  footer,
+  header,
+  showBorder,
+}: Props): ReactNode {
```
AlbertCarreras, line 31: why not just 

`}: SideNavigationProps): ReactNode {
`

?

`type Props = { ...SideNavigationProps };
`seems redundant, no?

AlbertCarreras (approved):
LGTM. It works well. 

Added some API suggestions that if you  want to accept/change should not be in followup  PRs as then we'll need codemod and breaking change.

The new icons also should preceed as well for simplicity.

But it works and there's good refactors. The code is organized, and structured. Good new patterns that we can explore to see if we can replace other existing ones more convoluted. 

It's hard to 100% review all the code line by line and get what's happening on each line, but overall looks good and the tests work.

I think we can revisit and try to dry more but as is works so good for now.

We should add tests and maybe one visual screenshot

packages/gestalt/src/SideNavigation/ItemsEllipsis.js
diyorbek: Yes. I added it so that icon doesn't get squished or smaller when ellipsis has border and border shrinks the content width.

packages/gestalt/src/SideNavigation/NavigationContent.js
diyorbek: good question. I will make a follow up PR for this.

packages/gestalt/src/SideNavigation/NavigationContent.js
diyorbek: `sideNavigationCollapsed` means it is in collapsed state even when it is expanded for preview. Preview and collapsed are two different things. Preview state cannot happen unless the sidenav is collapsed.
Here the logic applied only when sidenav is collapsed but the timeout is related to preview state.

packages/gestalt/src/SideNavigation/NavigationContent.js
diyorbek: this applies the whole sidenav. there is an issue of getting behind some conent which has higher z-index. But as long as the sidenav is correctly positioned, It will be displayed correctly.

I was really hesitant to use z-index here but we have no choice. We don't have a good stacking context management.

packages/gestalt/src/SideNavigation/navigationChildrenUtils.js
diyorbek: sections or items collapse based on icons. sidenav collapses as empty when there is no icons. that's why we need to count icons.

packages/gestalt/src/SideNavigationTopItem.js
diyorbek: we can but not now. please😄

packages/gestalt/src/utils/flattenChildren.js
```diff
@@ -0,0 +1,45 @@
...
+import { Children, cloneElement, Fragment, isValidElement, type Node as ReactNode } from 'react';
+
+export type ReactChildArray = ReturnType<typeof Children.toArray>;
+
+// Function is copied from "react-keyed-flatten-children" and modified for simplicity
+// https://github.com/grrowl/react-keyed-flatten-children/blob/master/index.ts
+export function flattenChildrenWithKeys(
+  children: ReactNode,
+  depth: number = 0,
+  keys: $ReadOnlyArray<string | number> = [],
```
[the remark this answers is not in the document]
diyorbek: sure

packages/gestalt/src/utils/flattenChildren.js
diyorbek: if we don't render we use flattenChildren. If we have to render then we need keys, so we use `flattenChildrenWithKeys`. You can see the usages in `navigationChildrenUtils.js`

packages/gestalt/src/SideNavigation/NavigationContent.js
diyorbek: Sorry, I think I had some more additional props there but forgot to simplify after removing them.

packages/gestalt/src/SideNavigation.js
diyorbek: @AlbertCarreras We should be opinonated on this. 
The above components have expanded state because they can be exapanded. By default they are NOT expanded. So it makes sense to control that behavior with `expanded` prop. Because expanded state is a change relative to its default state.

SideNavigation, on the other hand, is not an expandible component. It is in expanded state by default. But it can BECOME collapsible. So it makes sense to control that with `collapsed` prop. Because collapsed state is a change relative to its default state.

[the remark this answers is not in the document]
diyorbek: @AlbertCarreras Preview is not only for communicating for the header and footer. Sorry if my comment misled you. Besides the header and footer, there are ellipsis items with contain multiple items, group items which need to be expanded when clicked. Those require sidenav to be in its normal width so they are displayed properly. That's why we need preview state.
I would suggest you to check out the Figma handoff to better understand the preview state.

diyorbek:
@AlbertCarreras thanks for the review. I appreciate the help! Especially the screenshots! I wanted to touch less code as possible. But ended up with some dry code. SideNavigation inherently has some repeated logic in its subcomponents. We should definetely come back to this and make it as compact as possible.

packages/gestalt/src/SideNavigation.js
AlbertCarreras: Didn't thought of it that way.  MMM Based on that ... when enabling this feature, isn't the default state "collapsed", then we expanded when interacting. But I guess you mean that SideNavigation needs to be expanded as a default in order to be able to work. But aint the same as Accordion? Accordion default state should be expanded...

haha

idk. 

whatever you decide.

docs/pages/web/sidenavigation.js
diyorbek: descriptions are added

packages/gestalt/src/SideNavigation.css
diyorbek: done

packages/gestalt/src/SideNavigation.js
diyorbek: > when enabling this feature, isn't the default state "collapsed"

That actually my point. You would enable this feature by adding `collapsed` prop and sidenav will appear as collapsed. :D

packages/gestalt/src/SideNavigation/NavigationContent.js
diyorbek: @AlbertCarreras No need to follow up. Because SideNavigation is not responsive. It has fixed 280px width when it's collapsible. So resizing doesn't cause content to shift.

packages/gestalt/src/SideNavigation/NavigationContent.js
diyorbek: mouse can enter while sidenav is in preview mode. when that happens, we should remove the timer, so it doesn't close while the mouse is still on sidenavigation.

packages/gestalt/src/contexts/SideNavigationProvider.js
rlingineni: Would this make sense to put in our tokens library and import it from there?

packages/gestalt/src/utils/flattenChildren.js
```diff
@@ -0,0 +1,45 @@
+// @flow strict
+import { Children, cloneElement, Fragment, isValidElement, type Node as ReactNode } from 'react';
+
+export type ReactChildArray = ReturnType<typeof Children.toArray>;
+
+// Function is copied from "react-keyed-flatten-children" and modified for simplicity
+// https://github.com/grrowl/react-keyed-flatten-children/blob/master/index.ts
+export function flattenChildrenWithKeys(
```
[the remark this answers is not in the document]
rlingineni: I don't follow what flattening children means?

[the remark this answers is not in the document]
rlingineni: Can we re-render the children with a new list of elements

PR #3478, opened 2024-03-11
Enable use of heights with 0 on two column masonry logic

manvoid:
### Summary

Enable use of heights with 0 on two column masonry logic

#### What changed?

...

#### Why?

...

### DIFF
--- a/packages/gestalt/src/Masonry/defaultTwoColumnModuleLayout.js
+++ b/packages/gestalt/src/Masonry/defaultTwoColumnModuleLayout.js
@@ -25,7 +25,7 @@ function getPositionsOnly<T>(
 function getAdjacentColumnHeightDeltas(heights: $ReadOnlyArray<number>): $ReadOnlyArray<number> {
   return heights.reduce((acc, height, index) => {
     const adjacentColumnHeight = heights[index + 1];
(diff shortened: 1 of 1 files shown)


liuyenwei (approved):

packages/gestalt/src/Masonry/defaultTwoColumnModuleLayout.js
```diff
@@ -29,7 +29,7 @@ function getPositionsOnly<T>(
 function getAdjacentColumnHeightDeltas(heights: $ReadOnlyArray<number>): $ReadOnlyArray<number> {
   return heights.reduce((acc, height, index) => {
     const adjacentColumnHeight = heights[index + 1];
-    if (adjacentColumnHeight) {
+    if (adjacentColumnHeight >= 0) {
```
willsmanley, line 32: why do we need this check exactly? could adjacentColumnHeight really be negative?

manvoid: No negative but it can be `undefined` because we are looking the current height + 1, the check is only for that but it was also filtering the `0` value. We could add other types of checks but we want the height to be a positive including 0.

PR #3482, opened 2024-03-14
Masonry: Adds a data attribute for two column module on ssr

manvoid:
### Summary

To enable the use of a two column module on Masonry using SSR we need to:

- Identify on CSS and early scripts if an item is two column. This PR adds a new attribute to the ssr structure called `data-column-span` that can hold the column span value of the item.
- Add logic to set the correct witdh for n column modules when rendered for mearurements
- Change the logic of the split index on large batches so tcm can be moved to the first position when applicable

### DIFF
--- a/packages/gestalt/src/Masonry.js
+++ b/packages/gestalt/src/Masonry.js
@@ -541,6 +541,8 @@ export default class Masonry<T: { ... }> extends ReactComponent<Props<T>, State<
             <div // keep this in sync with renderMasonryComponent
               className="static"
               data-grid-item
+              // $FlowFixMe[prop-missing] We're assuming `columnSpan` exists
+              data-column-span={item.columnSpan}
               // eslint-disable-next-line react/no-array-index-key
               key={i}
               ref={(el) => {
--- a/packages/gestalt/src/Masonry.js
+++ b/packages/gestalt/src/Masonry.js
@@ -482,10 +482,6 @@ export default class Masonry<T: { +[string]: mixed }> extends ReactComponent<Pro
     } = this.props;
     const { hasPendingMeasurements, measurementStore, width } = this.state;
     const { positionStore } = this;
-    const twoColumnWidth =
-      typeof columnWidth === 'number' && typeof gutter === 'number'
-        ? columnWidth * 2 + gutter
-        : columnWidth;
 
     let getPositions;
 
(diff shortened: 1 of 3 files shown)


packages/gestalt/src/Masonry.js
```diff
@@ -541,6 +541,8 @@ export default class Masonry<T: { ... }> extends ReactComponent<Props<T>, State<
             <div // keep this in sync with renderMasonryComponent
               className="static"
               data-grid-item
+              // $FlowFixMe[prop-missing] We're assuming `columnSpan` exists
+              data-column-span={item.columnSpan}
```
liuyenwei, line 545: nit: `item.columnSpan ?? 1`

also, could we just update the flow type for item?

packages/gestalt/src/Masonry.js
```diff
@@ -560,8 +565,10 @@ export default class Masonry<T: { ... }> extends ReactComponent<Props<T>, State<
                 WebkitTransform: 'translateX(0px) translateY(0px)',
                 width:
                   layout === 'flexible' || layout === 'serverRenderedFlexible'
-                    ? undefined
-                    : layoutNumberToCssDimension(columnWidth), // we can't set a width for server rendered flexible items
+                    ? undefined // we can't set a width for server rendered flexible items
+                    : layoutNumberToCssDimension(
+                        item.columnSpan === 2 ? twoColumnWidth : columnWidth,
```
liuyenwei, line 570: what about just
```
columnWidth * (item.columnSpan ?? 1)
```

packages/gestalt/src/Masonry/defaultTwoColumnModuleLayout.js
```diff
@@ -234,7 +251,9 @@ const defaultTwoColumnModuleLayout = <T>({
 
   return (items): $ReadOnlyArray<Position> => {
     if (isNil(width) || !items.every((item) => measurementCache.has(item))) {
-      return items.map(() => offscreen(columnWidth));
+      return items.map((item) =>
+        offscreen(item.columnSpan === 2 ? columnWidth * 2 + gutter : columnWidth),
```
liuyenwei, line 255: same comment here

liuyenwei (approved):

PR #3486, opened 2024-03-20
PopoverEducational: fix zIndex bug + add example

AlbertCarreras:
PopoverEducational: fix zIndex bug + add example

Zindex was not properly working. Zindex was not being applied to absolute positioned Contents component. Removed unnecessary Box.

BEFORE
Brave Browser - PopoverEducational - Gestalt 2024-03-19 at 11 13 18 PM

AFTER
Brave Browser - PopoverEducational - Gestalt 2024-03-19 at 11 14 24 PM

### DIFF
--- a/packages/gestalt/src/PopoverEducational.js
+++ b/packages/gestalt/src/PopoverEducational.js
@@ -170,35 +170,34 @@ export default function PopoverEducational({
 
   if (!isInExperiment)
     return (
-      <Box position={zIndex ? 'relative' : undefined} zIndex={zIndex}>
-        <LegacyInternalPopover
-          accessibilityLabel={accessibilityLabel}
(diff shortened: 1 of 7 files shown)


packages/gestalt/src/PopoverEducational.js
```diff
@@ -170,35 +170,34 @@ export default function PopoverEducational({
 
   if (!isInExperiment)
     return (
-      <Box zIndex={zIndex} position={zIndex ? 'relative' : undefined}>
```
AlbertCarreras, line 173: Moving zIndex from this Box to Contents component with position absolute.

<Box zIndex={zIndex} position={zIndex ? 'relative' : undefined}>

willsmanley (approved):

PR #3488, opened 2024-03-20
Masonry: Render tcm in position on first line

manvoid:
### Summary

Our current implementation on ssr for the two column modules position the item on the index defined on the array, f.e. if tcm index is 3 and it fits on the first row then it is rendered on that index. But this is not the case when doing a resizing, since instead of having a large batch we have the batches of 6 items and the tcm is sent to the first index of the batch always for the first line.

Since this two behaviors are different we can see that the position of the tcm changes greatly when doing resizing.

This PR adds a check to see if the heights are 0 and the tcm fits on the first line then it creates a batch starting on the tcm and skips the graph logic since it is not used, this way we can keep the position for the item on the first row.

### DIFF
--- a/packages/gestalt/src/Masonry/defaultTwoColumnModuleLayout.js
+++ b/packages/gestalt/src/Masonry/defaultTwoColumnModuleLayout.js
@@ -40,10 +40,10 @@ function calculateTwoColumnModuleWidth(columnWidth: number, gutter: number): num
   return columnWidth * 2 + gutter;
 }
 
-function calculateSplitIndex(itemsWithoutPositionsLength: number, twoColumnIndex: number): ?number {
+function calculateSplitIndex(itemsWithoutPositionsLength: number, twoColumnIndex: number): number {
   // If the items length is the same as the batch size we don't set a split index
   if (itemsWithoutPositionsLength <= TWO_COL_ITEMS_MEASURE_BATCH_SIZE) {
-    return null;
+    return 0;
   }
 
   return twoColumnIndex + TWO_COL_ITEMS_MEASURE_BATCH_SIZE > itemsWithoutPositionsLength
@@ -307,7 +307,7 @@ const defaultTwoColumnModuleLayout = <T: { +[string]: mixed }>({
       const splitIndex = skipGraph
         ? twoColumnIndex
         : calculateSplitIndex(itemsWithoutPositions.length, twoColumnIndex);
-      const pre = splitIndex ? itemsWithoutPositions.slice(0, splitIndex) : [];
+      const pre = itemsWithoutPositions.slice(0, splitIndex);
       const batchWithTwoColumnItems = splitIndex
         ? itemsWithoutPositions.slice(splitIndex, splitIndex + TWO_COL_ITEMS_MEASURE_BATCH_SIZE)
         : itemsWithoutPositions;
--- a/packages/gestalt/src/Masonry/defaultTwoColumnModuleLayout.js
+++ b/packages/gestalt/src/Masonry/defaultTwoColumnModuleLayout.js
@@ -40,23 +40,6 @@ function calculateTwoColumnModuleWidth(columnWidth: number, gutter: number): num
   return columnWidth * 2 + gutter;
 }
 
-function calculateSplitIndex<T: { +[string]: mixed }>(
-  itemsWithoutPositions: $ReadOnlyArray<T>,
-): number {
-  // Currently we only support one two column item at the same time, more items will be supporped soon
-  const twoColumnIndex = itemsWithoutPositions.findIndex((item) => item.columnSpan === 2);
-
-  if (twoColumnIndex < TWO_COL_ITEMS_MEASURE_BATCH_SIZE) {
-    return 0;
-  }
-
-  if (twoColumnIndex + TWO_COL_ITEMS_MEASURE_BATCH_SIZE > itemsWithoutPositions.length) {
-    return itemsWithoutPositions.length - TWO_COL_ITEMS_MEASURE_BATCH_SIZE;
-  }
-
-  return twoColumnIndex;
-}
-
 function initializeHeightsArray<T>({
   centerOffset,
   columnCount,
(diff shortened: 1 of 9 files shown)


packages/gestalt/src/Masonry/defaultTwoColumnModuleLayout.js
```diff
@@ -302,8 +302,17 @@ const defaultTwoColumnModuleLayout = <T: { +[string]: mixed }>({
     if (hasTwoColumnItems) {
       // If the number of items to position is greater that the batch size
       // we identify the batch with the two column item and apply the graph only to those items
-      const splitIndex = calculateSplitIndex(itemsWithoutPositions);
-      const shouldBatchItems = itemsWithoutPositions.length > TWO_COL_ITEMS_MEASURE_BATCH_SIZE;
+      const twoColumnIndex = itemsWithoutPositions.indexOf(twoColumnItems[0]);
+
+      const skipGraph =
+        heights.every((height) => height === 0) && twoColumnIndex < heights.length - 1;
+      const shouldBatchItems =
```
liuyenwei, line 309: if skipGraph is true do we still need to batch items? or is this just to keep the code simple since it doesn't really matter anyways

packages/gestalt/src/Masonry/defaultTwoColumnModuleLayout.js
```diff
@@ -302,8 +302,17 @@ const defaultTwoColumnModuleLayout = <T: { +[string]: mixed }>({
     if (hasTwoColumnItems) {
       // If the number of items to position is greater that the batch size
       // we identify the batch with the two column item and apply the graph only to those items
-      const splitIndex = calculateSplitIndex(itemsWithoutPositions);
-      const shouldBatchItems = itemsWithoutPositions.length > TWO_COL_ITEMS_MEASURE_BATCH_SIZE;
+      const twoColumnIndex = itemsWithoutPositions.indexOf(twoColumnItems[0]);
+
+      const skipGraph =
+        heights.every((height) => height === 0) && twoColumnIndex < heights.length - 1;
```
liuyenwei, line 308: more of an optimization but i don't think height specifically needs to be 0 here (e.g. we could say something like skip graph and just place where it is if the deviation of the heights in the array is less than the size of the gutter)

okay as a followup though. also, i don't think we need the check for twoColumnIndex < heights.lenght - 1?

packages/gestalt/src/Masonry/defaultTwoColumnModuleLayout.js
```diff
@@ -402,15 +411,17 @@ const defaultTwoColumnModuleLayout = <T: { +[string]: mixed }>({
...
 
       // For each unpainted item, start generating possible layouts
-      oneColumnItems.forEach((item, i, arr) => {
-        addPossibleLayout({
-          item,
-          i,
-          arr,
-          heightsArr: paintedItemHeights,
-          prevNode: startNodeData,
+      if (!skipGraph) {
```
liuyenwei, line 414: so just clarifying, we're still creating the graph with what's been painted so far but just not adding any layouts?

packages/gestalt/src/Masonry/defaultTwoColumnModuleLayout.js
manvoid: Yeah, I was thinking on how to apply the rule on a more general way but this is more accurate on what we want to solve at the moment. I can create a ticket to follow this as an algorithm improvement.

packages/gestalt/src/Masonry/defaultTwoColumnModuleLayout.js
manvoid: Exactly, its not really needed but we would execute the same steps as when there is a batch, I did it like this so we are not repeating code but that is another option on how to handle this.

packages/gestalt/src/Masonry/defaultTwoColumnModuleLayout.js
manvoid: Yeah, we create the graph and add the first node that is no layout. This is the one used as winning node and that puts the tcm right after painted items.

packages/gestalt/src/Masonry/defaultTwoColumnModuleLayout.js
liuyenwei: okay that sounds good - can we leave a comment here so we don't lose track of this? also, can you confirm we should be able to remove the `twoColumnIndex < heights.lenght - 1`

packages/gestalt/src/Masonry/defaultTwoColumnModuleLayout.js
liuyenwei: sg, can you leave a comment here around this?

liuyenwei (approved):

packages/gestalt/src/Masonry/defaultTwoColumnModuleLayout.js
manvoid: We can remove it, the side effect is that the item is going to be positioned as the next item always without calculating a better position, tbh that check doesn't solve that either so we could remove it. We would have to implement something to better position on that case.

packages/gestalt/src/Masonry/defaultTwoColumnModuleLayout.test.js
```diff
@@ -388,7 +388,7 @@ describe('two column layout test cases', () => {
     const positionCache = new MeasurementStore<{ ... }, Position>();
     const heightsCache = new HeightsStore();
     const items = [
-      { 'name': 'Pin 0', 'height': 200, 'color': '#E230BA', 'columnSpan': 2 },
+      { 'name': 'Pin 0', 'height': 200, 'color': '#E230BA' },
```
liuyenwei, line 391: why was this changed?

manvoid: To test more positions, the same structure as the previous test, I'm testing when tcm is on index 0 and on index 3

packages/gestalt/src/Masonry/defaultTwoColumnModuleLayout.js
```diff
@@ -40,6 +40,17 @@ function calculateTwoColumnModuleWidth(columnWidth: number, gutter: number): num
   return columnWidth * 2 + gutter;
 }
 
+function calculateSplitIndex(itemsWithoutPositionsLength: number, twoColumnIndex: number): ?number {
+  // If the items length is the same as the batch size we don't set a split index
+  if (itemsWithoutPositionsLength <= TWO_COL_ITEMS_MEASURE_BATCH_SIZE) {
+    return null;
```
liuyenwei, line 46: nit: just return 0 here so return type can always be a number. no need to return null since 0 is also falsey

packages/gestalt/src/Masonry/defaultTwoColumnModuleLayout.js
```diff
@@ -283,20 +294,21 @@ const defaultTwoColumnModuleLayout = <T: { +[string]: mixed }>({
...
-      const splitIndex =
-        twoColumnIndex + TWO_COL_ITEMS_MEASURE_BATCH_SIZE > itemsWithoutPositions.length
-          ? itemsWithoutPositions.length - TWO_COL_ITEMS_MEASURE_BATCH_SIZE
-          : twoColumnIndex;
-      const pre = shouldBatchItems ? itemsWithoutPositions.slice(0, splitIndex) : [];
-      const batchWithTwoColumnItems = shouldBatchItems
+      const splitIndex = skipGraph
+        ? twoColumnIndex
+        : calculateSplitIndex(itemsWithoutPositions.length, twoColumnIndex);
+      const pre = splitIndex ? itemsWithoutPositions.slice(0, splitIndex) : [];
```
liuyenwei, line 310: if you change above to return 0, this can also just be `itemsWithoutPositions.slice(0, splitIndex)` since split(0,0) is the same as []

liuyenwei (approved):

PR #3490, opened 2024-03-21
TileData: Fixing width Overlap issues

rlingineni:
### Summary

TagData was not behaving well with multiple elements wrapping in a low. The title was overflowing the parent container, and it was causing overlap.

image

image

It seems like this was caused due to our break word change here:
https://github.com/pinterest/gestalt/pull/3136

This seems to work fine in the original [TagData](https://deploy-preview-2844--gestalt.netlify.app/web/tiledata?devexamples=true) PR.

Original TagData:
image

After:
image

Note the `word-break` change.  We swapped line clamps `word-break: break-word` in favor of `word-wrap: break-word`. 

This fixes it by adding an explicit max-width to the text-area. This may be the [issue](https://stackoverflow.com/questions/66724509/overflow-wrap-break-word-vs-word-break-break-word)

135 as a maxWidth is picked because it is ~70% and it seems to be a max-safe-area for the title without overflowing. We may choose to adjust this value.

#### What changed?

Setting a maxWidth on the TileData title

### Links

- [TDD](link to Paper doc)
- [Figma](link to Figma file)

### Checklist


AlbertCarreras (approved):
Thanks

PR #3491, opened 2024-03-21
Masonry: Fill in first row when multi-column item can't fit

liuyenwei:
### Summary
The approach we're taking to server render a multi-column item in the first row is:
- If the multi-column item can fit in the first row, render it in place
- If the multi-column item cannot fit in the first row, hide it and rendering single column modules in its place

For the second scenario, we currently cannot guarantee that the single column item which end up filling in the first row are the same as the server rendered layout. This PR addresses this by having the `defaultTwoColumnLayout` take this case into consideration and apply the following logic:
- Split items into two batches: `pre`( items before the multi column module), `batchWithTwoColumnItems` (multi-column item + other items)
- For first row only (e.g. 0 heights), if the multi column module cannot fit in the first row, iterate through `batchWithTwoColumnItems` and move enough single column items into `pre` to fill out the first row.

### DIFF
--- a/packages/gestalt/src/Masonry/defaultTwoColumnModuleLayout.js
+++ b/packages/gestalt/src/Masonry/defaultTwoColumnModuleLayout.js
@@ -296,11 +296,11 @@ const defaultTwoColumnModuleLayout = <T: { +[string]: mixed }>({
     if (hasTwoColumnItems) {
       // Currently we only support one two column item at the same time, more items will be supporped soon
       const twoColumnIndex = itemsWithoutPositions.indexOf(twoColumnItems[0]);
+      const isFirstRow = heights.every((height) => height === 0);
 
       // Skip the graph logic if the two column item batch is on the first line
       const skipGraph =
-        heights.every((height) => height === 0) &&
-        itemsWithoutPositions.length <= TWO_COL_ITEMS_MEASURE_BATCH_SIZE;
+        isFirstRow && itemsWithoutPositions.length <= TWO_COL_ITEMS_MEASURE_BATCH_SIZE;
 
       // If the number of items to position is greater that the batch size
       // we identify the batch with the two column item and apply the graph only to those items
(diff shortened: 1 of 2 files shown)


packages/gestalt/src/Masonry/defaultTwoColumnModuleLayout.js
```diff
@@ -296,11 +296,11 @@ const defaultTwoColumnModuleLayout = <T: { +[string]: mixed }>({
     if (hasTwoColumnItems) {
       // Currently we only support one two column item at the same time, more items will be supporped soon
       const twoColumnIndex = itemsWithoutPositions.indexOf(twoColumnItems[0]);
+      const isFirstRow = heights.every((height) => height === 0);
 
       // Skip the graph logic if the two column item batch is on the first line
       const skipGraph =
-        heights.every((height) => height === 0) &&
-        itemsWithoutPositions.length <= TWO_COL_ITEMS_MEASURE_BATCH_SIZE;
+        isFirstRow && itemsWithoutPositions.length <= TWO_COL_ITEMS_MEASURE_BATCH_SIZE;
```
liuyenwei, line 303: unrelated to this PR but @manvoid this logic doesn't entirely seem correct. shouldn't this instead be something like
```
const skipGraph = isFirstRow && canFitOnFirstRow?
```

manvoid: Yes, I need to change that, in fact I have another bug if the two column index is on the second batch that means that not all the heights are 0 because the first batch was already positioned, but the two column item is still going to be rendered on the first line. We would need to change the `isFirstRow` to only check if any height is 0 and then check if the tcm can fit there as you mention.

packages/gestalt/src/Masonry/defaultTwoColumnModuleLayout.js
```diff
@@ -312,6 +312,25 @@ const defaultTwoColumnModuleLayout = <T: { +[string]: mixed }>({
...
+          // if we're in the first row but we can't render the multi column item in the first row, fill pre with single column items to ensure we have enough to cover the initial row
+          // this is specifically for the first row to ensure we match server rendering styling
+          for (let i = 0; i < remainingColumnsToFill; i += 1) {
+            const singleColumnItemIndex = batchWithTwoColumnItems.findIndex(
+              (item) => !item.columnSpan || item.columnSpan === 1,
+            );
+            if (singleColumnItemIndex > -1) {
+              // if we find a single column module, remove it from the two column batch and add it to the pre array
+              pre.push(batchWithTwoColumnItems[singleColumnItemIndex]);
+              batchWithTwoColumnItems.splice(singleColumnItemIndex, 1);
```
manvoid, line 328: This is effectively reducing the number of elements on `batchWithTwoColumnItems` so the two column module will have less elements to calculate the optimal position.

liuyenwei: ah good callout. let me update

packages/gestalt/src/Masonry/defaultTwoColumnModuleLayout.js
liuyenwei: i can update this line as part of this PR if you want

PR #3492, opened 2024-03-21
Masonry: Positions tcm correctly for last item on first row

manvoid:
### Summary

When we render the items of the first row if there is a two column item on the last spot of the row we have to replace it with another one column item, we can handle that on ssr but when hydration happens this item can swap places with other item since it is part of the graph of the two column item.

This PR handles this taking advantage of the concepts of pre items (items to be rendered before the two column module) and introducing a new array `graphBatch` this two represent what comes before the two column item and what is used to create the graph to position the two column item.

We can change the `splitIndex` so we can have more items on the `pre` depending on if the two column module fits or not and have a consistent experience with ssr. 

#### Important notes

So the original position of the two column item don't affect this calculations the pr removes it from the batch array, because of this the items on the batch array are in fact one less than before that's why I changed the `TWO_COL_ITEMS_MEASURE_BATCH_SIZE` constant to 5.

### DIFF
--- a/packages/gestalt/src/Masonry/defaultTwoColumnModuleLayout.js
+++ b/packages/gestalt/src/Masonry/defaultTwoColumnModuleLayout.js
@@ -40,11 +40,17 @@ function calculateTwoColumnModuleWidth(columnWidth: number, gutter: number): num
   return columnWidth * 2 + gutter;
 }
 
-function calculateSplitIndex(itemsWithoutPositionsLength: number, twoColumnIndex: number): number {
+function calculateSplitIndex(
+  itemsWithoutPositionsLength: number,
+  twoColumnIndex: number,
+  isFirstRow: ?number,
+  fitsFirstRow: ?number,
+): number {
   // If the items length is the same as the batch size we don't set a split index
-  if (itemsWithoutPositionsLength <= TWO_COL_ITEMS_MEASURE_BATCH_SIZE) {
-    return 0;
-  }
+  // if (itemsWithoutPositionsLength <= TWO_COL_ITEMS_MEASURE_BATCH_SIZE - 1) {
+
+  //   return 0;
+  // }
 
   return twoColumnIndex + TWO_COL_ITEMS_MEASURE_BATCH_SIZE > itemsWithoutPositionsLength
     ? itemsWithoutPositionsLength - TWO_COL_ITEMS_MEASURE_BATCH_SIZE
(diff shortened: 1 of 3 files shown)


packages/gestalt/src/Masonry/defaultTwoColumnModuleLayout.js
```diff
@@ -7,7 +7,7 @@ import { type NodeData, type Position } from './types';
 
 // When there's a 2-col item in the most recently fetched batch of items, we need to measure more items to ensure we have enough possible layouts to minimize whitespace above the 2-col item
 // This may need to be tweaked to balance the tradeoff of delayed rendering vs having enough possible layouts
-export const TWO_COL_ITEMS_MEASURE_BATCH_SIZE = 6;
+export const TWO_COL_ITEMS_MEASURE_BATCH_SIZE = 5;
```
liuyenwei, line 10: why are we changing the batch size in this PR?

packages/gestalt/src/Masonry/defaultTwoColumnModuleLayout.js
```diff
@@ -40,15 +40,29 @@ function calculateTwoColumnModuleWidth(columnWidth: number, gutter: number): num
...
-function calculateSplitIndex(itemsWithoutPositionsLength: number, twoColumnIndex: number): number {
+function calculateSplitIndex(
+  oneColumnItemsLength: number,
+  twoColumnIndex: number,
+  isFirstRow: ?boolean,
+  fitsFirstRow: ?boolean,
+): number {
+  // We add one more item to pre so it is positioned instead of two column item
+  if (isFirstRow && !fitsFirstRow) {
+    return twoColumnIndex + 1;
```
liuyenwei, line 51: let's pass in availableSlotsOnFirstRow so we make this dynamic
```
if (isFirstRow && !fitsFirstRow) {
  return twoColumnIndex + availableSlotsOnFirstRow;
}
```

packages/gestalt/src/Masonry/defaultTwoColumnModuleLayout.js
```diff
@@ -40,15 +40,29 @@ function calculateTwoColumnModuleWidth(columnWidth: number, gutter: number): num
...
 
-function calculateSplitIndex(itemsWithoutPositionsLength: number, twoColumnIndex: number): number {
+function calculateSplitIndex(
+  oneColumnItemsLength: number,
+  twoColumnIndex: number,
+  isFirstRow: ?boolean,
+  fitsFirstRow: ?boolean,
+): number {
+  // We add one more item to pre so it is positioned instead of two column item
+  if (isFirstRow && !fitsFirstRow) {
```
liuyenwei, line 50: i don't think this is right actually. this will be true even if the twoColumnItem is not eligible to be in the first row

should this be, instead, something like
```
if (isFirstRow && twoColumnIndex < columnCount && !fitsFirstRow)
```

tested locally and that seems to result in the tests passing

packages/gestalt/src/Masonry/defaultTwoColumnModuleLayout.js
manvoid: The correct logic would be:

```
if (isFirstRow && twoColumnIndex < columnCount && !fitsFirstRow) {
  return availableSlotsOnFirstRow;
}
```

since we want the split to be exactly the number of elements we want to insert on the first row

liuyenwei (approved):

PR #3496, opened 2024-03-25
Masonry: Multi column feature, fix first row items position when batch is small

manvoid:
### Summary

When the position for items on the first row are calculated with a multi column item and the batch is small the correct postition was not set correctly because we returned `0` as split index although we have to maintain al least the position of the elements on the first row.

To fix this intead of returning 0 we return the number of available slots on first row to fill them correctly. This means that the multi column item will have less items to calculate the optimal position but in my testing is not that bad and we need to keep the items on the first row.

#### Testing

Unit test added

#### Other changes

- We were incorrectly passing batches of only 5 elements to the layout
- We can change `columnCount` with `availableSlotsOnFirstRow` to calculate split index since that is more accuarate

### DIFF
--- a/packages/gestalt/src/Masonry/defaultTwoColumnModuleLayout.js
+++ b/packages/gestalt/src/Masonry/defaultTwoColumnModuleLayout.js
@@ -334,14 +334,16 @@ const defaultTwoColumnModuleLayout = <T: { +[string]: mixed }>({
 
       // Calculate how many items are on pre array and how many on graphBatch
       // pre items are positioned before the two column item
-      const splitIndex = calculateSplitIndex({
-        oneColumnItemsLength: oneColumnItems.length,
-        twoColumnIndex,
-        emptyColumns,
-        replaceWithOneColItems,
-      });
+      const splitIndex = skipGraph
+        ? twoColumnIndex
+        : calculateSplitIndex({
+            oneColumnItemsLength: oneColumnItems.length,
+            twoColumnIndex,
+            emptyColumns,
+            replaceWithOneColItems,
+          });
 
-      const pre = oneColumnItems.slice(0, skipGraph ? twoColumnIndex : splitIndex);
+      const pre = oneColumnItems.slice(0, splitIndex);
       const graphBatch = skipGraph
         ? []
         : oneColumnItems.slice(splitIndex, splitIndex + TWO_COL_ITEMS_MEASURE_BATCH_SIZE);
--- a/packages/gestalt/src/Masonry/defaultTwoColumnModuleLayout.js
+++ b/packages/gestalt/src/Masonry/defaultTwoColumnModuleLayout.js
@@ -44,11 +44,17 @@ function calculateSplitIndex({
   oneColumnItemsLength,
   twoColumnIndex,
   emptyColumns,
+  replaceWithOneColItems,
 }: {
   oneColumnItemsLength: number,
   twoColumnIndex: number,
   emptyColumns: number,
+  replaceWithOneColItems: boolean,
 }): number {
+  if (replaceWithOneColItems) {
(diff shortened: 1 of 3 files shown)


packages/gestalt/src/Masonry/defaultTwoColumnModuleLayout.js
```diff
@@ -46,29 +46,26 @@ function calculateSplitIndex({
...
   oneColumnItemsLength: number,
   twoColumnIndex: number,
   isFirstRow: boolean,
   fitsFirstRow: boolean,
   availableSlotsOnFirstRow: number,
-  columnCount: number,
 }): number {
   // We add one more item to pre so it is positioned instead of two column item
-  if (isFirstRow && twoColumnIndex < columnCount && !fitsFirstRow) {
+  if (isFirstRow && twoColumnIndex < availableSlotsOnFirstRow && !fitsFirstRow) {
```
liuyenwei, line 57: slightly confused here.  `availableSlotsOnFirstRow` should be the number of columns that we have available to fit a module in right? wouldn't it make sense to still compare `index` to total count? 

e.g.if we have 6 column grid with two columns available in the first row and the two column module is in index 3, we would be returning 2 here as the split index

packages/gestalt/src/Masonry/defaultTwoColumnModuleLayout.js
```diff
@@ -323,34 +360,50 @@ const defaultTwoColumnModuleLayout = <T: { +[string]: mixed }>({
...
+      const emptyColumns = heights.reduce((acc, height) => (height === 0 ? acc + 1 : acc), 0);
+
+      const multiColumnItemColumnSpan = parseInt(twoColumnItems[0].columnSpan, 10);
+      const fitsFirstRow = emptyColumns >= multiColumnItemColumnSpan + twoColumnIndex;
 
-      // Skip the graph logic if the two column item batch is on the first line and fits
-      const skipGraph = isFirstRow && fitsFirstRow;
+      // Skip the graph logic if the two column item can be displayed on the first row
+      if (fitsFirstRow) {
+        const { positions: finalPositions, heights: finalHeights } = getSkipGraphPositions({
```
liuyenwei, line 372: i like that we're being explicit about the skipGraph case but i'm not sure this is better than the previous/current approach of just having an empty batch for the graph since this ultimately adds a bit more duplicated logic (e.g. updating the position and heights caches, getting the one column item positions, etc)

this is also something we'll likely need to tune a bit more in the future so rather than anchoring on `fitsFirstRow` I think keeping the current logic as-is makes more sense

packages/gestalt/src/Masonry/defaultTwoColumnModuleLayout.js
```diff
@@ -323,34 +360,50 @@ const defaultTwoColumnModuleLayout = <T: { +[string]: mixed }>({
...
-        (acc, height) => (height === 0 ? acc + 1 : acc),
-        0,
-      );
-      const fitsFirstRow = availableSlotsOnFirstRow >= multiColumnItemWidth + twoColumnIndex;
+      // The empty columns can be different from columnCount if there are
+      // items already positioned from previous batches
+      const emptyColumns = heights.reduce((acc, height) => (height === 0 ? acc + 1 : acc), 0);
+
+      const multiColumnItemColumnSpan = parseInt(twoColumnItems[0].columnSpan, 10);
+      const fitsFirstRow = emptyColumns >= multiColumnItemColumnSpan + twoColumnIndex;
```
liuyenwei, line 368: this makes more sense thank you - i honestly think if we just applied ^ changes to the last commit, that would be enough/more clear already

liuyenwei (approved):

PR #3505, opened 2024-04-02
Masonry: Experimental Masonry Refactor

liuyenwei:
## Summary
This PR introduces an experimental `MasonryV2` component which is an attempt to rewrite `Masonry` and bring it up to date with modern React best practices. 

## Why
`Masonry` is a fairly convoluted component and can be difficult to reason about, especially for engineers who are unfamiliar with it. There are various lifecycles and out of band updates (e.g. rAF, scroll, resize events) which make it difficult to understand what actually happens in a render.

The last time we seriously took a look at `Masonry` was 2018, when we shipped a similar rewrite. Since then, much has changed with React and, unfortunately, not much has changed with `Masonry`.  My goals with this PR are to make Masonry easier to reason about and also incorporate newer React features/best practices. 

## Changes
There are quite a few changes, but I'll highlight the most important ones

### Function vs Class component
The most noticeable change is that `MasonryV2` is now a function component vs a class component. When doing this refactor, I prioritized a few things:
- less state
- less effects/"lifecycles"
- fewer renders

Where it made sense, I also broke out functionality into hooks (e.g. `useLayout`, `useViewport`, etc)

### Rendering the `gridBody`
The current Masonry has three conditions in its render function:
- no width + hasPendingMeasurements (serverRender or hydration)
- no width (remount)
- width available( full layout possible)

Because of this, we actually render a different dom structure between the initial hydration pass and the subsequent full layout. This PR addresses this by consolidating the render logic so we always return the same DOM structure, but tweaked accordingly depending on the condition (e.g. styles are different between server render + full layout). This helps to simplify things but also ensure consistency between renders

### Measurement vs Layout
The current Masonry has two "sections" - one div for rendering items to measure, one div for rendering positioned items. The general flow is:
- render item to measure inside the "measuring" div
- measure item in ref callback
- during the next render pass, position the item and render it in the "layout" div

The downside of this approach is that it forces us to unmount + remount the item. This PR changes the approach by always rendering `items` in the order they're received. For items that haven't yet been measured, we first position them "offscreen" and then, on the subsequent render, we update the position accordingly based on the layout. The benefit here is that we no longer have to unmount + remount an item but instead, can just treat it as a re-render. 

### useTransition vs rAF
The current Masonry performs an out-of-band update by repeatedly scheduling a setState instead a requestAnimationFrame. The general idea here was to reduce scroll jank by scheduling rendering inside of an animation frame. 

With [concurrent rendering](https://react.dev/blog/2022/03/29/react-v18#what-is-concurrent-react) this is now something that is possible directly through React. This PR updates the approach and removes the requestAnimationFrame in favor of [useTransition](https://react.dev/reference/react/useTransition) instead. 

## Test Plan
Lots of testing both on the masonry test harness and in Pinboard
- Integration tests pass
- Server side rendering works
- Desktop + mobile web works
- Homefeed, search, Pin, Board all work as expected

### DIFF
--- a/packages/gestalt/src/MasonryV2.js
+++ b/packages/gestalt/src/MasonryV2.js
@@ -1,0 +1,544 @@
+// @flow strict
+import {
+  Component as ReactComponent,
+  forwardRef,
+  type Node as ReactNode,
+  Suspense,
+  useCallback,
+  useEffect,
+  useImperativeHandle,
+  useLayoutEffect,
+  useMemo,
+  useReducer,
+  useRef,
+  useState,
+  useSyncExternalStore,
+  useTransition,
+} from 'react';
+import debounce, { type DebounceReturn } from './debounce';
+import FetchItems from './FetchItems';
+import styles from './Masonry.css';
+import { type Cache } from './Masonry/Cache';
+import defaultLayout from './Masonry/defaultLayout';
+import defaultTwoColumnModuleLayout from './Masonry/defaultTwoColumnModuleLayout';
+import fullWidthLayout from './Masonry/fullWidthLayout';
+import HeightsStore, { type HeightsStoreInterface } from './Masonry/HeightsStore';
+import MeasurementStore from './Masonry/MeasurementStore';
+import ScrollContainer from './Masonry/ScrollContainer';
+import { getElementHeight, getRelativeScrollTop, getScrollPos } from './Masonry/scrollUtils';
+import { type Position } from './Masonry/types';
+import uniformRowLayout from './Masonry/uniformRowLayout';
+import throttle, { type ThrottleReturn } from './throttle';
+
+const RESIZE_DEBOUNCE = 300;
+
+// When there's a 2-col item in the most recently fetched batch of items, we need to measure more items to ensure we have enough possible layouts to minimize whitespace above the 2-col item
+// This may need to be tweaked to balance the tradeoff of delayed rendering vs having enough possible layouts
+const TWO_COL_ITEMS_MEASURE_BATCH_SIZE = 6;
+
+const layoutNumberToCssDimension = (n: ?number) => (n !== Infinity ? n : undefined);
+
+type Layout = 'basic' | 'basicCentered' | 'flexible' | 'serverRenderedFlexible' | 'uniformRow';
+
+type Props<T> = {
+  /**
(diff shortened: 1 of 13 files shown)


packages/gestalt/src/MasonryV2.js
```diff
@@ -0,0 +1,654 @@
...
+  }, [gridWrapper, scrollContainer]);
+
+  return {
+    containerHeight,
+    containerOffset,
+    scrollTop,
+  };
+}
+
+function useFetchOnScroll({
```
liuyenwei, line 210: FYI: this functionality mostly lives in the `FetchItems` component in current Masonry

packages/gestalt/src/MasonryV2.js
```diff
@@ -0,0 +1,654 @@
...
+
+    startTransition(() => {
+      forceUpdate();
+    });
+  };
+
+  // these are all for backwards compatibility with the old Masonry
+  // will work on removing these once this lands
+  useImperativeHandle(ref, () => ({
+    handleResize: () => reflow(),
```
liuyenwei, line 480: there are some use cases in pinboard which currently reference the Masonry class instance - these were the ones I found, so currently exposing on the ref for backwards compatibility. will lock down what's exposed on the ref on the Pinboard side and also remove these whenever this ships

packages/gestalt/src/MasonryV2.js
```diff
@@ -0,0 +1,654 @@
...
+  const hasPendingMeasurements = itemMeasurementsCount < items.length;
+  const canPerformLayout = width != null;
+
+  const positions: $ReadOnlyArray<?Position> = useMemo(() => {
+    if (!canPerformLayout) {
+      return [];
+    }
+    // we currently calculate positions separately for items that have been measured and items that haven't
+    // maintain this paradigm for now for parity but ideally we should just able to pass all items to the layout function
+    const itemsWithMeasurements = items.filter((item) => measurementStore.has(item));
```
liuyenwei, line 323: i initially tried just passing all items into the layout functions (vs separating out items w/ measurements vs items w/out measurements) and there were some cases where weird things happened (Specifically with multi column).

Breaking this apart for now but would like to revisit this later. we should be able to just give the layout function all items and it should properly handle the two cases internally

manvoid (approved):

willsmanley (approved):

PR #3527, opened 2024-04-11
Masonry: Implement graph early bailout feature

manvoid:
### Summary

Tho determine the best position for the two column item the current behavior creates all the possible combinations of items on the graphBatch and then looks for the one with the minimum whitespace, although this works is very inneficient since we could have a good enough answer. To improve it this PR implements a new feature to exit early the algorithm when a certain whitespace threshold is found. 

#### Tests

Manual tests showed a really awesome improvement using a threshold of gutter*2 px, reducing the number of iterationr from 326 to ~4* (limited manual tests)

Unit test added

#### Notes

- To enable this a `whitespaceThreshold` must be set, but this PR doesn't set one, that would be on another pr to determine the best threshold
- The future of this feature is very related to the implementation of a dynamic batch size and used together to create a better positioning
- The process to track metrics with the improvements will come on a future PR

### DIFF
--- a/packages/gestalt/src/Masonry/defaultTwoColumnModuleLayout.js
+++ b/packages/gestalt/src/Masonry/defaultTwoColumnModuleLayout.js
@@ -240,6 +240,137 @@ function getTwoColItemPosition<T>({
   };
 }
 
+function getGraphPositions<T>({
+  items,
+  positions,
+  heights,
+  whitespaceThreshold,
+  ...commonGetPositionArgs
+}: {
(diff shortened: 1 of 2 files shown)


willsmanley (approved):

rlingineni:
@manvoid - this just caught my eye.

It was fun to read - what's the difference between good enough vs best?

Also, to run through all 324 combinations, could the operation of pre-sorting be done server-side to be faster?

PR #3530, opened 2024-04-16
Masonry: Enable n-column modules on first row

manvoid:
### Summary

This is the first PR that enables an initial support for n-modules on masonry (3+ columnSpan). This PR enables the positioning of the multi colmn items on the first row, this does't require any change on the graph logic, mainly this changes the logic on the following functions:

- `getAdjacentColumnHeightDeltas`: If two columns returns the deltas between columns, if multi column returns the averages of the deltas required to position the multi column item, f.e. a 3 column item will return the avg of two deltas. This number is the used to determine the best position.
- `getMultiColItemPosition`: Updated to consider that a delta can imply more than two colums, to determine the tallest column it creates a split with the heights for each possible placement
- `calculateMultiColumnModuleWidth`: Updated to return multi column width
- Chage var and function names from `twoCol` to `multiCol`

#### Notes

- This only enables multi column if it fits on first row, multi column on any other behavior is not supported
- All the two colum features are the same
- The flags to enable two col modules are the same

#### Tests

A unit test was added, the visual result of the added tests are:

### DIFF
--- a/packages/gestalt/src/Masonry.js
+++ b/packages/gestalt/src/Masonry.js
@@ -6,7 +6,7 @@ import styles from './Masonry.css';
 import { type Cache } from './Masonry/Cache';
 import defaultLayout from './Masonry/defaultLayout';
 import defaultTwoColumnModuleLayout, {
-  TWO_COL_ITEMS_MEASURE_BATCH_SIZE,
+  MULTI_COL_ITEMS_MEASURE_BATCH_SIZE,
 } from './Masonry/defaultTwoColumnModuleLayout';
 import fullWidthLayout from './Masonry/fullWidthLayout';
 import HeightsStore, { type HeightsStoreInterface } from './Masonry/HeightsStore';
@@ -576,13 +576,16 @@ export default class Masonry<T: { +[string]: mixed }> extends ReactComponent<Pro
       // Full layout is possible
       const itemsToRender = items.filter((item) => item && measurementStore.has(item));
       const itemsWithoutPositions = items.filter((item) => item && !positionStore.has(item));
-      const hasTwoColumnItems =
-        _twoColItems && itemsWithoutPositions.some((item) => item.columnSpan === 2);
+      const hasMultiColumnItems =
+        _twoColItems &&
+        itemsWithoutPositions.some(
+          (item) => typeof item.columnSpan === 'number' && item.columnSpan > 2,
+        );
 
       // If there are 2-col items, we need to measure more items to ensure we have enough possible layouts to find a suitable one
       // we need the batch size (number of one column items for the graph) + 1 (two column item)
-      const itemsToMeasureCount = hasTwoColumnItems
-        ? TWO_COL_ITEMS_MEASURE_BATCH_SIZE + 1
+      const itemsToMeasureCount = hasMultiColumnItems
+        ? MULTI_COL_ITEMS_MEASURE_BATCH_SIZE + 1
         : minCols;
       const itemsToMeasure = items
         .filter((item) => item && !measurementStore.has(item))
--- a/packages/gestalt/src/Masonry/defaultTwoColumnModuleLayout.js
+++ b/packages/gestalt/src/Masonry/defaultTwoColumnModuleLayout.js
@@ -7,7 +7,7 @@ import { type NodeData, type Position } from './types';
 
 // When there's a 2-col item in the most recently fetched batch of items, we need to measure more items to ensure we have enough possible layouts to minimize whitespace above the 2-col item
 // This may need to be tweaked to balance the tradeoff of delayed rendering vs having enough possible layouts
-export const TWO_COL_ITEMS_MEASURE_BATCH_SIZE = 5;
+export const MULTI_COL_ITEMS_MEASURE_BATCH_SIZE = 5;
 
 function isNil(value: mixed): boolean %checks {
   return value === null || value === undefined;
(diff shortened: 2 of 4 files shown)


packages/gestalt/src/Masonry.js
```diff
@@ -576,13 +576,16 @@ export default class Masonry<T: { +[string]: mixed }> extends ReactComponent<Pro
       // Full layout is possible
       const itemsToRender = items.filter((item) => item && measurementStore.has(item));
       const itemsWithoutPositions = items.filter((item) => item && !positionStore.has(item));
-      const hasTwoColumnItems =
-        _twoColItems && itemsWithoutPositions.some((item) => item.columnSpan === 2);
+      const hasMultiColumnItems =
+        _twoColItems &&
+        itemsWithoutPositions.some(
+          (item) => typeof item.columnSpan === 'number' && item.columnSpan > 2,
```
willsmanley, line 582: why is this not >= 2?

packages/gestalt/src/Masonry/defaultTwoColumnModuleLayout.js
```diff
@@ -26,36 +26,58 @@ function getPositionsOnly<T>(
...
+  heights: $ReadOnlyArray<number>,
+  columnSpan: number,
+): $ReadOnlyArray<number> {
+  const adjacentHeightDeltas = heights.reduce((acc: $ReadOnlyArray<number>, height, index) => {
     const adjacentColumnHeight = heights[index + 1];
     if (adjacentColumnHeight >= 0) {
       return [...acc, Math.abs(height - adjacentColumnHeight)];
     }
     return acc;
   }, []);
```
willsmanley, line 39: unrelated to this PR but this is an opportunity for a performance improvement:

```
const adjacentHeightDeltas = [];
for (let i = 0; i < heights.length - 1; i++) {
  adjacentHeightDeltas.push(Math.abs(heights[i] - heights[i + 1]));
}
```

packages/gestalt/src/Masonry/defaultTwoColumnModuleLayout.js
```diff
@@ -26,36 +26,58 @@ function getPositionsOnly<T>(
...
+      });
+      return [...acc, sum / (columnSpan - 1)];
+    }
+    return acc;
+  }, []);
 }
 
-function calculateTwoColumnModuleWidth(columnWidth: number, gutter: number): number {
-  return columnWidth * 2 + gutter;
+function calculateMultiColumnModuleWidth(
```
willsmanley, line 57: unrelated to this PR but also it might be good to test to see if memoizing this function might help to avoid recalculations. not urgent.

packages/gestalt/src/Masonry/defaultTwoColumnModuleLayout.js
```diff
@@ -209,32 +234,38 @@ function getTwoColItemPosition<T>({
...
   const left = lowestAdjacentColumnHeightDeltaIndex * columnWidthAndGutter + centerOffset;
 
   // Increase the heights of both adjacent columns
-  heights[tallestColumn] += heightAndGutter;
-  heights[leftIsTaller ? tallestColumn + 1 : tallestColumn - 1] = heights[tallestColumn];
+  const tallestColumnFinalHeight = heights[tallestColumn] + heightAndGutter;
+
+  Array.from({ length: columnSpan }).forEach((_, index) => {
+    heights[index + lowestAdjacentColumnHeightDeltaIndex] = tallestColumnFinalHeight;
+  });
```
willsmanley, line 260: another perf improvement opportunity
```
for (let i = 0; i < columnSpan; i++) {
  heights[i + lowestAdjacentColumnHeightDeltaIndex] = tallestColumnFinalHeight;
}
```

packages/gestalt/src/Masonry/defaultTwoColumnModuleLayout.js
```diff
@@ -245,12 +276,14 @@ function getGraphPositions<T>({
   positions,
   heights,
   whitespaceThreshold,
+  columnSpan,
   ...commonGetPositionArgs
 }: {
```
willsmanley, line 281: another performance improvement: instead of using deep recursion, we can do the same thing with a queue. that way there is less overhead in the stack so we can do faster computations:
```
function getGraphPositions<T>({
  items,
  positions,
  heights,
  whitespaceThreshold,
  columnSpan,
  ...commonGetPositionArgs
}: {
  items: Array<T>,
  heights: Array<number>,
  positions: Array<{ item: T, position: Position }>,
  whitespaceThreshold?: number,
  columnSpan: number,
  centerOffset: number,
  columnWidth: number,
  columnWidthAndGutter: number,
  gutter: number,
  measurementCache: Cache<T, number>,
  positionCache?: Cache<T, Position>,
}): {
  winningNode: NodeData<T>,
  additionalWhitespace: number,
} {
  const graph = new Graph<T>();
  const queue = [{ nodeData: { id: 'start', heights, positions }, index: 0 }];
  let bestNode = { nodeData: queue[0].nodeData, score: Infinity };

  while (queue.length > 0) {
    const { nodeData, index } = queue.shift();
    const currentPositions = nodeData.positions;
    const currentHeights = nodeData.heights;

    for (let i = index; i < items.length; i++) {
      const item = items[i];
      const itemHeight = measurementCache.get(item);
      const { top, left, updatedHeights } = updateColumnHeightsAndPositions({
        centerOffset: commonGetPositionArgs.centerOffset,
        columnWidthAndGutter: commonGetPositionArgs.columnWidthAndGutter,
        gutter: commonGetPositionArgs.gutter,
        heights: currentHeights,
        columnStartIndex: mindex(currentHeights), // mindex finds the minimum index
        columnSpan: 1, // For single-column items
        itemHeight,
      });

      const newPosition = { item, position: { top, left, width: commonGetPositionArgs.columnWidth, height: itemHeight } };
      const newPositions = [...currentPositions, newPosition];
      const newNodeData = { id: item, heights: updatedHeights, positions: newPositions };
      graph.addNode(newNodeData);
      graph.addEdge(nodeData, newNodeData, Math.min(...updatedHeights));

      if (whitespaceThreshold && Math.min(...updatedHeights) < whitespaceThreshold) {
        if (Math.min(...updatedHeights) < bestNode.score) {
          bestNode = { nodeData: newNodeData, score: Math.min(...updatedHeights) };
        }
        break; // Break early if a good enough layout is found
      }

      queue.push({ nodeData: newNodeData, index: i + 1 });
    }
  }

  return bestNode.nodeData.positions.length > 0 ? { winningNode: bestNode.nodeData, additionalWhitespace: bestNode.score } : { winningNode: graph.findLowestScoreNode('start'), additionalWhitespace: Infinity };
}

```

willsmanley (approved):

packages/gestalt/src/Masonry.js
manvoid: Good catch, thanks

packages/gestalt/src/Masonry/defaultTwoColumnModuleLayout.js
manvoid: We can verify that later

packages/gestalt/src/Masonry/defaultTwoColumnModuleLayout.js
manvoid: Interesting, I could try this but since its a bigger change I would like to test it independently

PR #3541, opened 2024-04-29
Masonry: Bug fix, enable layout not returning positions of all items

manvoid:
### Summary

We found a bug after enabling multiple multi column items on masonry here https://github.com/pinterest/gestalt/pull/3537, this was because masonry expects that the layout functions return all the positions in one run, but for multiple multi column items we do not calculate the position of the items after the second multi column item. This broke masonry although strangely enough the bug was not consistent, that's why it was not immediately obvious. The bug was different on V1 and on V2.

On V1 there was an exception because it tried to render something without a position, to fix it we filter the items with positions before render.

On V2 the layout function was not triggered again after positioning the first multi column item because there was no a condition for that. To fix it we added the multi column items count as dependency.

### DIFF
--- a/packages/gestalt/src/MasonryV2.js
+++ b/packages/gestalt/src/MasonryV2.js
@@ -322,12 +322,13 @@ function useLayout<T: { +[string]: mixed }>({
   positions: $ReadOnlyArray<?Position>,
   updateMeasurement: (T, number) => void,
 } {
-  const hasMultiColumnItems =
-    _twoColItems &&
-    items
-      .filter((item) => item && !positionStore.has(item))
-      .some((item) => typeof item.columnSpan === 'number' && item.columnSpan > 1);
-  const itemToMeasureCount = hasMultiColumnItems ? MULTI_COL_ITEMS_MEASURE_BATCH_SIZE : minCols;
+  const multiColumnItemsCount = _twoColItems
+    ? items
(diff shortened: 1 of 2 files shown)


packages/gestalt/src/MasonryV2.js
```diff
@@ -322,12 +322,13 @@ function useLayout<T: { +[string]: mixed }>({
   positions: $ReadOnlyArray<?Position>,
   updateMeasurement: (T, number) => void,
 } {
-  const hasMultiColumnItems =
-    _twoColItems &&
-    items
-      .filter((item) => item && !positionStore.has(item))
-      .some((item) => typeof item.columnSpan === 'number' && item.columnSpan > 1);
-  const itemToMeasureCount = hasMultiColumnItems ? MULTI_COL_ITEMS_MEASURE_BATCH_SIZE : minCols;
+  const multiColumnItemsCount = _twoColItems
```
liuyenwei, line 325: hmmm this doesn't seem right- in the case where we measure all items and start positioning them, there's actually nothing that would trigger a re-render after the first "batch" has been positioned

is there a reason the layout function still couldn't just return the positions for all items? e.g. split the graph into multiple batches but still return the full set?

manvoid:
We did a refactor of the feature so this PR is no longer required

PR #3544, opened 2024-04-30
Masonry: Remove heights cache

liuyenwei:
## Summary
Currently, we have a `heightCache` that is only used for the `defaultTwoColumnModuleLayout`.  This PR makes two changes around this:
- it removes the height cache completely
- it updates the initial height calculation for the `defaultTwoColumnModuleLayout` to account for multi-column items. Currently it only supports 1-2 column items.

I honestly cannot remember why height cache was originally added when two column support was first deployed but taking a closer, there isn't any reason we specifically need to do this just for the two column layout given given that the calculation is cheap.

More specifically, my motivation for deleting this is because it introduces a strange edge case when we reflow the grid because:
- the measurement cache, position caches are externally cache
- we clear the measurement and position caches, but not the height store.

I initially thought of just clearing the height cache on reflow but after doing some investigation, decided just to remove this to simplify things.


manvoid (commented):
Could you also update the `initializeHeightsArray` function to handle the multi column width correctly? Previously I didn't do it because this code was not triggered on my testing, I was going to ask you about that.

packages/gestalt/src/Masonry/defaultTwoColumnModuleLayout.js
```diff
@@ -117,15 +118,22 @@ function initializeHeightsArray<T>({
...
-      const heightToAdd = position.height + gutter;
-      heights[col] += heightToAdd;
-      if (isTwoColumnModule) {
-        // if position width is greater than columnWidth
-        // increment height of the neighboring column as well
-        heights[col + 1] += heightToAdd;
+      const columnSpan = getItemColumnSpan(item);
+      const moduleHeight = position.height + gutter;
+      if (position.top < heights[col]) {
+        // handles the case where the column height is already taller because of a previously added multi-column module
```
liuyenwei, line 124: this is specifically for the case where we have a multi-column item _before_ a single column

image

in this case, we would skip adding the height for 201 since the column height would already be taller

packages/gestalt/src/Masonry/defaultTwoColumnModuleLayout.js
```diff
@@ -117,15 +118,22 @@ function initializeHeightsArray<T>({
...
+        return;
+      }
+      if (columnSpan > 1) {
+        // find tallest column
+        const colHeights = heights.slice(col, col + columnSpan);
```
liuyenwei, line 129: e.g.
if adding a three column item into col: 3:
- find tallest column between 3, 4, 5
- add module to tallest column
- set heights of other columns to be the same

liuyenwei:
> Could you also update the `initializeHeightsArray` function to handle the multi column width correctly? Previously I didn't do it because this code was not triggered on my testing, I was going to ask you about that.

updated!

manvoid (approved):

packages/gestalt/src/Masonry/defaultTwoColumnModuleLayout.js
manvoid: Nice edge case finding!

manvoid (approved):

willsmanley (approved):

PR #3545, opened 2024-04-30
Enable multiple multi column items2

manvoid:
# Summary

We didn't have support for multiple multi column items on one pass of the layout logic, the impact is that in the case of multiple multi column items only the first one has the correct width since all others are treated as one col items.

Previously I fixed this by only processing the items up to the second multi column item and the layout function only returned those, this was an iddue because masonry expects to receive the positions for all the items sent.

This new approach calculates all the positions by extracting the multi column item logic and using a loop.

## Notes

- I also changed a little bit the `getMultiColItemPosition` to support multiple multi column items on the first row

## Tests

Added unit test for new case, manually tested with pinboard on v1 and v2

### DIFF
--- a/packages/gestalt/src/Masonry/defaultTwoColumnModuleLayout.js
+++ b/packages/gestalt/src/Masonry/defaultTwoColumnModuleLayout.js
@@ -485,13 +485,8 @@ const defaultTwoColumnModuleLayout = <T: { +[string]: mixed }>({
 
     if (hasMultiColumnItems) {
       // Currently we only support one two column item at the same time, more items will be supporped soon
-      const itemsToPosition =
-        multiColumnItems.length > 1
-          ? itemsWithoutPositions.slice(0, itemsWithoutPositions.indexOf(multiColumnItems[1]))
-          : itemsWithoutPositions;
-
-      const multiColumnIndex = itemsToPosition.indexOf(multiColumnItems[0]);
-      const oneColumnItems = itemsToPosition.filter(
+      const multiColumnIndex = itemsWithoutPositions.indexOf(multiColumnItems[0]);
+      const oneColumnItems = itemsWithoutPositions.filter(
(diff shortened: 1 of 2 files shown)


packages/gestalt/src/Masonry/defaultTwoColumnModuleLayout.js
```diff
@@ -483,112 +614,41 @@ const defaultTwoColumnModuleLayout = <T: { +[string]: mixed }>({
...
 
-      // Adding the extra prev column items to the position cache
-      if (paintedItemPositions.length > itemsWithPositions.length) {
-        paintedItemPositions.forEach(({ item, position }) => {
-          positionCache.set(item, position);
-        });
+      let currentPositions = paintedItemPositions;
+      let currentHeights = paintedItemHeights;
+
+      for (let i = 0; i < multiColumnItems.length; i += 1) {
```
liuyenwei, line 628: would it make sense to break these up into batches instead? i think a few things that are confusing here:
- we always filter the full set of items for each iteration of the loop
- _what_ actually gets filtered is dependent on the previous iteration placing more items in the position cache

instead, we could also do:
```
const batchSize = multiColumnItems.length;
const batches = Array.from({ length: batchSize }, () => ([])).map((batch, i) => {
    const startIndex = i === 0 ? 0 : items.indexOf(multiColumnItems[i]);
    const endIndex = i + 1 === multiColumnItems.length ? items.length : items.indexOf(multiColumnItems[i + 1])
    return items.slice(startIndex, endIndex);
});
const { heights: currentHeights, positions: currentPositions } = batches.reduce((acc, itemsToPosition) => {
  return getPositionsWithMultiColumnItem({
    ....
  })
}, { heights:paintedItemHeights, positions: paintedItemPositions });
```

unless i'm mistaken, this should yield the same result since each loop is still using the updated positions and heights from the previous batch. the only difference is that the dependency is a bit clearer (imo)

PR #3552, opened 2024-05-02
Accordion: support borderStyles prop

AlbertCarreras:
Accordion, support borderStyles prop

Screenshot by Dropbox Capture

They supports requests like 
image


hectoid:
this is good @AlbertCarreras ! nice-to-have, but can be a future iteration: border-bottom only for when people stack these. I think that would mean no corner rounding too.

willsmanley (approved):

PR #3555, opened 2024-05-07
Masonry: Fix heights init

liuyenwei:
# Summary
This PR fixes an issue introduced by https://github.com/pinterest/gestalt/pull/3544 that was causing overlapping positions in the `defaultTwoColumnModuleLayout` when a multi column item was rendered towards the end of the first page. 
image
image

The overlap happens in the situation when `initializeHeightsArray` is called and the following is true:
- there is a two column item
- the two column item is positioned _before_ single column items that come after it in the array

i.e.
[A, B, C, D, E, F, G] - assuming E is the multi column item and the layout looks like:

|  A  |  B  | C |
|  D  |  F  | G |
|  _   | E  |  E |

In this scenario, the current logic will think that the column height for column 1, 2 are B + E and C + E respectively. 

# Fix
This PR fixes this scenario by simplifying the logic. Ultimately, the height of each column is determine by the lowest position module in that column so we can change this to just be:
(for each item):
- determine which column the item is positioned in
- calculate its absolute height (top + height + gutter)
- for each column the module spans, update the column height if the absolute height is greater

# Test Plan
Tested this in pinboard - I managed to find an example that consistently repro'd
Also added unit tests around the heights logic using the same repro ^
image

### DIFF
--- a/packages/gestalt/src/Masonry/defaultTwoColumnModuleLayout.test.js
+++ b/packages/gestalt/src/Masonry/defaultTwoColumnModuleLayout.test.js
@@ -859,98 +859,29 @@ describe('initializeHeightsArray', () => {
     const measurementStore = new MeasurementStore<{ ... }, number>();
     const positionCache = new MeasurementStore<{ ... }, Position>();
     const items: $ReadOnlyArray<Item> = [
-      {
-        'name': 'Pin 0',
-        'height': 476,
-      },
-      {
-        'name': 'Pin 1',
-        'height': 381,
-      },
-      {
-        'name': 'Pin 2',
-        'height': 274,
-      },
-      {
-        'name': 'Pin 3',
-        'height': 303,
-      },
-      {
-        'name': 'Pin 4',
-        'height': 475,
-      },
-      {
-        'name': 'Pin 5',
-        'height': 496,
-      },
-      {
-        'name': 'Pin 6',
-        'height': 177,
-      },
-      {
-        'name': 'Pin 7',
-        'height': 440,
-      },
-      {
-        'name': 'Pin 8',
-        'height': 497,
-      },
-      {
-        'name': 'Pin 9',
-        'height': 430,
-      },
-      {
-        'name': 'Pin 10',
-        'height': 409,
-      },
-      {
-        'name': 'Pin 11',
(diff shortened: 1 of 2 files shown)


packages/gestalt/src/Masonry/defaultTwoColumnModuleLayout.test.js
```diff
@@ -849,3 +851,391 @@ describe('multi column layout test cases', () => {
...
 });
+
+describe('initializeHeightsArray', () => {
+  test('correctly determines column heights before laying out new items (default layout)', () => {
+    const gutter = 16;
+    const columnWidth = 236;
+    const measurementStore = new MeasurementStore<{ ... }, number>();
+    const positionCache = new MeasurementStore<{ ... }, Position>();
+    const items: $ReadOnlyArray<Item> = [
+      {
```
manvoid, line 862: nit: Can we move this to be one-liners so its easier to read?

packages/gestalt/src/Masonry/defaultTwoColumnModuleLayout.test.js
```diff
@@ -849,3 +851,391 @@ describe('multi column layout test cases', () => {
...
+        if (!acc[column]) {
+          acc[column] = [];
+        }
+        acc[column].push(position);
+        return acc;
+      }, [])
+      .map((column) => column.sort((a, b) => (a?.top ?? 0) - (b?.top ?? 0)));
+
+    const newTops = newPositionsByColumns.map((column) => column[0].top);
+    expect(newTops).toEqual(heights);
```
manvoid, line 1239: I see that you are verifying that after adding new items the new items start were the previous items ended (the previous heights), just not super sure why this would be necesary, isn't the issue present on the first heightsInitialize? Maybe a little bit of description here could help

liuyenwei: done. lmk if this is clearer. basically what i'm doing here is:
- layout
- validate heights based on initial items are correct
- layout
- validate positions are correct (based on previously validated heights)

manvoid (approved):

AlbertCarreras (approved):

PR #3558, opened 2024-05-09
Masonry: Bug fix, "basic" layout not left aligning 

CalvinChhour:
#### Issue
When layout is set to `basic`, the docs state that the grid will be "Left-aligned, fixed-column-width masonry layout." This wasn't the case, due to `gutterWidth` pushing the items away from the left side. When `gutterWidth` is set to `0`, it is appropriately left aligned, but then you have no gutter spacing between pins. 

Additionally, center alignment was miscalculated due to the inclusion of an extra gutter (e.g. when you have 4 columns, you should only calculate 3 `gutterWidths` for the `contentWidth`)

#### What changed?

I've introduced a new optional prop `justify` which controls the horizontal alignment of items within the Masonry grid.
This change affects both Masonry and MasonryV2

## Center Alignment
Center alignment was slightly off, due to an extra `gutter` being calculated
| Before | After |
|--------|--------|
| Kapture 2024-05-13 at 15 20 07 | Kapture 2024-05-13 at 15 17 52 | 

## Start Alignment
We were center aligning before, we are now properly start aligning when `justify='start'` is set
| Before | After |
|--------|--------|
| Kapture 2024-05-13 at 15 20 07| Kapture 2024-05-13 at 15 22 06|

## End Alignment
End alignment didn't exist previously
| Before | After |
|--------|--------|
| Did not exist | Kapture 2024-05-13 at 15 23 56 | 

#### Why?

This change is needed for situations where we need the Masonry to be left aligned to match other Pin Grid formats (e.g. aligning Modules with PinReps with the Masonry).

This image highlights an issue we have, where the left aligned title looks disjointed from the Masonry grid.
image

### Links

- [Jira](https://jira.pinadmin.com/browse/CLOSEUP-1896)
- [Figma](https://www.figma.com/design/oB4UJiu8pONgAakK1tgPte/%F0%9F%90%B6-dWeb-Closeup-N-Column-Grid?node-id=159-96421&t=Ovx4hMTU2958Ot89-0)

### Checklist

### DIFF
--- a/packages/gestalt/src/Masonry/defaultLayout.js
+++ b/packages/gestalt/src/Masonry/defaultLayout.js
@@ -1,5 +1,5 @@
 // @flow strict
-import { type Justify } from 'gestalt/src//Masonry/types';
+import { type Align, type Layout } from 'gestalt/src//Masonry/types';
 import { type Cache } from './Cache';
 import mindex from './mindex';
 import { type Position } from './types';
@@ -11,19 +11,51 @@ const offscreen = (width: number, height: number = Infinity) => ({
   height,
 });
 
+const calculateCenterOffset = ({
+  columnCount,
+  columnWidthAndGutter,
+  gutter,
+  justify,
+  layout,
+  rawItemCount,
+  width,
+}: {
+  columnCount: number,
+  columnWidthAndGutter: number,
+  gutter: number,
+  justify: Align,
+  layout: Layout,
+  rawItemCount: number,
+  width: number,
+}): number => {
+  if (layout === 'basicCentered') {
+    const contentWidth = Math.min(rawItemCount, columnCount) * columnWidthAndGutter + gutter;
+    return Math.max(Math.floor((width - contentWidth) / 2), 0);
+  }
+  if (justify === 'center') {
+    return Math.max(Math.floor((width - columnWidthAndGutter * columnCount + gutter) / 2), 0);
+  }
+  if (justify === 'end') {
+    return width - (columnWidthAndGutter * columnCount - gutter);
+  }
+  return 0;
+};
+
 const defaultLayout =
   <T>({
+    justify,
     cache,
     columnWidth = 236,
     gutter = 14,
-    justify,
+    layout,
     minCols = 2,
     rawItemCount,
     width,
   }: {
     columnWidth?: number,
     gutter?: number,
-    justify: Justify,
+    justify: Align,
+    layout: Layout,
     cache: Cache<T, number>,
     minCols?: number,
     rawItemCount: number,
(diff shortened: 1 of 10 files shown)


liuyenwei (changes requested):
I think this is a matter of the prop (and documentation) being confusing. We currently assume that Masonry is _always_ "centered" within the container. What the `justify` prop does is control where items start lining up from (https://github.com/pinterest/gestalt/pull/929 is perhaps a more clear example of this).  The confusing bit is that this only impacts the situation where a grid does not contain enough items to fill up the container width. 

image
vs
image

This PR as-is would break quite a few things since it would cause all grids on pinterest to shift left
image

I think one way we can solve this is to introduce  `justify` prop to Masonry, something like `justify: start|center|end` which we can pass in to all layouts and control whether the grid is actually rendered left, center, or right-aligned.  For `basicCentered` - I think we can leave as-is for now,  and revisit separately

packages/gestalt/src/Masonry/defaultLayout.test.js
```diff
@@ -75,27 +75,46 @@ test('centers grid within the viewport', () => {
     width: 8000,
   });
   expect(layout(items)).toEqual([
-    { top: 0, height: 100, left: 7, width: 236 },
```
liuyenwei, line 78: its unexpected to me that these tests are different. this might just be a git thing, but even looking below at the `centers grid within the viewport` test, the positions are very different

packages/gestalt/src/Masonry/defaultLayout.js
```diff
@@ -10,19 +11,51 @@ const offscreen = (width: number, height: number = Infinity) => ({
...
-    justify,
+    layout,
     minCols = 2,
     rawItemCount,
     width,
   }: {
     columnWidth?: number,
     gutter?: number,
-    justify: 'center' | 'start',
+    justify: Align,
```
CalvinChhour, line 57: Let me know if this isn't okay, felt like we could just keep this prop as `justify`. @liuyenwei

jackhsu978:
@CalvinChhour sorry for the short notice, but I am planning on landing https://github.com/pinterest/gestalt/pull/3567 this week for TypeScript migration, and this PR will conflict with that. Can you check with @liuyenwei and determine if you'll be able to merge this PR tomorrow (5/16)? If so, I will wait and apply the same changes to the [TS branch](https://github.com/jackhsu978/gestalt/tree/ts).

packages/gestalt/src/Masonry/defaultLayout.js
liuyenwei: let's keep them consistent

liuyenwei (approved):
one comment about keeping the wording consistent between align vs justify but otherwise lgtm. thank you for making the changes!

PR #3559, opened 2024-05-09
Tokens: Adding Theme VR Web/Semantic Tokens

rlingineni:
This PR adds a set of tokens under each theme.

The outputs are non-breaking, but a few things have been re-organized.

1. All source have been moved from `main-theme` to a `classic` folder
2. `alias-dark/lightMode` files have been renamed to `sema-light/darkMode` (this is just a file name change)
3. Outputs for a theme are now in grouped in folders in dist (e.g. `vrweb/variable.css` instead of `vrweb_variables.css` ). Classic will still output directly, so nothing should break

Things to note in the **VR Theme**:

**Base and Semantic Token Clashes**
`Data-Visualization` naming is used in both base and semantic token (this causes a clash). Maybe we should just define these at the semantic level.  So the semantic tokens are renamed to `data-viz` to avoid this. 
  
**Missing Line Heights**
Line Heights have not been added to the font tokens. There are some pending [comments](https://docs.google.com/spreadsheets/d/1IwQ527sxsfqgspxKKPsWfWvQz5I-v2F4gRdomOgRsxY/edit#gid=1757426051). 

**Rounding Circle Clash**
`rounding-circle` similar to data-viz tokens is both a base and sema token. So ideally, the base token would be named differently.

**Elevations values are incorrect**
This was a direct copy from the spread-sheet, but the values for Elevation may be incorrect and may have to be updated. The names should be as expected.

We may need to format like this to make [elevation values](https://github.com/amzn/style-dictionary/issues/456) more cross platform compatible.

**Missing Negative spacing values**
Once negative spacing values are added, they will have to be added to the VR themes
       
**Next steps to figure out:**

1. How to prefix a token with Base, Sema, Comp.. (each token set is already in a file called `sema.json` or `base.json`). I think we can do it with a transformGroup.

2. Any missing component token stuff as needed

3. Double check the expected values are correct (e.g elevation values)

### Links

- [TDD](link to Paper doc)
- [Figma](link to Figma file)

### Checklist


packages/gestalt-design-tokens/build.js
```diff
@@ -848,17 +857,19 @@ const platformFileMap = {
   ios: ['ios', 'ios-swift', 'ios-swift-separate-enums'],
 };
 
-['main-theme', 'vr-theme'].forEach((theme) =>
+['classic', 'vr-theme', 'vrweb-theme'].forEach((theme) =>
   ['light', 'dark'].forEach((mode) => {
-    if (theme === 'main-theme') {
+    if (theme !== 'vrweb-theme') {
```
rlingineni, line 862: generates all the themes, except vrweb-theme

packages/gestalt-design-tokens/tokens/rounding/vrweb-theme/base.json
```diff
@@ -0,0 +1,37 @@
...
+    "24": {
+      "value": "24px"
+    },
+    "28": {
+      "value": "28px"
+    },
+    "32": {
+      "value": "32px"
+    },
+    "pill": {
```
AlbertCarreras, line 30: this was 999

PR #1769, opened 2021-11-05
Internal: Upgrade to Node.js 16

christianvuerings:
### Summary

Node.js 16 is now the active Node.js version: https://nodejs.org/en/about/releases/

image

Also fixes this exception:
```
error eslint@8.1.0: The engine "node" is incompatible with this module. Expected version "^12.22.0 || ^14.17.0 || >=16.0.0". Got "14.15.4"
error Found incompatible module.
```

PR #3560, opened 2024-05-09
Internal: Upgrade to Node 20 [take 1]

jackhsu978:
## What changed?

Upgrade from Node 16 to Node 20

## Why?

Node 16 is no longer maintained. See [release schedule](https://nodejs.org/en/about/previous-releases).
schedule

Also, [TypeScript ESLint](https://typescript-eslint.io/) requires Node 18, so upgrading is necessary for the TypeScript migration.

## Links

See also [PR](https://github.com/pinterest/gestalt/pull/1769) that upgrades from Node 14 to Node 16


AlbertCarreras (approved):

AlbertCarreras:
@jackhsu978 tests seem to not work.

diyorbek (approved):

PR #3561, opened 2024-05-10
SideNavigation: Use TransitionEvent

diyorbek:
### Summary

Replaced `setTimeout` logic with `transitionend` event. So that the transition duration is independent from JS code.
This also ensures that `transitioning` is set to `true` only when the transition is indeed ended (not when the timer signals so).

#### Why?

So that the transition duration is kept only in the CSS code.

### Links

- [TDD](link to Paper doc)
- [Figma](link to Figma file)

### Checklist


AlbertCarreras:
Screenshot by Dropbox Capture
Transition in  css file. right?

PR #3564, opened 2024-05-11
Masonry: Fix multi-column issues

liuyenwei:
# Summary
This PR addresses two issues I noticed with the multi-column layout, specifically on MasonryV2

## Positioning was inconsistent with V1
In https://github.com/pinterest/gestalt/pull/3530/, we updated the measure batch size for multi-column to be `MULTI_COL_ITEMS_MEASURE_BATCH_SIZE` + 1 for `Masonry` but did not apply the same change to V2. This was causing multi-column positioning to be different between the two

## Incorrect layout on resize
I also noticed that https://github.com/pinterest/gestalt/pull/3545 introduced a scenario where the layout function can return _more_ positions than items, depending on what's already in the position cache. 
image

This is not a big deal for Masonry but it does affect V2 since we assume that if there are no items in the positionCache, we fall back to position by index. This PR addresses the issue by filtering out items that are already in the position cache - this was actually the original intent, but was lost after the post-code review refactor: https://github.com/pinterest/gestalt/pull/3545/files/c5b9ed65188629c036fa5db23af3eb88d1cc727c

image

### DIFF
--- a/packages/gestalt/src/Masonry/defaultTwoColumnModuleLayout.js
+++ b/packages/gestalt/src/Masonry/defaultTwoColumnModuleLayout.js
@@ -637,7 +637,7 @@ const defaultTwoColumnModuleLayout = <T: { +[string]: mixed }>({
         (acc, itemsInBatch, i) =>
           getPositionsWithMultiColumnItem({
             multiColumnItem: multiColumnItems[i],
-            itemsToPosition: itemsInBatch.filter((item) => !positionCache.get(item)),
+            itemsToPosition: itemsInBatch.filter((item) => !positionCache.has(item)),
             heights: acc.heights,
             prevPositions: acc.positions,
             whitespaceThreshold,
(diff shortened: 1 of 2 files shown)


packages/gestalt/src/Masonry/defaultTwoColumnModuleLayout.js
```diff
@@ -634,10 +634,10 @@ const defaultTwoColumnModuleLayout = <T: { +[string]: mixed }>({
         heights: $ReadOnlyArray<number>,
         positions: $ReadOnlyArray<{ item: T, position: Position }>,
       } = batches.reduce(
-        (acc, itemsToPosition, i) =>
+        (acc, itemsInBatch, i) =>
           getPositionsWithMultiColumnItem({
             multiColumnItem: multiColumnItems[i],
-            itemsToPosition,
+            itemsToPosition: itemsInBatch.filter((item) => !positionCache.has(item)),
```
manvoid, line 640: Although this works I think the main problem is that we are not filtering the positioned items when we generate the batches, I think it makes more sense just to change `items` to `itemsWithoutPositions` on the batches logic.

```
      const batchSize = multiColumnItems.length;
      const batches = Array.from({ length: batchSize }, (): $ReadOnlyArray<T> => []).map(
        (batch, i) => {
          const startIndex = i === 0 ? 0 : itemsWithoutPositions.indexOf(multiColumnItems[i]);
          const endIndex =
            i + 1 === multiColumnItems.length
              ? itemsWithoutPositions.length
              : itemsWithoutPositions.indexOf(multiColumnItems[i + 1]);
          return itemsWithoutPositions.slice(startIndex, endIndex);
        },
      );
```

manvoid (approved):

willsmanley (approved):

PR #3567, opened 2024-05-14
Internal: TypeScript Migration

jackhsu978:
This PR migrates from Flow-typed JavaScript code to TypeScript code.

## Breaking Change

We are **dropping the Flow typing support** for Gestalt!

Starting in Gestalt v148, **Flow typings will _not_ be exported**. We will only export TypeScript typings going forward.

### Why migrate codebase to TypeScript?
Because we have decided to make TypeScript our standard language for all Pinterest web projects.

### Why drop Flow support?
Maintaining both TypeScript and Flow typings requires significant amount of manual work. Since our active internal web clients consuming Gestalt have already been migrated from Flow to TypeScript, the additional work is not worth it.

### Which directories are migrated?

- `/docs/`
- `/packages/gestalt/src/`
- `/packages/gestalt-charts/src/`
- `/packages/gestalt-datepicker/src/`

### Why not migrate all directories?

While it's our goal to convert all Flow-typed JavaScript files to TypeScript files, it's low priority to migrate the internal code and potentially removable code (such as Gestalt upgrade codemod).

### Why do we still have `flow-typed` directory?

This is because not all JavaScript code is migrated to TypeScript. Nonetheless, this PR removes most of the Flow library definitions.

### Why change `Element<typeof X>` to `ReactElement`?

In TypeScript, it's not possible to limit a render prop to be a React element of a specific React component. See [this doc](https://react.dev/learn/typescript#typing-children)
> Note, that you cannot use TypeScript to describe that the children are a certain type of JSX elements, so you cannot use the type-system to describe a component which only accepts `<li>` children.

### Why upgrade `date-fns` / `@mui/x-date-picker`?

1. We are using `date-fns` v2 in some areas and v3 in some other areas. Upgrading makes it consistent.
2. TS Typings is wrong in v2.16.1 and is only fixed in v2.17 https://github.com/date-fns/date-fns/pull/2339
3. Upgrading `date-fns` requires `@mui/x-date-picker` to also be upgraded. See https://github.com/mui/mui-x/issues/12144

### Why add `acornInjectPlugins` to rollup config?

See https://www.npmjs.com/package/@rollup/plugin-typescript#preserving-jsx-output

### Why `docs/tsconfig.json` and `docs/next-env.d.ts`?

The two files are actually auto-generated when I run `yarn start` locally. These files should not be manually edited.

### Why check-in `tsMigrate.sh`?

The file isn't actually necessary and can be removed. I am just leaving it in for future reference.


packages/gestalt-datepicker/src/DateField.tsx
```diff
@@ -1,41 +1,8 @@
 import { useEffect } from 'react';
+import { Locale } from 'date-fns/locale';
 import { useGlobalEventsHandler } from 'gestalt';
 import InternalDateField from './DateField/InternalDateField';
 
-// LocaleData type from https://github.com/date-fns/date-fns/blob/81ab18785146405ca2ae28710cdfbb13a294ec50/src/locale/af/index.js.flow
```
AlbertCarreras: this is soooo nice

.flowconfig
```diff
@@ -5,16 +5,14 @@
...
 <PROJECT_ROOT>/node_modules/react-docgen/node_modules
+<PROJECT_ROOT>/node_modules/@rollup/plugin-typescript/node_modules
 
 [options]
 exact_by_default=true
 include_warnings=true
 react.runtime=automatic
 sharedmemory.hash_table_pow=21
 
-[libs]
```
AlbertCarreras, line 15: eye candy commit

jackhsu978: @AlbertCarreras Actually, removing this doesn't mean that we don't use `flow-typed` anymore. In fact, we are still using because small part of the codebase (like eslint linter rules and codemod) still use Flow. It's just that the default value for libs is already `flow-typed` (see docs https://flow.org/en/docs/config/libs/), so we shouldn't need to declare it here.

With that said, I did get to remove majority of the flow-typed files, so it's a big win (even though this line doesn't actually covey that).

AlbertCarreras (approved):
CANT WAIT

PR #3568, opened 2024-05-14
Internal: add temporary files to ignore lists

jackhsu978:
## What changed?

Add temporary files to ignore list:

- `coverage/`
- `.rollup.cache/`
- `*.tsbuildinfo`

## Why?

These files get auto-generated when we run CI commands (or `tsc` for TypeScript type checking) locally, and we should generally ignore these files (when we lint, format, commit, ...etc)

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,3 +1,4 @@
+coverage
 docs/.next
 docs/docs-components/metadata.js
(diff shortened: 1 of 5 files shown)


AlbertCarreras (approved):
thanks

PR #3569, opened 2024-05-14
Internal: update npm module dependencies

jackhsu978:
## What changed?

Install dependencies that will be needed for TypeScript migration, based on [this package.json change](https://github.com/pinterest/gestalt/commit/4796b665d1beba91c73477d647c5c029ae4a8f3a#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519)

## Why?

TypeScript migration


jackhsu978:
Abandoned b/c this was only created for investigation purpose on the failure https://github.com/pinterest/gestalt/actions/runs/9086651086/job/24972788266?pr=3567

The reason for the failure is because our private registry was used instead when I run `yarn add` inside of `docs/`, since `.npmrc` only exists in the root directory, not in `docs/`.

PR #3570, opened 2024-05-15
Internal: Fix some issues caught in TypeScript branch

jackhsu978:
## What changed?

Fix some errors that are caught in the TypeScript branch https://github.com/pinterest/gestalt/pull/3567

Errors include:

- If `<Parent>` uses `<Child>`, `<Child>` should be declared before `<Parent>`. This is violated and is caught by [no-use-before-define](https://eslint.org/docs/latest/rules/no-use-before-define) ESLint rule in TS.
- fix missing a11y props in documentation. This is caught by TS type checking.
- makes a test handle the case where `document.querySelector` returns a`HTMLElement` Node. This is caught by TS type checking.

## Why?

Minimize the scope of https://github.com/pinterest/gestalt/pull/3567


packages/gestalt/src/Masonry.js
```diff
@@ -181,9 +181,7 @@ export default class Masonry<T: { +[string]: mixed }> extends ReactComponent<Pro
 
   positionStore: Cache<T, Position>;
 
-  insertAnimationFrame: AnimationFrameID;
-
-  measureTimeout: TimeoutID;
```
AlbertCarreras, line 186: @jackhsu978 what happened to this `measureTimeout`

AlbertCarreras: Was it dead code?

jackhsu978: yes, it's dead code

PR #3574, opened 2024-05-16
Masonry: Add multi column item support to flexible masonry

manvoid:
### Summary

This PR adds support for multi column items on the masonry flexible layout.

#### How

The logic of the default and full width layouts without the multi column items is in fact really simple and very similar between both of them. The difference is on how we calculate the number of columns, the column width and the offset. With this in mind the multi column code is in fact the same so the approach used in this PR is to extract all the shared logic into a separate function called `multiColumnLayout` (This file is almost the same as the previous `defaulTwoColumnModuleLayout` without some logic) and create the specific layout file that calculates the column count, column width and center offset to then call the `multinColumnLayout` function.

#### Notes

- Unit tests are WIP
- I need to refactor this once the typescript migration is completed
- There was a bug on the full width layout because the `columnWidth` was in fact `columnWidth + gutter`, this was fixed 
- We calculate the offscreen is slightly different ways on both layouts `-9999` vs `Infinity`, I suppose this doesn't have an important impact but have to verify

#### Manual tests

https://github.com/pinterest/gestalt/assets/700818/2ce51137-f2e5-43b3-b8f7-4a8a7fa1ce2f


packages/gestalt/src/Masonry/fullWidthMultiColumLayout.js
```diff
@@ -0,0 +1,52 @@
...
+      }));
+  }
+  // "This is kind of crazy!" - you
+  // Yes, indeed. The "guessing" here is meant to replicate the pass that the
+  // original implementation takes with CSS.
+  const colguess = Math.floor(width / idealColumnWidth);
+  const columnCount = Math.max(Math.floor((width - colguess * gutter) / idealColumnWidth), minCols);
+  const columnWidth = Math.floor(width / columnCount) - gutter;
+
+  return (items: $ReadOnlyArray<T>) =>
```
liuyenwei, line 39: instead of a forked layout, it seems like we could move this into `fullWidthLayout` as a conditional path. what do you thinka bout doing that? i think the ultimate goal was always to have the `multiColumnLayout` logic live inside the "base" layouts so awesome to see this

packages/gestalt/src/Masonry/defaultMultiColumnLayout.js
```diff
@@ -0,0 +1,67 @@
...
+
+    const columnWidthAndGutter = columnWidth + gutter;
+    const columnCount = Math.max(Math.floor((width + gutter) / columnWidthAndGutter), minCols);
+
+    let centerOffset;
+    if (justify === 'center') {
+      const contentWidth = Math.min(rawItemCount, columnCount) * columnWidthAndGutter + gutter;
+
+      centerOffset = Math.max(Math.floor((width - contentWidth) / 2), 0);
+    } else {
```
liuyenwei, line 48: same comment/question here about just moving this into the base layout function

totally fine if you'd rather be more cautious initially

packages/gestalt/src/Masonry/fullWidthLayout.js
```diff
@@ -52,13 +53,13 @@ const fullWidthLayout = <T>({
       } else {
         const col = mindex(heights);
         const top = heights[col];
-        const left = col * columnWidth + gutter / 2;
+        const left = col * columnWidthAndGutter + gutter / 2;
 
         heights[col] += height;
```
liuyenwei, line 58: just FYI - i checked and we do currently apply a slight gutter on mobile web so if we were to roll out the fullWidthMultiColumnLayout, it would cause an increased gap. it might be good to address that first (in a separate PR) and then land this one.

PR #3576, opened 2024-05-18
Tokens: VR tokens (final vr-theme files) for Android and iOS consumption (web mapping layer not included)

AlbertCarreras:
Tokens: Prepare VR tokens (final vr-theme files) for Android and iOS consumption (web mapping layer not included)

## Breaking change
from 
`gestalt-design-tokens/dist/<format>/
`to 
`gestalt-design-tokens/dist/<format>/classic/`


packages/gestalt-design-tokens/build.js
```diff
@@ -1,96 +1,315 @@
 const StyleDictionary = require('style-dictionary');
```
AlbertCarreras, line 1: The bulk of changes are here... hard to review!

PR #3597, opened 2024-05-28
Internal: Auto-generate .d.ts type declaration files

liuyenwei:
# Summary
Currently, we maintain separate type declaration files (*.d.ts) that are separate from the source code. These types are also bundled together with the published release to provide type information for consumers. This approach is problematic for two reasons:
1) It requires extra overhead for developers to manually maintain a set of types
2) The type declaration is disconnected from the actual source code and, as a result, can quickly become out of sync. Most recently, I encountered this when adding new Masonry props but forgetting to update the type declaration file

This PR removes the manual type declaration files in favor of having the typescript compiler emit declaration files. To enable this, there were two main changes made:

## tsConfig updates
There is currently one tsConfig file for the entire repo. This PR updates the configuration so that there is now:
- one "global" tsConfig file 
- one tsConfig file in each package, which extends the global tsConfig. 

The package-specific tsConfig files allow us to update the `compilerOptions` to emit declaration files for specific packages (gestalt, gestalt-charts, gestalt-datepicker). 
```
"compilerOptions": {
    "composite": true,
    "noEmit": false,
    "emitDeclarationOnly": true,
    "declaration": true,
    "outDir": "./dist",
    "rootDir": "./src"
  },
```

The global tsConfig file takes care of repo-wide settings. To tie everything together, we leverage Typescript's [project references feature](https://www.typescriptlang.org/docs/handbook/project-references.html#overall-structure) which allows us to reference other "projects" that contain a `.tsConfig` file. This also enables the compiler to understand the dependency graph so it can sequence things correctly when run with the `--build` flag. 
```
"references": [
    { "path": "./packages/gestalt" },
    { "path": "./packages/eslint-plugin-gestalt" },
    { "path": "./packages/gestalt-codemods" },
    { "path": "./packages/gestalt-datepicker" },
    { "path": "./packages/gestalt-charts" },
    { "path": "./docs" }
  ]
```

## rollup updates
The main change to the rollup configuration is that each project will now be built with its respective `tsConfig.json`. 
With the change to emit declaration files, this means that the `dist` folder will now contain type declaration files that match the structure of the source code
image

## Additional changes
The other changes in this PR are mostly around removing/updating type suppressions as a result of the auto generated type information being different from the manual types. 

# Callouts
- I encountered a weird issue with the `global.d.ts` file not being applied to each package when running the compiler. I was able to workaround this by adding `global.d.ts` files to the packages that needed them (gestalt + gestalt-datepicker) but would like to followup on this.
- Because this new approach requires us to always call `tsc` with the `--build` flag, I added a `tsc` script in package.json which automatically adds the `--build` flag. 

# Testing
I tested all the following scenarios locally:
[x] yarn dev works (clean checkout)
[x] yarn dev works (after a previous run)
[x] yarn dev - type information is updated with incremental build
[x] yarn build works (no cache)
[x] yarn build works (cache)
[x] yarn tsc works (no cache)
[x] yarn tsc works (cache)


willsmanley (approved):

AlbertCarreras (approved):

jackhsu978 (approved):
This is awesome! 😍 ❤️ 💯

PR #3612, opened 2024-06-05
Image: Address onLoad not being called for SSR images

liuyenwei:
# Summary
`<Image />` accepts an `onLoad` callback which ends up being attached to the native `img` element's `load` event handler and is triggered once the image resource has finished loading. 

The problem with this approach is that the `onLoad` handler does not get triggered if the image resource has already finished loading _before_ the `<Image />` component is rendered. The most common scenario of this happening is server-side rendering since the image request will be kicked off via the server-rendered HTML and, very likely, be completed by the time client hydration happens.

# Change
This PR introduces a fix for this by attached a ref callback to the native `img` callback which will do the following:
- check that `_fixCompletedOnLoad` is true, the [img complete attribute](https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/complete) is true, and the `onLoad` callback has not yet been triggered
- if the above is true, we will call `onLoad` will a fake load Event. Because the load event has already happened, we need to construct our own.

## Why is this in an experiment
I discovered this issue when debugging an impression drop with my Masonry V2 experiment. I found that, because my experiment resulted in the initial server rendered Pins not being unmounted/re-mounted, the image onLoad callbacks were never being triggered and consequently, we were not logging impressions.

I'm planning to test this change across both control and enabled groups of the Masonry V2 experiment to validate this and in parallel, will chat with metrics quality about rolling this out since it _could_ have impression impact.

### DIFF
--- a/packages/gestalt/src/Image.tsx
+++ b/packages/gestalt/src/Image.tsx
@@ -78,6 +78,10 @@ type Props = {
    * A comma-separated list of one or more strings indicating a set of possible image sources for the user agent to use.
    */
   srcSet?: string;
+  /**
+   * Experimental prop: Fixes an issue where onLoad is not triggered if the image is already loaded by the time the component renders (i.e. SSR)
+   */
+  _fixCompletedOnLoad?: boolean;
 };
 
 /**
@@ -102,6 +106,8 @@ export default class Image extends PureComponent<Props> {
 
   static displayName: string | null | undefined = 'Image';
 
+  onLoadCalled: boolean = false;
+
   componentDidMount() {
     if (shouldScaleImage(this.props.fit)) {
       this.loadImage();
@@ -116,6 +122,7 @@ export default class Image extends PureComponent<Props> {
   }
 
   handleLoad: (event: React.SyntheticEvent<HTMLImageElement>) => void = (event) => {
+    this.onLoadCalled = true;
     this.props.onLoad?.({ event });
   };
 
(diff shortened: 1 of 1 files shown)


packages/gestalt/src/Image.tsx
```diff
@@ -134,6 +141,30 @@ export default class Image extends PureComponent<Props> {
     }
   }
 
+  refCallback = (node: HTMLImageElement | null) => {
+    const { _fixCompletedOnLoad } = this.props;
+    // For certain scenarios, such as server-side rendering, the image may already be loaded by the time the component is rendered resulting in the onLoad event not being triggered.
+    // To address these, we can use a ref callback and check whether the image is already loaded - if it is, we trigger the onLoad event manually.
+    if (_fixCompletedOnLoad && node?.complete && !this.onLoadCalled) {
```
AlbertCarreras, line 148: `!this.onLoadCalled `

so `handleLoad` when image loads and we set `this.onLoadCalled = true;`. Then, when it renders we set the ref and we check `onLoadCalled`

`handleLoad` gets sets manually if `this.onLoadCalled = true;`. 

should `&& !this.onLoadCalled ` be `&& this.onLoadCalled `

AlbertCarreras, line 148: `!this.onLoadCalled `

so `handleLoad` when image loads and we set `this.onLoadCalled = true;`. Then, when it renders we set the ref and we check `onLoadCalled`

`handleLoad` gets sets manually if `this.onLoadCalled = true;`. 

"if it is, we trigger the onLoad event manually."

should `&& !this.onLoadCalled ` be `&& this.onLoadCalled `

liuyenwei: `!this.onLoadCalled` is here to guard against a scenario where the native `img` load callback is fired (and `onLoad` called) before the ref callback

i think the only situation where this could happen is:
- image component renders
- image request completes very quickly
- ref callback is triggered

i think this would be an edge case but wanted to play it safe

AlbertCarreras (approved):

PR #3623, opened 2024-06-12
Tokens: Add line-height tokens

rlingineni:
### Summary
This PR adds line-height tokens. Line-height tokens are different by language, so they are organized by language type. The full [values](https://docs.google.com/spreadsheets/d/1IwQ527sxsfqgspxKKPsWfWvQz5I-v2F4gRdomOgRsxY/edit#gid=1232169872) can be seen here.

You will likely need to use a localization provider to set the line-height tokens correct by language. 

The structure for line-height tokens can be [seen here:](https://github.com/rlingineni/gestalt/commit/1d8c11755b308100806403c1d4022c0cc6050e4a#diff-e1eec7292049d26c6514be76abb90090b510acd62b5192ee13ef6f53ebffbb31)

I picked this folder structure for the outputs instead of suffixing the language at the end because this list may grow and there may be more language related tokens in the future. This helps keep them all sorted and organized by language. Each of these files also contain tokens with the same names.

```
language:
 -> default
   -> line-height-tokens-file
 -> CK (Chinese + Korean)
    -> line-height-tokens-file
 -> Tall
    -> line-height-tokens-file
 -> Ja (Japanese)
    -> line-height-tokens-file
 -> Th (Thai)
    -> line-height-tokens-file
 -> Vi (Vietnamese)
    -> line-height-tokens-file
```

image

### Links

- [Jira](https://jira.pinadmin.com/browse/GESTALT-8040)
- [Slack](https://pinterest.slack.com/archives/C014X9LTRCN/p1717523648002989)

### Checklist

- [X] Added unit tests


packages/gestalt-design-tokens/tokens/vr-theme/language/sema-line-height-th.json
```diff
@@ -0,0 +1,21 @@
...
+      "XXS": {
+        "value": "{font.line-height.130.value}"
+      },
+      "XS": {
+        "value": "{font.line-height.150.value}"
+      },
+      "MD": {
+        "value": "{font.line-height.170.value}"
+      },
+      "LG": {
```
AlbertCarreras, line 13: is this correct? LG bigger than XL?

packages/gestalt-design-tokens/tokens/vr-theme/language/sema-line-height-default.json
```diff
@@ -0,0 +1,21 @@
...
+      "XXS": {
+        "value": "{font.line-height.110.value}"
+      },
+      "XS": {
+        "value": "{font.line-height.120.value}"
+      },
+      "MD": {
+        "value": "{font.line-height.130.value}"
+      },
+      "LG": {
```
AlbertCarreras, line 13: hows that LGs are higher than XL?

packages/gestalt-design-tokens/build.js
```diff
@@ -256,6 +297,8 @@ function getSources({ theme, modeTheme, platform }) {
     `tokens/vr-theme/base-elevation${modeTheme}.json`,
     'tokens/vr-theme/sema-elevation.json',
     `tokens/vr-theme/sema-color${modeTheme}.json`,
+    `tokens/vr-theme/base-line-height.json`,
+    `tokens/vr-theme/language/sema-line-height-${language}.json`,
```
AlbertCarreras, line 301: can we just inline the files here rather than putting them below in the forEach? line 1080

```

    `tokens/vr-theme/language/sema-line-height-default.json`,
    `tokens/vr-theme/language/sema-line-height-js.json`,
...

```

I think it's easier to follow fro here than all the drilling of the languages

packages/gestalt-design-tokens/build.js
```diff
@@ -831,132 +910,106 @@ StyleDictionary.registerTransformGroup({
...
+     "destination": "GestaltTokensFontSize.h",
+      "format": "ios/static.h",
+      "_format_comment": "https://amzn.github.io/style-dictionary/#/formats?id=iosstringsh",
+      "className": "GestaltTokensFontSize",
+      "type": "GestaltTokensFontSizeName"
+    }
+  */
+  let iOSObjectiveCFiles = categories.flatMap((category) => {
+    const pascalName = category
+      .split('-')
```
AlbertCarreras, line 939: ` .split('-')` is necessary?

packages/gestalt-design-tokens/build.js
```diff
@@ -831,132 +910,106 @@ StyleDictionary.registerTransformGroup({
...
+        'className': `GestaltTokens${pascalName}${getTheme(theme)}`,
+        'type': `GestaltTokens${pascalName}Name${getTheme(theme)}`,
+        ...getFilter(category.split(' ')[0], category.split(' ')[1]),
+      };
+    });
+  });
+
+  let iOSSwiftFiles = categories.flatMap((category) => {
+    const pascalName = category
+      .split('-')
```
AlbertCarreras, line 961: .split('-') is necessary?

rlingineni: See above

packages/gestalt-design-tokens/build.js
rlingineni: Agreed, inline would look cleaner. 

But I'm not able to because `sema-line-height-ja` exports the same tokens as `sema-line-height-default`, running it in the for-each loop as a separate build allows us to avoid the conflict

Otherwise, there are conflicts because the exported names are the same

packages/gestalt-design-tokens/tokens/vr-theme/base-line-height.json
```diff
@@ -0,0 +1,30 @@
+{
+  "font": {
+    "line-height": {
```
AlbertCarreras, line 3: How do the "-" translates to token?
Screenshot by Dropbox Capture

It creates a deeper level no?

AlbertCarreras: Screenshot by Dropbox Capture



Should we just do lineHeight? we decided something similar with data-visualization, we went for dataviz instead 
Screenshot by Dropbox Capture

rlingineni: I'm okay with `line-height`. They're two separate words. I thought for data-viz one of the other issues was the sema and base naming conflict

This was the example from Android I saw, so I think it's expected:

image

AlbertCarreras (approved):

AlbertCarreras:
@rlingineni Can we get this merge by Monday? I want to implement next week.

rlingineni:
@AlbertCarreras  - it's in

PR #3637, opened 2024-06-18
Masonry: Add _loadingStateItems & _renderLoadingStateItems props

rebeccayeyi:
#### What changed?

- Add `_loadingStateItems` & `_renderLoadingStateItems` props
- Add `Masonry-renderLoadingItems` visual test page

#### Why?
- These props are needed in order to support the shimmering skeleton loading state in web app. The next steps would be to create a `SkeletonPin` in Webapp use these props in Webapp's Masonry (similar to how the visual test is using them).

### Links

- [Deploy preview](https://deploy-preview-3637--gestalt.netlify.app/visual-test/Masonry-renderLoadingItems)
- [Shimmering Skeleton Loading State](https://docs.google.com/document/d/1dTM_bzahuxgU8vVd_uRCkqfPsEoVa89l2AReh_qb9tA/edit#heading=h.vy2e1uh0x4m0)

### Screen Recording

https://github.com/pinterest/gestalt/assets/19737469/e63b7490-35e3-469f-b928-3e9fe0e75143

### Checklist

- [x] Added unit tests


docs/pages/visual-test/Masonry.tsx
```diff
@@ -1,45 +1,132 @@
...
+          ref={(ref) => {
+            gridRef.current = ref;
+          }}
+          columnWidth={170}
+          gutterWidth={20}
+          items={pins}
+          layout="basicCentered"
+          loadingStateItems={skeletonPins}
           renderItem={({ data }) => <GridComponent data={data} />}
+          renderLoadingItems={({ data }) => <SkeletonPin height={data.height} />}
```
rebeccayeyi, line 125: how configurable do we want this to be? @liuyenwei brought up a good point that this might make it easier to experiment different loading states

rebeccayeyi: discussed in https://docs.google.com/document/d/10wCnGJ6uPdtDAAkzLYFJCt-i8rmERE_8PCrFnGu9crQ/edit#heading=h.oc974dez4na

packages/gestalt/src/Masonry/defaultLayout.ts
```diff
@@ -100,7 +100,7 @@ const defaultLayout =
         })
       : items.reduce<Array<any>>((acc, item) => {
           const positions = acc;
-          const height = measurementCache.get(item);
+          const height = measurementCache.get(item) || (item as any).height;
```
rebeccayeyi, line 103: this is needed or else the loading page is empty -- wondering if we should update the type to make `height` required?

eg: `U extends { height: number }` (of course, this would mean updating a bunchhhh of types 😅 )

packages/gestalt/src/Masonry.tsx
```diff
@@ -120,14 +120,32 @@ type Props<T> = {
...
+   * A function that renders the item you would like displayed in the grid. This function is passed three props: the item's data, the item's index in the grid, and a flag indicating if Masonry is currently measuring the item.
+   *
+   * If present, `heightAdjustment` indicates the number of pixels this item needs to grow/shrink to accommodate a 2-column item in the grid. Items must respond to this prop by adjusting their height or layout issues will occur.
+   */
+  _loadingStateItems?: ReadonlyArray<U>;
+  /**
+   * Experimental prop to render a loading state
+   * A function that renders the item you would like displayed in the grid. This function is passed three props: the item's data, the item's index in the grid, and a flag indicating if Masonry is currently measuring the item.
+   *
+   * If present, `heightAdjustment` indicates the number of pixels this item needs to grow/shrink to accommodate a 2-column item in the grid. Items must respond to this prop by adjusting their height or layout issues will occur.
```
rebeccayeyi, line 135: oops needs to update these comments!

docs/pages/visual-test/Masonry-renderLoadingItems.tsx
```diff
@@ -0,0 +1,139 @@
...
+
+const styles = {
+  backgroundColor: `${TOKEN_COLOR_GRAY_ROBOFLOW_300}`,
+  backgroundSize: '200vw 100%',
+  content: '',
+  display: 'flex',
+  flexDirection: 'column',
+  position: 'relative',
+  width: '100%',
+} as const;
```
rebeccayeyi, line 64: noting that i didnt add the animation since this was just for testing purposes

packages/gestalt/src/Masonry.tsx
```diff
@@ -608,11 +640,9 @@ export default class Masonry<T> extends ReactComponent<Props<T>, State<T>> {
       const itemsToMeasureCount = hasMultiColumnItems
         ? MULTI_COL_ITEMS_MEASURE_BATCH_SIZE + 1
         : minCols;
-      const itemsToMeasure = items
-        .filter((item) => item && !measurementStore.has(item))
-        .slice(0, itemsToMeasureCount);
+      const itemsToMeasure = items.slice(0, itemsToMeasureCount);
```
manvoid, line 643: Is this correct? That we remove the filter for items that have not been measured?

packages/gestalt/src/Masonry/defaultLayout.ts
manvoid: So for the skeleton items we don't render them first to get the height? This is something that could be good to look into more detail.

docs/pages/visual-test/Masonry-renderLoadingItems.tsx
```diff
@@ -0,0 +1,139 @@
...
+      />
+    </Flex>
+  );
+}
+
+const randomPinHeight = () => Math.random() * 200 + 100;
+
+const skeletonPins = [...new Array(3)]
+  .map(() => [
+    { id: 1, height: randomPinHeight(), width: 474 },
```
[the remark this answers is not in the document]
rebeccayeyi: should be using id and width, thanks for the catch!

packages/gestalt/src/Masonry/defaultLayout.ts
rebeccayeyi: lets discuss at the next Masonry sync 👍

docs/pages/visual-test/Masonry-renderLoadingItems.tsx
```diff
@@ -0,0 +1,139 @@
...
+function getPins(): Promise<Array<Pin>> {
+  const pinList = [...new Array(3)].map(() => [...pins]).flat();
+  return Promise.resolve(pinList);
+}
+
+const styles = {
+  backgroundColor: `${TOKEN_COLOR_GRAY_ROBOFLOW_300}`,
+  backgroundSize: '200vw 100%',
+  content: '',
+  display: 'flex',
```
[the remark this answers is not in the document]
rebeccayeyi: good call out

packages/gestalt/src/Masonry.tsx
rebeccayeyi: fixed 👍

packages/gestalt/src/Masonry.tsx
```diff
@@ -120,14 +120,28 @@ type Props<T> = {
    *
    * This is an experimental prop and may be removed or changed in the future.
    */
-  _getColumnSpanConfig?: (item: T) => ColumnSpanConfig;
+  _getColumnSpanConfig?: (item: T | U) => ColumnSpanConfig;
+  /**
+   * An array of items to display that contains the data to be rendered by `_renderLoadingStateItems`.
+   */
+  _loadingStateItems?: ReadonlyArray<U>;
```
liuyenwei, line 127: do we need a generic type for `_loadingStateItems` or can we just be specific? aside from `height` being passed in, is there anything else we need?

packages/gestalt/src/Masonry.tsx
```diff
@@ -471,7 +493,9 @@ export default class Masonry<T> extends ReactComponent<Props<T>, State<T>> {
           height: layoutNumberToCssDimension(height),
         }}
       >
-        {renderItem({ data: itemData, itemIdx: idx, isMeasuring: false })}
+        {isLoading && _renderLoadingStateItems
+          ? _renderLoadingStateItems({ data: itemData as U, itemIdx: idx, isMeasuring: false })
```
liuyenwei, line 497: feels odd that `_renderLoadingstateItems` has the same API signature as `renderItem` since `isMeasuring` will always be false

packages/gestalt/src/Masonry.tsx
```diff
@@ -608,11 +637,12 @@ export default class Masonry<T> extends ReactComponent<Props<T>, State<T>> {
       const itemsToMeasureCount = hasMultiColumnItems
         ? MULTI_COL_ITEMS_MEASURE_BATCH_SIZE + 1
         : minCols;
-      const itemsToMeasure = items
+
+      const itemsToMeasure = (items.length === 0 ? _loadingStateItems : items)
```
liuyenwei, line 641: we should never be measuring loadingStateItems imo since we have the ability to assign the height ahead of time

packages/gestalt/src/Masonry.tsx
```diff
@@ -622,18 +652,27 @@ export default class Masonry<T> extends ReactComponent<Props<T>, State<T>> {
...
+                    idx,
+                    // If we have items in the positionStore (newer way of tracking positions used for 2-col support), use that. Otherwise fall back to the classic way of tracking positions
+                    // this is only required atm because the two column layout doesn't not return positions in their original item order
+                    position: positionStore.get(itemData) ?? positions[idx],
+                  }),
+                )}
           </div>
           <div className={styles.Masonry} style={{ width }}>
-            {itemsToMeasure.map((data, i) => {
+            {itemsToMeasure.map((data: T | U, i) => {
```
liuyenwei, line 675: same comment here but ideally we don't measure the loading state items since the loading state should be as cheap as possible to render

packages/gestalt/src/Masonry/defaultLayout.ts
rebeccayeyi: closing the loop here from our discussion:

- we should not render the items first since we will always know what the height is
  - make height required

docs/pages/visual-test/Masonry-renderLoadingItems.tsx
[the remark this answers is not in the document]
rebeccayeyi: updated to just have height

packages/gestalt/src/Masonry.tsx
rebeccayeyi: good callout, i dont think this needs to be generic since we will be the ones controlling this in webapp and we dont need anything besides `height` (makes the types easier too!)

packages/gestalt/src/Masonry.tsx
rebeccayeyi: updated!

packages/gestalt/src/Masonry.tsx
```diff
@@ -15,7 +15,14 @@ import throttle, { ThrottleReturn } from './throttle';
 
 const RESIZE_DEBOUNCE = 300;
 
-const layoutNumberToCssDimension = (n?: number | null) => (n !== Infinity ? n : undefined);
+const layoutNumberToCssDimension = (n?: number | null) => {
+  if (n) {
+    return n !== Infinity ? n : undefined;
+  }
+  return undefined;
+};
```
rebeccayeyi, line 23: updated to always return a number or undefined to remove the type errors

packages/gestalt/src/Masonry/defaultLayout.ts
```diff
@@ -100,9 +102,14 @@ const defaultLayout =
         })
       : items.reduce<Array<any>>((acc, item) => {
           const positions = acc;
-          const height = measurementCache.get(item);
+          let height = measurementCache.get(item);
           let position;
 
+          if (renderLoadingState) {
+            // Height is a required key for _loadingStateItems
+            height = (item as { height: number }).height;
```
rebeccayeyi, line 110: lmk your thoughts here -- we could update the generic to be `T extends { height?: number }`, but that would require updating Masonry's generic to be the same and this felt like the shortest solution

packages/gestalt/src/Masonry.tsx
```diff
@@ -595,6 +655,39 @@ export default class Masonry<T> extends ReactComponent<Props<T>, State<T>> {
...
+        gutter,
+        layout,
+        minCols,
+        rawItemCount: _loadingStateItems.length,
+        width,
+        logWhitespace: _logTwoColWhitespace,
+        _getColumnSpanConfig,
+        renderLoadingState,
+      });
+      const positions = getPositions(_loadingStateItems as any);
```
manvoid, line 673: One option could be to have two props on the getPositions function, and that the second prop is for loading items ({height: ...}), that way we can clearly separate both cases and usa that inside the layout functions to get the height from the cache in the normal items case but use th height prop on the loading items case.

packages/gestalt/src/Masonry/defaultLayout.ts
manvoid: As mentioned above to me it sounds better to add a new second prop to have (items, loadingItems), that way you can keep the T type for items and the new {height: ..} structure for the loadingItems prop. Aldo with that you could add logic here to differentiate the loading items from the normal items

packages/gestalt/src/Masonry.tsx
```diff
@@ -595,6 +655,39 @@ export default class Masonry<T> extends ReactComponent<Props<T>, State<T>> {
       // When the width is empty (usually after a re-mount) render an empty
       // div to collect the width for layout
       gridBody = <div ref={this.setGridWrapperRef} style={{ width: '100%' }} />;
+    } else if (renderLoadingState) {
+      getPositions = defaultLayout({
```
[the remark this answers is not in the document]
liuyenwei: +1 okay for now, but we can clean up later

packages/gestalt/src/Masonry.tsx
```diff
@@ -478,6 +497,42 @@ export default class Masonry<T> extends ReactComponent<Props<T>, State<T>> {
...
+      return (
+        <div
+          key={`item-${idx}`}
+          className={[styles.Masonry__Item, styles.Masonry__Item__Mounted].join(' ')}
+          data-grid-item
+          role="listitem"
+          style={{
+            top: 0,
+            left: 0,
+            transform: `translateX(${left}px) translateY(${top}px)`,
```
liuyenwei, line 522: iirc, the loading state animation didn't work correctly with transform since that messed with fixed positioning. we likely want to update this to just set top, left instead

packages/gestalt/src/Masonry.tsx
liuyenwei: if we update the function signature for one, we should do so for all the layout functions. while the inputs are different, the current expectation is that the return function signature is the same (which we shouldn't break). 

alternatively, we could also make the input type polymorphic `(items: ReaonlyArray<T> | ReadonlyArray<LoadingStateItem>)`

packages/gestalt/src/Masonry.tsx
```diff
@@ -121,6 +128,19 @@ type Props<T> = {
...
+  _loadingStateItems?: ReadonlyArray<LoadingStateItem>;
+  /**
+   * Experimental prop to render a loading state
+   *
+   * A function that renders the loading state items you would like displayed in the grid. This function is passed three props: the item's data, the item's index in the grid, and a flag indicating if Masonry is currently measuring the item.
+   */
+  _renderLoadingStateItems?: (arg1: {
+    readonly data: LoadingStateItem;
+    readonly itemIdx: number;
+  }) => ReactNode;
```
willsmanley, line 143: nit: comment might be out of date, depending on if the third prop is now removed

rebeccayeyi: thank you for the catch!

packages/gestalt/src/Masonry.tsx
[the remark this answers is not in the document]
rebeccayeyi: need to double check with product/design, but for now im assuming we want to only support the default layout for mvp so i thought it made sense to always use the default layout here 

we could add an extra condition in each if block (eg: `(layout === 'flexible' || layout === 'serverRenderedFlexible') && width !== null && !renderLoadingState`), but only using defaultLayout here meant that we could update defaultLayout's signature to accept a polymorphic parameter for items rather than updating all the other layout utils which would never get called with loading state items -- lmk if this makes sense

packages/gestalt/src/Masonry.tsx
rebeccayeyi: addressed in https://github.com/pinterest/gestalt/pull/3637#discussion_r1681456191

packages/gestalt/src/Masonry.tsx
rebeccayeyi: thank you for this! great memory

packages/gestalt/src/Masonry/defaultLayout.test.ts
```diff
@@ -247,3 +247,69 @@ describe.each([undefined, getColumnSpanConfig])('default layout tests', (_getCol
...
+  test('uses the measurementCache height when not a loadingStateitem', () => {
+    const measurementStore = new MeasurementStore<Record<any, any>, number>();
+    const positionCache = new MeasurementStore<Record<any, any>, Position>();
+    const items: ReadonlyArray<Item> = [
+      { 'name': 'Pin 0', 'height': 100 },
+      { 'name': 'Pin 1', 'height': 120 },
+      { 'name': 'Pin 2', 'height': 80 },
+      { 'name': 'Pin 3', 'height': 100 },
+    ];
+    items.forEach((item: any) => {
```
rebeccayeyi, line 290: added this test because i noticed that my `isLoadingStateItem` check would always return true if an item had a height key (even if it should not render a loading state item)

updated `isLoadingStateItem` to also check that `renderLoadingState` is true

packages/gestalt/src/Masonry/defaultLayout.ts
rebeccayeyi: marking this one as resolved since its a duplicate of https://github.com/pinterest/gestalt/pull/3637#discussion_r1680007162

liuyenwei (approved):

packages/gestalt/src/Masonry/defaultLayout.ts
```diff
@@ -100,8 +103,14 @@ const defaultLayout =
         })
       : items.reduce<Array<any>>((acc, item) => {
           const positions = acc;
-          const height = measurementCache.get(item);
           let position;
+          let height;
```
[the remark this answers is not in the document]
manvoid: +1

packages/gestalt/src/Masonry.tsx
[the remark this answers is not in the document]
manvoid: Yes, I also think is better if we solve this soon

manvoid (approved):

packages/gestalt/src/Masonry.tsx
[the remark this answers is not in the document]
rebeccayeyi: that makes sense, Karsten (design) is back on monday so i can check with him!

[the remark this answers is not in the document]
rebeccayeyi: alright so we have landed on a [decision](https://pinterest.slack.com/archives/C075Z345M37/p1721655708191709) to support flexible layout 👍 

i would like to merge this pr first though since i dont want it to get stale and i will create a new pr with flexible support

PR #3638, opened 2024-06-18
Checkbox, Modal, ModalAlert, TagData: Adding Data-test-id for those components

RicentCB:
#### What changed?

Some of the components do not have the data-test-id property, which makes it difficult to reference them in an integration test. For this reason, some of them have been updated and had it added

#### Why?

In integration tests, there are multiple ways to locate an element, such as using the ID, classes, or content. However, these approaches can change without prior notice and may also depend on the language in which the application is being executed.
Therefore, one of the best practices is to have a property that remains constant, something that does not affect the execution or performance and remains consistent at all times.

Thus, the data-test-id property is crucial for testing purposes, and it is now very helpful to have it in all components if feasible.

### Checklist

- [x] Added unit tests

### DIFF
--- a/packages/gestalt/src/Checkbox/InternalCheckbox.tsx
+++ b/packages/gestalt/src/Checkbox/InternalCheckbox.tsx
@@ -20,6 +20,7 @@ import useFocusVisible from '../useFocusVisible';
 
 type Props = {
   checked?: boolean;
+  dataTestId?: string;
   disabled?: boolean;
   errorMessage?: string;
   helperText?: string;
@@ -47,6 +48,7 @@ const InternalCheckboxWithForwardRef = forwardRef<HTMLInputElement, Props>(funct
   {
     checked = false,
     disabled = false,
+    dataTestId,
     errorMessage,
     helperText,
     id,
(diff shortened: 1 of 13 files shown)


AlbertCarreras:
Pending discussion. Closing for now.

packages/gestalt/src/Checkbox/InternalCheckbox.tsx
```diff
@@ -131,7 +133,7 @@ const InternalCheckboxWithForwardRef = forwardRef<HTMLInputElement, Props>(funct
   return (
     <Box>
       <Box alignItems="start" display="flex" justifyContent="start" marginEnd={-1} marginStart={-1}>
-        <Box paddingX={1} position="relative">
+        <Box data-test-id={dataTestId} paddingX={1} position="relative" >
```
AlbertCarreras, line 136: this should go in the input, not the outer box.

AlbertCarreras:
Modal, ModalAlert, TagData are random addditions, without having discussed a proposal or even the problem.

packages/gestalt/src/Checkbox/InternalCheckbox.tsx
RicentCB: Alberto, I included this here because in some tests, we need to interact with the input, while in others, we need to handle the label or the icon. It is easier to handle all these deeper elements with a single data-test-id in this manner:

```cy.get('[data-test-id="checkbox"] input')
cy.get('[data-test-id="checkbox"] label')
cy.get('[data-test-id="checkbox"] icon')
```

This way, we can manipulate the element we need. I apologize for any confusion in my previous explanation, and I will strive to provide clearer explanations in our future discussions.

PR #3640, opened 2024-06-18
Tokens:  small fixes to line height tokens

AlbertCarreras:
Token:  small fixes to line height tokens


rlingineni (approved):

packages/gestalt-design-tokens/tests/__snapshots__/web.test.js.snap
```diff
@@ -2100,11 +2100,11 @@ Object {
   --sema-opacity-light: var(--base-opacity-20);
   --sema-opacity-pin-wash: var(--base-opacity-4);
   --sema-opacity-transparent: var(--base-opacity-0);
-  --sema-font-lineheight-xl: var(--base-font-lineheight-140);
-  --sema-font-lineheight-lg: var(--base-font-lineheight-130);
-  --sema-font-lineheight-md: var(--base-font-lineheight-120);
+  --sema-font-lineheight-xl: var(--base-font-lineheight-150);
+  --sema-font-lineheight-lg: var(--base-font-lineheight-140);
```
rlingineni, line 2104: ideally, this wouldn't be output in the big css file. All the sema token files would be in the language files only. 

I think we can add a CSS Sema LineHeight Filter and toss these tokens from the main.css file

rlingineni (approved):

packages/gestalt-design-tokens/tests/__snapshots__/web.test.js.snap
AlbertCarreras: I think it's ok to have them there. You need the default set up to not require the provider injections.

PR #3642, opened 2024-06-20
SideNavigation: SideNavigationGroupItem with link

diyorbek:
### Summary

GroupItem subcomponent now can have links.
The example is added in the **Subcomponent composability** section of SideNaivgation docs page.

Screen Recording 2024-06-20 at 21 00 43 (1)

#### What changed?

> [!NOTE]
> Refactored and moved things around to make the implementation less DRY.

`SideNavigation.Group` and `SideNavigation.NestedGroup` now have following new props:
 - `href?: string`
 - `active?: 'page' | 'section'`
 - `onClick?: ClickEventHandler `

### Links

- [Jira](https://jira.pinadmin.com/browse/GESTALT-8032)
- [Figma](https://www.figma.com/design/vjhfBsOtHw0wVg67vqwz1v/branch/V3zxGsrUf68qHDFNmD7rKw/Gestalt-for-Web?m=auto&node-id=35045-2769&t=U54hte1Tx0N2FQvh-1)

### Checklist

- [ ] Added unit tests


packages/gestalt/src/SideNavigation/GroupExpandIconButton.tsx
```diff
@@ -0,0 +1,76 @@
...
+        />
+      </Box>
+    );
+  }
+
+  return (
+    <IconButton
+      accessibilityControls={expandIconButtonProps?.accessibilityControls}
+      accessibilityExpanded={expandIconButtonProps?.accessibilityExpanded}
+      accessibilityLabel=""
```
diyorbek, line 59: @hectoid i think we should label this. I want sure what to do here.

packages/gestalt/src/SideNavigationGroup.tsx
```diff
@@ -126,31 +136,37 @@ export default function SideNavigationGroup({
   validateChildren({ children: navigationChildren, filterLevel: 'nested' });
 
   const isExpandable = display === 'expandable';
+  const isUncontrolled = expandedProp === undefined;
+  const isLink = !!href;
 
-  const [expanded, setExpanded] = useState<boolean>(hasAnyActiveChild);
-  const [isExpanded, setIsExpanded] = useState<boolean>(hasAnyActiveChild);
```
diyorbek, line 131: @AlbertCarreras this was redundant, so i removed this state

packages/gestalt/src/SideNavigation/ItemContent.tsx
```diff
@@ -0,0 +1,258 @@
...
+        {!collapsed && primaryAction ? (
+          <Flex.Item flex="none">
+            {/* This is a workaround to announce the counter as it's replaced on focus */}
+            {counter ? (
+              <Box display="visuallyHidden">
+                {`, `}
+                <Box aria-label={counter?.accessibilityLabel} />
+              </Box>
+            ) : null}
+            <Box aria-hidden marginEnd={-2} rounding="circle">
```
diyorbek, line 238: @AlbertCarreras @hectoid I don't understand why PrimaryActionButton should be "aria-hidden". Anything I am missing here?

AlbertCarreras:
I think this PR is doing too much.

You're changing several components APIs beyond your SideNavigation changes.

I'd say we need onMouseEnter, onMouseLeave in all components if we decide we want to make them accessible, But All components not just 1.

Here we are opening the door to something that only TapArea was doing. 

I'd suggest,

1. Going TapArea rather than changing IconButton 
2. Proposing consistency onMouseEnter, onMouseLeave props for all Button like components... But it's a bigger discussion

diyorbek:
@hectoid @AlbertCarreras I made a change that when nested item is selected, ots parent will always have border. This is how i understood the specs. If it's not something intended, i'll remove

Screenshot 2024-06-21 at 11 30 18

packages/gestalt/src/SideNavigation/GroupContent.tsx
```diff
@@ -1,72 +1,30 @@
...
   expanded: boolean;
-  selectedItemId: string;
   itemId: string;
-  paddingStyle:
-    | {
-        paddingInlineStart: string | number | undefined;
-        paddingInlineEnd: string | number | undefined;
-      }
-    | Record<any, any>;
-  icon?: IconType;
```
AlbertCarreras, line 42: nice cleanup here. Typescript pro!

AlbertCarreras:
> @hectoid @AlbertCarreras I made a change that when nested item is selected, ots parent will always have border. This is how i understood the specs. If it's not something intended, i'll remove
> 
> Screenshot 2024-06-21 at 11 30 18

This prob a design spec. Whatever Hector says.

packages/gestalt/src/SideNavigation/ItemContent.tsx
AlbertCarreras: I did this. 

The IconButton still gets announced when you focus on it on keyboard navigation. 

aria-hidden prevents this Button to be read as part of the parent Button. Otherwise all the Buttons would be read in the parent one. You want to hide that Button on Buttons content, so it only announces the parent one, and the the children when focused. Makes sense?

AlbertCarreras:
> > @hectoid @AlbertCarreras I made a change that when nested item is selected, ots parent will always have border. This is how i understood the specs. If it's not something intended, i'll remove
> > Screenshot 2024-06-21 at 11 30 18
> 
> This prob a design spec. Whatever Hector says.

Screenshot by Dropbox Capture

I think for simplicity I'll leave as it was. If you play around, there's time you can get 3 selected items... not sure what's going on.

AlbertCarreras:
Can we create an example to just this? I cant test it on `Subcomponent compoasabiltiy`, dont see any link behavior.

packages/gestalt/src/TapArea.tsx
```diff
@@ -90,7 +90,7 @@ type Props = {
    */
   onFocus?: FocusEventHandler;
   /**
-   *
```
AlbertCarreras, line 93: thanks

packages/gestalt/src/SideNavigation/TapControl.tsx
```diff
@@ -0,0 +1,76 @@
+import { ComponentProps } from 'react';
+import TapArea from '../TapArea';
+import TapAreaLink from '../TapAreaLink';
+
+type TProps = Pick<
```
AlbertCarreras, line 5: Tprops?

packages/gestalt/src/SideNavigation/TapControl.tsx
```diff
@@ -0,0 +1,76 @@
+import { ComponentProps } from 'react';
```
AlbertCarreras, line 1: SideNavigationGroupItemTapControl << can we name the file like this?

packages/gestalt/src/SideNavigation/GroupExpandIconButton.tsx
danhendersonede: Is the GroupExpandIconButton the caret that represents if the group is expanded or collapsed? 

If so, it does not need a label as the icon's focus target is wrapped up with the group button (although makes me wonder why this component uses an IconButton, and not just Icon), the state of the group is already communicated via `aria-expended`, see VO output:

image

AlbertCarreras:
Can you document in description chnages in API? 

We are adding 3 props? it's hard to track among the whole changes.

packages/gestalt/src/SideNavigation/GroupExpandIconButton.tsx
AlbertCarreras: This icons are more representational than anything else . the real button is the component itself. 

HOWEVER!!! If the whole think becomes a link while keeping the show/hide children, then this IconButton for opening then they have to 1. get the proper label (from provider) and 2. have keyboard navigation (they dont right now.)

Brave Browser - SideNavigation - Gestalt 2024-06-21 at 10 49 28 AM

diyorbek: @AlbertCarreras yes, you're correct. Actually this is the case where parent item is a link. I am just not sure what should be the default label

packages/gestalt/src/SideNavigation/GroupExpandIconButton.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+  return (
+    <ItemIconButton
+      accessibilityControls={expandIconButtonProps?.accessibilityControls}
+      accessibilityExpanded={expandIconButtonProps?.accessibilityExpanded}
+      accessibilityLabel=""
+      icon={expanded ? 'arrow-up' : 'arrow-down'}
+      isItemActive={!!active}
+      onKeyDown={({ event }) => event.stopPropagation()}
+      onTap={(arg) => {
```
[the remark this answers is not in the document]
diyorbek: ofc)

packages/gestalt/src/SideNavigation/TapControl.tsx
diyorbek: oops, typo

packages/gestalt/src/SideNavigation/ItemContent.tsx
diyorbek: should we add comment to the code?

hectoid:
> @hectoid @AlbertCarreras I made a change that when nested item is selected, ots parent will always have border. This is how i understood the specs. If it's not something intended, i'll remove
> 
> Screenshot 2024-06-21 at 11 30 18

very nice work @diyorbek ! Sorry if this wasn't clear in the spec, but:

- For the groups that have a parent link, the parent does not need the border; it will look like the children with the reversed background when selected, and no background or border when unselected.
- For groups that don't have a parent link, then it needs the border when selected

hectoid:
Examples look good! One tweak to the subgroup with links session if possible. can be a follow up ticket instead depending on time.

When the entire link group is collapsed, show the parent link as selected with the border. The way we do in the collapsed icon-only version.
image

Otherwise, users lose track of where in the nav they are:
image

Sorry, you had the border selected before and I wasn't clear. Continue to show it when it's completely collapsed.

hectoid:
hey again @diyorbek ! sorry, a couple of nits/questions.

1. I'm seeing the icon shift slightly when selecting the parent link (see attached)
2. Is there a reason why the Group Display example doesn't show any selected states for the parent or children? https://deploy-preview-3642--gestalt.netlify.app/web/sidenavigation#Group-display


https://github.com/pinterest/gestalt/assets/96082362/39c23711-3814-4970-9548-93d2a1999496

diyorbek:
@hectoid I have been planning to make follow up PR for icon shift and other nits I found. This is PR is already big so I didn't want to rush it. Speaking of rushing, this PR is hanging for 2 weeks :D 

Regarding the examples, none of the examples work if you pay attention. Except the ones I created (Collapsible, Group link). If we want to fix the examples, we should consider that is not a small work. I would actually go ahead and fix all of the component examples in the docs site. Because, honestly, they are very, very poorly done.

hectoid:
@diyorbek from my end, this is approved! sorry, I can never find where to approve on github.

AlbertCarreras (approved):

PR #3646, opened 2024-06-24
AvatarGroup: enable actions in smaller sizes

AlbertCarreras:
AvatarGroup: enable actions in smaller sizes

BEFORE counter and addCollaborators were only available to md.
AFTER counter and addCollaborators are available to xs, sm, & md.

Brave Browser - AvatarGroup - Gestalt 2024-07-01 at 10 44 50 AM

### DIFF
--- a/docs/pages/web/avatargroup.tsx
+++ b/docs/pages/web/avatargroup.tsx
@@ -1,4 +1,5 @@
-import { AvatarGroup } from 'gestalt';
+import { useState } from 'react';
+import { AvatarGroup, Box, SelectList } from 'gestalt';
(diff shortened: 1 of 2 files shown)


diyorbek (approved):

diyorbek:
A clarifying question: So the size only applies to the avatar count? Cuz i didn't size prop being passed to any of inner components, or being used as width/height.

AlbertCarreras:
> A clarifying question: So the size only applies to the avatar count? Cuz i didn't size prop being passed to any of inner components, or being used as width/height.

Yes, the size existed but we were not rendering the count/add icon buttons.

PR #3652, opened 2024-07-01
Tokens: Support Token Studio Structure

rlingineni:
# Pull Request Instructions

This PR Adds support for Token Studio in the VR Theme. 

This introduces two main changes:

1. Moving to a new folder structure
-> e.g. `sema-color.json -> sema/color.json`

2. Changing `darkTheme` prefixing to just `dark` for consistency in theme naming

3. VR Theme now introduces `base` and  `sema` in the JSON structure. `VR-theme-web-mapping` has been updated accordingly

All tests pass, so it should have no impact on the outputs. We will make additional changes to the token names in a later PR.


packages/gestalt-design-tokens/build.js
```diff
@@ -190,6 +192,25 @@ const filterList = [
   filterFontWeight,
 ];
 
+const getPrefix = (filePath) => {
+  const filePaths = filePath.split('/');
+  const filename = path.parse(filePath).name;
+
+  if (filePaths.includes('sema') || filename.startsWith('sema-')) {
```
diyorbek, line 199: Does this mean `sema` token files can exist outside of `sema` folder?

packages/gestalt-design-tokens/build.js
```diff
@@ -393,12 +414,35 @@ StyleDictionary.registerFormat({
...
+    const attrNames = ['category', 'type', 'item', 'subitem', 'state'];
+    if (hasPrefix) {
+      attrNames.unshift('prefix');
+    }
+
+    const originalAttrs = prop.attributes || {};
+    const generatedAttrs = {};
+
+    for (let i = 0; i < prop.path.length && i < attrNames.length; i += 1) {
+      generatedAttrs[attrNames[i]] = prop.path[i];
```
diyorbek, line 433: Honestly, I am not sure how this works. If `prop.path` and `attrNames` may have different `length`s, how do we know they both follow the same sequence? Because you're maching them with indices.

diyorbek: @rlingineni also you said there were no major changes in the generated output code. Can you explaint how it works? 🤔

packages/gestalt-design-tokens/package.json
```diff
@@ -21,7 +21,8 @@
   "devDependencies": {
     "decomment": "^0.9.5",
     "style-dictionary": "^3.9.2",
-    "tinycolor2": "1.6.0"
+    "tinycolor2": "1.6.0",
+    "lodash.camelcase": "4.3.0"
```
AlbertCarreras, line 25: instead of adding a package (we are trying to deprecate lodash in Pinbaord...) can we just copy the function in a helper funcion file?

packages/gestalt-design-tokens/tokens/vr-theme/$metadata.json
```diff
@@ -0,0 +1,25 @@
+{
```
AlbertCarreras, line 1: is order required by some reason?

packages/gestalt-design-tokens/tokens/vr-theme/base/lineheight.json
```diff
@@ -0,0 +1,40 @@
+{
```
AlbertCarreras, line 1: can we add this under font? lineheight is actually a font token it's not a category by itself.

AlbertCarreras (approved):
just a comment on lodash

packages/gestalt-design-tokens/build.js
rlingineni: Yes, they can either be prefixed, or sema can be in the filepath. 

We have `classic` and `vr-theme` both of which are organized differently

packages/gestalt-design-tokens/package.json
rlingineni: I'm re-using a function from style-dictionary and they are using lo-dash.

This one is made from multiple files. This is a sub-package, so we are only importing that one function, not all of lo-dash. It's also a dev dependency, so it will have no impact on the bundle

packages/gestalt-design-tokens/tokens/vr-theme/$metadata.json
rlingineni: This is generated by token-studio, I am not quite sure why

packages/gestalt-design-tokens/tokens/vr-theme/base/lineheight.json
rlingineni: We will do this in the July 2024 release, for now this will keep things 1:1 with our existing repo

PR #3662, opened 2024-07-04
Tooltip: Release v2

diyorbek:
### Summary

- Removed all the Tooltip v2 experiment related code from Popover.
- Deleted all legacy code related to Popover and Tooltip.

### Links

- [Jira](https://jira.pinadmin.com/browse/GWEB-120)
- [TDD](https://docs.google.com/document/d/1GPzNVK9X-A06PgOh5Ha3ncRFb27Cxl0RRBckauFRaZ8/edit?usp=sharing)

### Checklist

- [x] Added unit tests


packages/gestalt/src/Popover/Contents.tsx
```diff
@@ -51,7 +51,7 @@ export default function Contents({
   hideWhenReferenceHidden,
   onPositioned,
   shouldTrapFocus,
-  overflow = 'auto',
```
AlbertCarreras, line 54: why are we removing this `auto`?

packages/gestalt/src/Popover/Contents.tsx
```diff
@@ -137,19 +137,16 @@ export default function Contents({
         <div
           aria-label={accessibilityLabel}
           className={classnames(layoutStyles.relative, styles.maxDimensions, {
-            [styles.minDimensions]: width !== null,
+            [styles.minDimensions]: width != null,
```
AlbertCarreras, line 140: why this change>\?

packages/gestalt/src/Popover/InternalPopover.tsx
```diff
@@ -53,12 +53,12 @@ export default function InternalPopover({
   hideWhenReferenceHidden,
   onPositioned,
   disableFocusTrap = false,
-  overflow,
+  overflow = 'auto',
```
AlbertCarreras, line 56: oh I see.https://github.com/pinterest/gestalt/pull/3662/files#r1669356514

packages/gestalt/src/Popover/InternalPopover.tsx
```diff
@@ -53,12 +53,12 @@ export default function InternalPopover({
...
 }: Props) {
   const { accessibilityDismissButtonLabel: accessibilityDismissButtonLabelDefault } =
     useDefaultLabelContext('Popover');
 
-  const dismissButtonRef = useRef<null | HTMLAnchorElement | HTMLButtonElement>(null);
```
AlbertCarreras, line 61: ha thanks

AlbertCarreras (approved):
LGTM!

packages/gestalt/src/Popover/Contents.tsx
diyorbek: To remove an incosistency and TS error. `width` is read from `size` prop. `size` accepted both `underifned` and `null`. This creates confusion. If `size` shouldn't be set, then it should be `undefined`, it doesn't have to be `null`.

PR #3670, opened 2024-07-10
Tokens: removing px units from font size tokens

aarshap:
#### What changed?

We've removed px units from font size tokens. 

#### Why?

There are two reasons we've made this change:

1. Tokens serve multiple platforms, not all of which use px values.

2. Tokens Studio can't perform math on numerical values that contain a unit. We need unitless values for typography tokens that will be part of an upcoming release.

@rlingineni will be updating this PR with an update to our transformer that will apply platform-specific values on compile.


rlingineni:
@diyorbek - this is a good example of a Style Dictionary transformer and how we're using it to add "px" in the outputs

packages/gestalt-design-tokens/build.js
```diff
@@ -771,6 +787,7 @@ StyleDictionary.registerTransformGroup({
     'name/cti/snake',
     'name/conflictFixing',
     'color/hex8android',
```
AlbertCarreras, line 789: Be careful , in Android we have `sp`
Screenshot by Dropbox Capture

AlbertCarreras: iOS uses px.

AlbertCarreras (changes requested):
see comment

AlbertCarreras (approved):

packages/gestalt-design-tokens/build.js
rlingineni: Thanks! Yup this converts to px first.

All snapshots pass with the transformer

PR #3679, opened 2024-07-15
Masonry: Initial proposal for dynamic heights on masonry

manvoid:
#### What changed?

Enables dynamic heights on masonry items. To achieve this I added a new resize observer for the items on masonry, on resize (change on height) keep track of all the columns that need to be pushed down and update the `top` values accordingly on the position cache.

#### Important notes on implementation:

-  The observer is added on the DOM element, but the item reference is a data object, we need a way to map the DOM node to the item data, for this I added a new prop `idxToDataMap`, this way we can assign an idx and use that on the resize handle to get the item data
- When an item height changes if all the items where one column this would be trivial, but since we can have multi column modules it complicates things, when a multi column item is pushed down it can push down more columns than the original item, we need to keep a map of all the affected areas. This is handled on the `recalcHeights` function using the concept of `area` to determine which columns should be moved

#### Other notes

- When items change heights and we have several multi column modules after it, it can happen that the whitespace is increased a lot, there is no easy way to fix that
- This has been only implemented on the legacy Masonry component

### Tests

https://github.com/user-attachments/assets/2fdbc50f-1fef-401f-8f76-9d0f8a4ab518


packages/gestalt/src/Masonry/dynamicHeightsUtils.ts
```diff
@@ -0,0 +1,96 @@
...
+  const changedItemPosition = positionStore.get(changedItem);
+
+  if (!changedItemPosition || changedItemPosition.height === newHeight) {
+    return;
+  }
+
+  measurementStore.set(changedItem, newHeight);
+
+  const { top, left, width, height } = changedItemPosition;
+  const itemsWithPositions = items.filter((item) => item && positionStore.has(item));
```
liuyenwei, line 44: we could simplify this a bit if we just filter out items with positions below the changed item upfront

`position.top > (changedItemPosition.top + changedItemPosition.height)`

packages/gestalt/src/Masonry/dynamicHeightsUtils.ts
```diff
@@ -0,0 +1,96 @@
...
+        .filter((item) => getColumnSpanConfig(item) !== 1)
+        .map((item) => {
+          const position = positionStore.get(item);
+          return position ? { item, position } : undefined;
+        })
+        .filter((itemPosition) => !!itemPosition)
+        .sort((a, b) => a.position.top - b.position.top)
+    : [];
+
+  const areas = multiColumnItemsPositions.reduce(
```
liuyenwei, line 56: can you explain why areas needs to be an array vs just a left + right bound?

packages/gestalt/src/Masonry.tsx
```diff
@@ -180,6 +186,32 @@ export default class Masonry<T> extends ReactComponent<Props<T>, State<T>> {
       props.measurementStore || Masonry.createMeasurementStore();
 
     this.positionStore = props.positionStore || Masonry.createMeasurementStore();
+    this.idxToDataMap = {};
```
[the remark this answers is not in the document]
manvoid: Idealy yes, the items go in order, although since the idx is set to the filtered `itemsToRender` it is not exactly 1-1, if for some reason there is an item on the array that doesn't have a measure but the others do then it is not going to match, if we don't want to use this map we could filter before getting the position index from the items array.

packages/gestalt/src/Masonry.tsx
```diff
@@ -180,6 +186,32 @@ export default class Masonry<T> extends ReactComponent<Props<T>, State<T>> {
...
+
+    this.resizeObserver =
+      /* eslint-disable-next-line no-underscore-dangle */
+      props._dynamicHeights && typeof window !== 'undefined' && this.positionStore
+        ? new ResizeObserver((entries) => {
+            for (let i = 0; i < entries.length; i += 1) {
+              const { target, contentRect } = entries[i];
+              const idx = Number(target.getAttribute('data-grid-item-idx'));
```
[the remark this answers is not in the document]
manvoid: If I do this then tsserver complains:   Property 'dataset' does not exist on type 'Element'. tsserver

packages/gestalt/src/Masonry.tsx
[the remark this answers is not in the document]
manvoid: I removed it for now but we can discuss if this is really an issue.

packages/gestalt/src/Masonry.tsx
```diff
@@ -180,6 +186,32 @@ export default class Masonry<T> extends ReactComponent<Props<T>, State<T>> {
...
+                recalcHeights({
+                  items: this.state.items,
+                  changedItem: item,
+                  newHeight: contentRect.height,
+                  positionStore: this.positionStore,
+                  measurementStore: this.state.measurementStore,
+                  /* eslint-disable-next-line no-underscore-dangle */
+                  getColumnSpanConfig: this.props._getColumnSpanConfig,
+                });
+                this.forceUpdate();
```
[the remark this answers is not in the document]
manvoid: Changed it, added a flag to trigger it only after all the entries have been handled

packages/gestalt/src/Masonry/ItemResizeObserverWrapper.tsx
```diff
@@ -0,0 +1,34 @@
...
+  idx,
+  children,
+}: {
+  resizeObserver: any;
+  idx: number;
+  children: ReactNode;
+}) {
+  const ref = useRef(null);
+
+  useEffect(() => {
```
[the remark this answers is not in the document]
manvoid: Awesome finding!! Thanks! Using layoutEffect fixes the problem

packages/gestalt/src/Masonry/dynamicHeightsUtils.ts
manvoid: I did the change to work only on the items that are after item that changed height. I only filtered by `position.top > changedItemPosition.top` because on the case that the height change is large it can happen that the other logic excludes some small items

packages/gestalt/src/Masonry/dynamicHeightsUtils.ts
manvoid: This is because the affected areas are tied to the height, in certain heights only certai area applies fe just bellow the item but on other heights the areas can grow.

packages/gestalt/src/Masonry.tsx
```diff
@@ -204,30 +204,43 @@ export default class Masonry<T> extends ReactComponent<Props<T>, State<T>> {
...
-                  getColumnSpanConfig: this.props._getColumnSpanConfig,
-                });
-                this.forceUpdate();
+                const item: T = this.state.items[idx];
+                const changedItemPosition = this.positionStore.get(item);
+                const newHeight = contentRect.height;
+
+                if (
+                  changedItemPosition &&
+                  newHeight > 0 &&
```
manvoid, line 224: Added this check just in case it could happen that a the resize is triggered for the item that is not mounted. This also fixed the vistualized issue but adding the layoutEffect also sounds good.

packages/gestalt/src/Masonry/dynamicHeightsUtils.ts
```diff
@@ -0,0 +1,96 @@
...
+  measurementStore: Cache<T, number>;
+  getColumnSpanConfig?: (item: T) => ColumnSpanConfig;
+}) {
+  measurementStore.set(changedItem, newHeight);
+
+  const { top, left, width, height } = changedItemPosition;
+
+  const itemsWithPositions = items.filter((item) => {
+    const position = positionStore.get(item);
+    return position && position.top > changedItemPosition.top;
```
liuyenwei, line 43: should this be `position.top > (changedItemPosition.top + changedItemPosition.height)`? otherwise you'd also be factoring in items which are not below the changed item

packages/gestalt/src/Masonry.tsx
liuyenwei: instead of the if here, can we have `recalcHeight` return whether a change was applied?

```
const didUpdate = recalcHeights();
if (didUpdate) {
 this.forceUpdate()
}
```

packages/gestalt/src/MasonryV2.tsx
```diff
@@ -664,6 +677,40 @@ function Masonry<T>(
...
+                    newHeight: contentRect.height,
+                    positionStore,
+                    measurementStore,
+                    getColumnSpanConfig: _getColumnSpanConfig,
+                  }) || triggerUpdate;
+              }
+            }
+            if (triggerUpdate) {
+              setHeightUpdateTrigger((prev) => prev + 1);
+              forceUpdate();
```
liuyenwei, line 707: we shouldn't need a forceUpdate here if we're already updating state

packages/gestalt/src/Masonry/dynamicHeightsUtils.ts
```diff
@@ -0,0 +1,68 @@
...
+          positionStore.set(item, { ...position, top: position.top + heightDelta });
+          return {
+            left: Math.min(area.left, position.left),
+            right: Math.max(area.right, position.left + position.width),
+          };
+        }
+        return area;
+      },
+      { left, right: left + width } as { left: number; right: number },
+    );
```
liuyenwei, line 60: much cleaner! thank you

liuyenwei (approved):

PR #3686, opened 2024-07-17
Switch: VR Implementation

eniomoura:
#### What changed?

Implemented VR changes to Switch

Light Unselected
image
Light Unselected Focus Ring
image
Light Unselected Hover
image
Light Unselected Pressed
image
Light Selected
image
Light Selected Focus Ring
image
Light Selected Hover
image
Light Selected Pressed
image
Light Unselected Disabled
image
Light Selected Disabled
image
Dark Selected
image
Dark Selected Focus Ring
image
Dark Selected Hover
image
Dark Selected Pressed
image
Dark Unselected
image
Dark Unselected Focus Ring
image
Dark Unselected Hover
image
Dark Unselected Pressed
image
Dark Selected Disabled
image
Dark Unselected Disabled
image
No Pressed Or Hover On Disabled ✅
No Focus Ring On Disabled ✅

#### Why?

To enable Web Visual Refresh.

### Links

- [Jira](https://jira.pinadmin.com/browse/GESTALT-7079)
- [Figma](https://www.figma.com/design/vjhfBsOtHw0wVg67vqwz1v/Gestalt-for-Web?node-id=34852-4411&m=dev)


AlbertCarreras:
Screenshot by Dropbox Capture

AlbertCarreras:
Checking size: seems correct. Screenshot by Dropbox Capture
Screenshot by Dropbox Capture

AlbertCarreras:
Can you add both on/off to this example, in a column. 
Screenshot by Dropbox Capture

AlbertCarreras:
Hover state is not working https://capture.dropbox.com/ES7LBk0KOnLhNx28

AlbertCarreras:
NOtice that the boackground and border all have color for each states 
Screenshot by Dropbox Capture

AlbertCarreras:
They don't look the same 
Screenshot by Dropbox Capture
Screenshot by Dropbox Capture

AlbertCarreras:
Before sending for review:
- manually check all states: regular, disabled, focus, hovered, pressed (if applies)

AlbertCarreras:
@eniomoura Can you check why this is showing a delay? 
Brave Browser - Switch - Gestalt 2024-07-22 at 12 02 56 PM

Compare with Badge, where this doesnt happen

packages/gestalt/src/Switch.tsx
```diff
@@ -62,6 +70,16 @@ export default function Switch({ disabled = false, id, name, onChange, switched
     !disabled ? styles.sliderDark : styles.sliderLight,
   );
 
+  const sliderVrStyles = classnames(
+    {
+      [styles.blackSlider]: !switched,
+      [styles.disabledSlider]: disabled && !switched,
+      [styles.slsiderVrOn]: switched,
```
AlbertCarreras, line 77: typo slider

packages/gestalt/src/Switch.tsx
```diff
@@ -62,6 +70,16 @@ export default function Switch({ disabled = false, id, name, onChange, switched
     !disabled ? styles.sliderDark : styles.sliderLight,
   );
 
+  const sliderVrStyles = classnames(
+    {
+      [styles.blackSlider]: !switched,
```
AlbertCarreras, line 75: instead of black can we say enabledSlider?

packages/gestalt/src/Switch.css
```diff
@@ -1,5 +1,7 @@
 :root {
   --g-switch-height: 24px;
+  --g-vr-slider-height-on: 18px;
```
AlbertCarreras, line 3: can we avoid vr if there's no conflicting name? we can do `--g-slider-height-on: 18px;` this prevents us from having to rename thigs later on

packages/gestalt-design-tokens/tokens/classic/comp-web-color-dark.json
```diff
@@ -428,6 +428,22 @@
           "selected": {
             "value": "#ababab"
           }
+        },
+        "hover": {
+          "selected": {
+            "value": "#E7E6DF"
```
AlbertCarreras, line 434: where are all this color coming from?

AlbertCarreras: Are all this these tokens?

AlbertCarreras: Screenshot by Dropbox Capture

packages/gestalt/src/Switch.css
eniomoura: Done, I was thinking of explicitly using VR names to make it easier to find which changes are VR later, but I suppose we can use commits for that

packages/gestalt/src/Switch.tsx
eniomoura: Fixed

packages/gestalt/src/Switch.tsx
eniomoura: Done

packages/gestalt-design-tokens/tokens/classic/comp-web-color-dark.json
eniomoura: From the token translation guide, I couldn't get the dark mode tokens from Figma, but I confirmed they were correct by looking at the component in the playground mode

eniomoura:
> I don't think there's an active state. Can you check if active is being triggered at all? If not we might be able to simplify....
> 
> https://developer.mozilla.org/en-US/docs/Web/CSS/:active

Turns out it was triggering, but hover is equal in priority and you have to declare the css on the "hover, active" order or it won't show up... TIL

PR #3689, opened 2024-07-18
Toast: VR Fixes

eniomoura:
#### What changed?

Fixed indicated errors in the VR implementation for Toast:
- Error text is not bold in VR, now normal text:
image
- Buttons should be white, not default color, now fixed (with notes below)
image
- RTL was broken with usage of paddingLeft/Right, now fixed with paddingInlineStart/End
image

It was also indicated a missing max width for the Toast text, but it seems to be working (716px, with a cutoff to 100% of the screen in mobile as per [Figma](https://www.figma.com/design/rVnMTnlB64QbKrF5c6KjUl/Gestalt-Toast-Design?node-id=16-402&m=dev)):
image
image

#### Why?

To ensure quality delivery of the VR implementation


packages/gestalt/src/Toast/PrimaryAction.tsx
```diff
@@ -39,6 +39,7 @@ export default function PrimaryAction({ accessibilityLabel, label, size = 'lg',
   return (
     <Button
       accessibilityLabel={accessibilityLabel}
+      color="white"
```
eniomoura, line 42: @AlbertCarreras This might be wrong -- The default color for Button is changing to white/dark on VR right? Because this 'white' prop is meant to be used in front of dark images, and will not work with dark mode.
I think the right approach is to leave as it was before and make the default white/black when we get to implementing Button VR.

AlbertCarreras: can we add white "conditionally to  VR"

packages/gestalt/src/sharedSubcomponents/OverridingSpan.tsx
```diff
@@ -31,7 +37,7 @@ export default function OverridingSpan({
       : styles.textColorOverrideLight;
   }
   // Error type enforces bold weight
-  if (isError) {
+  if (isError && !isInExperiment) {
```
AlbertCarreras, line 40: I think we still need the override to enforce the color when we pass Text rather than string

packages/gestalt/src/sharedSubcomponents/thumbnailSubcomponents.tsx
```diff
@@ -124,18 +131,9 @@ export function Message({
...
-  });
-
   return (
     <Box aria-hidden>
-      <Mask
-        height={isInExperiment ? SIZE_VR_THUMBNAIL : SIZE_THUMBNAIL}
-        rounding={2}
-        width={isInExperiment ? SIZE_VR_THUMBNAIL : SIZE_THUMBNAIL}
-      >
+      <Mask height={SIZE_THUMBNAIL} rounding={2} width={SIZE_THUMBNAIL}>
```
AlbertCarreras, line 136: why are we removing `isInExperiment ? SIZE_VR_THUMBNAIL : SIZE_THUMBNAIL`?

packages/gestalt/src/Toast/PrimaryAction.tsx
AlbertCarreras: color={ experiment ?  'white' : undefined}

packages/gestalt/src/sharedSubcomponents/OverridingSpan.tsx
AlbertCarreras: { experiment ? styles.VRtextColorOverrideError : styles.textColorOverrideError }

packages/gestalt/src/sharedSubcomponents/thumbnailSubcomponents.tsx
AlbertCarreras: OK.

packages/gestalt/src/Toast/PrimaryAction.tsx
eniomoura: DONE

packages/gestalt/src/sharedSubcomponents/OverridingSpan.tsx
eniomoura: DONE

AlbertCarreras (approved):

PR #3702, opened 2024-07-27
Masonry: fix importing types on Masonry/defaultLayout.ts

antoniopresto:
fix importing types on Masonry/defaultLayout.ts


packages/gestalt/src/Masonry/defaultLayout.ts
```diff
@@ -1,9 +1,8 @@
-import { Align, Layout, LoadingStateItem } from 'gestalt/src//Masonry/types';
```
antoniopresto, line 1: was using  `//` in ```src//Masonry```

AlbertCarreras (approved):

PR #3703, opened 2024-07-29
 TextArea: VR Implementation

AlbertCarreras:
TextArea: VR Implementation

## Classic
Screenshot by Dropbox Capture
## VR
Screenshot by Dropbox Capture
Screenshot by Dropbox Capture

## Classic
Screenshot by Dropbox Capture
## VR
Screenshot by Dropbox Capture
Screenshot by Dropbox Capture

## Classic
Screenshot by Dropbox Capture
## VR
Screenshot by Dropbox Capture
Screenshot by Dropbox Capture

### DIFF
--- a/packages/gestalt/src/TextArea.tsx
+++ b/packages/gestalt/src/TextArea.tsx
@@ -7,6 +7,8 @@ import FormErrorMessage from './sharedSubcomponents/FormErrorMessage';
 import FormHelperText from './sharedSubcomponents/FormHelperText';
 import FormLabel from './sharedSubcomponents/FormLabel';
 import styles from './TextArea.css';
+import VRTextArea from './TextArea/VRTextArea';
+import useInExperiment from './useInExperiment';
 
 const ROW_HEIGHT = 24;
 const INPUT_PADDING_WITH_TAGS = 20;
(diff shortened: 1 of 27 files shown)


eniomoura (approved):
LGTM

packages/gestalt/src/TextArea.tsx
```diff
@@ -215,7 +222,29 @@ const TextAreaWithForwardRef = forwardRef<HTMLTextAreaElement, Props>(function T
     maxHeight: rows * ROW_HEIGHT + INPUT_PADDING_WITH_TAGS,
   } as const;
 
-  return (
+  return isInVRExperiment ? (
```
rlingineni, line 225: nit: can we return the component and not use an `if,else`

Also, is this an area where can use spread `props` safely?

Since it's an internal component with the exact same props

Feels like it could look cleaner in 2 lines, if there's nothing special happening:
```
if(isInVRExperiement)
`return <VRTextArea  ...props/>`
```

rlingineni:
Are there playwright snapshots for VR?

packages/gestalt/src/TextArea.tsx
AlbertCarreras: I like being explicit with props, I'm not passing all props to all variants. I like destructuring at the very top and setting defaults there.

AlbertCarreras: The if/selse is being refactored in the follow up PR where I implement tags

AlbertCarreras:
> Are there playwright snapshots for VR?

Good question. I think we could do snapshots but there are so many variances that we need to take into account that the most reliable thing is the docs. 

I think we could take snapshots.

Maybe in a followup we could do that. 

Overall we need a better strategy for snapshots... we should cover al states etc,...

PR #3709, opened 2024-07-31
Toast: VR implementation changes

eniomoura:
#### What changed?

Changed default accessibility label for Toast to 'dismiss message' instead of 'dismiss toast'. Also changed VR color for the success icon to 'default', as a special case for Toast.

#### Why?

https://pinterest.slack.com/archives/C014X9LTRCN/p1722364602407439?thread_ts=1721244858.309079&cid=C014X9LTRCN

### DIFF
--- a/docs/pages/web/utilities/defaultlabelprovider.tsx
+++ b/docs/pages/web/utilities/defaultlabelprovider.tsx
@@ -275,7 +275,7 @@ function getLabelsTable(/* fallbackLabels: { [string]: { [string]: mixed } } */)
     {
       component: 'Toast',
       prop: 'accessibilityDismissButtonLabel',
-      label: 'Dismiss toast',
+      label: 'Dismiss message',
     },
     {
       component: 'Toast',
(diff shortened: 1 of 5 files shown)


rlingineni (approved):
Approved, but I think because Default Label Provider is changed, this is a major

docs/pages/web/utilities/defaultlabelprovider.tsx
```diff
@@ -275,7 +275,7 @@ function getLabelsTable(/* fallbackLabels: { [string]: { [string]: mixed } } */)
     {
       component: 'Toast',
       prop: 'accessibilityDismissButtonLabel',
-      label: 'Dismiss toast',
```
AlbertCarreras, line 278: let's make sure this gets updated in Pinboard!

packages/gestalt/src/sharedSubcomponents/thumbnailSubcomponents.tsx
```diff
@@ -186,7 +190,7 @@ export function TypeThumbnail({ type }: { type: 'default' | 'success' | 'error'
         >
           <Icon
             accessibilityLabel={accessibilityIconSuccessLabel}
-            color="success"
+            color={isInExperiment ? 'default' : 'success'}
```
AlbertCarreras, line 193: cool

AlbertCarreras (approved):

docs/pages/web/utilities/defaultlabelprovider.tsx
eniomoura: Doing it in https://github.com/pinternal/pinboard/pull/83102
Blocked by AI that insists I use Dismiss instead of Dismiss 🤖
I made a post on i18n to check about that.

PR #3721, opened 2024-08-08
Tokens: Hover token RGBA fixes

rlingineni:
# Pull Request Instructions

Adding a missing StyleDictionary transform to use RGBA values correctly in final transform outputs in CSS.

Previously, the tokens would not resolve correctly because hover tokens are set up as token studio rgba values, and they would output to hex values when referenced, but they need to be proper rgb values

image


AlbertCarreras (approved):

packages/gestalt-design-tokens/transformers/hexToRgba.js
```diff
@@ -0,0 +1,56 @@
...
+  if (val === undefined) return undefined;
+
+  const transformHEXRGBa = (tokVal) => {
+    const regex = /rgba\(\s*(?<hex>#.+?)\s*,\s*(?<alpha>\d*(\.\d*|%)*)\s*\)/g;
+    return tokVal.replace(regex, (match, hex, alpha) => {
+      try {
+        const { r, g, b } = tinycolor(hex).toRgb();
+        return `rgba(${r}, ${g}, ${b}, ${alpha})`;
+      } catch (e) {
+        console.warn(`Tried parsing "${hex}" as a hex value, but failed.`);
```
eniomoura, line 23: Remove or supress, then ci should pass

eniomoura (approved):

PR #3729, opened 2024-08-23
ButtonToggle: Add bounce animation

diyorbek:
### Summary

1. Added click animations to VR ButtonToggle. 
2. `compressStyle` from `useTapFeedback` is not applied.
3. Used CSS animations instead of transitions. Because clicks can happen within milliseconds and transitions may not be noticable (see "before" examples below).

#### What changed?

**BEFORE**
ezgif-1-7e543f1e09
ezgif-1-ec45a3a6fb

**AFTER**
ezgif-4-500140c784
ezgif-1-1ce8dc860c

### Links

- [Jira](https://jira.pinadmin.com/browse/GESTALT-8275)
- [TDD](https://docs.google.com/document/d/1J6QYhY81mnpZRc0e7wTW7EelUNW7i_u5C8iQsRO6soo/edit?usp=sharing)
- [Figma](https://www.figma.com/design/pMibC8csfCbNMU4jgLVUJh/Motion-Pilot-specs?node-id=2006-82&t=t1XAu3T8peiaefDL-1)

### Checklist

- [ ] Added unit tests

### DIFF
--- a/packages/gestalt/src/ButtonToggle.tsx
+++ b/packages/gestalt/src/ButtonToggle.tsx
@@ -180,6 +180,16 @@ const ButtonToggleWithForwardRef = forwardRef<HTMLButtonElement, Props>(function
         onFocus={(event) => {
           onFocus?.({ event });
         }}
+        onKeyDown={(e) => {
+          if (isInVRExperiment) {
+            buttonToggleAnimation.handleKeyDown(e);
+          }
+        }}
+        onKeyUp={(e) => {
+          if (isInVRExperiment) {
+            buttonToggleAnimation.handleKeyUp(e);
+          }
+        }}
         onMouseDown={() => {
           handleMouseDown();
           if (isInVRExperiment) buttonToggleAnimation.handleMouseDown();
@@ -290,6 +300,16 @@ const ButtonToggleWithForwardRef = forwardRef<HTMLButtonElement, Props>(function
       }}
       onFocus={(event) => {
         onFocus?.({ event });
+      }}
+      onKeyDown={(e) => {
+        if (isInVRExperiment) {
+          buttonToggleAnimation.handleKeyDown(e);
+        }
+      }}
+      onKeyUp={(e) => {
+        if (isInVRExperiment) {
+          buttonToggleAnimation.handleKeyUp(e);
+        }
       }}
       onMouseDown={() => {
         handleMouseDown();
--- a/packages/gestalt/src/ButtonToggle.tsx
+++ b/packages/gestalt/src/ButtonToggle.tsx
@@ -4,6 +4,7 @@ import borderStyles from './Borders.css';
 import styles from './ButtonToggle.css';
 import ColorPicker, { SkinColor } from './ButtonToggle/ColorPicker';
 import LabeledThumbnail from './ButtonToggle/LabeledThumbnail';
+import useButtonToggleAnimation from './ButtonToggle/useButtonToggleAnimation';
 import { useColorScheme } from './contexts/ColorSchemeProvider';
 import { useGlobalEventsHandlerContext } from './contexts/GlobalEventsHandlerProvider';
 import Flex from './Flex';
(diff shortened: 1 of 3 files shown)


packages/gestalt/src/ButtonToggle.tsx
```diff
@@ -274,8 +301,24 @@ const ButtonToggleWithForwardRef = forwardRef<HTMLButtonElement, Props>(function
       onFocus={(event) => {
         onFocus?.({ event });
       }}
-      onMouseDown={handleMouseDown}
-      onMouseUp={handleMouseUp}
+      onKeyDown={(e) => {
```
rlingineni, line 304: My only feedback is that this seems repeated, but as I thought about, I'm not sure if there is a better way to abstract this. Can you pass in the 'experiment' name or the flag, to the hook itself. 

e.g. in `useButtonToggleAnimation` you might be able to do the if statement there instead, so in the future for more VR animation experiments, we can do all of that in that one method

rlingineni (approved):

packages/gestalt/src/ButtonToggle/ButtonToggleAnimation.css
```diff
@@ -0,0 +1,34 @@
...
+  animation: scaleUp calc(var(--base-motion-duration-150) / 2)
+    var(--base-motion-easing-linear) forwards;
+}
+
+@keyframes scaleDown {
+  from {
+    transform: scale(1);
+  }
+  to {
+    transform: scale(0.95);
```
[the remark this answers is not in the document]
diyorbek: I think we do. Actually we also need component tokens for the easings and durations. I need to talk about this with Kate. But for now this is all experimental. We'll standardize the motion implementation better as we implement more.

packages/gestalt/src/ButtonToggle.tsx
```diff
@@ -204,7 +237,6 @@ const ButtonToggleWithForwardRef = forwardRef<HTMLButtonElement, Props>(function
     [styles.thumbnailMd]: size === 'md' && graphicSrc,
     [styles.thumbnailSm]: size === 'sm' && graphicSrc,
     [styles[color]]: !disabled && !selected,
-    [touchableStyles.tapCompress]: !disabled && isTapping,
```
AlbertCarreras, line 207: why did we removed this one? Should have stayed only in `!isInVRExperiment`

AlbertCarreras:
@diyorbek We want the animation in both mouse and keyboard click right?

In classic, it was only in mouse, now it's in both. No biggie, just curious.

Brave Browser - ButtonToggle - Gestalt 2024-08-27 at 12 27 03 PM
Brave Browser - ButtonToggle - Gestalt 2024-08-27 at 12 27 49 PM

diyorbek:
@AlbertCarreras I didn't have any specific instructions on keyboard clicks. I think the animation should be there regardless of the source of click.

packages/gestalt/src/ButtonToggle.tsx
diyorbek: It didn't have any effect. Tap compression is controlled by transform CSS property from `useTapFeedback`.

packages/gestalt/src/ButtonToggle.tsx
diyorbek: @AlbertCarreras What's your take on this? I just wanted to have the experiment related code to be in one file. So that when we eventually remove the experiment code, we only change one file instead of multiple.

AlbertCarreras: it's stylistic. it's not super dry as is, but it's very explicit in the component itself and less opaque. 

Technically this animations are VR only so we can gate outside.

If we gate inside, then we'll reuse it in more components and it will be more dry. 

It doesnt bother me either way. Not opinionated.

packages/gestalt/src/ButtonToggle.tsx
AlbertCarreras: Does this mean we can clean all the isTapping from other components?

diyorbek: i think so, yes

PR #3742, opened 2024-09-10
Masonry: Ensure zero height items dont have a gutter

liuyenwei:
### Summary
This PR updates Masonry to ensure that items with zero height do not actually occupy any space

At the moment, if Masonry renders a component with zero height (i.e. an item that returns `null`), it could still end up occupying space since we still add the `gutter` to the position. This results in subsequent items in the column being pushed down.

image

This PR addresses this by updating the `heightAndGutter` logic to only return height + gutter if height > 0. I also took the opportunity to clean up the code a tad


AlbertCarreras (approved):

AlbertCarreras:
prettier nit: 
Screenshot by Dropbox Capture

PR #3744, opened 2024-09-10
Popover, Dropdown, PopoverEducational: support `forcePosition` prop

AlbertCarreras:
BREAKING CHANGE:

We are deprecating idealPosition: forceDown

```
--component=Popover or Dropdown or PopoverEducational
--prop=idealDirection
--value=forceDown

```

Popover, Dropdown, PopoverEducational: support `forcePosition` prop

Test example here: https://deploy-preview-3744--gestalt.netlify.app/web/popover#Ideal-direction


packages/gestalt/src/Popover/InternalPopover.tsx
```diff
@@ -18,7 +18,7 @@ type Props = {
   color?: Color;
   onKeyDown?: (arg1: { event: React.KeyboardEvent<HTMLElement> }) => void;
   id?: string;
-  idealDirection?: 'up' | 'right' | 'down' | 'left' | 'forceDown';
+  idealDirection?: 'up' | 'right' | 'down' | 'left' | 'forceDown' | 'forceRight';
```
elnanni, line 21: Would this be a better approach if taken from packages/gestalt/src/utils/positioningTypes.ts using MainDirections?

diyorbek:
@AlbertCarreras I think we should reconsider the API for force positioning. What if we'll need forceLeft, forceUp in the future? I would suggest to introduce `forceDirection: boolean` prop which will just disable auto-flipping/positioning behavior. We could just get rid of "force" prefixed positions.

diyorbek:
The new prop doesn't have to be `forceDirection`. It could be `disableAutoFlip` or something like that. The point is this props will disable the auto-flipping beahvior of the popover.

packages/gestalt/src/Popover/Contents.tsx
```diff
@@ -54,12 +66,18 @@ export default function Contents({
   overflow,
 }: Props) {
   const caretRef = useRef<HTMLElement | null>(null);
-  const idealPlacement = idealDirection ? DIRECTIONS_MAP[idealDirection] : 'top';
+
+  let idealPlacement: 'top' | 'right' | 'bottom' | 'left' = 'top';
```
diyorbek, line 70: Maybe this is more accurate: 

```suggestion
  let idealPlacement: Side = 'top';
```

packages/gestalt/src/Popover/usePopover.ts
```diff
@@ -74,13 +71,20 @@ export default function usePopover({
...
   scrollBoundary,
   hideWhenReferenceHidden,
   onPositioned,
 }: Props): UseFloatingReturn {
-  const isForceDown = direction === 'forceDown';
-  const placement = isForceDown ? 'bottom' : direction;
+  let placement: Placement = direction ?? 'bottom';
+  const isRtl = typeof document === 'undefined' ? false : document?.dir === 'rtl';
+
+  if (isRtl && direction === 'left') {
```
diyorbek, line 83: @AlbertCarreras have you tested this bahvior? I though @floating-ui library already handled RTL. 🤔

diyorbek (approved):
LGTM

packages/gestalt/src/Popover/Contents.tsx
AlbertCarreras: DONE

packages/gestalt/src/Popover/usePopover.ts
AlbertCarreras: It might flip accordingly... but if we force behavior idk... I'll test it anyways.

AlbertCarreras: just tested it... no, when forced... it stays i the right/left, cos they don't do end/start and we read LRT/LTR from the document

PR #3771, opened 2024-09-20
Icons: Add badge-icons

rlingineni:
# Pull Request Instructions

Init add of new Badge Component Icons

### Checklist

- [ ] Added unit tests


AlbertCarreras:
This icons look incorrect. 
Screenshot by Dropbox Capture

AlbertCarreras:
They should be 24-24 and we can adjust size internally. This should be just moving them out to an internal folder and removing them from the index,

AlbertCarreras (approved):
Run yarn tsc please before merging. LGTM

packages/gestalt/src/icons-vr-theme/components/.eslintrc.json
```diff
@@ -0,0 +1,5 @@
+{
+  "rules": {
+    "max-len": 0
```
diyorbek, line 3: just curious, what does this rule do? (in the context of this PR)

diyorbek (approved):
LGTM

packages/gestalt/src/IconCompact.tsx
```diff
@@ -0,0 +1,101 @@
...
+  /**
+   * Properly positions Icon relative to an inline element, such as Text using the inline property.
+   */
+  inline?: boolean;
+  /**
+   * Use a number for pixel sizes or a string for percentage based sizes.
+   *
+   * See the [size](https://gestalt.pinterest.systems/web/icon#Size) variant to learn more.
+   */
+  size?: number | string;
```
AlbertCarreras, line 59: should we limit size here?

packages/gestalt/src/Icon/InternalIcon.tsx
```diff
@@ -0,0 +1,185 @@
...
+    styles[color],
+    styles.icon,
+    { [styles.iconBlock]: !inline },
+  );
+  const isInExperiment = useInExperiment({
+    webExperimentName: 'web_gestalt_visualRefresh',
+    mwebExperimentName: 'web_gestalt_visualRefresh',
+  });
+
+  const getIconPath = (iconToUse?: IconName) => {
```
AlbertCarreras, line 101: i think we can simplify a lot this logic. 

we can add a size='compact' | 'default' prop and set the viewBox to the right size

packages/gestalt/src/Icon/InternalIcon.tsx
```diff
@@ -0,0 +1,185 @@
...
+  | 'warning'
+  | 'info'
+  | 'recommendation'
+  | 'inverse'
+  | 'shopping'
+  | 'brandPrimary'
+  | 'light'
+  | 'dark';
+
+type IconName = keyof typeof icons | keyof typeof compactIconsVR | keyof typeof compactIconsClassic;
```
AlbertCarreras, line 24: type IconName = keyof typeof icons | keyof typeof compactIconsVR | keyof typeof compactIconsClassic;

do we need : ```| keyof typeof compactIconsClassic;```?

docs/pages/web/icon.tsx
```diff
@@ -222,6 +223,20 @@ Should be used sparingly and only in places where the UI is very dense and a lar
             title="Custom SVG icon"
           />
         </MainSection.Subsection>
+        <MainSection.Subsection
```
AlbertCarreras, line 226: Can we do a follow up PR adding a single page for IconCompact?

packages/gestalt/src/IconCompact.tsx
AlbertCarreras: they should be '16' the only accepted value for now

packages/gestalt/src/Icon/InternalIcon.tsx
rlingineni: I guess all VR icons will contain classic right? So maybe you are correct that we don't need this

docs/pages/web/icon.tsx
rlingineni: Yes - I think we should. I wanted to wait since there's not too many right now. and it's a replica of `Icon`

packages/gestalt/src/IconCompact.tsx
rlingineni: No I think this size is different afaik. This is the SVG height/width. But what we mean by compact Icons is that they are thinner (they can in theory be scaled up) to larger viewboxes.

image

rlingineni: But maybe you are right in that they should be fixed

rlingineni: Removed

PR #3774, opened 2024-09-23
Spinner: VR implementation changes

diyorbek:
### Summary

| classic | VR  |
|---|---|
| ezgif-7-1cc6dea540 | ezgif-7-6856616749

### Links

- [Jira](https://jira.pinadmin.com/browse/GESTALT-8377)
- [Figma](https://www.figma.com/design/pMibC8csfCbNMU4jgLVUJh/Motion-Pilot-specs?node-id=2118-2350&m=dev)

### Checklist

- [ ] Added unit tests


AlbertCarreras:
Is it meant to be transparent ? 
Screenshot by Dropbox Capture

AlbertCarreras:
There no dark mode version? 
Screenshot by Dropbox Capture

AlbertCarreras (approved):
Looks good. 

couple questions.

packages/gestalt/src/Spinner/VRSpinner.css
```diff
@@ -0,0 +1,111 @@
+.spinner {
+  --size: 48px;
```
AlbertCarreras, line 2: is this size redundant?

packages/gestalt/src/Spinner/VRSpinner.css
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  100% {
+    transform: scale(1);
+  }
+}
+
+@keyframes colors {
+  0% {
+    background-color: #d452d1;
```
AlbertCarreras, line 97: nit, not for now but we could component tokenize these colors

diyorbek:
> Is it meant to be transparent ? Screenshot by Dropbox Capture

yes, it is.

diyorbek:
> There no dark mode version? Screenshot by Dropbox Capture

AFAIK, no

packages/gestalt-design-tokens/tokens/vr-theme/comp/spinner.json
```diff
@@ -0,0 +1,34 @@
+{
+  "comp": {
```
AlbertCarreras, line 2: can we moved the colors ones to comp-web-color-light.json?


 
Screenshot by Dropbox Capture

AlbertCarreras (approved):

packages/gestalt-design-tokens/tokens/vr-theme/comp/spinner.json
```diff
@@ -0,0 +1,34 @@
...
+    "spinner": {
+      "size": {
```
AlbertCarreras, line 4: This sizes... I'd keep them hard coded cos we don't have sizes right now. We have spacing, but I see spacing more like paddings and margins, more than heights and width.

packages/gestalt/src/Spinner/VRSpinner.css
```diff
@@ -0,0 +1,111 @@
...
+  width: var(--g-dot-space);
+}
+
+.spinnerFrame > div {
+  animation-composition: add;
+  animation-duration: 800ms, 1.8s;
+  animation-iteration-count: infinite;
+  animation-name: scale, colors;
+  animation-timing-function: linear;
+  border-radius: 50%;
```
AlbertCarreras, line 38: can we tokenize this? is this equal to circle?

packages/gestalt/src/Spinner/VRSpinner.css
```diff
@@ -0,0 +1,111 @@
...
+.dot1 {
+  animation-delay: 0s;
+  left: 50%;
+  top: 0;
+  transform: translate3d(-50%, -50%, 0);
+}
+
+.dot2 {
+  animation-delay: 100ms, 0s;
+  bottom: 0;
```
AlbertCarreras, line 53: here we can do space-0 or sema-spacing-0

packages/gestalt/src/Spinner/VRSpinner.tsx
```diff
@@ -0,0 +1,31 @@
...
+  accessibilityLabel?: string;
+  delay?: boolean;
+  show: boolean;
+  size?: 'sm' | 'md' | 'lg';
+};
+
+export default function Spinner({ accessibilityLabel, delay = true, show, size = 'md' }: Props) {
+  const { accessibilityLabel: accessibilityLabelDefault } = useDefaultLabelContext('Spinner');
+
+  if (!show) return <div />;
```
AlbertCarreras, line 15: why not returning null?

.stylelintrc.json
```diff
@@ -21,7 +21,7 @@
         "ignoreProperties": ["composes"]
       }
     ],
-    "custom-property-pattern": "g|color|font|elevation|space|opacity|motion|font-.+",
+    "custom-property-pattern": "g|color|font|elevation|space|opacity|motion|font-.+|size",
```
AlbertCarreras, line 24: yeah I'd avoid this size one. as we don't have size categories yet!

rlingineni: Can we support a size category? 

@diyorbek are these size tokens provided by design

packages/gestalt/src/Spinner/VRSpinner.css
```diff
@@ -0,0 +1,111 @@
...
+.lg {
+  --g-size: var(--comp-spinner-size-lg);
+}
+
+.md {
+  --g-size: var(--comp-spinner-size-md);
+}
+
+.sm {
+  --g-size: var(--comp-spinner-size-sm);
```
rlingineni, line 18: what does --g-size mean? / where is it used?

rlingineni: Oh I see, you're using it above in the `.spinner` context

rlingineni: I think - to simplify this further -> ` --g-dot-size: calc(var(--g-size) / 3);`  is actually 3 variations of a token.

So in theory, what this is effectively making is almost 9 variations of the calculated set

```
--g-dot-size-sm
--g-dot-size-lg
--g-dot-size-md
...
```

# Option 1
So maybe this may not be right in one sense. Because that's a lot of tokens.

I don't think the token should have the size baked into it. Maybe the way to architect this is with singular tokens.

```
--comp-spinner-size-height
--comp-spinner-size-width
--comp-dot-size
--comp-dot-space
```

I think the ideal approach is we have sizing similar to dark mode - and we can override the tokens. We need a sizing provider, and it needs to handle sizes across everything. I'm not sure what our bandwith is to do that rn.

# Option 2

You might be able to move this value definition into JS instead, and then keep everything as is

```
const map = {md:COMP_SPINNER_SIZE_MD, lg:COMP_SPINNER_SIZE_LG, sm:COMP_SPINNER_SIZE_SM'}

 <div style={{
 '--g-size':{map[size]}
 }} />
    
```

packages/gestalt/src/Spinner/VRSpinner.css
```diff
@@ -0,0 +1,111 @@
...
+
+.dot1 {
+  animation-delay: 0s;
+  left: 50%;
+  top: 0;
+  transform: translate3d(-50%, -50%, 0);
+}
+
+.dot2 {
+  animation-delay: 100ms, 0s;
```
rlingineni, line 52: should this be a motion token?

packages/gestalt/src/Spinner/VRSpinner.css
diyorbek: i dont understand why we should tokenize the dots of spinner. these rules are specific to the component and private.

diyorbek: We could tokenize all the calculations in the file too. but i see no point. we could absolutely tokenize every single value here, but we don't do it, right?

packages/gestalt/src/Spinner/VRSpinner.css
diyorbek: the same reason as [here](https://github.com/pinterest/gestalt/pull/3774/files#r1781101993).

I see these values as private.

diyorbek: We were thinking about adding delay tokens but they don't appear as much. Maybe further down the line we can consider if there're many delayed animations, or some kind of patterns grow.

packages/gestalt/src/Spinner/VRSpinner.css
diyorbek: the same reason as [here](https://github.com/pinterest/gestalt/pull/3774/files#r1781101993)

packages/gestalt-design-tokens/tokens/vr-theme/comp/spinner.json
diyorbek: I kept the color tokens in the VR theme as component tokens. Then mapped to them in comp-web-color-light.json in VR mapping theme

packages/gestalt/src/Spinner/VRSpinner.css
diyorbek: THanks @rlingineni ! I went with option 2. 👍

packages/gestalt-design-tokens/tokens/classic/comp-web-color-dark.json
```diff
@@ -436,6 +436,19 @@
           }
         }
       },
+      "spinner": {
```
AlbertCarreras, line 439: you can remove these ones

packages/gestalt-design-tokens/tokens/vr-theme-web-mapping/comp-web-color-light.json
```diff
@@ -259,6 +259,19 @@
           }
         }
       },
+      "spinner": {
```
AlbertCarreras, line 262: you dont need this either

packages/gestalt-design-tokens/tokens/vr-theme/comp/spinner.json
rlingineni: the structure should be vr-theme/comp/spinner/default.json

packages/gestalt-design-tokens/tokens/vr-theme-web-mapping/comp-web-color-light.json
rlingineni: +1 - the VR tokens will be pulled in here

packages/gestalt/src/Spinner/VRSpinner.css
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  100% {
+    transform: scale(1);
+  }
+}
+
+@keyframes colors {
+  0% {
+    background-color: var(--color-background-spinner-color-pink);
```
rlingineni, line 80: I thought the naming for VR-Theme is like this:
```
-comp-spinner-color-background-default
```

https://docs.google.com/document/d/1SMLBtYQtPrIg7JDBV6o0jymkdQ2ItocmI_sOtNhOxIM/edit

Could we use the VR Naming

rlingineni (approved):
Approve - but can we only add these tokens only in the VR-Theme? 


I'm not seeing them used in the classic spinner

PR #3786, opened 2024-09-30
Tokens: Refactor Token Build Script

rlingineni:
# Pull Request Instructions

This is a refactor of our `build.js` token script which was previously hard to maintain and read. 

Everything now lives under the `design-tokens/src` directory.

| File  | Function  |
|---|---|
|  formatters/* |This is a place for all formatters. Formatters are used when generating files.    |  
|   platforms/* |  This is where the output file for each platform is defined |  
|  transformers/* |  This contains all transformer registrations and transformation groups that are utilized by each platform. |
|  utils/* | This contains any utility functions we use. For example, a hex to rgba function. |
|  filters.js | This contains all of the filters we use in the final build when restricting to specific token sets |
|build.js| This is the entrypoint where we build for each platform


packages/gestalt-design-tokens/src/build.js
```diff
@@ -0,0 +1,55 @@
...
+
+const { registerTokenTransformGroups } = require('./transformers/transformGroups');
+const { registerTokenTransforms } = require('./transformers/transforms');
+const { registerTokenFilters } = require('./filters');
+const { getWebConfig } = require('./platforms/web');
+const { getAndroidConfiguration } = require('./platforms/android');
+const { getIOSConfiguration } = require('./platforms/ios');
+const { registerWebFormats } = require('./formatters/registerWebFormats');
+
+StyleDictionary.registerFileHeader({
```
AlbertCarreras, line 11: shoudl we move this out too to keep everything isolated?

AlbertCarreras (approved):

packages/gestalt-design-tokens/src/build.js
rlingineni: Yes - good idea

PR #3797, opened 2024-10-07
RadioGroupButton: VR motion implementation

diyorbek:
### Summary

VR RadioGroupButton motion implementation WITHOUT other VR specific styles.

#### What changed?

**BEFORE**
ezgif-7-4b5c4062c9

**AFTER**
ezgif-1-a2e9bd38d7

### Links

- [Jira](https://jira.pinadmin.com/browse/GESTALT-8378)
- [Figma](https://www.figma.com/design/pMibC8csfCbNMU4jgLVUJh/Motion-Pilot-specs?node-id=2163-7303&t=rCSdnETB8qo1SoyQ-1)

### Checklist

- [ ] Added unit tests


packages/gestalt/src/ButtonToggle.tsx
```diff
@@ -186,7 +186,7 @@ const ButtonToggleWithForwardRef = forwardRef<HTMLButtonElement, Props>(function
   const isDarkModeRed = isDarkMode && color === 'red';
   const { isFocusVisible } = useFocusVisible();
 
-  const buttonToggleAnimation = useButtonToggleAnimation();
+  const buttonToggleAnimation = useTapScaleAnimation();
```
AlbertCarreras, line 189: is this expected?

AlbertCarreras: just renaming i guess

packages/gestalt/src/RadioGroupButton.tsx
```diff
@@ -130,27 +140,78 @@ const RadioGroupButtonWithForwardRef = forwardRef<HTMLInputElement, Props>(funct
...
+
   const styleSize = size === 'sm' ? controlStyles.sizeSm : controlStyles.sizeMd;
 
   const bgStyle = disabled && !checked ? styles.BgDisabled : styles.BgEnabled;
 
   const { isFocusVisible } = useFocusVisible();
 
   const { parentName } = useRadioGroupContext();
+
+  const tapScaleAnimation = useTapScaleAnimation();
```
AlbertCarreras, line 153: nit  

 const { handleMouseDown, handleMouseUp } = useTapScaleAnimation();

packages/gestalt/src/RadioGroupButton.tsx
```diff
@@ -130,27 +140,78 @@ const RadioGroupButtonWithForwardRef = forwardRef<HTMLInputElement, Props>(funct
...
+  const radioButtonStyles = isInVRExperiment
+    ? classnames(styleSize, layoutStyles.relative, borderStyles.circle, tapScaleAnimation.classes, {
+        [focusStyles.accessibilityOutlineFocus]: focused && isFocusVisible,
+      })
+    : classnames(bgStyle, borderColor, borderWidth, styleSize, styles.RadioButton, {
+        [focusStyles.accessibilityOutlineFocus]: focused && isFocusVisible,
+      });
+
+  const checkedStyles = classnames(
+    styles.VRRadioButton,
```
AlbertCarreras, line 170: can we dry this so we know which props are reused in unchecked and checked and which ones are for one or the other.

   p.e. all this are shared:
styles.VRRadioButton,
borderColor,
 bgStyle,
    styleSize,

AlbertCarreras:
i assume we don't need to listen to useReducedMotion here... can you confirm that?

AlbertCarreras (approved):
Few nits, but overall LGTM

packages/gestalt/src/ButtonToggle.tsx
diyorbek: yes, i renamed it cuz now they is shared

packages/gestalt/src/RadioGroupButton.tsx
diyorbek: I didn't destructure the object to keeps its context "visible" while reading the code. We may have other even handler functions in the future. But this one is just for the animation. That's why i think its better to keep it this way.

packages/gestalt/src/RadioButton.css
```diff
@@ -57,3 +57,35 @@
 .InputEnabled {
   composes: pointer from "./Cursor.css";
 }
+
```
rlingineni, line 60: Maybe add a comment here that these are VR related items

rlingineni (approved):

PR #3799, opened 2024-10-09
SearchGuide: Fixing issue with selected/focus and shrinking image

eniomoura:
#### What changed?

Fixed an issue where thumbnail images would shrink when selecting or focusing the SearchGuide in VR.

Before:
image

After:
Screenshot 2024-10-08 at 21 17 35
Screenshot 2024-10-08 at 21 17 59

#### Why?

To improve component quality.


AlbertCarreras:
Screenshot by Dropbox Capture


This Avatar seems out of padding now

packages/gestalt/src/SearchGuide.css
```diff
@@ -33,6 +34,7 @@ a.searchguideVr {
   composes: flex from "./Layout.css";
   composes: xsItemsCenter from "./Layout.css";
   composes: justifyCenter from "./Layout.css";
+  border-radius: var(--rounding-400);
```
AlbertCarreras, line 37: can we do sema token here?

packages/gestalt/src/SearchGuide.tsx
```diff
@@ -115,7 +114,13 @@ const SearchGuideWithForwardRef = forwardRef<HTMLButtonElement, Props>(function
         [focusStyles.hideOutline]: !isFocusVisible && !selected,
         [focusStyles.accessibilityOutline]: isFocusVisible,
       });
-  const childrenDivClasses = classnames(styles.childrenDiv);
+  const childrenDivClasses = classnames(
+    styles.childrenDiv,
+    isInVRExperiment && {
+      [styles[`color${color}`]]: !selected,
```
AlbertCarreras, line 120: can we do a map rather than `color${color}`. This usually breaks TS and doesnt detect breaking changes

packages/gestalt/src/SearchGuide.tsx
```diff
@@ -141,7 +146,16 @@ const SearchGuideWithForwardRef = forwardRef<HTMLButtonElement, Props>(function
           </Box>
         )}
         {'image' in thumbnail && (
-          <div className={isInVRExperiment ? styles.imageDivVr : styles.imageDiv}>
+          <div
+            className={
+              isInVRExperiment
+                ? classnames({
+                    [styles.imageDivVr]: true,
```
AlbertCarreras, line 153: should we do here:
[styles.imageDivVr]: !selected,
[styles.selectedImageDivVr]: selected,

packages/gestalt/src/SearchGuide.css
eniomoura: This is also used in classic

packages/gestalt/src/SearchGuide.tsx
eniomoura: done

packages/gestalt/src/SearchGuide.tsx
eniomoura: fixed, selectedImageDiv shouldn't be there anymore

AlbertCarreras (approved):

PR #3804, opened 2024-10-16
BannerCallout:  VR Implementation Changes

AlbertCarreras:
BannerCallout:  VR Implementation Changes

### DIFF
--- a/packages/gestalt/src/BannerCallout/VRBannerCallout.tsx
+++ b/packages/gestalt/src/BannerCallout/VRBannerCallout.tsx
@@ -231,8 +231,8 @@ export default function BannerCallout({
   };
 
   const content = (
-    <Fragment>
-      <Flex direction="column" gap={4}>
+    <Flex alignItems="center" width="100%" wrap>
+      <Flex.Item minWidth={0}>
         <Flex gap={3}>
           <Icon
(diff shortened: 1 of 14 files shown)


eniomoura (approved):
LGTM

packages/gestalt/src/BannerCallout/VRBannerCallout.tsx
```diff
@@ -0,0 +1,208 @@
...
+      </Box>
+      {/*
+            LG BREAKPOINT
+      */}
+      <Box
+        borderStyle={type === 'default' ? 'sm' : undefined}
+        color={backgroundColor}
+        dangerouslySetInlineStyle={{ __style: largePadding }}
+        display="none"
+        lgDisplay="block"
```
rlingineni, line 172: Is this the only thing that's changing? This seems like a lot of dupe

Can we move into a function or something called `getResponsiveSize(size:'sm|'lg|'md')`

rlingineni: I'm not sure, but also this approach may have React do 3X the work because the components would still be loaded into the virtual dom I think - since it's hidden by CSS. I think ideally this could be 3 separate smaller components, or abstracted at a different level

rlingineni (commented):
comments

PR #3812, opened 2024-10-18
DateRange: make onCancel and onSubmit events optional (web)

CarlosAvina:
### Summary

Makes `onCancel` and `onSubmit` events optional props. This will also hide cancel and submit buttons.

#### What changed?

Makes `onCancel` and `onSubmit` events optional props. If those props are not passed then the component is going to hide the cancel and submit buttons.

#### Why?

The current DatePicker used in analytics doesn't have an apply button, the user is accustomed to an action to occur when the endDate is selected, the data fetching occurs currently on the endDateChange event. There is no onCancel or onSubmit business logic on our side. For accomplish this we don't need to change the logic of the component, we just need to conditionally render those buttons if the props are present.

### Screenshots
Screenshot 2024-10-18 at 2 12 38 p m
Screenshot 2024-10-18 at 2 12 26 p m

### Links

- [TDD](link to Paper doc)
- [Figma](link to Figma file)

### Checklist

- [ ] Added unit tests

### DIFF
--- a/packages/gestalt-datepicker/src/DateRange.tsx
+++ b/packages/gestalt-datepicker/src/DateRange.tsx
@@ -35,7 +35,7 @@ type Props = {
   /**
    * Callback triggered when the user clicks the Cancel button to not persist the selected dates. It should be used to close DateRange. See the [controlled component variant](https://gestalt.pinterest.systems/web/daterange#Controlled-component) to learn more.
    */
-  onCancel: () => void;
+  onCancel?: () => void;
   /**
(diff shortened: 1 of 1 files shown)


AlbertCarreras:
https://gestalt.pinterest.systems/web/daterange#Controlled-component 

on this example, can you fix the on dismiss? and add setShowComponent(false)?
thasnk! it was innacessible that way


Screenshot by Dropbox Capture

Otherwise

AlbertCarreras (approved):

PR #3817, opened 2024-10-22
Masonry: Log number of iterations it takes to find optimal whitespace

bryamrrr:
### Summary

We want to add Masonry metrics to Statsboard. We want to have insights on how many iterations we make each time we position a multicolumn module.

#### What changed?

- logWhitespace now includes number of iterations to be used when logging
- Changed getGraphPositions to return winningNode and this new metric

https://github.com/user-attachments/assets/30b69a3c-3620-4337-9268-3053133abf27

Added the columnspan so we can use it as a tag when logging and potentially segment the data.

Enabling whitespace threshold to 16:

https://github.com/user-attachments/assets/62c2efd5-6add-47d4-95d9-6bf6a1c612dc

### DIFF
--- a/packages/gestalt/src/Masonry.tsx
+++ b/packages/gestalt/src/Masonry.tsx
@@ -114,11 +114,11 @@ type Props<T> = {
    */
   virtualize?: boolean;
   /**
-   * Experimental prop to log the additional whitespace shown above two-column items.
+   * Experimental prop to log the additional whitespace shown above multicolumn items and the number of iterations it took to position it.
    *
    * This is an experimental prop and may be removed in the future.
    */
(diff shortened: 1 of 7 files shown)


liuyenwei (approved):

packages/gestalt/src/Masonry.tsx
manvoid, line 1: I think you will need to also change this on MasonryV2

bryamrrr: you're right! ty

bryamrrr:
> @bryamrrr Some types seem to be failing Screenshot by Dropbox Capture

Fixed! ty

PR #3822, opened 2024-10-23
Box, Flex: extended margin/padding range with 0.25, 0.5 and 1.5

AlbertCarreras:
Box, Flex: extended margin/padding range with 0.25, 0.5 and 1.5

Screenshot by Dropbox Capture
Screenshot by Dropbox Capture
Screenshot by Dropbox Capture
Screenshot by Dropbox Capture
Screenshot by Dropbox Capture
Screenshot by Dropbox Capture


packages/gestalt/src/Flex.tsx
```diff
@@ -5,7 +5,27 @@ import FlexItem from './FlexItem';
...
+  | 7
+  | 8
+  | 9
+  | 10
+  | 11
+  | 12
+  | 13
+  | 14
+  | 15
+  | 16;
```
rlingineni, line 28: Just a thought - If the change is here - and we're not moving to the 100 scale in the actual TS props.. do we really need to migrate all of the internal classes to have a 100 in the naming? 

Isn't it all minified in the build - this PR seems to touch a lot of files, but I'm not sure if there is a clear benefit since the classes are auto-generated

AlbertCarreras: It's one step closer to tokens and o.25 becomes 250 rather than 25 or 025, and it follows the tokens underneath.

AlbertCarreras: i cannot have clannsmaes that are `margin0.25`

AlbertCarreras: Screenshot by Dropbox Capture


Next will be moving from 1 to 100 in the API. And all we be aligned.

PR #3825, opened 2024-10-24
Masonry: Enable whitespaceThreshold prop

bryamrrr:
#### What changed?

We are enabling whitespace threshold so it can be used to position multicolumn modules without excessive work to find an optimal whitespace.

#### Why?

This is a follow-up to https://github.com/pinterest/gestalt/pull/3817.


liuyenwei (approved):
note for anyone else reviewing: `whitespaceThreshold` already exists in the actual layout function(s). This just makes it externally configurable

manvoid (approved):

PR #3830, opened 2024-10-25
SearchGuideLink: Style discrepancy fixes

eniomoura:
#### What changed?

Fixed style differences between SearchGuide and SearchGuideLink

#### Why?

These components are supposed to be identical in design. This PR also fixes a focus shifting issue in SearchGuideLink that would cause issues on pinboard migration, and a spacing issue in the avatar variants that was not conforming to design.

### Links

- [Jira](https://jira.pinadmin.com/browse/CPW-7460)


AlbertCarreras:
Can we document the before and afters?

AlbertCarreras:
Can we match 
Screenshot by Dropbox Capture
?

AlbertCarreras:
We need to apply the border radius props to the image to get a good curve.
Screenshot by Dropbox Capture

We need to remove this trick, looks bad: 
Screenshot by Dropbox Capture

AlbertCarreras (approved):

PR #3868, opened 2024-11-13
Tokens: Remove percents from Android rounding

rlingineni:
This changes the value of the Android circle rounding output from `50%` to `0.5`. No actual token values have been updated.

### DIFF
--- a/packages/gestalt-design-tokens/src/transformers/transforms.js
+++ b/packages/gestalt-design-tokens/src/transformers/transforms.js
@@ -171,6 +171,23 @@ function registerTokenTransforms(sd) {
     },
   });
 
+  sd.registerTransform({
+    name: 'value/rounding/android',
+    type: 'value',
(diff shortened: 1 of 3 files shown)


packages/gestalt-design-tokens/src/transformers/transforms.js
```diff
@@ -171,6 +171,23 @@ function registerTokenTransforms(sd) {
...
   });
 
+  sd.registerTransform({
+    name: 'value/rounding/android',
+    type: 'value',
+    matcher(prop) {
+      return prop.attributes.category === 'rounding' && prop.value.endsWith('%');
+    },
+    transformer(prop) {
+      console.log('prop', prop);
```
AlbertCarreras, line 181: do we wanna consolelog this?

AlbertCarreras (approved):

PR #3872, opened 2024-11-14
Badge: VR fixes

AlbertCarreras:
Badge: VR fixes


packages/gestalt/src/Badge.css
```diff
@@ -3,10 +3,17 @@
   composes: rounding1 from "./Borders.css";
   align-content: center;
   display: flex;
-  padding: var(--space-0) 6px;
```
rlingineni, line 6: I think 6px should be a component token as `--badge-padding-left`

packages/gestalt/src/Badge.css
```diff
@@ -3,10 +3,17 @@
...
   white-space: nowrap;
 }
 
+.padding {
+  padding: var(--space-0) 6px;
```
rlingineni, line 10: I think 6px should be a component token as `--badge-padding-y` or something, so we can change it easier int he future

packages/gestalt/src/Badge.css
```diff
@@ -3,10 +3,17 @@
...
+}
+
+.paddingVR {
+  padding: var(--sema-space-0) var(--sema-space-100);
```
rlingineni, line 14: Ideally we use component tokens for these

packages/gestalt/src/Badge.css
rlingineni: Oh it seems like this is classic

packages/gestalt/src/icons-vr-theme/compact/index.ts
```diff
@@ -10,6 +10,7 @@ import compactChevronUp from './compact-chevron-up.svg';
 import compactDash from './compact-dash.svg';
 import compactInfoCircleFill from './compact-info-circle-fill.svg';
 import compactLock from './compact-lock.svg';
+import compactSparkle from './compact-sparkle.svg';
```
rlingineni, line 13: If it's not too much - could you also paste this into the Android package as well. Thank you

rlingineni (approved):
comments

packages/gestalt/src/Badge.css
AlbertCarreras: we don't have cmp token for spacing yet.. i think

packages/gestalt/src/Badge.css
AlbertCarreras: we don't have cmp token for spacing yet.. i think

PR #3875, opened 2024-11-15
Masonry: Enable dynamic batches

bryamrrr:
#### What changed?

This PR introduces `_getModulePositioningConfig`, an experimental prop to enable dynamic batch sizes (and Early bailout). This is a way to dynamically set the number of items used to position a multicolumn module.

`_getModulePositioningConfig` uses two params (gridSize and moduleSize) and uses them to return a `ModulePositioningConfig`, which could have these values:
- itemsBatchSize: Maximum number of items used to position a module
- whitespaceThreshold: "Good enough" whitespace number when positioning a module
- iterationsLimit: Optional hard limit for DAG iterations

#### Why?

This will allow us to reduce the whitespace from larger multicolumn modules.

Other changes:
- Refactor to enable a gutter at the Masonry and MasonryV2 level. This way, we can have the gutter logic in a single place and use it everywhere is needed.
- Refactor of `GetColumnCount` to move the logic to know how many columns a grid have in a single place

Images confirming we have the correct module and grid sizes.

Screenshot 2024-11-14 at 3 41 41 p m
Screenshot 2024-11-14 at 3 41 13 p m


packages/gestalt/src/Masonry.tsx
```diff
@@ -701,15 +709,29 @@ export default class Masonry<T> extends ReactComponent<Props<T>, State<T>> {
...
-      // If there are 2-col items, we need to measure more items to ensure we have enough possible layouts to find a suitable one
-      // we need the batch size (number of one column items for the graph) + 1 (two column item)
-      const itemsToMeasureCount = hasMultiColumnItems
-        ? MULTI_COL_ITEMS_MEASURE_BATCH_SIZE + 1
-        : minCols;
+      let batchSize;
+      if (nextMultiColumnItem) {
+        const gridSize = getColumnCount({ gutter, columnWidth, width, minCols, layout });
+
+        const moduleSize = calculateActualColumnSpan({
```
bryamrrr, line 720: This was the tricky part of this implementation. These two values are easy to find in `multiColumnLayout.ts`, but are not ready at this scope.

- For grid size, I created a whole new util that abstract the way we calculate the column count from both `DefaultLayout` and `FullWidthLayout` (updating those utils to avoid code duplicity too).
- For moduleSize, I exposed the `calculateActualColumnSpan` from  `multiColumnLayout.ts` to be able to use it here.

All of this happens also in MasonryV2.

PR #3877, opened 2024-11-15
Masonry: update layout logic to support rendering skeleton pins

rebeccayeyi:
#### What changed?

- Added a utility function to retrieve an item's height from the skeleton pin, using it instead of the measurement cache when applicable.

#### Why?

- This change is necessary to render skeleton pins as part of Masonry's `items` rather than `_loadingStateItems`.
  - This should help with overall performance since we are now having the real pins just replace the skeleton pins where currently, we render either the skeleton pins or the real pins
  - This also [unlocks infinite scrolling](https://github.com/pinternal/pinboard/pull/105265#discussion_r1834686408)

#### Test Plan
- Tested https://github.com/pinternal/pinboard/pull/105265 with local Gestalt 

#### Note
- If this change is okay, will also apply the same changes to Masonry v2!

### Checklist

- [x] Added unit tests

### DIFF
--- a/packages/gestalt/src/Masonry/defaultLayout.ts
+++ b/packages/gestalt/src/Masonry/defaultLayout.ts
@@ -1,6 +1,11 @@
 import { Cache } from './Cache';
 import { getHeightAndGutter, offscreen } from './layoutHelpers';
-import { isLoadingStateItem, isLoadingStateItems, isSkeletonPin } from './loadingStateUtils';
+import {
+  areSkeletonPins,
+  isLoadingStateItem,
+  isLoadingStateItems,
+  isSkeletonPin,
+} from './loadingStateUtils';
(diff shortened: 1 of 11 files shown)


packages/gestalt/src/Masonry/defaultLayout.ts
```diff
@@ -99,9 +99,10 @@ const defaultLayout =
           ...otherProps,
         })
       : items.map((item) => {
-          const height = isLoadingStateItem(item, renderLoadingState)
-            ? item.height
-            : measurementCache.get(item);
+          const height =
```
rebeccayeyi, line 102: thoughts on making sure that the pin is a skeleton pin? we could just check to see if `height` is available, but i wasnt sure if we wanted to expose that API

PR #3881, opened 2024-11-18
Spinner: VR version intro & outro animation

diyorbek:
### Summary

ezgif-3-caabed37f9

#### What changed?

- Replaced Spinner element with SVG to simplify the code.

### Links

- [Jira](https://jira.pinadmin.com/browse/GESTALT-8534)
- [Figma](https://www.figma.com/design/pMibC8csfCbNMU4jgLVUJh/Motion-Pilot-specs?node-id=2300-3727&m=dev)

### Checklist

- [ ] Added unit tests


packages/gestalt/src/Spinner/VRSpinner.tsx
```diff
@@ -6,40 +7,88 @@ import { useDefaultLabelContext } from '../contexts/DefaultLabelProvider';
...
-type Props = {
-  accessibilityLabel?: string;
-  delay?: boolean;
+type SpinnerBodyProps = {
+  accessibilityLabel: string;
+  delay: boolean;
   show: boolean;
-  size?: 'sm' | 'md' | 'lg';
+  size: 'sm' | 'md' | 'lg';
+  onExitAnimationEnd: () => void;
```
AlbertCarreras, line 18: https://gestalt.pinterest.systems/web/overlaypanel#Props-onAnimationEnd

Can we keep similar API as in sidenavigation?

onAnimationEnd | (arg1: { animationState: "out" }) => void
-- | --




onAnimationEnd | (arg1: { animationState: "in" \| "out" }) => void
-- | --

packages/gestalt-datepicker/src/__snapshots__/DatePicker.jsdom.test.tsx.snap
```diff
@@ -9,14 +9,14 @@ exports[`DatePicker Mobile Datepicker renders 1`] = `
       class="_gestalt"
     >
       <label
-        class="LTD kVc BG7"
+        class="label__LTD block__kVc pointer__BG7"
         for="fake_id"
       >
         <div
-          class="u8x zI7 iyn Hsu"
+          class="box__zI7 accessibilityOutline__iyn borderBox__Hsu marginBottom200__u8x"
```
AlbertCarreras, line 16: run yarn build and run the snapshots again

packages/gestalt-design-tokens/tokens/vr-theme/comp/spinner/default.json
```diff
@@ -3,17 +3,39 @@
     "spinner": {
       "color": {
         "background": {
-          "pink": {
```
AlbertCarreras, line 6: could we just do 1, 2, 3 and avoid token name changes? 

spinner-color-background-01
spinner-color-background-02
spinner-color-background-03

packages/gestalt/src/Spinner/VRSpinner.tsx
```diff
@@ -6,40 +7,88 @@ import { useDefaultLabelContext } from '../contexts/DefaultLabelProvider';
...
+  size?: 'sm' | 'md' | 'lg';
+};
+
+export default function Spinner({
+  accessibilityLabel,
+  delay = true,
+  show: showProp,
+  size = 'md',
+}: Props) {
+  const [show, setShow] = useState(showProp);
```
AlbertCarreras, line 70: i dont think this is the best API

1. the behavior is opaque, it's tied to the show prop
2. however, engineers dont use the show prop properly... Screenshot by Dropbox Capture 
3. the animations is not controlled and it can't be controlled externally or affect external state

diyorbek: In terms of the API, I don't think anything has changed. This change doesn't effect the existing instances regardless of the choice of use:

1. If developers are conditionally rendering it, they can keep doing so. Spinner will just not show the outro animation.
2. If developers do want to have outro animation, they will need to control it by `show` prop.
3. If developers want to control it by `show` prop but don't want the outro animation:
  a. The can change the usage to conditional rendering.
  b. We can add another prop to disable outro animation.

packages/gestalt/src/Spinner/VRSpinner.tsx
```diff
@@ -6,40 +7,88 @@ import { useDefaultLabelContext } from '../contexts/DefaultLabelProvider';
...
           } as React.CSSProperties
         }
       >
-        <div className={styles.spinnerFrame}>
-          <div className={styles.dot1} />
-          <div className={styles.dot2} />
-          <div className={styles.dot3} />
-        </div>
+        <svg fill="none" viewBox="0 0 56 56" xmlns="http://www.w3.org/2000/svg">
+          <g style={{ '--g-start-x': '14px', '--g-start-y': '-8px' } as React.CSSProperties}>
```
diyorbek, line 42: Apologies for the future maintiners for this mess. Our current CSS validator script did not allow me to re-declare variables inside different selectors (which could have made the code much much simpler and readable).

AlbertCarreras (approved):

packages/gestalt/src/Spinner/VRSpinner.tsx
```diff
@@ -6,40 +7,88 @@ import { useDefaultLabelContext } from '../contexts/DefaultLabelProvider';
...
+export default function Spinner({
+  accessibilityLabel,
+  delay = true,
+  show: showProp,
+  size = 'md',
+}: Props) {
+  const [show, setShow] = useState(showProp);
+  const { accessibilityLabel: accessibilityLabelDefault } = useDefaultLabelContext('Spinner');
+
+  const unmountSpinner = () => {
```
AlbertCarreras, line 73: you might want to use useCallback here?

packages/gestalt/src/Spinner/VRSpinner.tsx
```diff
@@ -6,40 +7,88 @@ import { useDefaultLabelContext } from '../contexts/DefaultLabelProvider';
...
+    if (!showProp) setShow(false);
+  };
+
+  useEffect(() => {
+    if (showProp) setShow(showProp);
```
AlbertCarreras, line 78: should we do here something like...

 if (showProp !== show) setShow(showProp);

?

AlbertCarreras (approved):

packages/gestalt/src/Spinner/VRSpinner.tsx
diyorbek: **decided to address it later

packages/gestalt/src/Spinner/VRSpinner.tsx
diyorbek: calback is not used inside hook and depends only on `showProp`. I don't think useCallback will benefit us here.

PR #3044, opened 2023-06-28
Callout: Support Text Type in Callout

rlingineni:
Based on a slack convo with a customer.

#### What changed?
Adding support for Rich Text Type in a Callout. message E.g. Bold, italics.

Before
image

After
image

#### Why?

Slimbanner supports a text prop for the message, but callout does not. Seems like it would be a consistency thing among the two message props. Pending design review. 

### Links

- [Jira](https://jira.pinadmin.com/browse/GESTALT-XXXX)
- [TDD](link to Paper doc)
- [Figma](link to Figma file)

### Checklist

- [X] Added unit and Flow Tests
- [X] Added documentation + accessibility tests
- [X] Verified accessibility: keyboard & screen reader interaction
- [X] Checked dark mode, responsiveness, and right-to-left support
- [X] Checked stakeholder feedback (e.g. Gestalt designers)

PR #3906, opened 2024-12-04
BannerCallout: Remove Children.only logic prone to break if message is undefined in runtime

AlbertCarreras:
BannerCallout: Remove Children.only logic prone to break if message is undefined in runtime

BEFORE
Screenshot by Dropbox Capture

AFTER
Screenshot by Dropbox Capture


rlingineni (approved):

rlingineni:
@AlbertCarreras  - we did this to support Rich Text in Callout:

https://github.com/pinterest/gestalt/pull/3044

So, this may break callouts in case this is being used

rlingineni:
If this renders everything in Message, that means anything can be passed as children. Is that alllowed? I recall someone did that with a flow suppression in the past and it was hard to get them to change

PR #3910, opened 2024-12-11
DateRange: Highlight secondary range in DateRange component (web)

CarlosAvina:
#### What changed?

When the secondary date range is enabled in the DateRange component, the inactive range will be highlighted so it makes it easier for the user to see the current selected dates.

#### Why?

Without the highlights, it is considered a big gap of experience since it's hard to visualize the current selected ranges, the user can only see one of them (the one is currently being selected). If we show both ranges at the same time it will clearer to the user which dates are being compared/used.

For this we use the prop `highlightDates` which allows you to pass an array of dates and a custom class name to style the dates inside the array. Link to the docs [here](https://reactdatepicker.com/#example-highlight-dates-with-custom-class-names-and-ranges).

#### Screenshots

| Before | After |
|--------|------|
| <video src=https://github.com/user-attachments/assets/6a87ce1b-3f3e-46fa-b021-c800cfb0eee4> | <video src=https://github.com/user-attachments/assets/c96ba642-b3ea-4d61-9e90-d16096e35de6> |

### Links

- [Feedback from bugbash](https://docs.google.com/document/d/1BapsxpjUYMncpZ9XVsJlNcf-VImSEbKccKpxS0hg38g/edit?disco=AAABYI9n8ww)

### Checklist

- [ ] Added unit tests


packages/gestalt-datepicker/src/DateRange.css
```diff
@@ -28,3 +28,24 @@ html[dir="rtl"] .borderRight {
   transform: translateY(-50%);
   width: 3px;
 }
+
+/* secondary range styles  */
+
+:global ._gestalt .react-datepicker__day--in-range-secondary {
```
rlingineni, line 34: any reason these are global? If we support custom classnames, can we just pass the classname

packages/gestalt-datepicker/src/DateRange/InternalDatePicker.tsx
```diff
@@ -55,6 +57,42 @@ const InternalDatePickerWithForwardRef = forwardRef<HTMLInputElement, ModifiedPr
...
+    }
+
+    function generateHighligths(
+      startDate: Date | null | undefined,
+      endDate: Date | null | undefined,
+    ) {
+      const datesArray = startDate && endDate ? getDatesArray(startDate, endDate) : [];
+      return [
+        {
+          'react-datepicker__day--in-range-secondary': datesArray,
```
rlingineni, line 91: You might be able to pull in styles like:
```
import styles from 'daterange.css`

 {`[styles.days-in-secondary]`:dateArray}
```

rlingineni: Can you also add a snapshot test that says `secondaryHighlight range appears when selected` so we can have good coverage

packages/gestalt-datepicker/src/DateRange/InternalDatePicker.tsx
```diff
@@ -55,6 +57,42 @@ const InternalDatePickerWithForwardRef = forwardRef<HTMLInputElement, ModifiedPr
...
+        // Push the current date to the array
+        dates.push(new Date(currentDate));
+        // Increment the current date by one day
+        currentDate.setDate(currentDate.getDate() + 1);
+      }
+
+      return dates;
+    }
+
+    function generateHighligths(
```
rlingineni, line 84: maybe typo - should be `generateHighlights`

packages/gestalt-datepicker/src/DateRange/InternalDatePicker.tsx
```diff
@@ -95,9 +141,17 @@ const InternalDatePickerWithForwardRef = forwardRef<HTMLInputElement, ModifiedPr
             previousMonthButtonLabel={
               <Icon accessibilityLabel="" color="default" icon="arrow-back" size={16} />
             }
-            selected={rangeStartDate ?? undefined}
+            selected={
+              selectedRange === DateRangeType.Primary
```
rlingineni, line 145: Why do we need this now, and we didn't have it earlier?

packages/gestalt-datepicker/src/DateRange/InternalDatePicker.tsx
```diff
@@ -73,9 +111,17 @@ const InternalDatePickerWithForwardRef = forwardRef<HTMLInputElement, ModifiedPr
...
+            endDate={
+              selectedRange === DateRangeType.Primary
+                ? rangeEndDate ?? undefined
+                : secondaryRangeEndDate ?? undefined
+            }
             excludeDates={excludeDates && [...excludeDates]}
-            highlightDates={initRangeHighlight ? [initRangeHighlight] : []}
+            highlightDates={
+              selectedRange === DateRangeType.Primary
+                ? generateHighligths(secondaryRangeStartDate, secondaryRangeEndDate)
```
rlingineni, line 122: Ah so you need the inverse of the selection for the highlights?

packages/gestalt-datepicker/src/DateRange.tsx
```diff
@@ -358,16 +358,11 @@ function DateRange({
                     onSecondaryDateChange({ value: startDate }, { value: endDate });
                   }
                 }}
-                rangeEndDate={
-                  selectedRange === DateRangeType.Primary
-                    ? dateValue.endDate
-                    : secondaryDateValue?.endDate
-                }
-                rangeStartDate={
-                  selectedRange === DateRangeType.Primary
```
rlingineni, line 367: Why does this move to the Internal component now?

rlingineni (commented):
Looks good - mostly need a better solution for the classnames and testing

packages/gestalt-datepicker/src/DateRange.css
CarlosAvina: you are right, we don't need the global here 👍

packages/gestalt-datepicker/src/DateRange/InternalDatePicker.tsx
CarlosAvina: sure 👍🏾

packages/gestalt-datepicker/src/DateRange/InternalDatePicker.tsx
CarlosAvina: This is because the logic for supporting both ranges (primary and secondary) was handled in the `DateRange.tsx` component itself, but now we also need to know which one is selected inside the InternalDatePicker in order to apply the correct styles to the "inactive range". Previously this component only had access to the "active" range.

packages/gestalt-datepicker/src/DateRange/InternalDatePicker.tsx
CarlosAvina: Yes, here we want to highlight the "inactive" range, so, if the "active" range is the `primary`, then we want to highlight the secondary, and vice versa.

packages/gestalt-datepicker/src/DateRange.tsx
CarlosAvina: For this same reason: https://github.com/pinterest/gestalt/pull/3910#discussion_r1882938621

rlingineni (approved):

PR #3964, opened 2025-01-30
Masonry: Experimental breakpoints for gridSize

IsraLopez26:
## Summary

This PR adds on Masonry's multi-column module new experimental breakpoints for grid size. 

For original Masonry's responsive modules please refer to: https://github.com/pinterest/gestalt/pull/3636

### Current State 

Current grid size breakpoints: 
- sm: 2 columns
- md: 3-4 columns
- lg: 5-8 columns
- xl: 9+ columns

### Changes introduced

New experimental breakpoints: 
- sm: 2 columns
- md: 3-4 columns
- _lg1: 5-6 columns (New)
- lg: 5-8 columns (Existing)
- xl: 9+ columns

This change also adds logic for retro compatibility, to avoid breaking current configurations. 
In cases where `_lg1` is not configured, it would fallback to use existing `lg`

## Test plan

Unit test has been updated. 
Also tested locally: 

Using following configuration, verified multicolumn module size for different grid sizes: 

```
_getColumnSpanConfig = () => {
    return {
        sm: 1,
        md: 2, 
        _lg1: 3
        lg: 4,
        xl: 6,
    }
}
```

md size grid: 
image

_lg1 size grid:
image

lg size grid:
image

xl size grid:
image

When `_lg1` is not defined, it falls back to `lg`: 
image

### DIFF
--- a/packages/gestalt/src/Masonry/multiColumnLayout.ts
+++ b/packages/gestalt/src/Masonry/multiColumnLayout.ts
@@ -8,7 +8,7 @@ import { GetGraphPositionsReturn, NodeData, Position } from './types';
 // This may need to be tweaked to balance the tradeoff of delayed rendering vs having enough possible layouts
 export const MULTI_COL_ITEMS_MEASURE_BATCH_SIZE = 5;
 
-type GridSizeConfig = { 'sm': number, 'md': number, '_lg1'?: number, 'lg': number, 'xl': number }
+type GridSizeConfig = { 'sm': number; 'md': number; '_lg1'?: number; 'lg': number; 'xl': number };
 type GridSize = keyof GridSizeConfig;
 
 export type ColumnSpanConfig = number | GridSizeConfig;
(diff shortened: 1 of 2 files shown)


packages/gestalt/src/Masonry/multiColumnLayout.ts
```diff
@@ -24,8 +26,11 @@ export function columnCountToGridSize(columnCount: number): GridSize {
   if (columnCount <= 4) {
     return 'md';
   }
+  if (columnCount <= 6) {
+    return '_lg1';
```
liuyenwei, line 30: actually looking at this more closely, let's just add `_lg1`.  my original thought was the column span config would be `{ _lg1, _lg2 } | { lg }` but i think we can just support `{ _lg1, lg } `

so if _lg1 is defined, then:
- lg1: 5-6
- lg: 7-8

otherwise lg would just be 5-8

IsraLopez26: Thanks for the comments, this is now updated in the last commit

liuyenwei (approved):

PR #3990, opened 2025-02-12
Button: Fixing height 

gabrielpereirapinheiro:
**Button**
Before:
[Button before](https://github.com/user-attachments/assets/64648055-71ac-416f-8a73-38aea7acd332)
After:
[Button after](https://github.com/user-attachments/assets/fc9c6d67-5809-4a48-a462-fd0d20eff15b)

Button link

Before:
[Button Link before changes](https://github.com/user-attachments/assets/4bea4b05-0b29-474c-ab82-84b699a7f4cf)
After:
[Button Link after changes](https://github.com/user-attachments/assets/b0a2c3ed-5851-4063-a0f6-ab91f5b25e07)


AlbertCarreras:
@gabrielpereirapinheiro 

I don't the Button & ButtonLink compressions look the same. Button seems internal, while ButtonLink all compresses.

Can you double check?

Also, the title is misleading cos is not fixing height but the compression behavior.

PR #3993, opened 2025-02-12
Masonry: Flexible width items

IsraLopez26:
# Summary

This PR builds on Masonry's multi-column to add support for _flexible width items_.

This change introduces a new Masonry's experimental prop to support such feature: `_getResponsiveModuleConfigForSecondItem`

# Current state

At a high level, the current multi-column items work using the `_getColumnSpanConfig` prop. This sets a static width column span, based on the grid size. For more details, please see https://github.com/pinterest/gestalt/pull/3636

# Flexible Width Items

By introducing flexible width items, the idea is that the **second item** in the Masonry grid can span to fill empty columns in the first row of the grid, within the limits configured by using `_getResponsiveModuleConfigForSecondItem`.

#### _getResponsiveModuleConfigForSecondItem

```javascript
getResponsiveModuleConfigForSecondItem<T>: (item: T) => number | { min: number, max: number } | undefined
```

`_getResponsiveModuleConfigForSecondItem` is a function which takes the item data (T) as an input and returns one of the following:

- `number`: Fixed column span (same behavior as today)
-  `{ min: number, max: number }`: An object which defines the minimum and maximum limits the flexible width item could span. 
-  `undefined`: This is an indicator to know that this prop hasn't been defined, and the functionality should fallback to current behavior. 

We've identified the following edge cases and defined fallback logic: 

- If the current item is not in the second position (items index 1), the `_getResponsiveModuleConfigForSecondItem` logic will be ignored, and fallback to the current logic. 
- If the flexible width item doesn't fit the first row (min limit is larger than the first row empty columns), the item will be placed on the second column using minimum width. 
- If `_getResponsiveModuleConfigForSecondItem` is undefined but `_getColumnSpanConfig` is defined for the second item, it should still behave as the existing multi-column items, whose width depends on the grid size. 
- If by any chance both `_getResponsiveModuleConfigForSecondItem` and `_getColumnSpanConfig` are set for the second item, the flexible width `_getResponsiveModuleConfigForSecondItem` behavior will have preference. 

# Test

Added Unit Test for new functionality

Tested locally: 

Using _getColumnSpanConfig for item1: 
```
_getColumnSpanConfig={(item) => {
  return {
    sm: 2,
    md: 2,
    lg: 3,
    xl: 4,
  }
}
```

and using _getResponsiveModuleConfigForSecondItem for item2  

```
_getResponsiveModuleConfigForSecondItem={(item) => {
  return {min: 2, max: 6};
}
```

5cols example: Multicol item (red) 3 cols. Flexible Width item (black) 2 cols.
image

6cols example: Multicol item (red) 3 cols. Flexible Width item (black) 3 cols.
image

Testing max limit: Multicol item (red) 4 cols. Flexible Width item (black) 6 cols.
image

Testing min limit: Flex width item (black) doesn't fit first row, so it gets pushed to second row using min limit width
image


manvoid (commented):
In the description it says "If the flexible width item doesn't fit the first row (min limit is larger than the first row empty columns), the item will be placed on the second column using minimum width." but that doesn't seem exact because we will teat the module as a "normal" module, hence it will have the 5 pin batch and could be placed on the second or third row (or more) depending on the pins arangement.

packages/gestalt/src/Masonry/multiColumnLayout.ts
```diff
@@ -488,16 +542,19 @@ function getGraphPositions<T>({
 function getPositionsWithMultiColumnItem<T>({
   multiColumnItem,
   itemsToPosition,
+  originalItems,
```
manvoid, line 545: I'm not sure that passing the whole array of items its the best to solve this, since we really only need the `columnSpan` to calculate the width of the responsive module. Another approach could be to have a simple custom hook that stores that data and we can call it inside of `calculateActualColumnSpan`, in fact with that we could also remove some added dependencies there and simplify a little bit the logic.

packages/gestalt/src/Masonry.tsx
```diff
@@ -131,6 +135,18 @@ type Props<T> = {
...
+   * This used to enable multi-column flexible width support ONLY ON SECOND ITEM OF THE ARRAY OF ITEMS.
+   * _getResponsiveModuleConfigForSecondItem is a function that takes an individual grid item as an input and returns a ResponsiveModuleConfig.
+   * ResponsiveModuleConfig can be one of two things:
+   * - A number, which indicates a static number of columns the item should span
+   * - An object, which sets the minimum and maximum limits a multi-column item could span filling the empty columns in the first row of the grid (flexible width).
+   * If this prop is set for the second item in the array of items, the flexible width will work. Otherwise, it will fallback to the _getColumnSpanConfig behavior.
+   *
+   * This is an experimental prop and may be removed or changed in the future.
+   */
+  _getResponsiveModuleConfigForSecondItem?: (item: T) => ResponsiveModuleConfig;
```
liuyenwei, line 149: suggestion: we can enforce that `_getResponsiveModuleConfigForSecondItem` can only be used if `_getColumnSpanConfig` is passed in. something like:

```
type MultiColumnProps = { _getColumnSpanConfig?: .... } | { _getColumnSpanConfig: ...., _getResponsiveModuleConfigForSecondItem: ...}
```

packages/gestalt/src/Masonry/defaultLayout.ts
```diff
@@ -84,15 +89,17 @@ const defaultLayout =
       width,
     });
 
-    return _getColumnSpanConfig
+    return _getColumnSpanConfig || _getResponsiveModuleConfigForSecondItem
```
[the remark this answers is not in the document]
liuyenwei: +1 - left a similar comment above. the other thing to consider here is the two API approach is (ideally) short term

packages/gestalt/src/Masonry/multiColumnLayout.ts
```diff
@@ -54,14 +55,59 @@ function getColumnSpanFromGridSize(columnSpanConfig: ColumnSpanConfig, gridSize:
...
+    );
+    return columnSpan;
+  }
+  return 1;
+}
+
 function calculateActualColumnSpan<T>(props: {
   columnCount: number;
   item: T;
+  items: ReadonlyArray<T>;
```
liuyenwei, line 79: this function is currently called while we're already iterating `items` so, instead of passing items in can we just pass the item index + the column span of the previous item since we would have already calculated it so no need to do it again?

packages/gestalt/src/Masonry/multiColumnLayout.ts
liuyenwei: +1 similar comment above

packages/gestalt/src/Masonry/defaultLayout.ts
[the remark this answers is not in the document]
IsraLopez26: Thanks, I'll fix it in the next commit

packages/gestalt/src/Masonry/multiColumnLayout.ts
IsraLopez26: I think we do need this to pass down this items array, since this function can be called while iterating `items`, `itemsWithoutPositions`, or `itemsToPosition` (which are variants of the original Items array). 

In such case, we want to assure that the FlexibleWidthItem is in second position (index===1) from the **original items** array. 

That's why I'm including the `items.indexOf(item) === 1` check. 
I could re-name the prop to `originalItems` to make it more clear

packages/gestalt/src/Masonry/multiColumnLayout.ts
IsraLopez26: As discussed in previous conversation, we will keep the originalItems array to keep reference of first and second item in for responsive modules

packages/gestalt/src/Masonry/multiColumnLayout.ts
```diff
@@ -54,15 +55,60 @@ function getColumnSpanFromGridSize(columnSpanConfig: ColumnSpanConfig, gridSize:
...
+    return columnSpan;
+  }
+  return 1;
+}
+
 function calculateActualColumnSpan<T>(props: {
   columnCount: number;
   item: T;
+  firstItem: T;
+  idxInOriginalItems: number;
```
manvoid, line 80: nit: rename to `originalIdx`

manvoid, line 80: nit: rename to `originalIdx`

packages/gestalt/src/Masonry/multiColumnLayout.ts
```diff
@@ -170,7 +220,14 @@ export function initializeHeightsArray<T>({
       // we do Math.round here because both position.left and columnWidthAndGutter can be floating point numbers
       // in the case of fullWidthLayout (i.e. fluid grid)
       const col = Math.round((position.left - centerOffset) / columnWidthAndGutter);
-      const columnSpan = calculateActualColumnSpan({ columnCount, item, _getColumnSpanConfig });
+      const columnSpan = calculateActualColumnSpan({
+        columnCount,
+        item,
+        firstItem: originalItems[0]!,
+        idxInOriginalItems: originalItems.indexOf(item),
```
manvoid, line 227: I'm not sure if doing this every time could have some impact on performance, please consult with Yen, in case it does we could verify the index only if the current item has a responsive module config and change the prop to something like: `isResponsiveModule:  _getResponsiveModuleConfigForSecondItem(item) && originalItems.indexOf(item) === 1`

IsraLopez26: Agreed, that could be done to avoid excessive index calculations. Fixed in the last commit

manvoid (approved):

packages/gestalt/src/Masonry/multiColumnLayout.ts
```diff
@@ -737,12 +835,14 @@ const multiColumnLayout = <T>({
         getPositionsWithMultiColumnItem({
           multiColumnItem: multiColumnItems[i]!,
           itemsToPosition,
+          originalItems: items,
```
liuyenwei, line 838: there's a lot of passing around `originalItems`, getting the `firstItem` and checking `originalItems.indexOf(item) === 1`

instead, can we just do that once here and pass `firstItem`, `secondItem` going forward?

similarly, we call `_getResponsiveModuleConfigForSecondItem(item)` quite a few times. can we just do this once?
i.e.
```
const secondItem = originalItems[1];
const responsiveModuleConfigForSecondItem = _getResponsiveModuleConfigForSecondItem?.(item);
```

packages/gestalt/src/Masonry/multiColumnLayout.ts
```diff
@@ -54,15 +55,59 @@ function getColumnSpanFromGridSize(columnSpanConfig: ColumnSpanConfig, gridSize:
...
+    isFlexibleWidthItem,
+    _getColumnSpanConfig,
+    _getResponsiveModuleConfigForSecondItem,
+  } = props;
   const columnSpanConfig = _getColumnSpanConfig(item);
   const gridSize = columnCountToGridSize(columnCount);
-  const columnSpan = getColumnSpanFromGridSize(columnSpanConfig, gridSize);
+
+  const responsiveModuleConfig = _getResponsiveModuleConfigForSecondItem(item);
+  // By design, a responsive module should always be placed at the second position in the grid
```
liuyenwei, line 96: did you mean flexible width module?

packages/gestalt/src/Masonry/multiColumnLayout.ts
```diff
@@ -54,15 +55,59 @@ function getColumnSpanFromGridSize(columnSpanConfig: ColumnSpanConfig, gridSize:
...
+    firstItem,
+    isFlexibleWidthItem,
+    _getColumnSpanConfig,
+    _getResponsiveModuleConfigForSecondItem,
+  } = props;
   const columnSpanConfig = _getColumnSpanConfig(item);
   const gridSize = columnCountToGridSize(columnCount);
-  const columnSpan = getColumnSpanFromGridSize(columnSpanConfig, gridSize);
+
+  const responsiveModuleConfig = _getResponsiveModuleConfigForSecondItem(item);
```
liuyenwei, line 95: should this be in the `if (isFlexibleWidthItem)` conditional?

packages/gestalt/src/Masonry/multiColumnLayout.ts
liuyenwei: maybe rephrasing my question - why couldn't we do this?
```
const multiColumnLayout = ({ items, originalItems }) => {
  const secondItem = originalItems[1];
  const responsiveModuleConfigForSecondItem = _getResponsiveModuleConfigForSecondItem?.(item);
  const checkIsFlexibleWidthItem = (item) => responsiveModuleConfigForSecondItem && item === secondItem;
  
  .....
  // somewhere later
  calculateActualColumnSpan({
    columnCount,
    item,
    isFlexibleWidthItem: checkIsFlexibleWidthItem(item)
```

i'm not saying remove the check for `_getResponsiveModuleConfigForSecondItem(item) && originalItems.indexOf(item) === 1` but there's no reason we need to duplicate the same logic so many times imo

IsraLopez26: yeap, that makes a lot of sense. Sorry I misunderstood, Thanks for clarifying! 
It's now fixed in the last commit

PR #4000, opened 2025-02-19
DatePicker: add private prop _overrideRangeDateFix to enable correct logic in date ranges

AlbertCarreras:
DatePicker: add private prop _overrideRangeDateFix to enable correct logic in date ranges     

This component contains logic that made the component uncontrolled and causes unexpected behavior when trying to define min an max dates. We need to deprecate when all instances of DatePicker set _overrideRangeDateFix to true


jaewhyun:
Thank you for the fix!

PR #4013, opened 2025-02-26
Image, Video: if image or track `src` is empty string "" then set as undefined

pabmacias:
# Pull Request Instructions

#### What changed?

If the src is empty in img or track inside the Image and Video component, then it sets it to undefined.

#### Why?

We are getting many errors like this one when upgrading to React 19: `An empty string ("") was passed to the src attribute`. I tested this on my devapp and this fixes this errors.

### Checklist

- [ ] Added unit tests

### DIFF
--- a/packages/gestalt/src/Image.tsx
+++ b/packages/gestalt/src/Image.tsx
@@ -226,7 +226,7 @@ export default class Image extends PureComponent<Props> {
           onLoad={this.handleLoad}
           role={role === 'presentation' ? 'presentation' : undefined}
           sizes={sizes}
-          src={src || undefined}
+          src={typeof src === 'string' && src !== '' ? src : undefined}
           srcSet={srcSet}
           {...conditionalProps}
         />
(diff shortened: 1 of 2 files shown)


packages/gestalt/src/Image.tsx
```diff
@@ -226,7 +226,7 @@ export default class Image extends PureComponent<Props> {
           onLoad={this.handleLoad}
           role={role === 'presentation' ? 'presentation' : undefined}
           sizes={sizes}
-          src={src}
+          src={typeof src === 'string' && src !== '' ? src : undefined}
```
AlbertCarreras, line 229: can we do 

`          src={typeof src === 'string' && src === '' ?  undefined : src}
`

AlbertCarreras: I just want to make sure src={undefined} also makes it

AlbertCarreras (approved):

PR #4029, opened 2025-03-06
Masonry: Fix Multi-Column Module Positioning Bug on Dynamic Heights V2 and Optimize Height Recalculation

JosAnGoCa:
# Summary

This pull request addresses a reported bug that occurs when two multi-column modules are stacked directly above each other and the bottom module is smaller in size than the top module. Additionally, it resolves an issue with height recalculation when item size changes are less than 1 pixel.

## What changed?

The bug stemmed from the method used by dynamic heights v2 to calculate affected areas and columns. The previous logic only considered situations where a multi-column item was greater than or expanded beyond the current area, leading to incorrect repositioning. Specifically, a multi-column item with a column span of 3 positioned directly above a multi-column item with a span of 2 would not trigger repositioning for the second item.

The updated algorithm now checks if any boundary of the second multi-column item falls within the area of the first multi-column item, ensuring proper repositioning.

# Checklist

- [x] Added unit tests
- [x] Fix bug
- [x] Remove unnecesary height reposition when the difference is less than 1 px

### DIFF
--- a/packages/gestalt/src/Masonry/dynamicHeightsUtils.test.ts
+++ b/packages/gestalt/src/Masonry/dynamicHeightsUtils.test.ts
@@ -349,4 +349,150 @@ describe('dynamic heights on masonry', () => {
       expect(newPos).toEqual(expectedPos[index]);
     });
   });
+
+  test('when the height delta is less than 1, it should not reposition the items', () => {
+    const measurementStore = new MeasurementStore<Record<any, any>, number>();
+    const positionCache = new MeasurementStore<Record<any, any>, Position>();
+    const items: readonly [Item, Item, ...Item[]] = [
+      { 'name': 'Pin 0', 'height': 230, 'color': '#E230BA' },
+      { 'name': 'Pin 1', 'height': 201, 'color': '#F67076' },
+      { 'name': 'Pin 2', 'height': 200, 'color': '#F67076' },
+      { 'name': 'Pin 3', 'height': 120, 'color': '#FAB032', columnSpan: 2 },
+      { 'name': 'Pin 4', 'height': 202, 'color': '#F67076' },
+      { 'name': 'Pin 5', 'height': 90, 'color': '#F67076' },
+      { 'name': 'Pin 6', 'height': 60, 'color': '#F67076' },
+      { 'name': 'Pin 7', 'height': 100, 'color': '#F67076', columnSpan: 2 },
+      { 'name': 'Pin 8', 'height': 400, 'color': '#F67076' },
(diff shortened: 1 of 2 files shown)


packages/gestalt/src/Masonry/dynamicHeightsUtils.test.ts
```diff
@@ -349,4 +349,150 @@ describe('dynamic heights on masonry', () => {
       expect(newPos).toEqual(expectedPos[index]);
     });
   });
+
+  test('when the height delta is less than 1, it should not reposition the items', () => {
```
manvoid, line 353: This only tests when the increase is positive but that worked before, we should add one that tests this for a negative delta.

packages/gestalt/src/Masonry/dynamicHeightsV2Utils.ts
```diff
@@ -110,7 +112,7 @@ function recalcHeights<T>({
   if (
     !changedItemPosition ||
     newHeight === 0 ||
-    Math.floor(changedItemPosition.height) === Math.floor(newHeight)
+    Math.floor(Math.abs(newHeight - changedItemPosition.height)) === 0
```
manvoid, line 115: This is a different change that should be added on a different PR since it could have unexpected side effects.

packages/gestalt/src/Masonry/dynamicHeightsUtils.test.ts
```diff
@@ -349,4 +349,150 @@ describe('dynamic heights on masonry', () => {
...
+  });
+
+  test('smaller module item that is above and in the area of a bigger module should be repositioned', () => {
+    const measurementStore = new MeasurementStore<Record<any, any>, number>();
+    const positionCache = new MeasurementStore<Record<any, any>, Position>();
+    const items: readonly [Item, Item, ...Item[]] = [
+      { 'name': 'Pin 0', 'height': 200, 'color': '#E230BA' },
+      { 'name': 'Pin 1', 'height': 201, 'color': '#F67076' },
+      { 'name': 'Pin 2', 'height': 202, 'color': '#FAB032' },
+      // { 'name': 'Pin 3', 'height': 203, 'color': '#EDF21D' },
```
manvoid, line 435: nit: remove commented lines

packages/gestalt/src/Masonry/dynamicHeightsUtils.test.ts
JosAnGoCa: When I run the test without the changes, it fails, does not matter if there was positive or negative, but i added other test to check that

manvoid (approved):

PR #4035, opened 2025-03-11
Masonry: Enable dynamic batches

JosAnGoCa:
## Summary

Updated the PR created by Bryam introducing dynamic batches [3875](https://github.com/pinterest/gestalt/pull/3875)

### What changed?

This PR introduces `_getModulePositioningConfig`, an experimental prop to enable dynamic batch sizes (and Early bailout). This is a way to dynamically set the number of items used to position a multicolumn module.

`_getModulePositioningConfig` uses two params (gridSize and moduleSize) and uses them to return a `ModulePositioningConfig`, which could have these values:

- itemsBatchSize: Maximum number of items used to position a module
- whitespaceThreshold: "Good enough" whitespace number when positioning a module
- iterationsLimit: Optional hard limit for DAG iterations

### Why?
This will allow us to reduce the whitespace from larger multicolumn modules.

### Other changes:

Refactor to enable a gutter at the Masonry and MasonryV2 level. This way, we can have the gutter logic in a single place and use it everywhere is needed.
Refactor of `GetColumnCount` to move the logic to know how many columns a grid have in a single place
Images confirming we have the correct module and grid sizes.

### Overview

When the batch size increments, it gives the graph more possibilities to place the items and this lead to a posible better positioning of the multi-column item resulting on better whitespace. The batch size could depend on the number of columns and the column span of the multi-column item, thats why the `_getModulePositioningConfig` has those two variables as parameters.

### Example
- Current state
image

- With a batch size of 7
image


packages/gestalt/src/Masonry.tsx
```diff
@@ -224,6 +231,13 @@ export default class Masonry<T> extends ReactComponent<Props<T>, State<T>> {
 
     this.positionStore = props.positionStore || Masonry.createMeasurementStore();
 
+    const { layout, gutterWidth } = props;
+    let defaultGutter = DEFAULT_LAYOUT_DEFAULT_GUTTER;
+    if ((layout && layout === 'flexible') || layout === 'serverRenderedFlexible') {
```
IsraLopez26, line 236: I think the `layout &&` check should be outside parenthesis

JosAnGoCa: You'r right, fixed

packages/gestalt/src/Masonry/getColumnCount.ts
```diff
@@ -0,0 +1,39 @@
...
+  columnWidth: number | undefined;
+  width: number;
+  minCols: number;
+  layout: Layout;
+}): number {
+  if (layout === 'flexible' || (layout === 'serverRenderedFlexible' && width !== null)) {
+    // "This is kind of crazy!" - you
+    // Yes, indeed. The "guessing" here is meant to replicate the pass that the
+    // original implementation takes with CSS.
+    const idealColumnWidth =
```
IsraLopez26, line 25: I think this could be simplified as: 

`const idealColumnWidth = columnWidth ?? FULL_WIDTH_LAYOUT_DEFAULT_IDEAL_COLUMN_WIDTH`

packages/gestalt/src/Masonry/getColumnCount.ts
```diff
@@ -0,0 +1,39 @@
...
+      typeof columnWidth === 'undefined'
+        ? FULL_WIDTH_LAYOUT_DEFAULT_IDEAL_COLUMN_WIDTH
+        : columnWidth;
+    const idealGutter = typeof gutter === 'undefined' ? FULL_WIDTH_DEFAULT_GUTTER : gutter;
```
IsraLopez26, line 29: Similar as above: 

`const idealGutter = gutter ?? FULL_WIDTH_DEFAULT_GUTTER ;
`

packages/gestalt/src/Masonry/getColumnCount.ts
```diff
@@ -0,0 +1,39 @@
...
+    const colguess = Math.floor(width / idealColumnWidth);
+    return Math.max(Math.floor((width - colguess * idealGutter) / idealColumnWidth), minCols);
+  }
+
+  const idealGutter = typeof gutter === 'undefined' ? DEFAULT_LAYOUT_DEFAULT_GUTTER : gutter;
```
IsraLopez26, line 34: Same simplifications as above :)

packages/gestalt/src/Masonry.tsx
```diff
@@ -674,15 +681,36 @@ export default class Masonry<T> extends ReactComponent<Props<T>, State<T>> {
...
+
+      let batchSize;
+      if (nextMultiColumnItem) {
+        const gridSize = getColumnCount({ gutter, columnWidth, width, minCols, layout });
+        const responsiveModuleConfigForSecondItem =
+          _getResponsiveModuleConfigForSecondItem && items[1]
+            ? _getResponsiveModuleConfigForSecondItem(items[1])
+            : undefined;
+
+        const moduleSize = calculateActualColumnSpan({
```
IsraLopez26, line 696: Optional: 
maybe I'm being picky, but just to be more descriptive, should we name this variable something like `itemColSpan`?

make the change only if there is any other feedback. Otherwise, do not create a new commit just for this

packages/gestalt/src/Masonry/getColumnCount.ts
```diff
@@ -0,0 +1,39 @@
...
+  minCols,
+  layout,
+}: {
+  gutter: number;
+  columnWidth: number | undefined;
+  width: number;
+  minCols: number;
+  layout: Layout;
+}): number {
+  if (layout === 'flexible' || (layout === 'serverRenderedFlexible' && width !== null)) {
```
IsraLopez26, line 21: It is unlikely that this condition changes, but this condition should always be in sync with the one under `getLayoutAlgorithm.ts`, so I think we could create a separate function that takes as arguments `layout` and `width`, to have this check logic centralized in one place. 

something like: 
```
export function getIsFlexibleLayout (layout, width) {
   return layout === 'flexible' || (layout === 'serverRenderedFlexible' && width !== null)
}
```

And use this function in `getColumnCount.ts` and `getLayoutAlgorithm.ts` such as: 

```
if (getIsFlexibleLayout) {
   ...
}
```

IsraLopez26: Also, I think we're missing the condition for `uniformRowLayout`
So I think we should do something like: 

```
if (getIsFlexibleLayout(layout, width) || layout === 'uniformRowFlexible') {
   ...
}
```

please let me know if this is clear

packages/gestalt/src/MasonryV2.tsx
```diff
@@ -341,7 +348,7 @@ function useFetchOnScroll({
       | undefined,
   ) => void;
   scrollTop: number;
-  scrollContainerElement: HTMLElement | Window | null | undefined;
+  scrollContainerElement: HTMLElement | Window | Window | null | undefined;
```
IsraLopez26, line 351: I think this is duplicate

packages/gestalt/src/MasonryV2.tsx
```diff
@@ -681,6 +714,14 @@ function Masonry<T>(
     }
   }, []);
 
+  const gutter: number = useMemo(() => {
+    let defaultGutter = DEFAULT_LAYOUT_DEFAULT_GUTTER;
+    if ((layout && layout === 'flexible') || layout === 'serverRenderedFlexible') {
```
IsraLopez26, line 719: Here we should add the new function created above, to always keep this condition in sync

packages/gestalt/src/MasonryV2.tsx
```diff
@@ -428,15 +436,40 @@ function useLayout<T>({
...
+      const gridSize = getColumnCount({ gutter, columnWidth, width, minCols, layout });
+      const responsiveModuleConfigForSecondItem =
+        _getResponsiveModuleConfigForSecondItem && items[1]
+          ? _getResponsiveModuleConfigForSecondItem(items[1])
+          : undefined;
+
+      const moduleSize = calculateActualColumnSpan({
+        columnCount: gridSize,
+        firstItem: items[0]!,
+        isFlexibleWidthItem: layout === 'flexible',
```
IsraLopez26, line 457: this condition is not correct, we should be doing something like: 

```
  const secondItem = originalItems[1]!;
  const responsiveModuleConfigForSecondItem = _getResponsiveModuleConfigForSecondItem(secondItem);
  const isFlexibleWidthItem = !!responsiveModuleConfigForSecondItem && nextMultiColumnItem === secondItem;
```

packages/gestalt/src/Masonry.tsx
```diff
@@ -674,15 +681,36 @@ export default class Masonry<T> extends ReactComponent<Props<T>, State<T>> {
...
+          columnCount: gridSize,
+          firstItem: items[0]!,
+          isFlexibleWidthItem: layout === 'flexible',
```
IsraLopez26, line 699: this condition is not correct, we should be doing something like: 

```
  const isFlexibleWidthItem = !!responsiveModuleConfigForSecondItem && nextMultiColumnItem === items[1];
```

packages/gestalt/src/Masonry/multiColumnLayout.ts
```diff
@@ -58,7 +70,7 @@ function getColumnSpanFromGridSize(columnSpanConfig: ColumnSpanConfig, gridSize:
 function getColumnSpanFromResponsiveModuleConfig(
   columnCount: number,
   firstItemColumnSpan: number,
-  responsiveModuleConfig: ResponsiveModuleConfig,
+  responsiveModuleConfig?: ResponsiveModuleConfig,
```
IsraLopez26, line 73: Why are we making this prop optional?

packages/gestalt/src/Masonry.tsx
JosAnGoCa: I think that we need to define how are we gonna call the multicolumn items, to be referenced and standardize the references

packages/gestalt/src/Masonry/getColumnCount.ts
JosAnGoCa: The condition was missing because this function is not used in `uniformRowLayout`, but I'll add the case and also use the function on `uniformRowLayout`

packages/gestalt/src/MasonryV2.tsx
JosAnGoCa: Yep, it was probably an error on the updating of the branch
thanks for noticing that

packages/gestalt/src/MasonryV2.tsx
JosAnGoCa: The problem is that the function uses the width as a parameter, so i can add the width as a dependency to the memo function but that will make the function to be called unnecessary times, so i would prefer to leave it like that

JosAnGoCa: But there is a fix on this condition, and its because this is different than the other cases, it should be like this: `layout && (layout === 'flexible' || layout === 'serverRenderedFlexible')`

packages/gestalt/src/MasonryV2.tsx
JosAnGoCa: You're right, I don't know why I did that

packages/gestalt/src/Masonry/multiColumnLayout.ts
JosAnGoCa: Another error that i left while I was making the PR update, thanks

manvoid (approved):

PR #4079, opened 2025-05-20
[Gestalt web][minor] Modal: Adding data test id prop

RicentCB:
#### What changed?

This PR introduces a new optional **dataTestId** prop for the **Modal** component. When provided, this prop is passed as a **`data-test-id`** attribute to the main container of the modal (the Box that wraps the header and body). This allows test frameworks to easily target and select specific modals for integration testing.

#### Why?

The purpose of this PR is to make it easier and more reliable to select modal components and their children in automated integration tests. Currently, it's tough to find the right modal in the DOM when running tests—especially for complex flows like the "parental passcode" feature—since there's no unique attribute to select. Relying on accessibility props isn't best practice for testing purposes. With the new **`dataTestId`** prop, engineers can confidently attach a custom identifier for testing without affecting accessibility or user experience.

### Links

- [TDD](https://docs.google.com/document/d/1wMGDoUdHiny09_nHzQ0JWf1R5QZDhbRgGEOY7dJrwzI/edit?tab=t.u8lpg069nfln)

### Checklist

- [X] Added unit tests

### DIFF
--- a/packages/gestalt/src/Modal.jsdom.test.tsx
+++ b/packages/gestalt/src/Modal.jsdom.test.tsx
@@ -109,4 +109,23 @@ describe('Modal', () => {
     expect(screen.getByLabelText('Bottom sheet')).toBeVisible();
     expect(screen.getByLabelText('Dismiss bottom sheet')).toBeVisible();
   });
+
+  test('Modal shows data test id when passed as a prop', ()=>{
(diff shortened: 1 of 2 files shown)


AlbertCarreras:
Repo is discontinued. It moved internally.

